Issue #17, opened 2024-10-22
Implement network behaviour


jking-aus:
Introduction of libp2p network behaviour. This behaviour needs to have the following protocols: 
- Gossipsub
- Identify 
- Ping

With the following supported transports: 
 - TCP 
 - Quic
 
With the following supported multiplexes: 
 - Yamux
 
With noise xx handshakes for encryption

diegomrsantos:
@jking-aus do you think we can consider this done?

PR #43, opened 2024-10-24
Stub `network` crate 

jimmygchen:
Addresses #17 partially.

Specification: https://github.com/ssvlabs/ssv-spec/tree/main/p2p


AgeManning (approved):
Great start! Nice work

AgeManning (approved):

jking-aus (approved):
looks good

anchor/network/src/keypair_utils.rs
```diff
@@ -0,0 +1,51 @@
...
+/// generated and is then saved to disk.
+///
+/// Currently only secp256k1 keys are allowed, as these are the only keys supported by discv5.
+pub fn load_private_key(network_dir: &PathBuf) -> Keypair {
+    // check for key from disk
+    let network_key_f = network_dir.join(NETWORK_KEY_FILENAME);
+    if let Ok(mut network_key_file) = File::open(network_key_f.clone()) {
+        let mut key_bytes: Vec<u8> = Vec::with_capacity(36);
+        match network_key_file.read_to_end(&mut key_bytes) {
+            Err(_) => debug!("Could not read network key file"),
```
diegomrsantos, line 19: Should we return an error if the key file exists but can't be read?

jimmygchen: Good question - the current behaviour is, if the file does not exist, is invalid or cannot be read, then we generate a new one. 

This is the same behaviour from Lighthouse - tbh I didn't really think about changing it when I copied this from LH. This node private key is mostly transparent to the user and used for node id, peer id and  noise protocol encryption. As it's transparent to the user, I think it's better UX to treat it as invalid if the file cannot be read, and regenerate a new one, instead of crashing and asking the user to fix it manually or delete the existing corrupted file.

PR #53, opened 2024-11-03
Add mermaid support for our mdbook

AgeManning:
## Issue Addressed

Just browsing our book and noticed we don't have mermaid support. 

This adds support for mermaid diagrams, however I've had to mess with the workflow and I've not tested the new build. 

This works locally, but it remains to be seen if this holds up in the CI


AgeManning:
Maybe we need to re-open this

jking-aus:
lgtm

Issue #52, opened 2024-11-01
Allow merge to stable when required


jking-aus:
CI checks to prevent merging PRs to stable branch which is intended behaviour generally, but do need to be able to allow this when required

magick93:
Hi @AgeManning 

I suspect this issue is due to some applied [rulesets](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-rulesets/about-rulesets) but I'm unable to see them due to permissions. 

Are you able to give me, at least, read permissions to this repo? I believe that is different from public, and should allow me to view the rules the repo.

AgeManning:
I've fixed the permissions. 

But I think this might be a non-issue. 

Our procedure should just be to merge to unstable with PRs, then manually merge to stable and we're all good

jking-aus:
closed with PR 54

PR #54, opened 2024-11-25
check when merging to stable

magick93:
## Issue Addressed

[52](https://github.com/sigp/anchor/issues/52)

## Proposed Changes

Fail only if merge is directed at `stable` and is not coming from `unstable`.


AgeManning (approved):
This seems good to me. 

Guess we can test it next merge :tada:

PR #55, opened 2024-11-26
Add mermaid support for our mdbook

AgeManning:
## Issue Addressed

Just browsing our book and noticed we don't have mermaid support.

This adds support for mermaid diagrams, however I've had to mess with the workflow and I've not tested the new build.

This works locally, but it remains to be seen if this holds up in the CI

Clone of #53


jking-aus (dismissed):
lgtm but confirm you want this merged to stable not unstable

AgeManning:
Nope, thanks for the catch.

Fixed

jking-aus (commented):
lgtm

jking-aus (approved):
lgtm

Issue #28, opened 2024-10-23
Docker CI for Anchor


jking-aus:
Create a Docker CI such that when we create a release it pushes to DockerHub

PR #56, opened 2024-11-27
Docker build and push - part 1

magick93:
## Issue Addressed

- Docker build and push only runs if tests and code quality passes
- Removed step to create multi-arch docker manifest. 
- Added labels to the docker image, including `git.revision` - for provenance. 

Which issue # does this PR address? See https://github.com/sigp/anchor/issues/28

NB: this is using the Hashicorp Vault Action to retrieve the docker username and key.

I'll followup in a less public forum regarding these details.

## Follow-up tasks

- [ ] Correctly configure HSC with correct details


.github/workflows/docker.yml
```diff
@@ -1,163 +1,134 @@
...
+                url: https://vault.sigp.io
+                method: github
+                githubToken: ${{ secrets.GH_TOKEN }}
+                secrets: |
+                    spesi_kv/data/dev/docker/anchor DOCKER_USERNAME ;
+                    spesi_kv/data/dev/docker/anchor DOCKER_PASSWORD
+
+            - name: Dockerhub login
+              run: |
+                  echo "${DOCKER_PASSWORD}" | docker login --username ${DOCKER_USERNAME} --password-stdin
```
AgeManning, line 91: I think these have changed. They are now `DH_ORG` and `DH_KEY` I think

.github/workflows/release.yml
```diff
@@ -1,251 +1,247 @@
...
+        tags:
+            - v*
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+env:
+    DOCKER_PASSWORD: ${{ secrets.DOCKER_PASSWORD }}
+    DOCKER_USERNAME: ${{ secrets.DOCKER_USERNAME }}
```
AgeManning, line 15: Here also

.github/workflows/release.yml
```diff
@@ -1,251 +1,247 @@
...
+
+            - name: Build Anchor for Windows
+              if:   matrix.arch == 'x86_64-windows'
+              run:  cargo install --path anchor --force --locked  --profile ${{ matrix.profile }}
+
+            - name: Configure GPG and create artifacts
+              if: startsWith(matrix.arch, 'x86_64-windows') != true
+              env:
+                  GPG_SIGNING_KEY: ${{ secrets.GPG_SIGNING_KEY }}
+                  GPG_PASSPHRASE: ${{ secrets.GPG_PASSPHRASE }}
```
AgeManning, line 102: I think i have to add these, or find someone who has them to add

.github/workflows/release.yml
```diff
@@ -1,251 +1,247 @@
...
+
+            - name: Create Release Draft
+              env:
+                  GITHUB_USER: ${{ github.repository_owner }}
+                  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+              # The formatting here is borrowed from OpenEthereum: https://github.com/openethereum/openethereum/blob/main/.github/workflows/build.yml
+              run: |
+                  body=$(cat <<- "ENDBODY"
+                  <Rick and Morty character>
```
AgeManning, line 190: @jking-aus - and others. We probably need to come up with a release naming scheme.

.github/workflows/release.yml
```diff
@@ -1,251 +1,247 @@
...
+
+                  ## Testing Checklist (DELETE ME)
+
+                  - [ ] Run on synced Holesky Sigma Prime nodes.
+                  - [ ] Run on synced Canary (mainnet) Sigma Prime nodes.
+                  - [ ] Resync a Holesky node.
+                  - [ ] Resync a mainnet node.
```
AgeManning, line 197: This is a draft release checklist. These will likely need to be modified at some point for anchor specifically. Things we want to check before a release. @jking-aus

.github/workflows/release.yml
```diff
@@ -1,251 +1,247 @@
...
+                  Priorities](https://anchor-book.sigmaprime.io/installation-priorities.html)
+                  more information about this table.*
+
+                  ## All Changes
+
+                  ${{ steps.changelog.outputs.CHANGELOG }}
+
+                  ## Binaries
+
+                  [See pre-built binaries documentation.](https://anchor-book.sigmaprime.io/installation-binaries.html)
```
AgeManning, line 230: We will need to make sure this URL exists, i.e add it to the md-book. @jking-aus

Cargo.lock
AgeManning, line 1: It looks like you may have done a `cargo update`. 

I think we should leave code changes to a separate PR and leave this one solely on the docker and devops changes. Will make things easy to reason about if a dependency is causing us issues.

Cargo.toml
```diff
@@ -35,7 +35,7 @@ clap = { version = "4.5.15", features = ["derive", "wrap_help"]}
 discv5 = "0.8.0"
 dirs = "5.0.1"
 either = "1.13.0"
-futures = "0.3.30"
+futures = "0.3.31"
```
AgeManning: Similarly here, will leave the cargo update for a code-based PR

Makefile
```diff
@@ -135,8 +136,9 @@ audit: install-audit audit-CI
 install-audit:
 	cargo install --force cargo-audit
 
+# TODO - remove when updated in LH(?)
 audit-CI:
-	cargo audit
+	cargo audit --ignore RUSTSEC-2024-0399
```
AgeManning: Are we failing this RUSTSEC?

Maybe we address this first in a separate PR?

anchor/src/environment.rs
```diff
@@ -8,6 +8,8 @@ use task_executor::{ShutdownReason, TaskExecutor};
 use tokio::runtime::{Builder as RuntimeBuilder, Runtime};
 use tracing::{error, info, warn};
 use tracing_subscriber::EnvFilter;
+
+#[cfg(target_family = "unix")]
```
AgeManning: Won't this break the application for non-unix machines? Usually we'd import non-unix crates that are compatible. 

I've not tested anchor on non-unix, but if this currently doesn't work on non-unix, we probably should address it correctly.

AgeManning (commented):
Nice. This looks great. 

Left a few comments about some code changes. 

The extra script files about getting the rust version (and splitting up the extract-version workflow) was this to avoid putting the architecture into the name? Or to handle the case that we don't yet have a tag or release? Just curious as to why we shifted from the original version we use in Lighthouse. 

Everything else looks good to me :)

.github/workflows/docker.yml
magick93: In these workflows we are not trusting github to responsibility protect the dockerhub token in perpetuity. Instead we give it on demand, and approval, via vault. 
While it does use `DOCKER_USERNAME` and `DOCKER_PASSWORD` vars internally, the only variable that needs to be set in Github is to add your org token. See https://www.notion.so/sigp/HashiCorp-Vault-12a56eefb55e8023bb58de9d57cfe2b8

Makefile
magick93: Will do.
Removed.

anchor/src/environment.rs
magick93: This was a mistake to include this change in this PR. 
The fix is done in https://github.com/sigp/anchor/pull/68

Cargo.lock
magick93: Removed and resolved.

magick93:
This `extract-workflow` is to reduce duplication. We need to get the version for both binary and docker releases. 
The arch is still in the name - see https://hub.docker.com/r/magick93/anchor/tags and https://github.com/magick93/anchor/releases.

The use of `.github/scripts/toml_reader.sh` was to simplify and for reuse. Also, I was thinking of using the rust version from the Cargo.toml as a value in the docker rust builder image. I'm currently not sure if this is needed - maybe `cross` takes care of this. I need to investigate further.

jking-aus:
hey anton can you just resolve the minor conflict so I can squash and merge? If I do it my review wont approve the pr

magick93:
> hey anton can you just resolve the minor conflict...

@jking-aus - done

jking-aus (approved):
lgtm

PR #57, opened 2024-11-27
Add Anchor Processor

### LINKED ISSUE #41: Implement Anchor Processor
The processor is a task that sits between the network and various client components.

As tasks can process items at different speeds, messages need to be prioritized over others and dropped in some cases to prevent backlogs of work. 

Furthermore, grouping all the heavy lifting blocking tasks (i.e signature verification) can be localized into logical task to optimize the bulk of the processing of the client. 

The processor schedules, queues and prioritizes work elements for the Anchor client. See the Architecture overview in the book for further high-level information. 

A similar concept is the [beacon processor](https://github.com/sigp/lighthouse/tree/stable/beacon_node/beacon_processor) in Lighthouse.

dknopik:
## Issue Addressed

- #41 

## Proposed Changes

Add a "processor" component, inspired by the beacon processor from lighthouse. It differs from Lighthouses' processor:
- Multiple input channels instead of a single input channel which is split into queues.
- No reprocessing queues. Could be added of course, but I think we should avoid those if possible to design around synchronisation primitives such as `Notify` etc instead.
- No "overflowing queues", which drop existing items instead of logging an error. Can be implemented.

## Additional Info

I've abandoned efforts to create a common base processor module used by both Lighthouse and Anchor, as this would basically replace glue with other glue. Also, Lighthouse has a lot of specialised behaviour that would be tough to abstract in a common module.

This is only a preliminary version we can use for development - we will of course need to reconsider and/or fine-tune later when we can actually test performance.

This is a draft, stuff like metrics is still missing.


Zacholme7:
My very initial thoughts

1. I like the distinction between async and blocking tasks, but im not sure on the  `WorkItem` actually containing the function to be executed. In my mind, a `WorkItem` is just a signal of something that has to be done. It is just message + state. It is then up to the processor to handle the correct routing of that item to the thing that will fulfill it. Something like registering a handler for each item. Also, we would have to `Box::pin` all the async functions which might get a little icky and theres some overhead with dynamic dispatch. 
2. Following up on 1, maybe we should consider if we want the processor to act as more of an central executor or a manager. As an central executor, the processor is receiving messages from the channels and then executing them. Maybe it makes sense for the processor to act as more of a manager with a set of workers it coordiantes/checks on, while the workers are actually receiving the messages and doing the heavy lifting. Just something I thought of on the fly and may have no relevance/be practical at all. 

Other than those two high level thoughts I like it. The priority via `biased` is nice. It is also difficult to design this without us knowing the full workload right now. Interested to see what others think too.

dknopik:
Thank you for looking through the PR :)

> 1. I like the distinction between async and blocking tasks, but im not sure on the  `WorkItem` actually containing the function to be executed. In my mind, a `WorkItem` is just a signal of something that has to be done. It is just message + state. It is then up to the processor to handle the correct routing of that item to the thing that will fulfill it. Something like registering a handler for each item. Also, we would have to `Box::pin` all the async functions which might get a little icky and theres some overhead with dynamic dispatch.

This is one of the aspects of the beacon processor it kept in. I like it because it saves us infrastructure (having to register a handler for each queue) and allows us to easily submit work items that are different but belong to the same priority category (=queue).

> 2. Following up on 1, maybe we should consider if we want the processor to act as more of an central executor or a manager. As an central executor, the processor is receiving messages from the channels and then executing them. Maybe it makes sense for the processor to act as more of a manager with a set of workers it coordiantes/checks on, while the workers are actually receiving the messages and doing the heavy lifting. Just something I thought of on the fly and may have no relevance/be practical at all.

What do you mean by worker? Basically a thread that continually runs and receives messages from the manager, servicing a single queue?

Zacholme7:
> This is one of the aspects of the beacon processor it kept in. I like it because it saves us infrastructure (having to register a handler for each queue) and allows us to easily submit work items that are different but belong to the same priority category (=queue).

Ah okay cool that makes sense. I have not dug too much into the beacon processor. 

> What do you mean by worker? Basically a thread that continually runs and receives messages from the manager, servicing a single queue?

Say we have a slot where multiple validators each have a duty to fulfill. The duty manager would send a message to the specialized workers  which in turn would do any specific processing needed and spin up their own QBFT instances and handle that entire life cycle. The processor here is then used to make sure everything is progressing as intended and the workers are still alive, etc. 

I suppose here we also have to make the distinction between having workers based on queue priority like is currently done or having workers that specialize in some duty such as attestations. But we would have to somehow work in priority for the latter.

Might make sense to just not over engineer and route it all through the processor and have that manage all the QBFT instances.

Just food for thought and thinking out loud.

eserilev:
re: a common base processor module: 

I've done some work in this [PR](https://github.com/sigp/lighthouse/pull/6448) to generalize aspects of the Lighthouse beacon processor. Mostly to experiment with a different scheduling algorithm. Its still very much a WIP and I'm not sure if these changes would be useful for Anchor, but mentioning it here just in case

dknopik:
@eserilev thanks! Being able to switch scheduling strategies is definitely valuable. However, your PR does not yet generalize over the possible work items. It motivated me again to think about a good way to do that though, would be nice to come up with something that supports both :)

@Zacholme7 ah, got you :) yeah, until now I was thinking about the processor purely as a scheduler for task admission. You are right, having it manage e.g. QBFT instances and "route" QBFT messages to the correct instances might be worth it, so I will explore that as well. Thanks for your input!

jking-aus:
just reviewed -- good start mate and I think it's on the right track. Fair bit of work to come in refining queues, scheduling and prioritisation imo but keen to hear where your head is at. 

On some of the discussion above: 
- Yes the intent is to have the processor instantiate and manage the QBFT instances based on inputs from the duties service, which is what I think Zach's comment is saying. 
- It should queue tasks/events from the duties service as work items as well as messages between qbft instances, the networking layer, and the crypto module. 
- The scheduling stuff in the beacon processor and Eitan's PR is worth looking at. We need some prioritisation logic eventually.

anchor/processor/src/lib.rs
```diff
@@ -0,0 +1,145 @@
...
+    executor.spawn(processor(config, receivers, executor.clone()), "processor");
+    senders
+}
+
+async fn processor(config: Config, mut receivers: Receivers, executor: TaskExecutor) {
+    // TODO: consider having separate limits for blocking and async?
+    let semaphore = Arc::new(Semaphore::new(config.max_workers));
+
+    loop {
+        let Ok(permit) = semaphore.clone().acquire_owned().await else {
```
AgeManning: Slowing the processor here makes sense. We could also add logic in this loop to reduce queues if we needed, i.e remove expired elements etc (we ideally don't want to do O(n) searchers tho). 

As you've commented above, the main thing I think we need to be weary of is the blocking tasks. Tokio already does a good job managing these for us with threads: https://docs.rs/tokio/latest/tokio/task/fn.spawn_blocking.html

As I think we might be passing some non-blocking async messages through here, we probably want to be as permissive as possible. i.e I think its fine to dump 1000 non-blocking tasks into the tokio executor and let it sort it out and not really worry about how many things are accessing state at once. 

However for blocking tasks we should definitely be more careful and limit these as you've done here. Perhaps we don't want the handful of blocking tasks to prevent us from processing lots of non-blocking tasks

anchor/processor/src/lib.rs
```diff
@@ -0,0 +1,145 @@
...
+
+    pub fn send_async(&mut self, future: AsyncFn) {
+        self.send_work_item(WorkItem::new_async(self.name, future));
+    }
+
+    pub fn send_blocking(&mut self, func: BlockingFn) {
+        self.send_work_item(WorkItem::new_blocking(self.name, func));
+    }
+
+    pub fn send_work_item(&mut self, item: WorkItem) {
```
AgeManning: At some point, we might need to return an error here to handle the case of the queue being full

AgeManning (commented):
Looks good!

I read some of the comments. In terms of routing messages to QBFT instances, we had envisaged each instance being tagged with a unique `InstanceId`. 

The Processor (or some state attached to it) would build a mapping of `<InstanceId, Sender>` and `<InstanceId, Receiver>` when creating them. 
So we can always just send/recv messages from each instance. 

I dont think we need to worry to much about checking if they are running or not. We can have metrics to see how many we have running at any given time, but in general I think its fine to assume they will always complete, either success or failure in at most X time. There is no scenario where a QBFT task cannot end after X time (this saves us checking on them, we can just spawn and forget).

anchor/processor/src/lib.rs
dknopik: Good idea, done

anchor/processor/src/lib.rs
dknopik: Yeah, sounds good. I implemented a queue for async and quick tasks, which does not need a permit from the semaphore, as well as work expiry (which is only checked on queue retrieval right now).

dknopik:
I think we should go with the simple variant for the processor instead of sharing code with Lighthouse for now - Lighthouse's needs are pretty complex, and Anchor's needs not fully known yet, so for time budget reasons it is probably better to start with the non-modularized version and later revisit an implementation using the modular processor @eserilev is cooking up. :)

anchor/processor/src/lib.rs
```diff
@@ -0,0 +1,356 @@
...
+        }
+    }
+
+    /// Create an immediate work task. Has access to the [`ProcessorState`], and is thus ideal for
+    /// triggering some process, e.g. via a queue retrieved from the state. Must *NEVER* block!
+    ///
+    /// The [`DropOnFinish`] should be dropped when the work is done, for proper permit accounting
+    /// and metrics. This includes any work triggered by the closure, so [`DropOnFinish`] should
+    /// be sent along if any other process such as a QBFT instance is messaged.
+    pub fn new_immediate(name: &'static str, func: ImmediateFn) -> Self {
```
jking-aus, line 201: this is a good addition for state change and queue management

jking-aus (dismissed):
good work mate, concise and logical. 

I think for todos:
- We should use a more structured type definition for the queues: i.e. 
`pub Struct QueueConfig {name: str, capacity: usize, permit: bool}` then a vec of queues or something to define the queue type. Would make a few things like monitoring, runtime config and queue creation easier. 
- Maybe graceful shutdown handling (nice to have) 
- Consider resource limits and performance monitoring (nice to have)

dknopik:
Thanks for the review!

>     * We should use a more structured type definition for the queues: i.e.
>       `pub Struct QueueConfig {name: str, capacity: usize, permit: bool}` then a vec of queues or something to define the queue type. Would make a few things like monitoring, runtime config and queue creation easier.

Yeah, that might make sense. I am not sure about the configurability of needing a permit though - This is currently very ingrained into the way I implemented the main `select!`. Also, it doesn't really add value to have multiple permitless input queues, as the semantic  meaning of that queue is basically "throw into tokio asap and let it deal with it". The tasks in there should therefore be very quick or very non-blocking (basically `await`ing most of the time) - and if we have to prioritize which task we add to tokio first (just add, not execute!), that would mean that we produce these tasks faster than we can add them, and then something is seriously wrong anyway :)

>     * Maybe graceful shutdown handling (nice to have)

Oh yeah - that's a good idea. I wonder if we need an explicit shutdown queue or if dropping the senders is enough as shutdown signal :thinking: probably not due to the risk of deadlocking?
 
>     * Consider resource limits and performance monitoring (nice to have)

We have the semaphore as a limit already. What kind of limit do you envision? Also, for monitoring, what do you have in mind in addition to the metrics?

jking-aus:
> Thanks for the review!
> 
> > ```
> > * We should use a more structured type definition for the queues: i.e.
> >   `pub Struct QueueConfig {name: str, capacity: usize, permit: bool}` then a vec of queues or something to define the queue type. Would make a few things like monitoring, runtime config and queue creation easier.
> > ```
> 
> Yeah, that might make sense. I am not sure about the configurability of needing a permit though - This is currently very ingrained into the way I implemented the main `select!`. Also, it doesn't really add value to have multiple permitless input queues...

yep all good - permitless queue could be a separate type and it be fine. Just made sense to me to have it all in one for ease of use, even if it almost always false. 

> 
> > ```
> > * Maybe graceful shutdown handling (nice to have)
> > ```
> 
> Oh yeah - that's a good idea. I wonder if we need an explicit shutdown queue or if dropping the senders is enough as shutdown signal 🤔 probably not due to the risk of deadlocking?
> 
I could be wrong but I think only deadlock risk is blocked tasks with indefinite awaits. For now dropping senders OK but I think we need to handle the blocking tasks more carefully in next ver. Could we just use Notify? 

> > ```
> > * Consider resource limits and performance monitoring (nice to have)
> > ```
> 
> We have the semaphore as a limit already. What kind of limit do you envision? Also, for monitoring, what do you have in mind in addition to the metrics?

was thinking more for in the future when we need to optimise throughput or check backpressure -- queue latency, processing time, resource usage, etc. Not for this version but an idea for when we've got mvp and are finetuning.

jking-aus (approved):
lgtm

PR #58, opened 2024-11-27
Create a dedicated `ssv_types` crate. 

Zacholme7:
## Issue Addressed

N/A

## Proposed Changes

This PR starts a `ssv_types` crate within `anchor/common`. It contains core SSV types in accordance with the [spec](https://github.com/ssvlabs/ssv-spec). It does not contain related methods as we still need to determine what is needed, but just having these types defined helps. Validator related types are imported via the `types` crate in lighthouse. 

## Additional Info

This is just an initial draft to get the ball rolling. It will change as we see fit. Also [waiting on](https://github.com/sigp/lighthouse/pull/6621) re-export of `ValidatorIndex` in lighthouse into unstable.


anchor/common/ssv_types/src/operator.rs
```diff
@@ -0,0 +1,14 @@
+/// Unique identifier for an Operator
+pub type OperatorID = u64;
+
+/// Operator RSA public key
+pub type OperatorPublicKey = [u8; 459];
```
AgeManning: If we are using RSA, we probably should use a specific library and its associated public key type. 

Each library will have an RSAPublicKey struct. This object will have been verified when created or deserialized. If we use this type, then there is no chance that we are carrying around invalid/malicious bytes in other types. 

I think we want to deserialize any public key we get from the network instantly or reject the message, then we carry around only verified and valid keys throughout our types. 

i.e our types shouldn't really be bytes anywhere, unless we have to have them.

anchor/common/ssv_types/src/operator.rs
```diff
@@ -0,0 +1,14 @@
+/// Unique identifier for an Operator
+pub type OperatorID = u64;
```
AgeManning: For things that we will be using a lot, we should use the new type idiom: https://doc.rust-lang.org/rust-by-example/generics/new_types.html

This enforces the type at compile time, so we can't get confused by accidentally putting another type that is a u64 in place of this. 

For the operator-id, I have already done this. See [here](https://github.com/sigp/anchor/blob/stable/anchor/qbft/src/types.rs#L53) - I used usize, (I didn't check the specs :p, if the specs say u64, then we should change it to that).

anchor/common/ssv_types/src/share.rs
```diff
@@ -0,0 +1,51 @@
...
+use std::time::SystemTime;
+use types::{Address, Domain, Graffiti, PublicKey};
+
+type ValidatorIndex = usize;
+
+/// Share of a key that a operator owns and accompanying metadata
+#[derive(Debug, Clone)]
+pub struct SSVShare {
+    pub share: Share,
+    pub metadata: Metadata,
```
AgeManning, line 16: You've been awesome with comments on all the fields, might as well be consistent here

anchor/common/ssv_types/src/committee.rs
```diff
@@ -0,0 +1,23 @@
+use crate::{Operator, OperatorID, OperatorPublicKey};
+use types::Domain;
+
+/// Unique identifier for a committee.
+pub type CommitteeID = u64;
```
AgeManning: Same here as operator id. 

Its a bit cumbersome, but it does pay dividends in the future

AgeManning (commented):
Nice! I had a few comments

anchor/common/ssv_types/src/operator.rs
Zacholme7: Follow up from internal message. Pulled in the [rsa crate](https://crates.io/crates/rsa). We are facing a decoding issue as the public key is formatted as a PKCS#8 key but is wrapped in a PKCS#1 header. If the header matched the key, we would be able to decode this fine but the rsa crate rejects the decoding as the header does not match the underlying key. The options to fix this is to replace the header as I have implemented, or I believe you can strip it and then parse the key. Thoughts on this approach?

AgeManning: Replacing the header I think is the way to go, as an interop compatibility layer. In the future if they decide to move to be compliant we can just remove the interop layer (i.e replacing the header)

jking-aus:
good work but I think we'll go with your other idea -- closing this pr

Issue #23, opened 2024-10-23
Sync Smart Contract Data


AgeManning:
Anchor needs to sync committee data from the execution layer. 

We need to sync from the execution layer the smart contract data for SSV. This is for all operators so that our client can funciton normally for all operators if the keys are switched. 

A process similar to the [lighthouse deposit sync](https://github.com/sigp/lighthouse/tree/stable/beacon_node/eth1) can be used. 

We will need to store our current sync state into our db to prevent resync on restart.

Once synced, we need to listen to new events to update our view of the committee data.

An example of handling these events is [here](https://github.com/ssvlabs/ssv/blob/main/eth/eventhandler/handlers.go) and an example of syncing is [here](https://github.com/ssvlabs/ssv/blob/main/eth/eventsyncer/event_syncer.go)

Issue #20, opened 2024-10-22
Execution layer 


jking-aus:
# Integration layer
This is a major task to track completion of network integration for the anchor client

## Issues
- [ ] #23

PR #59, opened 2024-11-27
Execution integration

Zacholme7:
## Issue Addressed

- #20 
- #23 

## Proposed Changes

This PR tracks the integration of the execution layer into anchor. The process can be thought of as the following: 
1. Fetch all relevant contract events from the chain
2. Perform verification and parsing on the events and then store them into the database.
3. If we are caught up to the chain and following along, determine if some action needs to be taken in response to the event. 

## Additional Info

Types and the database are being worked on alongside this as they all mesh together.


anchor/eth/src/sync.rs
```diff
@@ -0,0 +1,248 @@
...
+        // start the live sync, options
+        // 1) spawn the sync off in its own long running task and return
+        // 2) transition into live sync and signal AtomicBool to coordinator
+        self.live_sync().await?;
+        todo!()
+    }
+
+    /// Perform a historical sync from the contract deployment block to catch up to the current
+    /// state of the SSV network
+    async fn historical_sync(&self) -> Result<(), String> {
```
jimmygchen: Would be nice to add some logging to this function to show sync progress.

anchor/eth/src/sync.rs
```diff
@@ -0,0 +1,248 @@
...
+            // Chunk the start and end block range into a set of ranges of size BATCH_SIZE
+            // and construct a future to fetch the logs in each range
+            let tasks: Vec<_> = (start_block..=current_block)
+                .step_by(BATCH_SIZE as usize)
+                .map(|start| {
+                    let (start, end) =
+                        (start, std::cmp::min(start + BATCH_SIZE - 1, current_block));
+                    self.fetch_logs(start, end)
+                })
+                .collect();
```
jimmygchen, line 225: Is it worth processing the batches (sequentially) as soon as we receive them? 
In the first iteration of the loop, this would collect all the logs since `start_block` until `current_block` into `tasks`, which _may_ eventually grow quite big right?

anchor/eth/src/sync.rs
```diff
@@ -0,0 +1,248 @@
...
+            for log in event_logs {
+                let block_num = log.block_number.ok_or("Log is missing block number")?;
+                ordered_event_logs.entry(block_num).or_default().push(log);
+            }
+
+            // join them back to a vec in ordered format
+            let ordered_event_logs: Vec<Log> = ordered_event_logs.into_values().flatten().collect();
+
+            // Logs are all fetched from the chain and in order, process them but do not send off to
+            // be processed
```
jimmygchen: I don't quite understand this bit:
> process them but do not send off to be processed

anchor/eth/src/sync.rs
```diff
@@ -0,0 +1,248 @@
...
+});
+
+/// Contract deployment address
+/// https://etherscan.io/address/0xDD9BC35aE942eF0cFa76930954a156B3fF30a4E1
+static CONTRACT_DEPLOYMENT_ADDRESS: LazyLock<Address> =
+    LazyLock::new(|| address!("DD9BC35aE942eF0cFa76930954a156B3fF30a4E1"));
+
+/// Contract deployment block on Ethereum Mainnet
+/// https://etherscan.io/tx/0x4a11a560d3c2f693e96f98abb1feb447646b01b36203ecab0a96a1cf45fd650b
+const CONTRACT_DEPLOYMENT_BLOCK: u64 = 17507487;
```
jimmygchen: Would be nice making this configurable so we can test it on local devnet and Holesky too.

anchor/eth/src/abi/ssv_contract.json
```diff
@@ -0,0 +1,2125 @@
+[
```
jimmygchen: Where is this contract from?
It looks slightly different to this one:
https://github.com/ssvlabs/ssv/blob/main/eth/contract/contract.abi

anchor/eth/src/sync.rs
```diff
@@ -0,0 +1,248 @@
...
+            Ok(sub) => sub.into_stream(),
+            Err(_) => todo!(), // have some reconnect mechansim
+        };
+
+        // Stream in new block headers
+        while let Some(block_header) = stream.next().await {
+            // fetch the logs and process with execute
+            let relevant_block = block_header.number - FOLLOW_DISTANCE;
+            let logs = self.fetch_logs(relevant_block, relevant_block).await?;
+            self.event_processor.process_logs(logs, true)?;
```
jimmygchen: Just a note that we'll eventually need some error handling here, and also worth adding some logging to show progress.

anchor/eth/src/sync.rs
```diff
@@ -0,0 +1,248 @@
...
+        // event ValidatorRemoved(address indexed owner, uint64[] operatorIds, bytes publicKey, Cluster cluster);
+        SSVContract::ValidatorRemoved::SIGNATURE_HASH,
+        // event ClusterLiquidated(address indexed owner, uint64[] operatorIds, Cluster cluster);
+        SSVContract::ClusterLiquidated::SIGNATURE_HASH,
+        // event ClusterReactivated(address indexed owner, uint64[] operatorIds, Cluster cluster);
+        SSVContract::ClusterReactivated::SIGNATURE_HASH,
+        // event FeeRecipientAddressUpdated(address indexed owner, address recipientAddress);
+        SSVContract::FeeRecipientAddressUpdated::SIGNATURE_HASH,
+        // event ValidatorExited(address indexed owner, uint64[] operatorIds, bytes publicKey);
+        SSVContract::ValidatorExited::SIGNATURE_HASH,
```
jimmygchen: I'm not very familiar with the execution side of things - but there is a longer term plan to remove log bloom [EIP-7558](https://eips.ethereum.org/EIPS/eip-7668), would this impact reading these logs from the EL? Is it worth considering adding an abstraction for this in the future?

jimmygchen (commented):
Nice! great progress @Zacholme7  :clap: 

I know this is WIP, but I thought some early feedback after a quick read might be helpful.

I've added some comments and questions above, and here are a few more:
- Would be useful to add more logging, both for informative and debug purpose (e.g. sync progress, live events received, fetch or processing errors etc)
- I'd love to see some unit & integration tests where it make sense. It doesn't have to be this PR, but I wanted to mention it since we recently started adding tests in the validator client, and it's been a little painful. Adding tests while the codebase is fresh can make things a bit more testable by design and save some headaches later.

I'm not familiar with SSV, so my comments are mostly generic, but I'd like to catch up and review this again later when it's ready.

Overall it's very tidy and nice to read, great job 👍

anchor/eth/src/sync.rs
Zacholme7: Potentially yes. I have not confirmed yet but I do not believe the current total amount of events emitted by the contract is very much. Although, as the network gains popularity and continues on over the years this may be a problem so it is probably best to make this change. Thanks!

anchor/eth/src/sync.rs
Zacholme7: ha this is very non-descriptive comment, i will change. When a node starts it will sync the history to store in the database. This is all old data that we just care about knowing. The node will then transition to a live sync where it is caught up to the network and needs to react to new information. So if we are in historical sync and get a `ValidatorRemoved` event we just want to save it and continue as it is old information. If we are in live sync and get a `ValidatorRemoved` event, we want to send this event to the processor to potentially take some action. Here it would be removing the validator from our working set if we are an operator for it.

anchor/eth/src/abi/ssv_contract.json
Zacholme7: From [here](https://etherscan.io/address/0xa2f1dadbb9e836b7ec47330ff9e5947d2f36fc35#code). I believe the one in the repo is old but the only thing that matters is the event signatures which they cannot change. I may also just remove the entire json and just define all the events with `sol!` since that is all that matters anyways.

anchor/eth/src/sync.rs
Zacholme7: ~~Oh wow, yea something we definitely have to consider. The entire node hinges on the events so may let the SSV team know and see if they have put some thought into it~~

Update I actually don't think this matters. The bloom filter is just an optimization for querying logs but the logs are still stored in the receipt. I will keep this in mind and make sure though.

anchor/eth/src/sync.rs
```diff
@@ -0,0 +1,411 @@
...
+
+/// RPC and WS clients types
+type RpcClient = RootProvider<Http<Client>>;
+type WsClient = RootProvider<PubSubFrontend>;
+
+/// Retry information for log fetching
+const MAX_RETRIES: i32 = 5;
+
+// Block follow distance
+const FOLLOW_DISTANCE: u64 = 8;
```
jking-aus, line 72: check with ssv on reason for 8 block follow distance

jking-aus (approved):
reviewed in call on 14/1 -- nice work zac good to get this one finished

PR #60, opened 2024-12-03
fix lint `unnecessary_lazy_evaluations` in `build_profile_name`

dknopik:
This clippy warning has been bothering me. I used the opportunity to also remove one (1) allocation.

PR #66, opened 2024-12-06
Unwrap clippy

jking-aus:
fix clippy in client/cli.rs


dknopik (commented):
This shows as having no files changed after merging - therefore, lgtm :D

But I think what happened here is that you also wanted to fix the same thing as I did in #60, which is merged by now - so I think we can close this?

jking-aus:
yep you fixed it in 60 my bad

Issue #21, opened 2024-10-23
Database


jking-aus:
Need to create a database in SQLite that will store: 
- State retrieved from the execution layer
- Peer data
- Any other information that must persist between sessions

PR #67, opened 2024-12-06
`NetworkDatabase` impl with new `ssv_types` system

Zacholme7:
## Issue Addressed

- #21 

## Proposed Changes

This PR introduces ~~a new `ssv_types` system and associated~~ a `NetworkDatabase`. 

Edit: New `ssv_types` split into #69 

The `NetworkDatabase` portion of this PR is a MVP of the database which holds all of the types information above. The database is just simple CRUD with in memory stores for relevant information. It will store all of the information during execution event sync and turn it into in memory stores that we deem relevant. These stores will be recreated via the DB upon node restart.


jking-aus (approved):
great work here zac, happy with it based on our walkthrough review yesterday. Let's try and break these down a bit in future so we can merge faster and get more collaborative with the other guys. sick effort though

anchor/common/ssv_types/src/cluster.rs
```diff
@@ -1,36 +1,40 @@
...
+    /// The Eth1 fee address for all validators in the cluster
+    pub fee_recipient: Address,
     /// The number of faulty operator in the Cluster
     pub faulty: u64,
     /// If the Cluster is liquidated or active
     pub liquidated: bool,
-    /// Metadata about the validator this committee represents
-    pub validator_metadata: ValidatorMetadata,
+    /// Operators in this cluster
+    pub cluster_members: HashSet<OperatorId>,
```
diegomrsantos, line 27: Should this be a Committee?

Zacholme7: I think committee is too broad of a term here. A committee just refers to a general set of operators while here the `cluster_members` are meant to be the direct members of the cluster. Its trying to enforce the relationship. But, there is also an argument that committee is also applicable.

PR #69, opened 2024-12-09
New `ssv_types`

Zacholme7:
## Issue Addressed

N/A

## Proposed Changes

This PR introduces a new `ssv_types` system.

The types follow a hierarchical structure of roughly the following. This cleanly represents all SSV specific information.
- Operator
    - Operator ID
    - Operator Public Key
- Cluster
    - Cluster ID
    - Validator Metadata
    - Cluster Members
        - Identifiers
        - Share

It does not contain related methods as we still need to determine what is needed, but this is an initial step to get the ball rolling.

### DIFF
--- a/anchor/common/ssv_types/src/cluster.rs
+++ b/anchor/common/ssv_types/src/cluster.rs
@@ -1,0 +1,53 @@
+use crate::OperatorId;
+use crate::Share;
+use derive_more::{Deref, From};
+use types::{Address, Graffiti, PublicKey};
+
+/// Unique identifier for a cluster
+#[derive(Clone, Copy, Debug, Default, Eq, PartialEq, Hash, From, Deref)]
+pub struct ClusterId(pub u64);
+
+/// A Cluster is a group of Operators that are acting on behalf of a Validator
+#[derive(Debug, Clone)]
+pub struct Cluster {
+    /// Unique identifier for a Cluster
+    pub cluster_id: ClusterId,
+    /// All of the members of this Cluster
+    pub cluster_members: Vec<ClusterMember>,
+    /// The number of faulty operator in the Cluster
+    pub faulty: u64,
+    /// If the Cluster is liquidated or active
+    pub liquidated: bool,
+    /// Metadata about the validator this committee represents
+    pub validator_metadata: ValidatorMetadata,
+}
+
+/// A member of a Cluster. This is just an Operator that holds onto a share of the Validator key
+#[derive(Debug, Clone)]
+pub struct ClusterMember {
+    /// Unique identifier for the Operator this member represents
+    pub operator_id: OperatorId,
+    /// Unique identifier for the Cluster this member is a part of
+    pub cluster_id: ClusterId,
+    /// The Share this member is responsible for
+    pub share: Share,
+}
+
+/// Index of the validator in the validator registry.
+#[derive(Clone, Copy, Debug, Default, Eq, PartialEq, Hash, From, Deref)]
+pub struct ValidatorIndex(pub usize);
+
+/// General Metadata about a Validator
+#[derive(Debug, Clone)]
+pub struct ValidatorMetadata {
+    /// Index of the validator
+    pub validator_index: ValidatorIndex,
+    /// Public key of the validator
+    pub validator_pubkey: PublicKey,
+    /// Eth1 fee address
+    pub fee_recipient: Address,
+    /// Graffiti
+    pub graffiti: Graffiti,
+    /// The owner of the validator
+    pub owner: Address,
+}
--- a/anchor/common/ssv_types/src/cluster.rs
+++ b/anchor/common/ssv_types/src/cluster.rs
@@ -8,7 +8,7 @@ use types::{Address, Graffiti, PublicKey};
 pub struct ClusterId(pub u64);
 
 /// A Cluster is a group of Operators that are acting on behalf of a Validator
-#[derive(Debug, Clone)]
+#[derive(Clone)]
 pub struct Cluster {
     /// Unique identifier for a Cluster
     pub cluster_id: ClusterId,
@@ -23,7 +23,7 @@ pub struct Cluster {
 }
 
 /// A member of a Cluster. This is just an Operator that holds onto a share of the Validator key
-#[derive(Debug, Clone)]
+#[derive(Clone)]
 pub struct ClusterMember {
     /// Unique identifier for the Operator this member represents
     pub operator_id: OperatorId,
--- a/anchor/common/ssv_types/src/cluster.rs
+++ b/anchor/common/ssv_types/src/cluster.rs
@@ -8,7 +8,7 @@ use types::{Address, Graffiti, PublicKey};
 pub struct ClusterId(pub u64);
 
 /// A Cluster is a group of Operators that are acting on behalf of a Validator
-#[derive(Clone)]
+#[derive(Debug, Clone)]
 pub struct Cluster {
     /// Unique identifier for a Cluster
     pub cluster_id: ClusterId,
@@ -23,7 +23,7 @@ pub struct Cluster {
 }
 
 /// A member of a Cluster. This is just an Operator that holds onto a share of the Validator key
-#[derive(Clone)]
+#[derive(Debug, Clone)]
 pub struct ClusterMember {
     /// Unique identifier for the Operator this member represents
     pub operator_id: OperatorId,
--- a/anchor/common/ssv_types/src/operator.rs
+++ b/anchor/common/ssv_types/src/operator.rs
@@ -1,0 +1,60 @@
+use crate::util::parse_rsa;
+use derive_more::{Deref, From};
+use openssl::pkey::Public;
+use openssl::rsa::Rsa;
+use std::cmp::Eq;
+use std::fmt::Debug;
+use std::hash::Hash;
+use types::Address;
+
+/// Unique identifier for an Operator.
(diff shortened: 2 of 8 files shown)


anchor/common/ssv_types/src/cluster.rs
```diff
@@ -0,0 +1,53 @@
...
+#[derive(Debug, Clone)]
+pub struct Cluster {
+    /// Unique identifier for a Cluster
+    pub cluster_id: ClusterId,
+    /// All of the members of this Cluster
+    pub cluster_members: Vec<ClusterMember>,
+    /// The number of faulty operator in the Cluster
+    pub faulty: u64,
+    /// If the Cluster is liquidated or active
+    pub liquidated: bool,
```
jking-aus, line 20: I don't see a liquidated flag in the specs but can understand why it would be useful. Is there no analogous implementation in the go client?

anchor/common/ssv_types/src/cluster.rs
```diff
@@ -0,0 +1,53 @@
...
+use derive_more::{Deref, From};
+use types::{Address, Graffiti, PublicKey};
+
+/// Unique identifier for a cluster
+#[derive(Clone, Copy, Debug, Default, Eq, PartialEq, Hash, From, Deref)]
+pub struct ClusterId(pub u64);
+
+/// A Cluster is a group of Operators that are acting on behalf of a Validator
+#[derive(Debug, Clone)]
+pub struct Cluster {
```
jking-aus, line 12: I think domain needs to be included on the Cluster unless you had another plan for that implementation

anchor/common/ssv_types/src/cluster.rs
```diff
@@ -0,0 +1,53 @@
...
+    pub validator_metadata: ValidatorMetadata,
+}
+
+/// A member of a Cluster. This is just an Operator that holds onto a share of the Validator key
+#[derive(Debug, Clone)]
+pub struct ClusterMember {
+    /// Unique identifier for the Operator this member represents
+    pub operator_id: OperatorId,
+    /// Unique identifier for the Cluster this member is a part of
+    pub cluster_id: ClusterId,
```
jking-aus, line 31: I dont think we need clusterid here since ClusterMember is already contained within a Cluster -- suggest drop unless there's a reason

anchor/common/ssv_types/src/share.rs
```diff
@@ -0,0 +1,8 @@
+use types::PublicKey;
+
+/// One of N shares of a split validator key.
+#[derive(Debug, Clone)]
+pub struct Share {
+    /// The public key of this Share
+    pub share_pubkey: PublicKey,
```
jking-aus, line 7: would we not need operatorID here?

anchor/common/ssv_types/src/cluster.rs
```diff
@@ -0,0 +1,53 @@
+use crate::OperatorId;
+use crate::Share;
+use derive_more::{Deref, From};
+use types::{Address, Graffiti, PublicKey};
+
+/// Unique identifier for a cluster
+#[derive(Clone, Copy, Debug, Default, Eq, PartialEq, Hash, From, Deref)]
+pub struct ClusterId(pub u64);
```
jking-aus, line 8: I know there is some terminology overlap but it's probably better we stick with "committee" for now so we're using the same terminology as the other client team

anchor/common/ssv_types/src/cluster.rs
```diff
@@ -0,0 +1,53 @@
...
+    pub share: Share,
+}
+
+/// Index of the validator in the validator registry.
+#[derive(Clone, Copy, Debug, Default, Eq, PartialEq, Hash, From, Deref)]
+pub struct ValidatorIndex(pub usize);
+
+/// General Metadata about a Validator
+#[derive(Debug, Clone)]
+pub struct ValidatorMetadata {
```
jking-aus, line 42: is there a reason we've pulled out the committee ID and domaintype from here?

anchor/common/ssv_types/src/operator.rs
```diff
@@ -0,0 +1,60 @@
...
+use std::hash::Hash;
+use types::Address;
+
+/// Unique identifier for an Operator.
+#[derive(Clone, Copy, Debug, Default, Eq, PartialEq, Hash, From, Deref)]
+pub struct OperatorId(pub u64);
+
+/// Client responsible for maintaining the overall health of the network.
+#[derive(Debug, Clone)]
+pub struct Operator {
```
jking-aus, line 16: is there any benefit to tracking our cluster memberships here or is that being handled in the DB?

jking-aus (commented):
I think this is the way to go - let's get it passing CI and have a chat about a few things then merge it down

edit: seeing the CI flags are dependencies --  let's just have a chat on the below and we'll merge anyway

anchor/common/ssv_types/src/cluster.rs
dknopik: An argument for "cluster" is that "committee" might be confusable with the committees from Ethereum consensus

anchor/common/ssv_types/src/share.rs
dknopik: The `Share` is part of `ClusterMember`, and the operator ID can be found there

anchor/common/ssv_types/src/share.rs
```diff
@@ -0,0 +1,8 @@
+use types::PublicKey;
+
+/// One of N shares of a split validator key.
+#[derive(Debug, Clone)]
+pub struct Share {
```
dknopik, line 5: Should we add the (encrypted) private key here?

anchor/common/ssv_types/src/operator.rs
dknopik: IMO that should be handled in the db

anchor/common/ssv_types/src/cluster.rs
dknopik: I think this type is supposed to basically represent the metadata of the validator operated by the cluster, not the cluster itself, so I think I like leaving it out here.

anchor/common/ssv_types/src/cluster.rs
Zacholme7: Its on the metadata [here](https://github.com/ssvlabs/ssv/blob/2315559a59c7233b37a55f716b626243cbbb5074/protocol/v2/types/ssvshare.go#L151) in the client.

anchor/common/ssv_types/src/cluster.rs
Zacholme7: I left domain and some other specific Validator data off since a lot of it is going to be included via the lighthouse types. If we realize it makes more sense to add it here we can do that.

anchor/common/ssv_types/src/cluster.rs
Zacholme7: Yep Daniel mentioned my idea behind Cluster. Im good to change it but in that case we should name it something like SSVCommittee just to be explicit about the difference.

anchor/common/ssv_types/src/cluster.rs
Zacholme7: Good point, but I think this depends on how we want to structure the maps for the database. If we want to do something like OperatorID => Vec of ClusterMembers, then this ClusterID is important. If we do ClusterID => Vec of ClusterMembers, then it is not important. The design of the database quite tightly influences the design of the types and it's hard to anticipate what the right structure is at this point.

anchor/common/ssv_types/src/operator.rs
Zacholme7: Yep, I recommend keeping Operator self contained and handling this in the DB

anchor/common/ssv_types/src/share.rs
Zacholme7: I had it in but removed it right before the PR as I convinced myself it wasn't needed in the Share for some reason. Thinking on it now it should probably be here. I'll add it back in thanks!

anchor/common/ssv_types/src/share.rs
Zacholme7: Something I considered but this is also a it depends on the structure of the database. We do have the access to the OperatorID from the ClusterMember. My idea was to go from ClusterID => Share where the Share is owned by the current operator, so storing the ID on the share is not needed in this case.

anchor/common/ssv_types/src/cluster.rs
Zacholme7: Yep, this is just about the Validator the cluster is acting on behalf of. Can be stored in database as ClusterID => ValidatorMetadata so this info still gets included. As for the domain, I left it off because I believe it is being included via lighthouse types, but I have to double check to confirm what it gets included through.

jking-aus (approved):
lgtm

anchor/common/ssv_types/src/operator.rs
```diff
@@ -0,0 +1,60 @@
...
+    /// ID to uniquely identify this operator
+    pub id: OperatorId,
+    /// Base-64 encoded PEM RSA public key
+    pub rsa_pubkey: Rsa<Public>,
+    /// Owner of the operator
+    pub owner: Address,
```
diegomrsantos, line 22: Is this necessary? I think I didn't see it in the go code

Zacholme7: The public key is used for initial operator identification. When you first join as an operator and do not yet know your id, the public key is used to match to event logs and figure out which id is yours. 

The owner address is where the SSV you get paid for running an operator is sent to

PR #70, opened 2024-12-11
Add gossipsub behaviour

diegomrsantos:
## Issue Addressed

Adds the Gossipsub Behavior for https://github.com/sigp/anchor/issues/17.

## Proposed Changes

Adds and configures Gossipsub Behavior according to the spec https://github.com/ssvlabs/ssv-spec/blob/main/p2p/CONFIG.md#pubsub-parameters.

## Additional Info

- `Outbound Queue Size`, `Validation Queue Size` and `Validation Throttle` defined in https://github.com/ssvlabs/ssv-spec/blob/main/p2p/CONFIG.md#pubsub-parameters don't have a clear equivalent to public Gossipsub parameters. They might be related to internal parameters and are ignored for now.
- `Topic Message Validator` and `Subscription Filter` are ignored for now.


anchor/network/src/network.rs
```diff
@@ -74,8 +79,18 @@ impl Network {
...
-                _swarm_message = self.swarm.select_next_some() => {
-                    // TODO handle and match swarm messages
+                swarm_message = self.swarm.select_next_some() => {
+                    match swarm_message {
+                        SwarmEvent::Behaviour(behaviour_event) => match behaviour_event {
+                            AnchorBehaviourEvent::Gossipsub(_ge) => {
+                                // TODO handle gossipsub events
+                            },
+                            // TODO handle other behaviour events
+                            _ => todo!()
```
jxs: nit:
```suggestion
                            _ => {}
```
otherwise this may panic as soon as it runs, as there will be other events. Same applies for the `todo!()` below for the other swarm events

anchor/network/src/network.rs
```diff
@@ -94,10 +108,40 @@ fn build_anchor_behaviour(local_keypair: Keypair) -> AnchorBehaviour {
...
+        .message_id_fn(gossip_message_id)
+        .flood_publish(false)
+        .validation_mode(ValidationMode::Strict)
+        .mesh_n(8) //D
+        .mesh_n_low(6) // Dlo
+        .mesh_n_high(12) // Dhi
+        .mesh_outbound_min(4) // Dout
+        .heartbeat_interval(Duration::from_millis(700))
+        .history_length(6)
+        .history_gossip(4)
```
jxs, line 136: should we already introduce the `NetworkLoad` `lighthouse` [uses](https://github.com/sigp/lighthouse/blob/0d90135047519f4c2ee586d50e560f7bb2ff9b10/beacon_node/lighthouse_network/src/config.rs#L366-L375)?

anchor/network/src/network.rs
```diff
@@ -94,10 +108,40 @@ fn build_anchor_behaviour(local_keypair: Keypair) -> AnchorBehaviour {
         identify::Behaviour::new(identify_config)
     };
 
+    // TODO those values might need to be parameterized based on the network
+    let slots_per_epoch = 32;
+    let seconds_per_slot = 12;
+    let duplicate_cache_time = Duration::from_secs(slots_per_epoch * seconds_per_slot); // 6.4 min
+
+    let gossip_message_id = move |message: &gossipsub::Message| {
+        gossipsub::MessageId::from(&Sha256::digest(&message.data)[..20])
```
jxs, line 121: we should also have the [network related](https://github.com/sigp/lighthouse/blob/0d90135047519f4c2ee586d50e560f7bb2ff9b10/beacon_node/lighthouse_network/src/config.rs#L444-L467) prefixes right?

jxs (commented):
Thanks for starting this Diego, left a couple comments. CC @AgeManning who probably knows more about the questions raised

anchor/network/src/network.rs
diegomrsantos: I believe the spec doesn't mention it https://github.com/ssvlabs/ssv-spec/blob/main/p2p/CONFIG.md#msg-id

anchor/network/src/network.rs
diegomrsantos: Good catch. I had this problem when adding discv5 and I used a log. I applied it here.

anchor/network/src/network.rs
AgeManning: The network load isn't necessary. It was a nice thing to have to tweak lighthouse (and go off-specification). We probably don't need it in Anchor, unless we are testing network parameters

jking-aus (approved):
lgtm

PR #71, opened 2024-12-12
Enforce alphabetically ordered cargo deps

dknopik:
Shamelessly stolen from @macladson's PR here: https://github.com/sigp/lighthouse/pull/6678


.github/workflows/test-suite.yml
```diff
@@ -192,3 +192,18 @@ jobs:
 #          cache-target: release
 #      - name: Run Makefile to trigger the bash script
 #        run: make cli
+  cargo-sort:
```
diegomrsantos, line 195: What do you think about moving this job and the `check-code` to a new workflow as they are more about code quality than testing?

dknopik: Don't really have an opinion on that. @magick93, what do you think?

magick93: IMO splitting it wont add too much value, though I'm happy to do it. 

The best value - in the case of test and code qual. comes from giving rapid feedback to the dev. It its not too computationally intensive maybe a pre-commit hook may be better.

jking-aus (approved):
lgtm

.github/workflows/test-suite.yml
jking-aus: good comment diego, let's merge for now but as workflows become more refined we can revisit

Issue #35, opened 2024-10-23
Discovery


AgeManning:
Add `discv5` to Anchor. 

This allows us to discover peers and dial them via libp2p. 

We should store ENRs in our database for bootstrapping the node when we restart.

diegomrsantos:
> We should store ENRs in our database for bootstrapping the node when we restart.

Is it the Peer Manager's responsibility?

PR #72, opened 2024-12-13
Add discv5

diegomrsantos:
## Issue Addressed

Add discv5 and partially implement https://github.com/sigp/anchor/issues/35.

## Proposed Changes

- Updated the discv5 dependency from version 0.8.0 to 0.9.0.
- Added a new CLI argument boot_nodes_enr for specifying base64-encoded ENRs to bootstrap the p2p network.
- Enhanced the configuration handling to include the new boot_nodes_enr argument and parse it correctly.
- Introduced the Discovery module to handle peer discovery using the discv5 protocol.
- Modified the network behavior to integrate the Discovery module, enabling peer discovery and connection management.
- Added a configuration option disable_discovery to control the activation of the discovery protocol.

## Additional Info

- This is mostly copying the discovery code from LH for now while we figure out the best way to modularise and reuse this code.
- This is the first step toward a full discovery implementation. More features must be added, like saving ENRs to the database.


anchor/network/src/discovery.rs
```diff
@@ -0,0 +1,460 @@
...
+        // Start the discv5 service and obtain an event stream
+        let event_stream = if !network_config.disable_discovery {
+            discv5.start().map_err(|e| e.to_string()).await?;
+            // TODO debug!(log, "Discovery service started");
+            EventStream::Awaiting(Box::pin(discv5.event_stream()))
+        } else {
+            EventStream::InActive
+        };
+
+        if !network_config.boot_nodes_multiaddr.is_empty() {
```
diegomrsantos, line 151: Bootnodes MA might not be necessary. See https://github.com/ssvlabs/ssv-spec/blob/main/p2p/SPEC.md#discovery

anchor/network/src/network.rs
```diff
@@ -85,12 +89,21 @@ impl Network {
...
+                                //self.peer_manager_mut().peers_discovered(peers);
+                                log::debug!("Discovered peers: {:?}", peers);
+                            }
                             // TODO handle other behaviour events
                             _ => {
                                 log::debug!("Unhandled behaviour event: {:?}", behaviour_event);
                             }
                         },
                         // TODO handle other swarm events
+                        SwarmEvent::NewListenAddr { .. } => {},
```
jxs: this isn't required for now right?

anchor/network/src/discovery.rs
diegomrsantos: Or maybe we should accept a UDP MA for discv5

anchor/network/src/network.rs
```diff
@@ -140,12 +156,24 @@ fn build_anchor_behaviour(local_keypair: Keypair) -> AnchorBehaviour {
...
     let gossipsub =
-        gossipsub::Behaviour::new(MessageAuthenticity::Signed(local_keypair), config).unwrap();
+        gossipsub::Behaviour::new(MessageAuthenticity::Signed(local_keypair.clone()), config)
+            .unwrap();
+
+    let discovery = {
+        // Build and start the discovery sub-behaviour
+        let mut discovery = Discovery::new(local_keypair.clone(), network_config)
+            .await
+            .unwrap();
```
jxs, line 175: nit, either:
```suggestion
            .expect("Discovery should be able to start");
```
or `?` to bubble up the error and treat it in the function caller

diegomrsantos:
I removed the hardcoded bootnode ENR. It's possible to use the holesky bootnode running: `run --package anchor --bin anchor -- --boot-nodes-enr=enr:-Li4QFIQzamdvTxGJhvcXG_DFmCeyggSffDnllY5DiU47pd_K_1MRnSaJimWtfKJ-MD46jUX9TwgW5Jqe0t4pH41RYWGAYuFnlyth2F0dG5ldHOIAAAAAAAAAACEZXRoMpD1pf1CAAAAAP__________gmlkgnY0gmlwhCLdu_SJc2VjcDI1NmsxoQN4v-N9zFYwEqzGPBBX37q24QPFvAVUtokIo1fblIsmTIN0Y3CCE4uDdWRwgg-j`

anchor/client/src/cli.rs
```diff
@@ -329,6 +329,14 @@ pub struct Anchor {
         help_heading = FLAG_HEADER
     )]
     help: Option<bool>,
+    #[clap(
+        long,
+        global = true,
+        value_delimiter = ',',
+        help = "One or more comma-delimited base64-encoded ENR's to bootstrap the p2p network",
+        display_order = 0
```
jxs, line 337: We can declare the default boot nodes somewhere (like `anchor/network/src/config.rs`) with:
```rust
const DEFAULT_BOOTNODES: [&'static str; 3] = ["ENR1", "ENR2", "ENR3"];
```
and then import it and use it here: 
```suggestion
        display_order = 0
        default_values_t = DEFAULT_BOOTNODES.into_iter().map(Into::into)
```

anchor/network/src/discovery.rs
```diff
@@ -0,0 +1,460 @@
...
+                //     log,
+                //     "Could not add peer to the local routing table";
+                //     "addr" => repr,
+                //     "error" => e.to_string(),
+                // )
+            });
+        }
+
+        // Start the discv5 service and obtain an event stream
+        let event_stream = if !network_config.disable_discovery {
```
jxs, line 143: will we want to disable discovery? cc @age @jking-aus

anchor/network/src/discovery.rs
```diff
@@ -0,0 +1,460 @@
...
+    pub async fn new(local_keypair: Keypair, network_config: &Config) -> Result<Self, String> {
+        let _enr_dir = match network_config.network_dir.to_str() {
+            Some(path) => String::from(path),
+            None => String::from(""),
+        };
+
+        // TODO info!(log, "ENR Initialised"; "enr" => local_enr.to_base64(), "seq" => local_enr.seq(), "id"=> %local_enr.node_id(),
+        //       "ip4" => ?local_enr.ip4(), "udp4"=> ?local_enr.udp4(), "tcp4" => ?local_enr.tcp4(), "tcp6" => ?local_enr.tcp6(), "udp6" => ?local_enr.udp6(),
+        //       "quic4" => ?local_enr.quic4(), "quic6" => ?local_enr.quic6()
+        // );
```
jxs: this can be removed right?

anchor/client/src/config.rs
```diff
@@ -131,6 +131,25 @@ pub fn from_cli(cli_args: &Anchor) -> Result<Config, String> {
...
+                // if !multi.iter().any(|proto| matches!(proto, Protocol::Udp(_))) {
+                //     slog::error!(log, "Missing UDP in Multiaddr {}", multi.to_string());
+                // }
+                // if !multi.iter().any(|proto| matches!(proto, Protocol::P2p(_))) {
+                //     slog::error!(log, "Missing P2P in Multiaddr {}", multi.to_string());
+                // }
+                // multiaddrs.push(multi);
+            }
+        }
+    }
```
jxs, line 151: we are currently iterating the bootnode list twice, first here and then when adding to `discv5`, we can parse them when adding to `discv5` and avoid double iteration

anchor/network/src/discovery.rs
```diff
@@ -0,0 +1,460 @@
...
+        let enr = build_enr(&enr_key, network_config).unwrap();
+        let mut discv5 = Discv5::new(enr, enr_key, discv5_config)
+            .map_err(|e| format!("Discv5 service failed. Error: {:?}", e))?;
+
+        // Add bootnodes to routing table
+        for bootnode_enr in network_config.boot_nodes_enr.clone() {
+            // TODO if bootnode_enr.node_id() == local_node_id {
+            //     // If we are a boot node, ignore adding it to the routing table
+            //     continue;
+            // }
```
jxs, line 121: is anchor planned to also be run as bootnode? If so this can be uncommented right?

anchor/network/src/discovery.rs
```diff
@@ -0,0 +1,460 @@
...
+            // TODO debug!(
+            //     log,
+            //     "Adding node to routing table";
+            //     "node_id" => %bootnode_enr.node_id(),
+            //     "peer_id" => %bootnode_enr.peer_id(),
+            //     "ip" => ?bootnode_enr.ip4(),
+            //     "udp" => ?bootnode_enr.udp4(),
+            //     "tcp" => ?bootnode_enr.tcp4(),
+            //     "quic" => ?bootnode_enr.quic4()
+            // );
```
jxs: this can be uncommented right?

anchor/network/src/discovery.rs
```diff
@@ -0,0 +1,460 @@
...
+
+            //let repr = bootnode_enr.to_string();
+            let _ = discv5.add_enr(bootnode_enr).map_err(|_e| {
+                // TODO  error!(
+                //     log,
+                //     "Could not add peer to the local routing table";
+                //     "addr" => repr,
+                //     "error" => e.to_string(),
+                // )
+            });
```
jxs: we are currently not handling the error in case of error adding the enr to `discv5` I suggest we at least log it, ie
```rust
if let Err(err) = discv5.add_enr(bootnode_enr) {
    error!("Could not add peer to the local routing table";  "addr" => repr, "error" => e.to_string());
}
```

anchor/network/src/discovery.rs
```diff
@@ -0,0 +1,460 @@
...
+            let result = self.process_completed_queries(query_result);
+            if result.is_some() {
+                return result;
+            }
+        }
+        None
+    }
+}
+
+impl NetworkBehaviour for Discovery {
```
jxs, line 333: this can probably be moved into upstream `discv5` feature gated by `libp2p` (like the `enr` crate does) as its shared by both `lighthouse` and `anchor` wdyt @AgeManning

anchor/network/src/discovery.rs
```diff
@@ -0,0 +1,470 @@
...
+        // If we are listening on ipv6, add the quic ipv6 port.
+        if let Some(quic6_port) = config.enr_quic6_port.or_else(|| {
+            config
+                .listen_addresses
+                .v6()
+                .and_then(|v6_addr| v6_addr.quic_port.try_into().ok())
+        }) {
+            builder.add_value(QUIC6_ENR_KEY, &quic6_port.get());
+        }
+    }
```
jxs, line 442: the spec only mentions TCP as a transport, and the `go` `ssv` client doesn't seem to support it, do we plan to support it nonetheless?

jxs (commented):
Thanks for starting this Diego! Left some comments, mainly I think we should not merge commented instructions without proper explaining on the why

anchor/network/src/network.rs
diegomrsantos: Good catch, removed.

anchor/client/src/cli.rs
diegomrsantos: It seems LH defines them in `yaml` files https://github.com/sigp/lighthouse/blob/stable/common/eth2_network_config/built_in_network_configs/holesky/boot_enr.yaml

anchor/client/src/config.rs
diegomrsantos: This is adding the ENRs from the cli into the network config. LH does the same https://github.com/sigp/lighthouse/blob/stable/beacon_node/src/config.rs#L1159

anchor/network/src/discovery.rs
diegomrsantos: Waiting for input here as I don't know the answer

anchor/network/src/discovery.rs
diegomrsantos: Done

anchor/network/src/discovery.rs
diegomrsantos: done

anchor/network/src/discovery.rs
diegomrsantos: We could support `QUIC` for communication between rust nodes.

anchor/network/src/network.rs
diegomrsantos: we'll need to refactor the function and return a `Result` instead.

diegomrsantos: Could we create an issue and address it in a new PR?

anchor/network/src/discovery.rs
diegomrsantos: We probably need a `local_enr`, but I don't understand how it works yet.

anchor/network/src/discovery.rs
AgeManning: hmm not sure. It is mainly used in testing, where you can set the bootnodes and only connect to that node. It is kinda handy. Maybe worth keeping

anchor/network/src/discovery.rs
AgeManning: Yeah, thats a good idea

AgeManning: oh except the types might be annoying. The `NetworkBehaviour` and libp2p traits, will come from the upstream libp2p dependency and then this won't work if Anchor sets its own libp2p version. 

i.e We will have to always be doing releases for discv5 to keep up to libp2p, which could get annoying

anchor/network/src/discovery.rs
AgeManning: Yeah, lets support it, and hopefully they follow suit

anchor/client/src/cli.rs
jxs: yeah can also be, but then we should get them by default, if you prefer it can be done in a subsequent PR

anchor/network/src/discovery.rs
jxs: the `NetworkBehaviour` traits (which they all depend) are contained in `libp2p-swarm` which updates roughly once a year, the last update  to [`0.45`](https://crates.io/crates/libp2p-swarm/versions) was due to the `Transport` trait update. There isn't an update like that in sight wrt that I think we are ok

anchor/network/src/discovery.rs
jxs: ok nice! Then we should probably create a PR upstream to manifest that

anchor/network/src/network.rs
jxs: yeah ofc :)

anchor/client/src/config.rs
jxs: LH doing something shouldn't be used as the only source of truth to follow for `anchor`. 
In this case specifically `lighthouse` is not using `clap` derive yet (it's an ongoing process, see [1](https://github.com/sigp/lighthouse/pull/6300), [2](https://github.com/sigp/lighthouse/pull/6493)).
Anchor is using clap derive from the beginning so we don't need to iterate the cli arguments as a string separated by commas, we already have a `Vec<String>`.
What I meant in the previous comment was that we can still have `config.network.boot_nodes_enr` be `Vec<String>`, and just call
```rust
config.network.boot_nodes_enr.append(cli_args.boot_nodes_enr)
```
and then on https://github.com/sigp/anchor/pull/72/files#diff-377ff7ebcbab6b7243aedbcf1c9139b355dd5294398ca8b0b218c32ccf2f1387R121 we parse them, so that we we only iterate the enr bootnodes list once

anchor/network/src/discovery.rs
diegomrsantos: Added a TODO

anchor/client/src/cli.rs
diegomrsantos: Yes, it's probably better in another PR.

anchor/client/src/config.rs
diegomrsantos: Thanks for clarifying, I understand now. In this case, the vector will only contain a few values. Would this optimization have a significant impact on performance?

anchor/network/src/network.rs
diegomrsantos: https://github.com/sigp/anchor/issues/75

anchor/client/src/config.rs
jxs: no worries Diego. probably it won't have a big impact on the performance but it isn't only about performance, it also makes the code more elegant at no cost

diegomrsantos: There's also the validation perspective. It's probably a good idea to parse and check if the ENR is valid here and add only valid ones to the network config.

anchor/network/src/discovery.rs
jking-aus: I think so yep -- would be good to be able to provide anchor based bootnodes so not reliant on the go client for entrypoints

anchor/network/src/network.rs
```diff
@@ -85,6 +91,20 @@ impl Network {
...
+                            //
+                            // The peer manager will subsequently decide which peers need to be dialed and then dial
+                            // them.
+                            AnchorBehaviourEvent::Discovery(DiscoveredPeers { peers }) => {
+                                //self.peer_manager_mut().peers_discovered(peers);
+                                log::debug!("Discovered peers: {:?}", peers);
+                                for (enr, _) in peers {
+                                    for tcp in enr.multiaddr_tcp() {
+                                        log::debug!("Dialing peer: {:?}", tcp);
+                                        self.swarm.dial(tcp).unwrap();
```
jking-aus: I realise this is first draft but can we do something like the below to dodge the unwrap please: 


```suggestion
if let Err(e) = self.swarm.dial(tcp.clone()) {
 log::error!("Error dialing peer {}: {}", tcp,  e);
 }                                       
```

anchor/network/src/network.rs
```diff
@@ -85,6 +91,20 @@ impl Network {
...
+                            // Inform the peer manager about discovered peers.
+                            //
+                            // The peer manager will subsequently decide which peers need to be dialed and then dial
+                            // them.
+                            AnchorBehaviourEvent::Discovery(DiscoveredPeers { peers }) => {
+                                //self.peer_manager_mut().peers_discovered(peers);
+                                log::debug!("Discovered peers: {:?}", peers);
+                                for (enr, _) in peers {
+                                    for tcp in enr.multiaddr_tcp() {
+                                        log::debug!("Dialing peer: {:?}", tcp);
```
jking-aus:
```suggestion
                                        log::trace!("Dialing peer: {:?}", tcp);
```

jking-aus (changes requested):
one small change then let's merge

jking-aus (approved):

Issue #27, opened 2024-10-23
Validator Duties Service


AgeManning:
Anchor needs to know what and when to sign particular messages. 

The validator client has this service and we should be able to utilize this as a dependency as best we can. It is located inside the validator client, its not linked here, as it currently undergoing a change of location

PR #76, opened 2024-12-18
create signature collector and make processor stateless

dknopik:
## Issue Addressed

- Implemented as part of #27 
- Improves upon #41 

## Proposed Changes

- Do not manage any longer running stuff in `processor`, as that provokes dependency cycles and weird locations for types.
- Create the signature collector, which... collects signatures. When all signature parts are received, we interpolate it into a full signature to be used for validator services. Currently, the actual cryptography of interpolating is mocked out. 

## Additional Info

The signature collector is the first component using the processor, which is why the small processor refactor is bundled with the signature collector in this PR.


dknopik:
FYI: if this feels familiar to you, the changes are also included in #77, as this PR (more specifically the processor changes) is a prerequisite for #77. I'll rebase #77 once this is merged.

jking-aus (approved):
lgtm -- great work Daniel

PR #77, opened 2024-12-20
Add QBFT manager

dknopik:
Refactor the current `qbft` crate to be sync and general purpose and create `qbft_manager` which integrates into the client architecture such as the processor. This module is going to be used by the validator store (to trigger QBFT instances) and the network (to pass on network messages related to QBFT). 

There is still quite a bit to do to clean this up, but this initial version should be able to get us going. 

Depends on #76.


anchor/common/qbft/src/types.rs
```diff
@@ -82,31 +86,17 @@ pub enum InstanceState {
...
-    Commit(OperatorId, ConsensusData<D>),
+    Commit(OperatorId, ConsensusData<D::Hash>),
     /// Round change message received from network
-    RoundChange(OperatorId, Round, Option<ConsensusData<D>>),
+    RoundChange(OperatorId, Round, Option<ConsensusData<D::Hash>>),
 }
 
-/// Messages that may be sent to the message_out channel from the instance to the client processor
-#[derive(Debug, Clone)]
-pub enum OutMessage<D: Debug + Clone + Eq + Hash> {
```
jking-aus, line 98: We had separate enums for in and out message with the intention of appending operatorId to all outbound messages. Also allowed us to send a completed message to the processor channel so we could queue the qbft instance to die. I'm OK to collapse these as I can see you're handling this now in the qbft_manager - nice.

anchor/common/qbft/src/types.rs
```diff
@@ -26,20 +28,29 @@ impl LeaderFunction for DefaultLeaderFunction {
...
         round: Round,
         instance_height: InstanceHeight,
-        committee_size: usize,
+        committee: &[OperatorId],
     ) -> bool {
-        *operator_id == ((*round + *instance_height) % committee_size).into()
+        *operator_id
+            == *committee
+                .get(((round.get() - Round::default().get()) + *instance_height) % committee.len())
+                .expect("slice bounds kept by modulo length")
```
jking-aus, line 39: could this panic if committee.len = 0? assume there is a sanity check somewhere to prevent this

anchor/common/qbft/src/config.rs
```diff
@@ -13,8 +12,7 @@ where
     pub instance_height: InstanceHeight,
     pub round: Round,
     pub pr: usize,
-    pub committee_size: usize,
-    pub committee_members: HashSet<OperatorId>,
+    pub committee_members: Vec<OperatorId>,
```
jking-aus: Why Vec instead of Hashset?

jking-aus (commented):
This is excellent work mate -- tonnes of great additions, in particular I like the idea of splitting out the instance manager from the consensus mechanism. Probably two or three PRs of work here tbh

anchor/common/qbft/src/config.rs
dknopik: Because order matters for leader selection. I considered using `BTreeSet` instead, as it is ordered as well, but decided to use `Vec` as in my opinion the simplicity of a `Vec` is a greater benefit than the guaranteed uniqueness of elements in a set.

anchor/common/qbft/src/types.rs
dknopik: Actually - yes, modulo/remainder panics if the `len` is 0. And no, I don't think this is currently checked :sweat_smile:  - I will extend the validation done by the config builder.

anchor/common/qbft/src/types.rs
dknopik: Yeah, I collapsed them because I found it to be simpler - especially as the manager should not really care about the *contents* of a message. Just that something needs to be broadcasted.

Separating `completed` out also allows us to fully separate the concerns: network messages vs consensus result.

anchor/common/qbft/src/config.rs
diegomrsantos: Are you referring to simplicity from an API perspective? If duplications are not allowed, using a data structure that enforces this constraint is beneficial. This approach inherently communicates the restriction without the need for additional checks.

dknopik: I meant implementation simplicity, not API simplicity.

Order matters in a sense that the order of the committee members is (as far as I know) not necessarily the order of their sorted `OperatorId`s. E.g. a committee might be ordered: (4, 2, 1, 3). 

A `BTreeSet` would therefore require some wrapper like this:

```rust
struct QbftOperator {
    index: usize,
    id: OperatorId,
}

impl PartialOrd for QbftOperator {
    // omitted, order by index
}
```

Actually, while writing this I realize that this would not even work for asserting uniqueness of OperatorId, as we now only make sure that the index is unique. We could use a `HashMap<OperatorId, usize>`, mapping from operator id to its index in the committee, but this is (in my opinion) ugly and not really "simple" API.

In principle, I agree that containers enforcing constraints improve APIs - but in this case, I believe the tradeoffs warrant a `Vec`. :)

anchor/common/qbft/src/config.rs
```diff
@@ -0,0 +1,178 @@
+use super::error::ConfigBuilderError;
```
diegomrsantos, line 1: Has this file been moved from `qbft/src/`? If so, it seems Git has not recognized this as a rename/move operation, likely because the file was substantially modified at the same time. To improve clarity and ensure accurate tracking of file movements, it might be beneficial to move files and make modifications in separate PRs in the future. What do you think about this approach?

dknopik: Hmm, I intentionally first moved those files in commit [6bc9471](https://github.com/sigp/anchor/pull/77/commits/6bc94719ebb53c6cd30e95191feb007b23d730bc) for this exact reason. I guess Github renders the full view off all changes (i.e. how they would look like after a squash). So yeah, we probably should do that move in it's own PR. Good catch, will do!

anchor/common/qbft/src/config.rs
diegomrsantos: I thought it'd be more straightforward like in Java that has a DS like that in the std https://docs.oracle.com/javase/8/docs/api/java/util/LinkedHashSet.html. It seems Rust doesn't have an equivalent one, but I found this crate https://docs.rs/indexmap/latest/indexmap/set/struct.IndexSet.html.

anchor/common/qbft/src/config.rs
dknopik: Will rebase this PR once #76 is merged, then it should be fine

anchor/common/qbft/src/types.rs
```diff
@@ -50,19 +61,12 @@ impl Round {
 
 /// The operator that is participating in the consensus instance.
 #[derive(Clone, Copy, Debug, Default, Eq, PartialEq, Hash, From, Deref)]
-pub struct OperatorId(usize);
+pub struct OperatorId(pub usize);
```
Zacholme7: Does it have to be usize here? Or can you just use the `OpeartorId` from the `ssv_types` crate

dknopik: good point - this predated `ssv_types`, didn't think of migrating it, will do

dknopik: actually, I am not 100% sure if we should do that. It would be kind of nice to have `qbft` as standalone and reusable as possible. what is everyone's opinion on that?

diegomrsantos: Are there more overlapping types?

anchor/common/ssv_types/src/message.rs
```diff
@@ -0,0 +1,190 @@
...
+    AggregateAndProof, BeaconBlock, BlindedBeaconBlock, Checkpoint, CommitteeIndex, EthSpec,
+    Hash256, PublicKeyBytes, Signature, Slot, SyncCommitteeContribution, VariableList,
+};
+// todo - dear reader, this mainly serves as plain translation of the types found in the go code
+// there are a lot of byte[] there, and that got confusing, below should be more readable.
+// it needs some work to actually serialize to the same stuff on wire, and I feel like we can name
+// the fields better
+
+#[derive(Clone, Debug)]
+pub struct SignedSsvMessage<E: EthSpec> {
```
diegomrsantos, line 16: I see there's some overlapping with https://github.com/sigp/anchor/pull/80/files#diff-46abd9dbbf9cc4366e3258211d1bf4ff7fce2df918e1e870e607150429fc97a2

anchor/common/ssv_types/src/message.rs
```diff
@@ -0,0 +1,190 @@
...
+    pub signatures: Vec<[u8; 256]>,
+    pub operator_ids: Vec<OperatorId>,
+    pub ssv_message: SsvMessage<E>,
+    pub full_data: Option<FullData<E>>,
+}
+
+#[derive(Clone, Debug)]
+pub struct SsvMessage<E: EthSpec> {
```
diegomrsantos, line 24: Why do we need the type parameter and the `EthSpec` bound?

anchor/common/ssv_types/src/message.rs
```diff
@@ -0,0 +1,190 @@
...
+// there are a lot of byte[] there, and that got confusing, below should be more readable.
+// it needs some work to actually serialize to the same stuff on wire, and I feel like we can name
+// the fields better
+
+#[derive(Clone, Debug)]
+pub struct SignedSsvMessage<E: EthSpec> {
+    pub signatures: Vec<[u8; 256]>,
+    pub operator_ids: Vec<OperatorId>,
+    pub ssv_message: SsvMessage<E>,
+    pub full_data: Option<FullData<E>>,
```
diegomrsantos, line 20: is an `Option` required here?

anchor/common/ssv_types/src/message.rs
dknopik: Yeah, this was just a rough draft for my understanding as the comment says - we should probably switch to yours :)

anchor/common/ssv_types/src/message.rs
dknopik: Because of the `BeaconBlock`, which I embedded directly for clarity instead of using a `byte[]` like in go SSV.

anchor/common/ssv_types/src/message.rs
dknopik: Yes, as the full data is not always set. E.g. only on PROPOSE qbft messages, not on subsequent messages such as PREPARE or COMMIT

anchor/common/qbft/src/types.rs
dknopik: In the `qbft` crate? I don't think so

diegomrsantos: So, I'm not sure it's worth it.

anchor/common/ssv_types/src/message.rs
diegomrsantos: There are better things in your code, like using an Enum for the raw data. We could keep both for now, but think about a way of merging them and where it should be in the code. Should we use the same type for the data parsed from the network and the one used in application logic?

anchor/common/qbft/src/config.rs
dknopik: done :)

anchor/common/ssv_types/src/message.rs
dknopik: We should probably use them in application code as well. In some cases, we rely on them being the same as networking messages - for example, PREPARE and COMMIT messages refer to the full data by hash, so we need the messages to be identical to the format used by go-ssv so that the hashes agree.

I propose we both merge our versions and then align them in an own PR. Intuitively, I would put them in `ssv_types`, but we can figure that out then :)

diegomrsantos: What about the `full_data` is passed as raw bytes to the application and the decoding is done there?

dknopik: In my opinion, the network stack should verify that `full_data` is actually a valid object, so that illegal messages with junk data are not further gossiped. But I am not sure if that is actually done by go-ssv...

dknopik: But yeah, I think we will (unfortunately) need some kind of type separation - we can only decode as soon as we know the data version (i.e. fork) that is specified in the message itself

diegomrsantos: It's possible to create a valid object with junk data anyway. I believe the more comprehensive validation in Gossipsub is done by the application.

diegomrsantos: This is possibly related https://docs.rs/libp2p-gossipsub/latest/libp2p_gossipsub/struct.Behaviour.html#method.report_message_validation_result

anchor/common/qbft/src/config.rs
```diff
@@ -61,83 +50,130 @@ impl<F: Clone + LeaderFunction> Config<F> {
...
-                committee_members: HashSet::new(),
-                quorum_size: 4,
-                round: Round::default(),
-                pr: 0,
-                round_time: Duration::new(2, 0),
-                max_rounds: 4,
-                leader_fn: DefaultLeaderFunction {},
-            },
+            operator_id: None,
+            instance_height: None,
```
diegomrsantos: Would it be better to keep the calls to the type default function?

diegomrsantos: Thinking more about it, is there a valid use case to create a configuration with those values as None?

diegomrsantos: Ok, this is handled in the build. Perhaps we could follow this pattern? https://blogs.oracle.com/javamagazine/post/exploring-joshua-blochs-builder-design-pattern-in-java

anchor/common/qbft/src/config.rs
```diff
@@ -12,9 +12,7 @@ where
     pub operator_id: OperatorId,
     pub instance_height: InstanceHeight,
     pub round: Round,
-    pub pr: usize,
-    pub committee_size: usize,
-    pub committee_members: HashSet<OperatorId>,
+    pub committee_members: IndexSet<OperatorId>,
```
diegomrsantos: Should those fields be private?

anchor/common/qbft/src/config.rs
```diff
@@ -27,21 +25,12 @@ impl<F: Clone + LeaderFunction> Config<F> {
...
-    }
-
-    pub fn commmittee_members(&self) -> HashSet<OperatorId> {
-        self.committee_members.clone()
-    }
 
-    /// The quorum size required for the committee to reach consensus
-    pub fn quorum_size(&self) -> usize {
-        self.quorum_size
+    pub fn commmittee_members(&self) -> &IndexSet<OperatorId> {
```
diegomrsantos: We could add this to the beginning of this `impl`

```rust
    /// Private constructor so it can only be built by our `ConfigBuilder`.
    fn from_builder(builder: &ConfigBuilder<F>) -> Self {
        Self {
            operator_id: builder.operator_id,
            instance_height: builder.instance_height,
            committee_members: builder.committee_members.clone(),
            round: builder.round,
            round_time: builder.round_time,
            max_rounds: builder.max_rounds,
            quorum_size: builder.quorum_size,
            leader_fn: builder.leader_fn.clone(),
        }
    }
```

anchor/common/qbft/src/config.rs
diegomrsantos: and here something like this
```rust
/// A "Joshua Bloch–style" Builder for `Config`.
#[derive(Clone, Debug)]
pub struct ConfigBuilder<F>
where
    F: LeaderFunction + Clone,
{
    // Mandatory fields
    operator_id: OperatorId,
    instance_height: InstanceHeight,
    committee_members: IndexSet<OperatorId>,

    // Optional fields with defaults set in the constructor
    round: Round,
    round_time: Duration,
    max_rounds: usize,
    quorum_size: usize,
    leader_fn: F,
}

impl<F> ConfigBuilder<F>
where
    F: LeaderFunction + Clone,
{
    pub fn new(
        operator_id: OperatorId,
        instance_height: InstanceHeight,
        committee_members: IndexSet<OperatorId>,
    ) -> Self {
        let committee_size = committee_members.len();
        let f = get_f(committee_size);
        let default_quorum = committee_size.saturating_sub(f);

        ConfigBuilder {
            operator_id,
            instance_height,
            committee_members,
            round: Round::default(),
            round_time: Duration::new(2, 0),
            max_rounds: 4,
            quorum_size: default_quorum,
            leader_fn: DefaultLeaderFunction {},
        }
    }
}

impl<F> ConfigBuilder<F>
where
    F: LeaderFunction + Clone,
{
    // Chained setter methods for optional fields to override defaults
    pub fn round(mut self, round: Round) -> Self {
        self.round = round;
        self
    }

    pub fn round_time(mut self, round_time: Duration) -> Self {
        self.round_time = round_time;
        self
    }

    pub fn max_rounds(mut self, max_rounds: usize) -> Self {
        self.max_rounds = max_rounds;
        self
    }

    pub fn quorum_size(mut self, quorum_size: usize) -> Self {
        self.quorum_size = quorum_size;
        self
    }

    pub fn leader_fn(mut self, leader_fn: F) -> Self {
        self.leader_fn = leader_fn;
        self
    }

    pub fn build(self) -> Result<Config<F>, ConfigBuilderError> {
        // Validate mandatory fields
        let committee_size = self.committee_members.len();
        if committee_size == 0 {
            return Err(ConfigBuilderError::NoParticipants);
        }
        if !self.committee_members.contains(&self.operator_id) {
            return Err(ConfigBuilderError::MissingOperatorIdInCommittee);
        }

        // Validate `quorum_size`
        let f = get_f(committee_size);
        if self.quorum_size < f * 2 + 1 || self.quorum_size > committee_size - f {
            return Err(ConfigBuilderError::InvalidQuorumSize);
        }

        // Validate `max_rounds`
        if self.max_rounds == 0 {
            return Err(ConfigBuilderError::ZeroMaxRounds);
        }

        // Validate `round`
        if self.round.get() > self.max_rounds {
            return Err(ConfigBuilderError::ExceedingStartingRound);
        }

        // If everything is okay, build and return a `Config`.
        Ok(Config::from_builder(&self))
    }
}
```

anchor/common/qbft/src/lib.rs
```diff
@@ -20,156 +20,98 @@ mod validation;
...
             round_change_messages: HashMap::with_capacity(estimated_map_size),
-            message_out,
-            message_in,
+            send_message,
             state: InstanceState::AwaitingProposal,
+            completed: None,
         };
-
-        (in_sender, out_receiver, instance)
+        qbft.start_round();
```
diegomrsantos, line 95: I noticed that the constructor is doing more than just initializing data structures - start_round() is changing state and sending messages. It might be cleaner and easier to work with if construction and execution were kept separate. This way, the caller can control when the QBFT process starts, and it could also make testing simpler. What do you think?

anchor/common/qbft/src/lib.rs
```diff
@@ -20,156 +20,98 @@ mod validation;
 #[cfg(test)]
 mod tests;
 
-type RoundChangeMap<D> = HashMap<OperatorId, Option<ConsensusData<ValidatedData<D>>>>;
+type RoundChangeMap<D> = HashMap<OperatorId, Option<ConsensusData<D>>>;
+
+pub trait Data: Debug + Clone {
```
diegomrsantos, line 25:
```rust
pub trait Hashable: Debug + Clone {
    type Digest: Debug + Clone + Eq + Hash;
    
    fn digest(&self) -> Self::Digest;
}
```

What do you think about it? The motivation is to make the trait name more descriptive and avoid using `Hash` twice - it seems like a recursive definition.

anchor/common/qbft/src/lib.rs
```diff
@@ -20,156 +20,98 @@ mod validation;
...
-                "Receiver channel closed. Terminating"
-            );
-            self.state = InstanceState::Complete
-        }
-    }
-
     /// Once we have achieved consensus on a PREPARE round, we add the data to mapping to match
     /// against later.
-    fn insert_consensus(&mut self, round: Round, data: ValidatedData<D>) {
+    fn insert_consensus(&mut self, round: Round, data: D::Hash) {
```
diegomrsantos: Should it be `hash` instead of `data`?

anchor/common/qbft/src/lib.rs
```diff
@@ -246,24 +188,45 @@ where
             // We are the leader
             debug!("Current leader");
             // Check justification of round change quorum
-            if let Some(validated_data) = self.justify_round_change_quorum().cloned() {
+            let hash = if let Some(validated_data) = self.justify_round_change_quorum().cloned() {
```
diegomrsantos:
```suggestion
            let hash = if let Some(validated_data) = self.justify_round_change_quorum() {
```

anchor/common/qbft/src/lib.rs
```diff
@@ -246,24 +188,45 @@ where
...
                     old_data = ?validated_data,
                     "Using consensus data from a previous round");
-                self.send_proposal(validated_data.clone());
-                self.send_prepare(validated_data);
+                validated_data
             } else {
                 debug!("Using initialised data");
-                self.send_proposal(self.start_data.clone());
-                self.send_prepare(self.start_data.clone());
+                self.start_data.clone()
```
diegomrsantos:
```suggestion
                &self.start_data
```

anchor/common/qbft/src/lib.rs
```diff
@@ -246,24 +188,45 @@ where
...
+            };
+            if let Some(data) = self.data.get(&hash).cloned() {
```
diegomrsantos:
```suggestion
            if let Some(data) = self.data.get(hash).cloned() {
```

anchor/common/qbft/src/config.rs
dknopik: No, IMO users should be able to use the `Config` directly instead of using the builder

dknopik: (but then we should probably move the validation)

anchor/common/qbft/src/config.rs
dknopik: sure, why not, feel free to commit that

anchor/common/qbft/src/lib.rs
dknopik: I named it `Data` intentionally, as it does more than hashing through the bounds on `Debug` and `Clone`.

I don't have a strong opinion on `Digest` vs `Hash`,  so feel free to change it.

anchor/common/qbft/src/lib.rs
dknopik: Not calling `start_round` leaves the instance in an invalid state, inviting misuse. I'd prefer to leave it here.

anchor/common/qbft/src/config.rs
diegomrsantos: > No, IMO users should be able to use the `Config` directly instead of using the builder

In this way, I don't see the point of the builder.

anchor/common/qbft/src/lib.rs
diegomrsantos: Why does the constructor create an instance in an invalid state?

anchor/common/qbft/src/config.rs
dknopik: The Builder is for ease of use when default values are desired. As the defaults are dependent on other values, a `Default` impl does not suffice.

diegomrsantos: The purpose of this builder pattern I mentioned is to facilitate the creation of an immutable config instance making it clearer what fields are required and optional, as well as centralizing the validation.

dknopik: Okay, fair point :+1:

anchor/common/qbft/src/lib.rs
dknopik: ah, sorry, got confused, you are right! should be fine without, even if messages come in without calling it. so yeah, we can change it.

anchor/common/qbft/src/config.rs
```diff
@@ -61,83 +58,201 @@ impl<F: Clone + LeaderFunction> Config<F> {
...
+    }
+
+    pub fn max_rounds(&self) -> usize {
+        self.max_rounds
+    }
+
+    pub fn leader_fn(&self) -> &F {
+        &self.leader_fn
+    }
+
```
diegomrsantos, line 187: do we need those?

anchor/common/qbft/src/types.rs
jking-aus: if it's just the one leave as is

anchor/common/qbft/src/lib.rs
jking-aus: valid comment but let's merge this in first

jking-aus (approved):
great work daniel - lots of good discussion as well

anchor/common/qbft/src/config.rs
dknopik: yeah, at least the test need to access the builder data (now that it is impossible to update the `Config`, see [c742d1d](https://github.com/sigp/anchor/pull/77/commits/c742d1d9d9fc56784ebcb2f6b66464e4828b87c7))

I am unsure if we now ended up with the most ergonomic/clean setup, but I think it's fine for now. We can revisit it later.

anchor/common/qbft/src/lib.rs
```diff
@@ -557,53 +521,62 @@ where
...
+        let consensus_data = ConsensusData {
             round: self.current_round,
             data: data.data,
-        }));
-        self.state = InstanceState::Prepare;
-        debug!(?self.state, "State Changed");
+        };
+        let operator_id = self.config.operator_id();
+        (self.send_message)(Message::Propose(operator_id, consensus_data.clone()));
+        self.received_propose(operator_id, consensus_data);
```
Zacholme7, line 549: why are we calling the corresponding receive after each send?

anchor/common/qbft/src/lib.rs
diegomrsantos: Is it worth creating an issue to address that?

PR #79, opened 2024-12-23
Subnet search

diegomrsantos:
## Issue Addressed

Related to https://github.com/sigp/anchor/issues/35

## Proposed Changes

- **Dependencies:** Updated dependencies in Cargo.toml including ethereum_serde_utils, ethereum_ssz, and ssz_types.
- **Discovery Module:** Added support for subnet queries, including new enums and functions for handling subnets.
- **ENR Modifications:** Enhanced build_enr to include a "subnets" field and added related bitfield functions.
- **Network Initialization:** Included a call to start_subnet_query during network behavior setup.
- **Protocol Identity:** Introduced ProtocolId struct, implemented ProtocolIdentity for it, updated Discovery to use Discv5<ProtocolId>, replaced ssv_node_predicate with tcp_predicate, and refined debug statements. The discv5's ProtocolID field is an identifier that nodes use to diverge away from other discv5 nodes (such as Ethereum). In the ssv network, it's set to 'ssvdv5' so that we find only SSV nodes in discovery right now. The `ssv` field in the ENR isn't necessary anymore.

## Additional Info

It advertises and searches for subnet number 9. This is an arbitrary choice for test purposes and does not have any specific significance.


anchor/network/src/discovery.rs
```diff
@@ -304,14 +364,35 @@ impl Discovery {
...
                 }
             }
-            _ => {
+            QueryType::Subnet(queries) => {
+                let subnets_searched_for: Vec<SSVSubnet> =
+                    queries.iter().map(|query| query.subnet).collect();
+
+                match query.result {
+                    Ok(r) if r.is_empty() => {
+                        debug!(subnets_searched_for = ?subnets_searched_for, "Grouped subnet discovery query yielded no results.");
```
dknopik:
```suggestion
                        debug!(
                            subnets_searched_for = ?subnets_searched_for,
                            "Grouped subnet discovery query yielded no results.",
                        );
```

just a nitpick as `rustfmt` doesn't split long macro invocations

anchor/network/src/discovery.rs
```diff
@@ -304,14 +364,35 @@ impl Discovery {
...
+                        // TODO queries.iter().for_each(|query| {
+                        //     self.add_subnet_query(query.subnet, query.min_ttl, query.retries + 1);
+                        // })
+                    }
+                    Ok(r) => {
+                        let results = r.into_iter().map(|enr| (enr, None)).collect();
+                        debug!(peers = ?results, subnets_searched_for = ?subnets_searched_for, "Peer grouped subnet discovery request completed");
```
dknopik:
```suggestion
                        debug!(
                            peers = ?results,
                            subnets_searched_for = ?subnets_searched_for,
                            "Peer grouped subnet discovery request completed",
                        );
```

just a nitpick as `rustfmt` doesn't split long macro invocations

anchor/network/src/discovery.rs
```diff
@@ -462,7 +550,45 @@ pub fn build_enr(enr_key: &CombinedKey, config: &Config) -> Result<Enr, String>
...
         .map_err(|e| format!("Could not build Local ENR: {:?}", e))
 }
+
+fn committee_bitfield(enr: &Enr) -> Result<Bitfield<Fixed<U128>>, &'static str> {
+    let bitfield_bytes: Bytes = enr
+        .get_decodable("subnets")
+        .ok_or("ENR subnet bitfield non-existent")?
+        .map_err(|_| "Invalid RLP Encoding")?;
+
+    BitVector::<U128>::from_ssz_bytes(&bitfield_bytes)
```
AgeManning, line 570: Are there 128 subnets?

anchor/network/src/discovery.rs
```diff
@@ -21,23 +23,63 @@ use libp2p::swarm::{
...
-
+use lighthouse_network::EnrExt;
+use serde::{Deserialize, Serialize};
+use ssz::{Decode, Encode};
+use ssz_types::length::Fixed;
+use ssz_types::typenum::U128;
+use ssz_types::{BitVector, Bitfield};
+
+/// Target number of peers to search for given a grouped subnet query.
+const TARGET_PEERS_FOR_GROUPED_QUERY: usize = 6;
```
AgeManning, line 39: We may be able to simplify some of this. I don't think the SSV subnets are dynamic, like if we have X,Y,Z operators, do we need to constantly search for subnets? Perhaps we can just subscribe to them forever?

AgeManning (commented):
Nice. 

I notice the subnet queries are still a `TODO`, so just noting that. 

I don't actually know about the SSV subnet specification. We might be able to simplify some things by having all the subnets long-lived (i.e we dont have to try and rotate or find peers for various subnets given a fixed operator set). 

In Lighthouse we've recently changed to deterministic subnets, where the first few bits of a node-id is used to determine who should listen on which subnet. This way we can search for a node-id with a given prefix (so the XOR metric in kad) is efficient at finding them, rather than a random walk. We probably should suggest doing the same for SSV. 

The other thing we need to know is how the backbone structure is formed. i.e which nodes are supposed to be permantently subscribed to which subnet. 

All of these things are just spec related, maybe we already know. This PR looks good to me as base structure to move on from. Nice :)

diegomrsantos:
> I notice the subnet queries are still a TODO, so just noting that.

What part exactly? In this PR we can search for peers in a subnet, some comments may be misleading. I left them as there are more things to do, this is the bare minimum.

anchor/network/src/discovery.rs
diegomrsantos: I think so https://github.com/ssvlabs/ssv/blob/main/network/p2p/p2p.go#L369

diegomrsantos: This is more reliable https://github.com/ssvlabs/ssv/blob/main/network/commons/common.go#L18

anchor/network/src/discovery.rs
diegomrsantos: How often we subscribe will depend on how frequently new committees are created for a peer. However, after subscribing, the subnet membership includes all peers assigned to that subnet number (CommitteeID % 128), not just peers from the same committee.

anchor/network/src/discovery.rs
```diff
@@ -304,14 +364,42 @@ impl Discovery {
...
+                let subnets_searched_for: Vec<SSVSubnet> =
+                    queries.iter().map(|query| query.subnet).collect();
+
+                match query.result {
+                    Ok(r) if r.is_empty() => {
+                        debug!(
+                            subnets_searched_for = ?subnets_searched_for,
+                            "Grouped subnet discovery query yielded no results.",
+                        );
+                        // TODO queries.iter().for_each(|query| {
```
AgeManning, line 385: I was referring to these todos

anchor/network/src/discovery.rs
```diff
@@ -304,14 +364,42 @@ impl Discovery {
...
+                            "Peer grouped subnet discovery request completed",
+                        );
+
+                        return Some(results);
+                    }
+                    Err(e) => {
+                        warn!(error = %e, "Subnet query failed");
+                    }
+                }
                 // TODO handle subnet queries
```
AgeManning, line 403: And this one

anchor/network/src/discovery.rs
diegomrsantos: It's because it's based on LH code and there're more things that I don't understand yet and I'm not sure it's necessary now. What do you think?

jking-aus (approved):
lgtm

Issue #74, opened 2024-12-16
Gossipsub topic subscription and message handling


jking-aus:
SSV Gossipsub topics need to be defined and handled. 

We've been given the below for reference but may need to seek further clarity: 
https://github.com/ssvlabs/ssv-spec/pull/491/files#diff-be5eaf872c86a9d95593ce73abfb790acd892c1278613ffa4b4c90580447ec01

diegomrsantos:
@jking-aus could you please clarify? I couldn't find topic info in the link mentioned.

diegomrsantos:
I found this https://github.com/ssvlabs/ssv-spec/blob/main/p2p/SPEC.md#subnets

PR #80, opened 2024-12-24
Subscribe to topic and decode SignedSSVMessage

diegomrsantos:
## Issue Addressed

Initial implementation for https://github.com/sigp/anchor/issues/74. The aim of this PR is to able to subscribe to a topic, receive and decode SignedSSVMessage's.

## Proposed Changes

- **Dependencies:** Updated dependencies in Cargo.toml including ethereum_ssz, ethereum_ssz_derive, and hex.
- **Network Types:** Added ssv_message module to src/types.rs.
- **Network Implementation:**
  - Enhanced network.rs to include ssv_message::SignedSSVMessage and ssz::Decode.
  - Subscribed to `ssv.v2.9` in the `run` method.
  - Handled GossipSub events to process and log SignedSSVMessage.
- **Gossipsub Configuration:** Changed validation_mode from Strict to Permissive. An error happened when running the system on Holesky using the previous mode.
- **New Files:** Added new file ssv_message.rs in anchor/network/src/types/.

## Additional Info

As in the previous subnet search PR, subnet 9 is an arbitrary choice for test purposes and has no specific significance.

### DIFF
--- a/anchor/network/src/network.rs
+++ b/anchor/network/src/network.rs
@@ -13,7 +13,7 @@ use libp2p::{futures, gossipsub, identify, ping, PeerId, Swarm, SwarmBuilder};
 use lighthouse_network::discovery::DiscoveredPeers;
 use lighthouse_network::discv5::enr::k256::sha2::{Digest, Sha256};
 use task_executor::TaskExecutor;
-use tracing::{debug, info, log, warn};
+use tracing::{debug, error, info, trace, warn};
 
 use crate::behaviour::AnchorBehaviour;
 use crate::behaviour::AnchorBehaviourEvent;
@@ -108,24 +108,23 @@ impl Network {
                                         message_id,
                                         message,
                                     } => {
-                                        log::debug!(
-                                            "Received message from {:?} with id {:?}: {:?}",
-                                            propagation_source,
-                                            message_id,
-                                            &message
+                                        debug!(
+                                            source = ?propagation_source,
+                                            id = ?message_id,
+                                            "Received SignedSSVMessage"
                                         );
                                         match SignedSSVMessage::from_ssz_bytes(&message.data) {
                                             Ok(deserialized_message) => {
-                                                log::debug!("SSVMessage: {:?}", deserialized_message);
+                                                debug!(msg = ?deserialized_message, "SignedSSVMessage deserialized");
                                             }
                                             Err(e) => {
-                                                log::error!("Error deserializing SSVMessage: {:?}", e);
+                                                error!("error" = ?e, "Failed to deserialize SignedSSVMessage");
                                             }
                                         }
                                     }
                                     // TODO handle gossipsub events
                                     _ => {
-                                        log::debug!("Unhandled gossipsub event: {:?}", ge);
+                                        debug!(event = ?ge, "Unhandled gossipsub event");
                                     }
                                 }
                                 // TODO handle gossipsub events
@@ -136,24 +135,24 @@ impl Network {
                             // them.
                             AnchorBehaviourEvent::Discovery(DiscoveredPeers { peers }) => {
                                 //self.peer_manager_mut().peers_discovered(peers);
-                                log::debug!("Discovered peers: {:?}", peers);
+                                debug!(peers =  ?peers, "Peers discovered");
                                 for (enr, _) in peers {
                                     for tcp in enr.multiaddr_tcp() {
-                                        log::trace!("Dialing peer: {:?}", tcp);
+                                        trace!(address = ?tcp, "Dialing peer");
                                         if let Err(e) = self.swarm.dial(tcp.clone()) {
-                                            log::error!("Error dialing peer {}: {}", tcp,  e);
+                                            error!(address = ?tcp, error = ?e, "Error dialing peer");
                                         }
                                     }
                                 }
                             }
                             // TODO handle other behaviour events
                             _ => {
-                                log::debug!("Unhandled behaviour event: {:?}", behaviour_event);
+                                debug!(event = ?behaviour_event, "Unhandled behaviour event");
                             }
                         },
                         // TODO handle other swarm events
                         _ => {
-                            log::debug!("Unhandled swarm event: {:?}", swarm_message);
+                            debug!(event = ?swarm_message, "Unhandled swarm event");
                         }
                     }
                 }
--- a/anchor/network/src/network.rs
+++ b/anchor/network/src/network.rs
@@ -5,7 +5,7 @@ use std::time::Duration;
 use futures::StreamExt;
 use libp2p::core::muxing::StreamMuxerBox;
 use libp2p::core::transport::Boxed;
-use libp2p::gossipsub::{MessageAuthenticity, ValidationMode};
+use libp2p::gossipsub::{IdentTopic, MessageAuthenticity, ValidationMode};
 use libp2p::identity::Keypair;
 use libp2p::multiaddr::Protocol;
 use libp2p::swarm::SwarmEvent;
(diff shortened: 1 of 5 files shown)


anchor/network/src/network.rs
```diff
@@ -83,12 +85,49 @@ impl Network {
...
+                                            propagation_source,
+                                            message_id,
+                                            &message
+                                        );
+                                        match SignedSSVMessage::from_ssz_bytes(&message.data) {
+                                            Ok(deserialized_message) => {
+                                                log::debug!("SSVMessage: {:?}", deserialized_message);
+                                            }
+                                            Err(e) => {
+                                                log::error!("Error deserializing SSVMessage: {:?}", e);
```
dknopik: IMO it would be nicer to use `tracing::error` instead of `tracing::log::error` and set the error field instead of embedding it in the message, similar to the logging at the beginning of this function. Same for other occurrences of `log::*`

anchor/network/src/types/ssv_message.rs
```diff
@@ -0,0 +1,633 @@
...
+            return Err(DecodeError::InvalidByteLength {
+                len: bytes.len(),
+                expected: MESSAGE_ID_LEN,
+            });
+        }
+        let mut id = [0u8; MESSAGE_ID_LEN];
+        id.copy_from_slice(bytes);
+        Ok(MessageID(id))
+    }
+}
```
dknopik, line 73: `impl`s look good, but it's unfortunate that this is necessary, using the derive macro would be nicer. I'll see if we can improve the situation upstream :+1:

anchor/network/src/types/ssv_message.rs
```diff
@@ -0,0 +1,633 @@
...
+
+impl MsgType {
+    /// Tries to convert a `u64` into a `MsgType`.
+    pub fn try_from_u64(value: u64) -> Result<Self, DecodeError> {
+        match value {
+            0 => Ok(MsgType::SSVConsensusMsgType),
+            1 => Ok(MsgType::SSVPartialSignatureMsgType),
+            _ => Err(DecodeError::NoMatchingVariant),
+        }
+    }
```
dknopik: I would suggest to `impl TryFrom<u64> for MsgType` instead :)

anchor/network/src/types/ssv_message.rs
```diff
@@ -0,0 +1,633 @@
...
+        if bytes.len() != U64_SIZE {
+            return Err(DecodeError::InvalidByteLength {
+                len: bytes.len(),
+                expected: U64_SIZE,
+            });
+        }
+        let value = u64::from_le_bytes(bytes.try_into().unwrap());
+        MsgType::try_from_u64(value)
+    }
+}
```
dknopik: Same here, would be nice for the macro to support generating this if there is `#[repr(u*)]` on an enum, will check upstream. Otherwise, lgtm.

anchor/network/src/types/ssv_message.rs
```diff
@@ -0,0 +1,633 @@
...
+
+/// Represents an Operator ID as a 64-bit unsigned integer.
+pub type OperatorID = u64;
+
+/// Represents an SSV Message with type, ID, and data.
+#[derive(Encode, Decode, Debug, Clone, PartialEq, Eq)]
+pub struct SSVMessage {
+    msg_type: MsgType,
+    msg_id: MessageID, // Fixed-size [u8; 56]
+    data: Vec<u8>,     // Variable-length byte array
```
dknopik, line 155: I do not like the `Vec<u8>` here - but unfortunately, it's probably the best option.

The other option would be an enum with `ssz_derive`'s `#[ssz(enum_behaviour = "transparent")]`, but this is inefficient (as it blindly tries to decode each variant until one variant does not fail to decode).

Longer term, IMO the SSV spec should switch to a [SSZ union](https://github.com/ethereum/consensus-specs/blob/dev/ssz/simple-serialize.md#union) - or separate topics for separate message types.

anchor/network/src/types/ssv_message.rs
```diff
@@ -0,0 +1,633 @@
...
+    }
+}
+
+/// Represents a signed SSV Message with signatures, operator IDs, the message itself, and full data.
+#[derive(Encode, Decode, Debug, Clone, PartialEq, Eq)]
+pub struct SignedSSVMessage {
+    signatures: Vec<Vec<u8>>, // Vec of Vec<u8>, max 13 elements, each up to 256 bytes
+    operator_ids: Vec<OperatorID>, // Vec of OperatorID (u64), max 13 elements
+    ssv_message: SSVMessage,  // SSVMessage: Required field
+    full_data: Vec<u8>,       // Variable-length byte array, max 4,194,532 bytes
```
dknopik, line 204: I think it's preferable to use [`ssz_types::VariableList`](https://docs.rs/ssz_types/0.10.0/ssz_types/struct.VariableList.html) (basically a `Vec` wrapper) here to enforce the maximum length.

dknopik (commented):
Left some remarks. Looking good!

anchor/network/src/network.rs
diegomrsantos: Good catch, thanks. Something like this?
```rust
error!("error" = ?e, "Failed to deserialize SSVMessage");
```

anchor/network/src/types/ssv_message.rs
diegomrsantos: Yeah, it fails with an error. Should we add a comment explaining why this is necessary?

anchor/network/src/network.rs
dknopik: yep, looks good!

diegomrsantos: Done. These log levels are mostly to help me debug the implementation and will probably change for production.

anchor/network/src/types/ssv_message.rs
diegomrsantos: Much better, thanks!

anchor/network/src/types/ssv_message.rs
diegomrsantos: I agree now. My argument of decoupling this type from the application doesn't make much as the `msg_type: MsgType` already has application knowledge, doesn't it? Also, not sure that decoding this at the application level is the right place for that. Would this inefficiency severely affect performance? No way around that? But another possible issue is the `EthSpec` everywhere in the code.

diegomrsantos: Not decoding it at the networking layer would make more sense if the `msg_type` was encoded inside the data. In this way, we would know nothing about the data and just deliver it to the application. If messages are created/deleted/modified nothing would change at this layer.

Zacholme7: I think we are pretty flexible in being able to push `msg_type` down into data. As long as qbft can work with serialized data without having to introduce `EthSpec` I dont think we have a ton of hard constraints. Some internal abstraction layer will be needed regardless. Something like
                                                                         
Network -> decode -> Validator store -> Qbft manager/Signature Collector -> Validator Store -> encoder -> Network

diegomrsantos: It's even simpler. We would just not decode `msg_type` and pass it as a number. It'd make even more sense with the current code, I guess.

anchor/network/src/types/ssv_message.rs
diegomrsantos: The docs seem to say the list is truncated if it's bigger than the max.

diegomrsantos: There's a validation in the current code checking the length.

dknopik: > The docs seem to say the list is truncated if it's bigger than the max.

Yeah, the `From` implementation unfortunately truncates for some reason. However, the `Decode` impl does check: https://docs.rs/ssz_types/0.10.0/src/ssz_types/variable_list.rs.html#274-313

jking-aus (approved):
nice -- good  discussion

PR #83, opened 2025-01-07
Add lockbud to CI

eserilev:
Adds [lockbud](https://github.com/BurtonQin/lockbud) to CI w/ one caveat

The caveat is the CI job never actually fails. Even if a deadlock is found, it'll just log the deadlock in CI, but wont return an error code. We have an issue up in LH [here](https://github.com/sigp/lighthouse/issues/6637) to eventually fix this

So yeah, not the most ideal CI job... but still might be nice to have until we get the job to fail correctly


magick93:
> Even if a deadlock is found, it'll just log the deadlock in CI, but wont return an error code. We have an issue up in LH [here](https://github.com/sigp/lighthouse/issues/6637) to eventually fix this

@eserilev do you have a way to reproduce this - as in some sample rust that will trigger output from lockbud?

If yes, I can try get it to fail the workflow.

eserilev:
@magick93 

you should be able to force a deadlock with the following snippet
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=89b90214fbcb364bd9ac5ccaf4ff7d01

If its easier you can revert the changes in this commit here in LH:
https://github.com/sigp/lighthouse/pull/6470/commits/101bf81c70d2a4445d54237d343f2abbc6a8c61b

lockbud will report a deadlock

dknopik:
@magick93, I have introduced such a failure: https://github.com/sigp/anchor/actions/runs/12806469752/job/35705033682?pr=83#step:5:1312 - hope it helps!

dknopik (approved):
LGTM

jking-aus (approved):

Issue #26, opened 2024-10-23
HTTP API


AgeManning:
Implement the various HTTP API Endpoints. 

We may be able to borrow a few from the [validator client](https://github.com/sigp/lighthouse/tree/stable/validator_client/src/http_api)

We are using axum, so the routes need to be updated

magick93:
Will this conform the same openapi specs as https://api.ssv.network/documentation/ ?

AgeManning:
I think partially, yes. 
Although that spec doesn't really tell us much apart from the endpoints. It doesn't give return values or types for the ones I checked, so it would be very hard to implement. 

I think we can use some of these endpoints that make sense, and then actually specify the return types if we are going to use it.

PR #84, opened 2025-01-08
Http api and http testing

magick93:
## Issue Addressed

This is related to #26 and related testing issues. 

## Proposed Changes

Proposal to use identical paths / endpoints and method names as the ssv openapi spec. 

But, as these spec is incomplete in terms of response types, to improve it by being more rigorous with, in particular, response types.


anchor/http_api/src/api/handlers.rs
```diff
@@ -0,0 +1,742 @@
+// Base API template
+use serde::{Serialize, Deserialize};
+use utoipa::ToSchema;
+
+
+/// OperatorMetadataDto model
+#[derive(Debug, Serialize, Deserialize, ToSchema)]
+pub struct OperatorMetadataDto {
```
Zacholme7: I think the fields should stick with rust underscore convention and just use [serde rename](https://serde.rs/field-attrs.html). Something like 
`#[serde(rename = "operatorName")]
pub operator_name: Option<String>`

anchor/http_api/src/api/handlers.rs
```diff
@@ -0,0 +1,742 @@
...
+
+// Base API template
+use axum::{
+    Router,
+    routing::{get, post, put, delete},
+};
+use super::handlers::*;
+
+pub fn create_router() -> Router {
+    Router::new()
```
Zacholme7: might be nice to try to break these functions up into modules for a nice organizational hierarchy. 

Something like a root handler module, than operator module, network module, cluster module, etc. Then instead of the long function name you could do `handle::operator::get_operator`, `handle::cluster::get_id`. Doesn't have to be that structure, just some separation.

anchor/http_api/src/api/handlers.rs
```diff
@@ -0,0 +1,742 @@
...
+    
+}
+
+// Base API template
+use axum::{ extract::{Path, Query, Json}, response::Json as JsonResponse, http::StatusCode };
+use serde::{Serialize, Deserialize};
+use utoipa::ToSchema;
+
+
+#[utoipa::path(GET, path = "/api/v4/{network}/accounts",  tag = "Accounts",   params(("network", String),("page", String),("perPage", String),),
```
Zacholme7: these macros are quite verbose. im not very familiar with utopia so im not sure if this is the convention. is there any way to simplify these?

Zacholme7:
Nice start on this its good to get all these down and now we can start to nail down specifics! Left some early feedback with nitpick comments.

anchor/http_api/src/api/handlers.rs
magick93: Thanks @Zacholme7 - yes, was thinking the same.

anchor/http_api/src/api/handlers.rs
magick93: The `utoipa` macro seems to be the most common way to add openapi / swagger docs. 

I get it though, it is verbose. But these openapi annotations do make it easier for api consumers. 

These function signatures are generated from templates. So we can tweak the templates to our liking.

jking-aus:
closing this -- appreciate the effort but I think we should work on adding features as required for the http-api

PR #86, opened 2025-01-09
switch to `anchor` branch for LH dependencies for now

dknopik:
Switch to the `anchor` branch for Lighthouse dependencies, which is our "playground" for changes to be merged into unstable. 

This is neat because we can now merge changes waiting for proper reviews in Lighthouse!

Also, sort workspace dependencies in the root Cargo.toml (which is for some reason not done by `cargo sort`).

Also, do a `cargo update` while we're at it.

Also, this PR reduces our total dependency count from 1126 to 873 and a clean `make lint` from takes now 2m 7s (on my machine, and down from 2m 49s).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,26 +3,6 @@
 version = 3
 
 [[package]]
-name = "account_utils"
-version = "0.1.0"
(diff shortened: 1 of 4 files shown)


jking-aus (approved):
yes, this - thank you

Issue #36, opened 2024-10-23
Unit tests for QBFT


jking-aus:
Write a series of unit tests for the QBFT consensus mechanism that cover at minimum: 
- Happy case and success
- Recovery from <= F malicious operators 
- Consensus failure 
- Simple attacks

PR #94, opened 2025-01-13
Qbft tests

Zacholme7:
## Issue Addressed

#36 

## Proposed Changes

Few changes here
- Added more unit tests to ensure proper functionality
- Bugfix to prevent case where ROUNDCHANGE never reached quorum 
- Track the number of nodes that reached consensus for testing
- Add the ability to stop and restart nodes in the qbft instance. This simulates nodes going down and rejoining.

Update
- Fixed tests to integrate with changes from #77 

Todo
- The qbft manager drives round changes so there should be more rigorous end to end testing using that. This should be a separate issue to test both the ROUNDCHANGE and the manager.


jking-aus (dismissed):
nice catch zac -- good extension to tests as well

anchor/common/qbft/src/lib.rs
```diff
@@ -162,12 +167,21 @@ where
...
+            round = *self.current_round,
+            "Starting new round"
+        );
 
         // Remove round change messages that would be for previous rounds
         self.round_change_messages
             .retain(|&round, _value| round >= self.current_round);
 
+        // We are waiting for consensus on a round change, do not start the round yet
+        if matches!(self.state, InstanceState::SentRoundChange) {
```
diegomrsantos, line 181: What happens to the caller(s) in this case?

anchor/common/qbft/src/lib.rs
```diff
@@ -416,7 +442,13 @@ where
                 if operators.len() >= self.config.quorum_size()
                     && matches!(self.state, InstanceState::Commit)
                 {
+                    debug!(
+                        self = *self.config.operator_id(),
+                        round = *self.current_round,
+                        "Reached commit consensus"
+                    );
                     self.completed = Some(Completed::Success(data.clone()));
+                    debug!(self = *self.config.operator_id(), round = *self.current_round, data = ?self.completed, "Consensus reached");
```
diegomrsantos, line 451: Are those two debugs necessary?

anchor/common/qbft/src/tests.rs
```diff
@@ -87,31 +91,52 @@ fn construct_and_run_committee<D: Data + Default + 'static>(
...
     TestQBFTCommittee {
         msg_queue,
         instances,
+        active_instances,
     }
 }
 
 impl<D: Default + Data, S: FnMut(Message<D>)> TestQBFTCommittee<D, S> {
-    fn wait_until_end(mut self) {
+    fn wait_until_end(mut self) -> i32 {
```
diegomrsantos, line 105: What about a more descriptive name?

```suggestion
    fn process_msgs_and_count_consensus(mut self) -> i32 {
```

diegomrsantos, line 105: What about a more descriptive name like `process_msgs_and_count_consensus`?

diegomrsantos: Would it be beneficial to return the operator's IDs instead?

anchor/common/qbft/src/tests.rs
```diff
@@ -126,11 +151,150 @@ impl Data for TestData {
...
+
+    // Then restart it
+    test_instance.restart_instance(&OperatorId::from(0));
+
+    let num_consensus = test_instance.wait_until_end();
+    assert_eq!(num_consensus, 5); // Should reach full consensus after recovery
+}
+
+#[test]
+fn test_duplicate_proposals() {
```
diegomrsantos, line 190: Is it a Byzantine fault?

anchor/common/qbft/src/lib.rs
Zacholme7: If we hit this state, that means that we have received some round change messages but have not received a quorum yet, so nothing happens until the node reaches a round change quorum and can start the next round.

anchor/common/qbft/src/lib.rs
Zacholme7: Nope, was just debugging tests. Will remove

anchor/common/qbft/src/tests.rs
Zacholme7: name change is good. operator ids is also a good change because then we can check which specific operators were able to reach consensus. thanks

anchor/common/qbft/src/tests.rs
Zacholme7: Do you mean is this function testing a byzantine fault? If so, yes since non leaders are sending propose messages.

anchor/common/qbft/src/lib.rs
diegomrsantos: I guess I'm just wondering about the API. In this case, we call `start_round`but nothing happens and it's not communicated to the caller.

anchor/common/qbft/src/tests.rs
diegomrsantos: When a node is byzantine, it still reaches consensus?

anchor/common/qbft/src/tests.rs
```diff
@@ -126,11 +151,150 @@ impl Data for TestData {
...
+            .instances
+            .get_mut(&OperatorId::from(id))
+            .unwrap();
+        instance.receive(msg.clone());
+        instance.receive(msg.clone());
+        instance.receive(msg.clone());
+    }
+
+    let num_consensus = test_instance.wait_until_end();
+    assert_eq!(num_consensus, 5); // Should still reach consensus despite duplicates
```
diegomrsantos, line 214: Maybe we could make `5` a constant imported from `TestQBFTCommitteeBuilder`?

anchor/common/qbft/src/lib.rs
Zacholme7: Ahh got it good point. I dont think it really matters. The caller can assume that when it makes some call to qbft, if there is a message that has to be sent it will receive it. In this case, we have no message to send since there is no progress to be made. [here](https://github.com/ssvlabs/ssv-spec/blob/770e204aeabfd2e37c4d3c2b652a43efec69aead/qbft/docs/qbft_bpmn.drawio.png) it just spins. I will think more on this.

anchor/common/qbft/src/tests.rs
Zacholme7: Here, yes. This is just a test for a specific type of byzantine behavior which is proposals from non-leaders. The protocol can handle and protect against this.

anchor/common/qbft/src/lib.rs
diegomrsantos: It somehow feels strange to call `start_round`, nothing happens, and there's no error. Does the caller find out about it in another way? Does it call it again later?

Zacholme7: Yep I get what you are saying. Calling `start_round` makes it seem like the round will start. Maybe `try_start_round` is better? Im hesitant to call this an error because it is functioning correctly, it just cant make any progress. Maybe it is better try to put this logic into the proposal code. 

The caller will keep calling it until there is a quorum of round change messages and it can make progress, or right away when it is round 0. There should be errors added so that the caller can handle clear errors such as incorrect messages, but in this case there is really nothing for the caller to do. All that can be done is to wait for more round change messages to reach quorum.

jking-aus (approved):
great work mate

Issue #88, opened 2025-01-10
`cargo sort` does not sort the workspace dependencies


dknopik:
## Description

`cargo sort` does not sort the workspace dependencies. This causes merges to accidentally shuffle around dependencies, which then causes trouble on futher merges.

## Expected Behaviour

`cargo sort` should also sort workspace dependencies.

## Steps to resolve

I don't know. Probably an upstream issue, this is just to track this.

Zacholme7:
taplo looks like it works. Just running `taplo fmt` in the root will sort all of the deps. https://crates.io/crates/taplo-cli. could try this out

dknopik:
> taplo looks like it works. Just running `taplo fmt` in the root will sort all of the deps. https://crates.io/crates/taplo-cli. could try this out

Doesn't seem like it's sorting out-of-the-box for me :thinking:

PR #97, opened 2025-01-14
Ensure the "workspace.dependencies" table is sorted

dknopik:
## Issue Addressed

- #88

## Proposed Changes

As a workaround, use Taplo to check during CI whether the dependencies are sorted. 

## Additional Info

Taplo allows empty lines for different sorted subsections, so I categorized the dependencies: Anchor internal, Lighthouse dependencies, other dependencies

Yes, this is annoying, but it ensures easy and clean merges once we are all on the same page.

### DIFF
--- a/.github/workflows/test-suite.yml
+++ b/.github/workflows/test-suite.yml
@@ -210,6 +210,6 @@ jobs:
         with:
           channel: stable
           cache-target: release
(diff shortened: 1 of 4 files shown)


jking-aus (approved):
nice lgtm

PR #105, opened 2025-01-15
Remove `ssv_types` dependency on `qbft`

Zacholme7:
## Proposed Changes

`ssv_types` was importing the `qbft` crate which prevented us from using those types in qbft as it created a cyclic dependency. Types should ideally not depend on any other internal crates to prevent this from happening, so I moved the `Data` trait into types to keep it contained.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5619,6 +5619,7 @@ version = "0.1.0"
 dependencies = [
  "derive_more 1.0.0",
  "indexmap",
+ "ssv_types",
  "tracing",
  "tracing-subscriber",
 ]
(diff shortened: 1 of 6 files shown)


jking-aus (approved):

dknopik:
I understand the reasoning behind this change and am fine with it. Just wanted to note my reasoning here: I was hoping that `qbft` could stay rather independent, i.e. usable for QBFT consensus without being specific to SSV (which it now is though the dependency on `ssv_types`).

Zacholme7:
@dknopik okay yea that makes sense. Ill probably work on getting it 100% functional with the ssv type dependency which maps well to the spec and then once we have it working we can look at making it generic.

PR #109, opened 2025-01-17
wire up everything

dknopik:
## Proposed Changes

- add config for network selection
- read or generate SSV private key
- init database and sync
- fill validator store from database

## Additional Info

:tada: :tada: :tada:

There is some room for optimisation/alignment between the components, but we can leave that for later.

### DIFF
--- a/anchor/client/src/config.rs
+++ b/anchor/client/src/config.rs
@@ -2,9 +2,9 @@
 // use clap_utils::{flags::DISABLE_MALLOC_TUNING_FLAG, parse_optional, parse_required};
 
 use crate::cli::Anchor;
+use eth2_network_config::Eth2NetworkConfig;
 use network::{ListenAddr, ListenAddress};
 use sensitive_url::SensitiveUrl;
-use serde::{Deserialize, Serialize};
 use std::fs;
 use std::net::IpAddr;
 use std::path::PathBuf;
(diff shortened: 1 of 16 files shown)


anchor/client/src/config.rs
```diff
@@ -2,30 +2,31 @@
...
 use sensitive_url::SensitiveUrl;
-use serde::{Deserialize, Serialize};
 use std::fs;
 use std::net::IpAddr;
 use std::path::PathBuf;
 use tracing::warn;
 
 pub const DEFAULT_BEACON_NODE: &str = "http://localhost:5052/";
 pub const DEFAULT_EXECUTION_NODE: &str = "http://localhost:8545/";
+pub const DEFAULT_EXECUTION_NODE_WS: &str = "wss://localhost:8545/";
```
Zacholme7: For default local nodes, I believe this should just be "ws"

anchor/client/src/lib.rs
```diff
@@ -276,26 +295,70 @@ impl Client {
...
+            async move {
+                if let Err(e) = syncer.sync().await {
+                    error!("Syncer failed: {e}");
+                }
+            },
+            "syncer",
+        );
 
-        // Create the processor-adjacent managers
+        // Wait until we have an operator id
```
Zacholme7: We need to make sure we are in live sync first before getting the operator id. For example, if we are operator 1 we will get our id right away and we will continue executing while we are still syncing.

The easiest way to do this is to probably work in an `AtomicBool` that is flipped when we enter live sync. Once the bool is true we know we are synced up and then can either fetch or wait on the id.

Zacholme7:
small comments, gj otherwise!

anchor/client/src/lib.rs
dknopik: Ah - realizing now that SSV supports removing a validator from the SSV network without exiting it :sweat_smile:. Yeah, then waiting to be fully live is necessary.

anchor/eth/src/sync.rs
```diff
@@ -106,6 +108,8 @@ pub struct SsvEventSyncer {
     event_processor: EventProcessor,
     /// The network the node is connected to
     network: Network,
+    /// Notify a channel as soon as the historical sync is done
+    historic_finished_notify: Option<Sender<()>>,
```
Zacholme7, line 112: Why is this an option? I think we will always want to wait for historical sync to finish first, right?

dknopik: We can only send a value through the `oneshot` sender once, consuming the sender (as it's taking ownership).

Therefore we need the option to be able to gain ownership via `take`.

Also, we might not need to wait for historic sync in e.g. testing code.

anchor/validator_store/src/lib.rs
```diff
@@ -56,69 +61,173 @@ use validator_store::{
...
+        for (cluster, validator) in db_clusters
+            .into_iter()
+            .filter_map(|id| self.database.clusters().get_by(id.key()).map(Arc::new))
+            .flat_map(|cluster| {
+                self.database
+                    .metadata()
+                    .get_all_by(&cluster.cluster_id)
+                    .unwrap_or_default()
+                    .into_iter()
+                    .map(move |metadata| (cluster.clone(), metadata))
```
Zacholme7, line 142: Note for future us, we should probably find some way to optimize/simplify this in the future :sweat_smile:

Zacholme7:
lgtm!

anchor/validator_store/src/lib.rs
diegomrsantos: I understand we're still learning and exploring, and I'm also doing the same, but something to keep in mind https://yiming.dev/clipping/2019/03/21/le-blanc's-law-a-k-a-later-equals-never/

dknopik: I'll read that later. :P

dknopik: But for real, I get what you mean. Still, I think it's worth "rushing" to an MVP, because then we can see where the problems are at, and optimize based on findings.

dknopik:
Kingy, can we please merge this? 🥺 @jking-aus

jking-aus (approved):
lgtm

PR #110, opened 2025-01-21
Update lh deps

dknopik:
Update to the newest version of the `anchor` branch over at Lighthouse and fix 1 compilation issue arising from it.

Also, specify LH dependencies to a specific commit to avoid accidental breaking dependency updates when doing `cargo update`.


jking-aus (approved):
lgtm

PR #115, opened 2025-01-23
Subnet service

dknopik:
## Proposed Changes

Small service that gets notified when a validator gets added or removed and passes on instructions to subscribe/unsubscribe/look for peers

### DIFF
--- a/anchor/client/src/lib.rs
+++ b/anchor/client/src/lib.rs
@@ -29,6 +29,7 @@ use std::net::SocketAddr;
 use std::path::Path;
 use std::sync::Arc;
 use std::time::{Duration, SystemTime, UNIX_EPOCH};
+use subnet_tracker::start_subnet_tracker;
 use task_executor::TaskExecutor;
 use tokio::net::TcpListener;
 use tokio::select;
@@ -135,16 +136,19 @@ impl Client {
             return Err("HTTP API Failed".to_string());
         }
 
-        // Start the p2p network
-        let network = Network::try_new(&config.network, executor.clone()).await?;
-        // Spawn the network listening task
-        executor.spawn(network.run(), "network");
-
         // Open database
         let database = Arc::new(
             NetworkDatabase::new(config.data_dir.join("anchor_db.sqlite").as_path(), &pubkey)
                 .map_err(|e| format!("Unable to open Anchor database: {e}"))?,
         );
+
+        let subnet_tracker =
+            start_subnet_tracker(database.watch(), network::SUBNET_COUNT, &executor);
+
+        // Start the p2p network
+        let network = Network::try_new(&config.network, subnet_tracker, executor.clone()).await?;
+        // Spawn the network listening task
+        executor.spawn(network.run(), "network");
 
         // Initialize slashing protection.
         let slashing_db_path = config.data_dir.join(SLASHING_PROTECTION_FILENAME);
--- a/anchor/network/src/network.rs
+++ b/anchor/network/src/network.rs
@@ -25,11 +25,12 @@ use crate::Config;
 use crate::types::ssv_message::SignedSSVMessage;
 use lighthouse_network::EnrExt;
 use ssz::Decode;
-use subnet_tracker::{SubnetEvent, SubnetId, SubnetTracker};
+use subnet_tracker::{SubnetEvent, SubnetId};
+use tokio::sync::mpsc;
 
 pub struct Network {
     swarm: Swarm<AnchorBehaviour>,
-    subnet_tracker: SubnetTracker,
+    subnet_tracker: mpsc::Receiver<SubnetEvent>,
     peer_id: PeerId,
 }
 
@@ -38,7 +39,7 @@ impl Network {
     // p2p network.
     pub async fn try_new(
         config: &Config,
-        subnet_tracker: SubnetTracker,
+        subnet_tracker: mpsc::Receiver<SubnetEvent>,
         executor: TaskExecutor,
     ) -> Result<Network, String> {
         let local_keypair: Keypair = load_private_key(&config.network_dir);
--- a/anchor/network/src/network.rs
+++ b/anchor/network/src/network.rs
@@ -25,16 +25,22 @@ use crate::Config;
 use crate::types::ssv_message::SignedSSVMessage;
 use lighthouse_network::EnrExt;
 use ssz::Decode;
+use subnet_tracker::{SubnetEvent, SubnetId, SubnetTracker};
 
 pub struct Network {
     swarm: Swarm<AnchorBehaviour>,
+    subnet_tracker: SubnetTracker,
     peer_id: PeerId,
 }
 
 impl Network {
     // Creates an instance of the Network struct to start sending and receiving information on the
     // p2p network.
-    pub async fn try_new(config: &Config, executor: TaskExecutor) -> Result<Network, String> {
+    pub async fn try_new(
+        config: &Config,
+        subnet_tracker: SubnetTracker,
+        executor: TaskExecutor,
+    ) -> Result<Network, String> {
         let local_keypair: Keypair = load_private_key(&config.network_dir);
         let transport = build_transport(local_keypair.clone(), !config.disable_quic_support);
         let behaviour = build_anchor_behaviour(local_keypair.clone(), config).await;
@@ -48,6 +54,7 @@ impl Network {
                 behaviour,
                 config,
             ),
+            subnet_tracker,
             peer_id,
         };
 
@@ -155,11 +162,53 @@ impl Network {
                             debug!(event = ?swarm_message, "Unhandled swarm event");
                         }
                     }
+                },
+                event = self.subnet_tracker.recv() => {
+                    match event {
+                        Some(event) => self.on_subnet_tracker_event(event),
+                        None => {
+                            error!("subnet tracker has quit");
+                            return;
+                        }
+                    }
                 }
                 // TODO match input channels
             }
         }
     }
+
+    fn on_subnet_tracker_event(&mut self, event: SubnetEvent) {
+        match event {
+            SubnetEvent::Join(subnet) => {
+                if let Err(err) = self
+                    .swarm
+                    .behaviour_mut()
+                    .gossipsub
+                    .subscribe(&subnet_to_topic(subnet))
+                {
+                    error!(?err, subnet = *subnet, "can't subscribe");
+                }
+                self.swarm
+                    .behaviour_mut()
+                    .discovery
+                    .start_subnet_query(vec![subnet]);
+            }
+            SubnetEvent::Leave(subnet) => {
+                if let Err(err) = self
+                    .swarm
+                    .behaviour_mut()
+                    .gossipsub
+                    .unsubscribe(&subnet_to_topic(subnet))
+                {
+                    error!(?err, subnet = *subnet, "can't unsubscribe");
+                }
+            }
+        }
+    }
+}
+
+fn subnet_to_topic(subnet: SubnetId) -> IdentTopic {
+    IdentTopic::new(format!("ssv.{}", *subnet))
 }
 
 async fn build_anchor_behaviour(
@@ -213,7 +262,6 @@ async fn build_anchor_behaviour(
             .unwrap();
         // start searching for peers
         discovery.discover_peers(FIND_NODE_QUERY_CLOSEST_PEERS);
-        discovery.start_subnet_query();
         discovery
     };
 
@@ -280,10 +328,11 @@ fn build_swarm(
 
 #[cfg(test)]
 mod test {
-    use task_executor::TaskExecutor;
-
     use crate::network::Network;
     use crate::Config;
+    use std::time::Duration;
+    use subnet_tracker::test_tracker;
+    use task_executor::TaskExecutor;
 
     #[tokio::test]
     async fn create_network() {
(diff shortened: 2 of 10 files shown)


anchor/network/src/network.rs
```diff
@@ -156,12 +167,53 @@ impl Network {
...
+            SubnetServiceEvent::Subscribe(subnet) => {
+                if let Err(err) = self
+                    .swarm
+                    .behaviour_mut()
+                    .gossipsub
+                    .subscribe(&subnet_to_topic(subnet))
+                {
+                    error!(?err, subnet = *subnet, "can't subscribe");
+                }
+                self.swarm
```
diegomrsantos, line 192: Interestingly, Lighthouse doesn't do that. There could be a check somewhere else (PeerManager?) that keeps the target number of peers for each subnet.

anchor/network/src/subnet.rs
```diff
@@ -0,0 +1,112 @@
...
+            Err(RecvError::Closed) => {
+                // should never happen, as we hold a handle to the db
+                error!("unexpected close of validator update channel");
+                return;
+            }
+        }
+    }
+}
+
+async fn catch_up(
```
diegomrsantos: Maybe a more descriptive name?

anchor/network/src/subnet.rs
```diff
@@ -0,0 +1,112 @@
...
+        match update {
+            Ok(ValidatorUpdate::ValidatorAdded { cluster, .. }) => {
+                let subnet = cluster_subnet(&cluster);
+                validator_counts[subnet] += 1;
+                if validator_counts[subnet] == 1 {
+                    let _ = sender
+                        .send(SubnetServiceEvent::Subscribe(SubnetId::from(subnet as u64)))
+                        .await;
+                }
+            }
```
diegomrsantos: Perhaps this logic belongs to the application layers instead. It could only be a matter of moving this file somewhere else. It could also make sense to define an unique entry point where the application talks to the network.

diegomrsantos (commented):
Thank you for starting this! I added some comments.

anchor/database/src/cluster_operations.rs
```diff
@@ -81,6 +81,13 @@ impl NetworkDatabase {
             validator.to_owned(),  // The metadata of the validator
         );
 
+        if our_share.is_some() {
+            let _ = self.validator_broadcast.send(ValidatorUpdate::ValidatorAdded {
+                pubkey: validator.public_key.clone(),
+                cluster: cluster.cluster_id,
+            });
+        }
```
Zacholme7: Can this be moved a few lines up to where we check `Some(share) = our_share`? Should probably also log a warning on send error.

dknopik: The reason why I put it at the end is that I want to avoid a race condition where a task receiving a notification tries to get the data of the new validator before the database wrote it to the map.

I did not add a warning log, because there might be no current subscribers - but I am open to add a warning anyway, since this should not usually be the case, except maybe during shutdown.

anchor/network/src/subnet.rs
dknopik: Where would you put it instead? A new crate, kind of like the separation of Lighthouses `network` vs `lighthouse_network` and `gossipsub`?

dknopik:
Pushed a new design based on #117.

Also pulled out the logic from `network` into its own crate, but kept the full logic (mapping updates to subnets) in one place. The reasoning behind that is that I think we should avoid bothering the `network` create with events as much as possible to avoid computation in the network service if possible.

anchor/subnet_tracker/src/lib.rs
```diff
@@ -0,0 +1,185 @@
...
+
+pub fn start_subnet_tracker(
+    db: watch::Receiver<NetworkState>,
+    subnet_count: usize,
+    executor: &TaskExecutor,
+) -> SubnetTracker {
+    // a channel capacity of 1 is fine - the subnet_tracker does not do anything else, it can wait.
+    let (tx, rx) = mpsc::channel(1);
+    executor.spawn(subnet_tracker(tx, db, subnet_count), "subnet_tracker");
+    SubnetTracker { events: rx }
```
diegomrsantos: What is `subnet_tracker` doing? It feels a bit confusing to me, it's being called before the `SubnetTracker`instance is created right after.

anchor/subnet_tracker/src/lib.rs
```diff
@@ -0,0 +1,185 @@
...
+
+#[derive(Clone, Debug)]
+enum JoinState {
+    Not,
+    Old,
+    New,
+    Still,
+}
+
+async fn subnet_tracker(
```
diegomrsantos, line 59: Maybe this could be a `start` function called on a `SubnetTracker` instance?

diegomrsantos: Something like

```rust
pub struct SubnetTracker {
    db: watch::Receiver<NetworkState>,
    subnet_count: usize,
    events_tx: mpsc::Sender<SubnetEvent>,
    events_rx: mpsc::Receiver<SubnetEvent>,
}

impl SubnetTracker {
    pub fn new(db: watch::Receiver<NetworkState>, subnet_count: usize) -> Self {
        let (tx, rx) = mpsc::channel(1);
        Self {
            db,
            subnet_count,
            events_tx: tx,
            events_rx: rx,
        }
    }

    /// Spawn the background task that processes network state updates.
    pub fn start(&self, executor: &TaskExecutor) {
        let db = self.db.clone();          // watchers can be cloned
        let tx = self.events_tx.clone();
        let subnet_count = self.subnet_count;

        executor.spawn(async move {
            run_subnet_tracker_loop(tx, db, subnet_count).await;
        }, "subnet_tracker");
    }

    /// Receive subnet events from the background task.
    pub async fn recv(&mut self) -> Option<SubnetEvent> {
        self.events_rx.recv().await
    }
}

async fn run_subnet_tracker_loop(
    tx: mpsc::Sender<SubnetEvent>,
    mut db: watch::Receiver<NetworkState>,
    subnet_count: usize
) {
    // The same logic as the existing `subnet_tracker` function
}
```

Wdyt? It might make things easier to understand and test. Or it could just be my OOP mind.

anchor/subnet_tracker/src/lib.rs
```diff
@@ -0,0 +1,185 @@
...
+        // do not await while holding lock!
+        // explicit scope needed because rustc cant handle equivalent drop(state)
+        {
+            let state = db.borrow();
+            for cluster_id in state.get_own_clusters() {
+                let Some(cluster) = state.clusters().get_by(cluster_id) else {
+                    continue;
+                };
+                let id = get_committee_id(&cluster);
+                let subnet = id % U256::from(subnet_count);
```
diegomrsantos: I'm unsure what `U256` means here, but do we need it? There are only 128 subnets.

anchor/subnet_tracker/src/lib.rs
```diff
@@ -0,0 +1,185 @@
...
+    Old,
+    New,
+    Still,
+}
+
+async fn subnet_tracker(
+    tx: mpsc::Sender<SubnetEvent>,
+    mut db: watch::Receiver<NetworkState>,
+    subnet_count: usize,
+) {
```
diegomrsantos, line 63: Would something like this work? Then we'd need only Join and Leave.

```rust
/// The main background task:
/// - Gathers the current subnets from `NetworkState`.
/// - Compares them to the previously-seen subnets.
/// - Emits `Join` events for newly-added subnets and `Leave` events for removed subnets.
async fn run_subnet_tracker_loop(
    tx: mpsc::Sender<SubnetEvent>,
    mut db: watch::Receiver<NetworkState>,
    subnet_count: usize,
) {
    // `previous_subnets` tracks which subnets were joined in the last iteration.
    let mut previous_subnets = HashSet::new();

    loop {
        // Acquire the current snapshot of the database state (this is synchronous).
        let state = db.borrow();

        // Build the `current_subnets` set by examining the clusters we own.
        let mut current_subnets = HashSet::new();
        for cluster_id in state.get_own_clusters() {
            if let Some(cluster) = state.clusters().get_by(cluster_id) {
                // Derive a numeric "committee ID" and convert to an index in [0..subnet_count].
                let id = get_committee_id(cluster);
                let index = (id % U128::from(subnet_count))
                    .try_into()
                    .expect("modulo must be < subnet_count");
                current_subnets.insert(index);
            }
        }

        // For every subnet that was previously joined but is no longer in `current_subnets`,
        // send a `Leave` event.
        for subnet in previous_subnets.difference(&current_subnets) {
            debug!(?subnet, "send leave");
            if tx
                .send(SubnetEvent::Leave(SubnetId(*subnet as u64)))
                .await
                .is_err()
            {
                warn!("Network no longer listening for subnets");
                return;
            }
        }

        // For every subnet that was not previously joined but is now in `current_subnets`,
        // send a `Join` event.
        for subnet in current_subnets.difference(&previous_subnets) {
            debug!(?subnet, "send join");
            if tx
                .send(SubnetEvent::Join(SubnetId(*subnet as u64)))
                .await
                .is_err()
            {
                warn!("Network no longer listening for subnets");
                return;
            }
        }

        // Update `previous_subnets` to reflect the current snapshot for the next iteration.
        previous_subnets = current_subnets;

        // Wait for the watch channel to signal a changed value before re-running the loop.
        if db.changed().await.is_err() {
            warn!("Database no longer provides updates");
            return;
        }
    }
}
```

anchor/subnet_tracker/src/lib.rs
dknopik: The committee id is an U256 (i.e. an unsigned integer with 256 bits), as it is a keccak256 hash reinterpreted as integer. We need to cast the subnet count here as U256 only implements the modulo operation for other U256.

We could also only use the least significant bits of the hash instead, but then the `subnet_count` is required to be a power of two for the result to be correct, which is also kind of ugly to assert there.

anchor/subnet_tracker/src/lib.rs
dknopik: Some arguments against this:

- Holding a sender unnecessarily will prevent us from detecting that the actual `subnet_tracker` has exited.
- All the data in the struct (except for the receiver) is unnecessary once the `subnet_tracker` has started, and then still carried around uselessly.
- I argue that this actually makes testing harder and increases coupling. The `test_tracker` at the bottom of the file shows nicely that it is trivial with the current setup to create a mock that does not require a `db: watch::Receiver<NetworkState>`. 

Some alternatives:
- Rename `SubnetTracker` to `SubnetTrackerHandle` or `SubnetTrackingReceiver` or something like that to make it a bit clearer
- No struct at all, just pass the `mpsc::Receiver<SubnetEvent>` around.

diegomrsantos: I agree, SubnetTracker isn't doing much currently, getting rid of it would probably make it simpler than my suggestion.

anchor/subnet_tracker/src/lib.rs
diegomrsantos: Makes sense, thanks for clarifying.

anchor/client/src/lib.rs
```diff
@@ -340,8 +344,9 @@ impl Client {
         );
 
         // Wait until we have an operator id and historical sync is done
-        let operator_id =
-            wait_for_operator_id_and_sync(&database, historic_finished_rx, &spec).await;
+        let operator_id = wait_for_operator_id_and_sync(&database, historic_finished_rx, &spec)
```
diegomrsantos: Is there a timeout for this?

dknopik: There is no timeout for this as we can only proceed meaningfully as soon as we have a operator id, and need to wait for historic sync to finish in order to not accidentally perform duties for validators that we are not meant to anymore. So a timeout would be incorrect here, as we can not continue and also don't want to crash.

diegomrsantos: But what happens if there's any error and this never finishes?

diegomrsantos: Also, I just realized that the function name says `sync` but there's an `await`. Ah, this isn't a sync/async, it's a historical sync. Maybe a bit more context in the comment could be useful.

dknopik: If the database or sync task crash irrecoverably, they exit, thereby closing the channel automatically. we then get an error here because the channel is now closed, and thus exit here ourselves because we can not continue.

If the database or sync task might recover, we also can just wait here for them to recover.

dknopik: Keep in mind that the operator might not be on chain yet. The user might launch anchor for the first time, creating a private key by doing so. Then the user might go on to immediately/after lunch break/after his weekend register the operator on chain. We want to start regular work as soon as he does so.

diegomrsantos: In `wait_for_operator_id_and_sync` there seems to exist an infinite loop that runs till the `ID`is inserted by some other code, is that correct? If so, what if the `ID` is never inserted by it, is it possible?

dknopik: Hmmmm. Yeah, while writing a response I realized that you are right. if live sync exits we do not catch it. 

I am not sure if this is a problem in practice, as live sync should never exit and retry instead, but we need to figure out how to tear down the process if it does anyway due to a bug

dknopik: Created https://github.com/sigp/anchor/issues/119 for this and related topics

anchor/network/src/network.rs
dknopik: Yeah, I think we should check the `peer_store` before trying to discover. Will add this in my simplified peer manager PR.

anchor/network/src/network.rs
```diff
@@ -25,16 +25,23 @@ use crate::Config;
 use crate::types::ssv_message::SignedSSVMessage;
 use lighthouse_network::EnrExt;
 use ssz::Decode;
+use subnet_tracker::{SubnetEvent, SubnetId};
+use tokio::sync::mpsc;
 
 pub struct Network {
     swarm: Swarm<AnchorBehaviour>,
+    subnet_tracker: mpsc::Receiver<SubnetEvent>,
```
diegomrsantos: how about sth like `subnet_event_receiver`?

diegomrsantos (dismissed):
Looks amazing!

diegomrsantos (approved):

PR #117, opened 2025-01-28
Lock database state globally and implement state watching

dknopik:
## Issue Addressed

Before this PR, the in-memory database was a bit spooky: As operations (such as adding/removing validators) touch several `DashMap`s without locking the whole state, readers might read inconsistent data across these maps. Reasoning about these maps and safe orders in accessing/updating them is kind of hard, especially as the client is still evolving rapidly.

Furthermore, other parts of the client are interested in DB updates, but currently have no way of receiving them except polling regularly.  

## Proposed Changes

Instead of relying on multiple `DashMap`s and `Atomic*`, we now lock the whole `NetworkState` until we are fully done modifying it. We do this by using `tokio::sync::watch`, which is basically a `Arc<RwLock<T>>` with some extra infrastructure for task notification if the value changed. We use this notification feature in `validator_store` to only load new validators if the state changed, and will use it in `network` to figure out the subnets we want to subscribe to.


anchor/database/src/lib.rs
```diff
@@ -100,23 +100,39 @@ impl NetworkDatabase {
...
+    /// the last call to this function.
     pub fn processed_block(&self, block_number: u64) -> Result<(), DatabaseError> {
         let conn = self.connection()?;
         conn.prepare_cached(SQL[&SqlStatement::UpdateBlockNumber])?
             .execute(params![block_number])?;
-        self.state
-            .single_state
-            .last_processed_block
-            .store(block_number, Ordering::Relaxed);
+        self.state.send_if_modified(|state| {
```
Zacholme7: Do we need this check to make sure it is modified? This function is only ever called if a new block has been processed so the state should always be modified.

anchor/database/src/lib.rs
```diff
@@ -161,4 +177,15 @@ impl NetworkDatabase {
     fn connection(&self) -> Result<PoolConn, DatabaseError> {
         Ok(self.conn_pool.get()?)
     }
+
+    /// Apply a modification to the state, and mark the state as modified
+    /// As soon as the end of block processing is signaled via [`processed_block`],
+    /// the notification to all watchers is triggered (instead of here)
+    fn modify_state(&self, f: impl FnOnce(&mut NetworkState)) {
```
Zacholme7, line 179: So this acts as the lock around the state? Or what is giving us the ability to perform mutable operations with shared references?

anchor/database/src/validator_operations.rs
```diff
@@ -10,21 +10,24 @@ impl NetworkDatabase {
...
-        // Make sure the cluster exists by getting the in memory entry
-        if let Some(mut cluster) = self.clusters().get_by(&owner) {
-            // Update the database
-            let conn = self.connection()?;
-            conn.prepare_cached(SQL[&SqlStatement::UpdateFeeRecipient])?
-                .execute(params![
-                    fee_recipient.to_string(), // New fee recipient address for entire cluster
-                    owner.to_string()          // Owner of the cluster
-                ])?;
+        // Update the database
```
Zacholme7, line 13: Do we want to make sure the cluster exists before trying to update in the database?

anchor/database/src/validator_operations.rs
```diff
@@ -34,20 +37,28 @@ impl NetworkDatabase {
...
-        // Make sure this validator exists by getting the in memory entry
-        if let Some(mut validator) = self.metadata().get_by(validator_pubkey) {
-            // Update the database
-            let conn = self.connection()?;
-            conn.prepare_cached(SQL[&SqlStatement::SetGraffiti])?
-                .execute(params![
-                    graffiti.0.as_slice(),        // New graffiti
-                    validator_pubkey.to_string()  // The public key of the validator
-                ])?;
+        // Update the database
```
Zacholme7, line 40: Same as above, should we check if it exists or just handle the error

Zacholme7:
small comments/questions, otherwise lgtm gj

anchor/database/src/lib.rs
dknopik: good point, will adjust!

anchor/database/src/lib.rs
dknopik: [`send_if_modified`](https://docs.rs/tokio/latest/tokio/sync/watch/struct.Sender.html#method.send_if_modified) and [`send_modify`](https://docs.rs/tokio/latest/tokio/sync/watch/struct.Sender.html#method.send_modify) of `tokio::sync::watch::Sender` lock the underlying `RwLock` mutably via [`write`](https://docs.rs/tokio/latest/tokio/sync/struct.RwLock.html#method.write) - during this time no reading locks can be acquired, ensuring that no one can read while we are in the passed closure.

anchor/database/src/validator_operations.rs
dknopik: In both cases, nothing happens. So I guess the question is, what in total saves more computation? Omitting the check, or avoiding the useless sqlite call and write lock? I don't really feel that strongly in any direction, it was just convenient to remove it while refactoring x) 

I'll put it back :)

Zacholme7: true really doesn't matter, I guess it just saves a warn log haha

dknopik: Would it even error? I assumed it would just do an update that touches 0 rows. or does that cause sqlite to return an error?

Zacholme7: Oh hm yea I think you're right.

dknopik: Ooops: that introduced a deadlock. https://github.com/sigp/anchor/actions/runs/13016197965/job/36305885019?pr=117#step:4:91

Can be avoided if we explicitly drop the borrowed state before mutating, but as discussed - both variants are equivalent, so I changed it back to my original proposal as I find it cleaner :)

Zacholme7 (approved):

PR #118, opened 2025-01-28
use PublicKeyBytes instead of PublicKey

dknopik:
`PublicKey` is the decompressed version of `PublicKeyBytes`, which is larger in memory and kind of costly to convert. We only need the decompressed version while verifying the signature during sync, so this PR switches to `PublicKeyBytes` everywhere else.


Zacholme7:
lgtm

jking-aus (approved):
lgtm

PR #121, opened 2025-02-03
add `ssv_network_config` for configurable networks

dknopik:
Analogous to Lighthouses `eth2_network_config`, implement `ssv_network_config`, which builds upon it. It has support for the hardcoded "holesky" and "mainnet" networks, and can load from extended LH testnet dirs via the `--testnet-dir` CLI option.

It includes some boot ENRs, which were taken from the go-ssv implementation.

Sync code is adjusted to use this config.


anchor/client/src/config.rs
```diff
@@ -59,12 +59,13 @@ impl Config {
     /// Build a new configuration from defaults.
     ///
     /// eth2_network: We pass this because it would be expensive to uselessly get a default eagerly.
```
Zacholme7: change comment to reflect param

anchor/eth/src/sync.rs
```diff
@@ -154,10 +131,8 @@ impl SsvEventSyncer {
         info!("Starting SSV event sync");
 
         // Get network specific contract information
-        let (contract_address, deployment_block) = match self.network {
-            Network::Mainnet => (*MAINNET_DEPLOYMENT_ADDRESS, MAINNET_DEPLOYMENT_BLOCK),
-            Network::Holesky => (*HOLESKY_DEPLOYMENT_ADDRESS, HOLESKY_DEPLOYMENT_BLOCK),
-        };
+        let contract_address = self.network.ssv_contract;
+        let deployment_block = self.network.ssv_contract_block;
```
Zacholme7, line 129: I think that these are better as constants in the code. Getting them from the config implies that they are configurable but they are not and then we can just match on the network which is configurable.

anchor/eth/src/sync.rs
```diff
@@ -39,23 +40,6 @@ static SSV_EVENTS: LazyLock<Vec<&str>> = LazyLock::new(|| {
...
-    LazyLock::new(|| address!("38A4794cCEd47d3baf7370CcC43B560D3a1beEFA"));
-
-/// Contract deployment block on Ethereum Mainnet
-/// Mainnet: https://etherscan.io/tx/0x4a11a560d3c2f693e96f98abb1feb447646b01b36203ecab0a96a1cf45fd650b
-const MAINNET_DEPLOYMENT_BLOCK: u64 = 17507487;
-
-/// Contract deployment block on the Holesky Network
-/// Holesky: https://holesky.etherscan.io/tx/0x998c38ff37b47e69e23c21a8079168b7e0e0ade7244781587b00be3f08a725c6
-const HOLESKY_DEPLOYMENT_BLOCK: u64 = 181612;
-
```
Zacholme7, line 56: Ties into comment above

anchor/common/ssv_network_config/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
```
Zacholme7, line 1: Some quick thoughts on how we want to handle this.

1) Personally, I am partial towards toml configs. I think [reth](https://reth.rs/run/config.html?highlight=enr#configuring-reth) does a really good job with theirs. If we are all good with yaml im happy with that. 
2) Ties into comments above about block and address deployment configs. 
3) Regardless of which approach we go with, I think having a central config file is the best rather than having a bunch people need to modify. 
4) So this is built in configs and then we will have a different config that people can modify? Or what is your idea behind this. I think that anything that is built in should be declared as constants, and anything that an end user can customize should be in a config. IMO the less a user can modify/mess with the better, to an extent, as they cant break as much.

Zacholme7 (changes requested):
Left a few thoughts, gj otherwise

anchor/eth/src/sync.rs
dknopik: They are configurable, and should be. For example local devnets, or other user-defined chains. Or even just a different set of contracts on holesky (or even mainnet). For those, we need to provide a way to specify the contract address externally.

anchor/common/ssv_network_config/Cargo.toml
dknopik: 1. I totally understand, usually I would also prefer toml. The reason why I chose yaml here is to keep close to the current CL testnet dir convention, which uses yaml. This format is actually supported by most CL clients: via `--testnet-dir` in Lighthouse.
2. addressed in other comment.
3. The main use case here development. For local devnets, people don't create these files manually, but use something like https://github.com/protolambda/eth2-testnet-genesis. Again, I chose to separate the files as this matches the existing approach. Regular users who want to use Anchor on holesky or mainnet will never need to edit these files or even know about their existence - they configure Anchor via the CLI, just like Lighthouse. My idea here was to extend these testnet dirs with the stuff we need for anchor - that way we can generate a single testnet dir and use it for all CL clients AND Anchor.
4. The built-in config will be baked into the binary, so the user can't mess with them. Again, this matches the approach taken by Lighthouse: https://github.com/sigp/lighthouse/tree/70194dfc6a3f4d10c9059610f889ff5a4e863a6a/common/eth2_network_config. That being said, I am open to just use `const`s for the defaults here, as we do not need large binary files like the CL does (the genesis state). But an advantage to the current structure is that the layout of the built in config folders matches the layout external testnet dirs would need, so it would be intuitive for devs to use their own testnet dirs: just copy the files from the repo and adjust them, or implement something that properly generates them. Let me know what you think!

Thanks for the review! :)

Zacholme7: Okay thanks that makes a lot more sense now. Im good with going with this approach.

Zacholme7 (approved):

Issue #106, opened 2025-01-15
QBFT round change justifications


jking-aus:
## Description

Our QBFT mechanism does not currently implement round change justifications. I believe we do round changes mostly correct with pr and pv but do not provide or verify the justifications. Early discussion with SSV team was that justifications are piggybacked on round change messages. 

It's worth reading the QBFT spec located here https://github.com/ssvlabs/ssv/blob/main/ibft/IBFT.md but snip below

"The content of a ROUND-CHANGE message sent by a node (that’s current round is r) is <ROUND-CHANGE, λ, r+1, pr, pv>, where pr indicates the highest round in which a quorum of PREPARE messages has been received, and pv indicates the value which was included in those PREPARE messages."

What they dont say here is we also need to piggyback the round change justification - as per the original paper here https://arxiv.org/pdf/2002.03613 (Section 4.4).

Here is, I think, the relevant section of the SSV go code:
https://github.com/ssvlabs/ssv/blob/main/protocol/v2/qbft/instance/round_change.go#L430

## Present Behaviour

Anchor sends round change messages and identifies pr and pv - it does not piggyback justifications. 

## Expected Behaviour

We should consider a quorum of round change messages to be justified if: 
- All of the messages in the quorum have prepared round and prepared value of none.
- Justification has a quorum of valid prepare messages provided with greatest pr. 

The leader must carry out verification of the justification message and then send this with the proposal. Other committee members must validate the justification before sending prepare. 

## Steps to resolve

The consensus logic needs to be implemented to achieve this but we will also have to make sure we're doing it the same way as go-ssv to prevent interop or logic issues.

Issue #98, opened 2025-01-14
Qbft Manager Unit tests


Zacholme7:
## Description

General functionality test for the QBFT manager and make sure that ROUNDCHANGE works appropriately.

PR #122, opened 2025-02-03
Qbft justifications

Zacholme7:
## Issue Addressed

#106 

## Proposed Changes

This is an initial implementation of the round change and prepare justifications for proper qbft behavior. The actual justification implementation is pretty small, but a lot of message validation checks have been added in from the spec. In the future, we can more rigorously sift through the checks to only keep what we need. 

## Additional Info

We do not have the the infra in place to properly test this. I have a follow up PR in progress that adds testing functionality (#98). The plan is to get this in and then follow up with another PR where all testing, bugfixes, and general productionizing occur so that the scope of this PR is limited. 

Update: There are a few bugs in this code that I have fixed in the qbft testing pr. Can also make the updates here if needed.


Zacholme7:
Do we know how to fix this failing CI?

dknopik:
> Do we know how to fix this failing CI?

I investigated a bit, see #123

anchor/common/qbft/src/lib.rs
```diff
@@ -367,27 +383,159 @@ where
...
+
+        // Make sure we have a quorum of round change messages
+        if msg.qbft_message.round_change_justification.len() < self.config.quorum_size() {
+            warn!("Did not recieve a quorum of round change messages");
+            return false;
+        }
+
+        // There was a quorum of round change justifications. We need to go though and verify each
+        // one. Each will be a SignedSSVMessage
+        for signed_round_change in &msg.qbft_message.round_change_justification {
```
dknopik, line 448: Are we checking somewhere whether the messages are actually originating from distinct operators?

anchor/common/qbft/src/lib.rs
```diff
@@ -367,27 +383,159 @@ where
...
+                match QbftMessage::from_ssz_bytes(signed_round_change.ssv_message().data()) {
+                    Ok(data) => data,
+                    Err(_) => return false,
+                };
+
+            // Make sure this is actually a round change message
+            if !matches!(round_change.qbft_message_type, QbftMessageType::RoundChange) {
+                warn!(message_type = ?round_change.qbft_message_type, "Message is not a ROUNDCHANGE message");
+                return false;
+            }
```
dknopik, line 461: It might be nice to do these well-formedness checks closer to the network layer, as we actually do not require any internal QBFT state here. Same for signature verification. Then we can decide quicker if we want to further propagate messages. 

But right now, the network code is not ready yet to actually implement this, so this should remain here for now. Just a remark.

anchor/common/qbft/src/lib.rs
```diff
@@ -529,34 +721,87 @@ where
...
+    //   round (u64) -> the message round
+    //   root (D::Hash) -> the hash of the proposed data, if necessary
+    //   full_data (Vec<u8>) -> serialized data that we are coming to conseusn on
+    // )
+    fn get_message_data(
+        &self,
+        msg_type: &QbftMessageType,
+        data_hash: D::Hash,
+    ) -> (u64, u64, D::Hash, Vec<u8>) {
+        // Get the fulldata for the data_hash
```
dknopik: At this point, we should introduce a struct for the return type instead IMHO.

anchor/common/qbft/src/lib.rs
```diff
@@ -565,26 +810,143 @@ where
...
         UnsignedSSVMessage {
             ssv_message,
             full_data,
         }
     }
 
+    // Get all of the round change jusitifcation messages
+    fn get_round_change_justifications(&self) -> Vec<SignedSSVMessage> {
+        // Make sure we are past the first round
+        if self.current_round > Round::default() {
```
dknopik: Nitpick and matter of taste: In cases where we have a check applying for the rest of the function, I think it's nicer to flip the condition and return early instead:

```rust
if self.current_round <= Round::default() {
    return vec![];
}
```

anchor/common/qbft/src/lib.rs
```diff
@@ -529,34 +721,87 @@ where
...
+        // Get the fulldata for the data_hash
+        let full_data = if let Some(data) = self.data.get(&data_hash) {
+            data.as_ssz_bytes()
+        } else {
+            vec![]
+        };
+
+        match msg_type {
+            QbftMessageType::Proposal | QbftMessageType::Prepare | QbftMessageType::Commit => {
+                // todo!() for prepare, does it leave fulldata blank??
```
dknopik: I was actually under the impression that it ONLY sets full data during propose :sweat_smile:

dknopik (commented):
Added some first thoughts.

> Update: There are a few bugs in this code that I have fixed in the qbft testing pr. Can also make the updates here if needed.

If it's not a lot of effort, it would be nice if you merge the fixes into this one.

There are a lot of TODOs in the comment, and one `todo!() ` invocation. Do you plan to address them, and do you want to discuss any specific ones?

anchor/common/qbft/src/lib.rs
Zacholme7: agreed, will change

anchor/common/qbft/src/lib.rs
Zacholme7: Yep you're right. Just referenced spec and it is only set during proposal and and round change.

anchor/common/qbft/src/lib.rs
Zacholme7: Yes the message container will ignore messages from an operator if it has already added one for that id + round. 

Also an argument for this check being pushed earlier into verification

anchor/common/qbft/src/lib.rs
```diff
@@ -488,28 +702,30 @@ where
...
+        } else {
+            // 2. If we receive f+1 round change messages, we need to send our own round-change message
+            let num_messages_for_round = self.round_change_container.num_messages_for_round(round);
+            if num_messages_for_round > self.config.get_f()
+                && !(matches!(self.state, InstanceState::SentRoundChange))
+            {
+                // Set the state so SendRoundChange so we include Round + 1 in message
+                self.state = InstanceState::SentRoundChange;
+
+                self.send_round_change(Hash256::default());
```
dknopik, line 723: Is `Hash256::default()` correct? I think it should be the hash of our start value, but I'm not 100% sure.

anchor/common/qbft/src/lib.rs
```diff
@@ -529,34 +745,67 @@ where
...
-        // Start a new round
-        self.current_round.set(next_round);
-        // Check if we have a prepared value, if so we want to send a round change proposing the
-        // value. Else, send a blank hash
-        let hash = self.last_prepared_value.unwrap_or_default();
-        self.send_round_change(hash);
+        // Set the state so SendRoundChange so we include Round + 1 in message
+        self.state = InstanceState::SentRoundChange;
+
+        self.send_round_change(Hash256::default());
```
dknopik, line 746: Same here.

anchor/common/qbft/src/lib.rs
```diff
@@ -333,23 +384,26 @@ where
...
-        let data_hash = wrapped_msg.signed_message.hash_fulldata();
-        if data_hash != wrapped_msg.qbft_message.root {
+        // Verify that the data root matches what was in the message
+        let data_hash = data.hash();
+        if data.hash() != wrapped_msg.qbft_message.root {
             warn!(from = ?operator_id, self=?self.config.operator_id(), "Data roots do not match");
             return;
         }
 
+        self.data.insert(wrapped_msg.qbft_message.root, data);
```
dknopik: We also store the data a few lines below - we should remove one of those.

anchor/common/qbft/src/lib.rs
Zacholme7: Referenced the spec. On round change if there is not a last prepared value/round the root and the fulldata are left empty.  So we check in the `get_message_data` if we have a value prepared we use that, otherwise we use `fulldata = vec![]` and `root = Hash256::default()`

Here is relevant spec code. [getting round change data](https://github.com/ssvlabs/ssv-spec/blob/0ee274958cbfdcec94bef9a0f0f969d0c02d52cc/qbft/round_change.go#L369) and [creating round change](https://github.com/ssvlabs/ssv-spec/blob/0ee274958cbfdcec94bef9a0f0f969d0c02d52cc/qbft/round_change.go#L401)

dknopik:
added #127 to track getting rid of some of these `expect`s

anchor/common/qbft/src/lib.rs
dknopik: Aaah, I see.

But then, why not pass the `last_prepared_value` into `new_unsigned_message` instead of the special case within `get_message_data`?

Zacholme7: Im not sure I understand. We will still have to check if `last_prepared_value` and `last_prepared_round` are Some and then update the `fulldata` and `data_round` accordingly, otherwise we just use the defaulted hash and empty vec.

dknopik: Ah, ok. I think I get it now. Thanks for your patience :sweat_smile:

dknopik (approved):
LGTM, good job!

There is the commit aggregation and some refactoring left to do as discussed, but let's go ahead with this to enable testing. Thanks!

Issue #123, opened 2025-02-04
Build script of dependency `eth2_network_config` is not run if building from cache


dknopik:
Affected run with debug output: https://github.com/sigp/anchor/actions/runs/13116079748/job/36637757405?pr=122

I think what's happening here is that the caching does not include the `/home/runner/.cargo/git` directory, so it is created freshly. Then, cargo does not run the `build.rs` of `eth2_network_config` for some reason, so the `genesis.ssz` is not extracted from the freshly cloned repo.

We will need to either cache that directory, or force cargo to run the `build.rs`.

As a workaround, I'll open a PR to disable the cache.

dknopik:
@magick93, replying here to your comment on the PR:

>@dknopik - could this solve the issue? https://stackoverflow.com/a/53728256/265119

These solutions look good. Most would require changes in Lighthouse, and we want to avoid unnecessary rebuilds in other cases. There is also a nice suggestion in there to `touch` the `build.rs` file, but the problem is that the exact location of the `build.rs` file will vary, as Cargo stores them on a path it somehow generates. We've got to experiment a bit to find a good solution I think.

PR #124, opened 2025-02-04
disable CI cache due to build issues on repeated builds

dknopik:
## Issue Addressed

- Workaround for #123.

## Proposed Changes

Disable cache for the two affected jobs.


magick93:
LG. 

We'll need to monitor this to see if the builds take substantially longer.

magick93 (approved):

magick93:
@dknopik - could this solve the issue? https://stackoverflow.com/a/53728256/265119

Issue #116, opened 2025-01-27
Implement ssv handshake


diegomrsantos:
From discussion on discord:

We need to look here https://github.com/ssvlabs/ssv/tree/main/network/peers/connections to see how handshake is implemented.
There's some exchange of node information on connections. The NodeInfo which is made of mostly a version and subnets you subscribe to.

diegomrsantos:
See https://github.com/ssvlabs/ssv-spec/blob/770e204aeabfd2e37c4d3c2b652a43efec69aead/p2p/SPEC.md#handshake-protocol

PR #125, opened 2025-02-04
Implement SSV Handshake Protocol

diegomrsantos:
## Issue Addressed

Implements https://github.com/sigp/anchor/issues/116

## Proposed Changes

- Dependencies:

  - Updated dependencies in Cargo.toml including async-trait, quick-protobuf, serde_json, and thiserror.

- Handshake Protocol:

  - Added a new module handshake with multiple new files including README.md, codec.rs, envelope.proto, envelope.rs, mod.rs, and node_info.rs.

- Network Behaviour:

  - Enhanced behaviour.rs to include handshake::Behaviour and added a new field handshake to AnchorBehaviour.

- Network Implementation:

  - Enhanced network.rs:
  - Updated to use the Handshake Behaviour
  - Created DefaultNodeInfoProvider to provide NodeInfo.

## Additional Info

There's a spec in README.md file


anchor/network/src/network.rs
```diff
@@ -217,11 +242,35 @@ async fn build_anchor_behaviour(
         discovery
     };
 
+    let peer_manager = PeerManager::new(
+        lighthouse_network::peer_manager::config::Config::default(),
+        vec![],
+        false,
+    );
+
+    let domain = format!("0x{}", hex::encode(vec![0x0, 0x0, 0x5, 0x2]));
```
diegomrsantos: This is based on https://github.com/ssvlabs/ssv/blob/main/network/p2p/p2p_setup.go#L181 and https://github.com/ssvlabs/ssv/blob/main/networkconfig/holesky.go#L14

anchor/network/Cargo.toml
```diff
@@ -23,15 +26,29 @@ libp2p = { version = "0.54", default-features = false, features = [
...
 ssz_types = "0.10"
+strum = "0.24.1"
 task_executor = { workspace = true }
 tokio = { workspace = true }
 tracing = { workspace = true }
 types = { workspace = true }
 version = { workspace = true }
+serde_json = "1.0.137"
+thiserror = "1.0.69"
+prost = "0.13.4"
```
jxs: we use `quick-protobuf` in `libp2p` which is also available in the dependency chain, so you shouldn't need to depend on `prost` which is also helpful as you don't need to specifiy `prost-build` in the build dependencies

anchor/network/src/behaviour.rs
```diff
@@ -12,4 +16,24 @@ pub struct AnchorBehaviour {
...
     /// Discv5 Discovery protocol.
     pub discovery: Discovery,
+    /// The peer manager that keeps track of peer's reputation and status.
+    pub peer_manager: PeerManager,
+
+    pub handshake: HandshakeBehaviour,
+}
+
+#[derive(Default)]
+struct DummyPeerInfoStore {}
```
jxs: this and the `PeerInfoStore` trait are not required right? As I see them commented in the `HandshakeBehaviour` impl

anchor/network/src/handshake/behaviour.rs
```diff
@@ -0,0 +1,313 @@
...
+use std::time::Instant;
+use tracing::debug;
+use crate::handshake::codec::EnvelopeCodec;
+use crate::handshake::record::envelope::Envelope;
+use crate::handshake::record::signing::{consume_envelope, seal_record};
+use crate::handshake::types::NodeInfo;
+
+/// Event emitted on handshake completion or failure.
+#[derive(Debug)]
+pub enum HandshakeEvent {
```
jxs: nit:
```suggestion
pub enum Event {
```
the module already gives out the namespace, no need for redundancy

anchor/network/src/handshake/behaviour.rs
```diff
@@ -0,0 +1,313 @@
...
+
+impl Subnets {
+    pub fn from_str(s: &str) -> Result<Self, Box<dyn Error>> {
+        // Parse subnets from string
+        Ok(Subnets)
+    }
+}
+
+/// Network behaviour handling the handshake protocol.
+pub struct HandshakeBehaviour {
```
jxs: nit:
```suggestion
pub struct Behaviour {
```

anchor/network/src/handshake/behaviour.rs
```diff
@@ -0,0 +1,313 @@
...
+    /// Request-response behaviour for the handshake protocol.
+    behaviour: Behaviour<EnvelopeCodec>,
+    /// Pending outgoing handshake requests.
+    pending_handshakes: HashMap<OutboundRequestId, PeerId>,
```
jxs: do we need to keep track of the established connections which haven't completed the handshake yet?

anchor/network/src/handshake/behaviour.rs
```diff
@@ -0,0 +1,313 @@
...
+    /// Keypair for signing envelopes.
+    keypair: Keypair,
```
jxs: I feel this should belong in the `Codec` to validate the handshake's there, if we cannot sign the envelope of an handshake then it is an invalid handshake and we can discard it right?

anchor/network/src/handshake/behaviour.rs
jxs: skimming through the `NetworkBehaviour` implementation, and without knowing much about the requirements I wonder if we can suffice our needs by just implementing the `Codec` trait and not needing to have `request-response` as a sub behaviour

anchor/network/src/handshake/error.rs
```diff
@@ -0,0 +1,25 @@
+use thiserror::Error;
+
+#[derive(Error, Debug)]
+pub enum HandshakeError {
```
jxs: nit:
```suggestion
pub enum Error {
```

jxs: do we need `thiserror::Error` for our `Error`s? 
We are probably fine having an `Enum` and impl'ing `Display` and `Error` for it like `request_response::OutboundFailure` does it for example.
Nonetheless since it's probably already in the dependency chain feel free to ignore this

anchor/network/src/handshake/codec.rs
```diff
@@ -0,0 +1,94 @@
...
+use async_trait::async_trait;
+use libp2p::StreamProtocol;
+use prost::bytes::BytesMut;
+use prost::encoding::{decode_varint, encode_varint, encoded_len_varint};
+use prost::Message;
+use tracing::debug;
+
+/// A `Codec` that reads/writes an **`Envelope`**
+#[derive(Clone, Debug, Default)]
+pub struct EnvelopeCodec;
```
jxs: nit:
```suggestion
pub struct Codec;
```

anchor/network/src/handshake/record/record.rs
```diff
@@ -0,0 +1,13 @@
+use std::error::Error;
+
+/// The `Record` trait parallels the idea in Go that each record knows how to:
+///   - Provide a domain (for signing separation)
+///   - Provide a "codec" (payload type)
+///   - Marshal to bytes, unmarshal from bytes
+pub trait Record {
```
jxs: this seems necessary right? As we are only using `NodeInfo`

jxs (commented):
Hi Diego, thanks for starting this.
Left some comments

anchor/network/src/handshake/behaviour.rs
diegomrsantos: What would this approach look like?

anchor/network/src/handshake/behaviour.rs
diegomrsantos: I'm not sure yet. The idea was to know if a response is for a previous request.

anchor/network/src/handshake/behaviour.rs
jxs: using the `request-response` `Behaviour` and then doing the validation envelope signing validation logic in the `Codec` implementation.
Does that make sense Diego?

diegomrsantos: Yes, having the validation in the codec is the way to go. About only one behavior, I'll need to think more about it.

anchor/network/src/handshake/behaviour.rs
diegomrsantos: Yes, I think it makes sense.

anchor/network/src/handshake/error.rs
diegomrsantos: This needs to be revisited, but It'll probably be removed.

anchor/network/src/handshake/record/record.rs
diegomrsantos: I'll probably remove it.

anchor/network/src/handshake/behaviour.rs
diegomrsantos: Actually, this isn't used in the validation, only when we seal the NodeInfo.

anchor/network/src/handshake/behaviour.rs
diegomrsantos: done

anchor/network/src/handshake/behaviour.rs
diegomrsantos: done

anchor/network/src/handshake/codec.rs
diegomrsantos: done

anchor/network/src/handshake/error.rs
diegomrsantos: Another advantage is the `#[from]` see https://github.com/sigp/anchor/pull/125/commits/dc79ba7fae2a0069f89211c7763b683aed927021

anchor/network/src/handshake/behaviour.rs
diegomrsantos: Removed as this is already handled by libp2p

anchor/network/src/handshake/record/record.rs
diegomrsantos: removed

anchor/network/Cargo.toml
```diff
@@ -22,15 +23,18 @@ libp2p = { version = "0.54", default-features = false, features = [
...
+  "request-response",
 ] }
 lighthouse_network = { workspace = true }
+quick-protobuf = "0.8.1"
 serde = { workspace = true }
+serde_json = "1.0.137"
 ssz_types = "0.10"
 subnet_tracker = { workspace = true }
 task_executor = { workspace = true }
+thiserror = "1.0.69"
```
dknopik, line 37: Should use workspace dependencies

anchor/network/src/behaviour.rs
```diff
@@ -12,4 +13,6 @@ pub struct AnchorBehaviour {
     pub gossipsub: gossipsub::Behaviour,
     /// Discv5 Discovery protocol.
     pub discovery: Discovery,
+
+    pub handshake: Behaviour,
```
dknopik: would be nicer to qualify which kind of Behaviour this is by referencing the module here: `handshake::Behaviour`, see the other fields above.

anchor/network/src/handshake/node_info.rs
```diff
@@ -0,0 +1,209 @@
...
+    /// Serialize `NodeInfo` to JSON bytes.
+    fn marshal(&self) -> Result<Vec<u8>, Error> {
+        let mut entries = vec![
+            "".to_string(),          // formerly forkVersion, now deprecated
+            self.network_id.clone(), // network id
+        ];
+
+        if let Some(meta) = &self.metadata {
+            let raw_meta = serde_json::to_vec(meta)?;
+            entries.push(String::from_utf8(raw_meta)?);
```
dknopik: We can use `serde_json::to_string` instead

anchor/network/src/network.rs
dknopik: This needs to be configurable per network - so it should be integrated into the `ssv_network_config` stuff

anchor/network/src/network.rs
```diff
@@ -327,6 +365,25 @@ fn build_swarm(
...
+    }
+}
+
+impl NodeInfoProvider for DefaultNodeInfoProvider {
+    fn get_node_info(&self) -> NodeInfo {
+        // In a real implementation, consider handling lock poisoning.
+        self.node_info.lock().unwrap().clone()
+    }
+}
+
```
dknopik: I guess this is a todo for another PR?

anchor/network/Cargo.toml
diegomrsantos: It's not defined in the root. Should it be defined there and inherited even if it's used only here for now?

anchor/network/src/handshake/node_info.rs
diegomrsantos: nice, thanks!

anchor/network/src/handshake/behaviour.rs
diegomrsantos: I think it's nice to keep the behavior and centralize everything related to the handshake there, exposing only the events to the main behavior.

anchor/network/src/handshake/error.rs
diegomrsantos: Done

anchor/network/Cargo.toml
diegomrsantos: I changed to `quick-protobuf`. But I found `prost` much easier to use and straightforward. It also seems to be more actively maintained.

anchor/network/Cargo.toml
dknopik: hmm, good question. We should align with the others on that

jxs: rule of thumb I use: if it's only used by a subcrate import it there, when it becomes used by two or more, move it to workspace

anchor/network/src/handshake/README.md
```diff
@@ -0,0 +1,272 @@
+# SSV NodeInfo Handshake Protocol Specification
```
jxs: this is awesome work Diego, thanks for the time putting in elaborating this document :heart: 
This feels a spec, what do you think about upstreaming it to the spec repo so that it lives there? 
So that we proper PR this document and it becomes canon

anchor/network/src/handshake/envelope/codec.rs
```diff
@@ -0,0 +1,103 @@
...
+
+const MAXIMUM_SIZE: u64 = 1024;
+
+impl From<envelope::Error> for io::Error {
+    fn from(err: envelope::Error) -> io::Error {
+        io::Error::new(io::ErrorKind::InvalidData, err)
+    }
+}
+
+/// A `Codec` that reads/writes an **`Envelope`**
```
jxs:
```suggestion
/// A `Codec` that reads/writes an **`Envelope`**.
```

anchor/network/src/handshake/envelope/codec.rs
```diff
@@ -0,0 +1,103 @@
...
+#[derive(Clone, Debug, Default)]
+pub struct Codec;
+
+#[async_trait]
+impl RequestResponseCodec for Codec {
```
jxs: nit:
```suggestion
impl request_response::Codec for Codec {
```

anchor/network/src/handshake/envelope/codec.rs
```diff
@@ -0,0 +1,103 @@
...
+use futures::{AsyncReadExt, AsyncWriteExt};
+use libp2p::futures::{AsyncRead, AsyncWrite};
+use libp2p::request_response::Codec as RequestResponseCodec;
+use libp2p::StreamProtocol;
+use std::io;
+use tracing::debug;
+
+const MAXIMUM_SIZE: u64 = 1024;
+
+impl From<envelope::Error> for io::Error {
```
jxs, line 12: do we need this conversion?

anchor/network/src/handshake/envelope/codec.rs
```diff
@@ -0,0 +1,103 @@
...
+
+    async fn read_request<T>(
+        &mut self,
+        _protocol: &Self::Protocol,
+        io: &mut T,
+    ) -> io::Result<Self::Request>
+    where
+        T: AsyncRead + Unpin + Send,
+    {
+        debug!("reading handsake request");
```
jxs, line 36: nit, we probably can lower these ones to `trace!` and leave a single `debug!` logging the request received

anchor/network/src/handshake/envelope/envelope.proto
```diff
@@ -0,0 +1,8 @@
+syntax = "proto3";
```
jxs: can we create a dir `generated` that contains the protobuf file and the generated rust code?
So that it follows the [upstream protocol](https://github.com/libp2p/rust-libp2p/tree/master/protocols) conventions?

anchor/network/src/network.rs
jxs: was going to ask similar, do we need the `NodeInfoProvider` trait for this? can't we just use `NodeInfo`?

anchor/network/src/handshake/mod.rs
```diff
@@ -0,0 +1,253 @@
...
+        &mut self,
+        peer_id: PeerId,
+        connection_id: ConnectionId,
+        event: THandlerOutEvent<Self>,
+    ) {
+        self.behaviour
+            .on_connection_handler_event(peer_id, connection_id, event);
+    }
+
+    fn poll(
```
jxs, line 190: implementing `poll` complicates things where I feel it's not needed.
following what I wrote on  to what I wrote in https://github.com/sigp/anchor/pull/125#discussion_r1943205976,  for example, by directly calling `request_response::poll` we need to bubble up all it's `ToSwarm` events, therefore we cannot have the `match` wildcard arm with empty curlies as it's being done.

Here we are only listening on `request_response::Message::Request` and `request_response::Message::Response`, then we bubble up `RequestResponseEvent::OutboundFailure` and `RequestResponseEvent::InboundFailure` re written as `Error`.

We can still isolate the validation and `Codec` logic here both if we:
 -  bubble up `request-response::Message` and call  `handshake::handle_handshake_request` and `handshake::handle_handshake_response` in  the `network::handle_handshake_event` that you already wrote
 
or

- add the `Keypair` and `NodeInfo`  to the `Codec` and have the validation logic there

anchor/network/src/network.rs
diegomrsantos: Done

anchor/network/src/behaviour.rs
diegomrsantos: Done

anchor/network/Cargo.toml
AgeManning: I've been taking the approach to keep it local if its unlikely to be used in other crates. 

If its a general crate, its generally best to put in the workspace dependencies, in case for some reason some other crate decides to use it. 

For this specific one, it seems fairly general and conceivable that another crate MIGHT use it in the future, so might be worth chucking in workspace deps.

anchor/network/src/network.rs
AgeManning: Might be better to do a read/write mutex if we are doing lots of reads. From this snippet of code, it looks like we can't deadlock.

anchor/network/src/handshake/mod.rs
```diff
@@ -0,0 +1,253 @@
...
+
+    fn handle_handshake_request(
+        &mut self,
+        peer_id: PeerId,
+        request: Envelope,
+        channel: ResponseChannel<Envelope>,
+    ) {
+        // Handle incoming request: send response then verify
+        let response = self.sealed_node_record();
+        let _ = self.behaviour.send_response(channel, response.clone()); // Any error here is handled by the InboundFailure handler
```
jxs, line 99: why are we responding here instead of on `Codec::write_response`? 

`Codec::write_response` is called right after we receive a request, see [here](https://github.com/libp2p/rust-libp2p/blob/5f0d3f2edfafac052068170ebf9e7f1e28db9600/protocols/request-response/src/handler.rs#L141-L155).

`write_response` accepts the response that you are sending here via `ResponseChannel` piggybacking on the suggestion above wrt to `poll` we can move the `sealed_node_record` function to the `Codec` wdyt?

jxs (commented):
Hi Diego, thanks for driving this! Left some comments

anchor/network/src/behaviour.rs
diegomrsantos: This has been removed

anchor/network/src/handshake/envelope/codec.rs
diegomrsantos: I agree. My idea for now is to leave everything as debug as it makes debugging while developing easier and when the system is more stable revisit the logs. What do you think?

anchor/client/src/config.rs
```diff
@@ -16,7 +16,7 @@ pub const DEFAULT_EXECUTION_NODE_WS: &str = "ws://localhost:8545/";
 /// The default Data directory, relative to the users home directory
 pub const DEFAULT_ROOT_DIR: &str = ".anchor";
 /// Default network, used to partition the data storage
-pub const DEFAULT_HARDCODED_NETWORK: &str = "mainnet";
+pub const DEFAULT_HARDCODED_NETWORK: &str = "holesky";
```
dknopik, line 19: We should not push this to the main branch

anchor/client/src/config.rs
```diff
@@ -103,8 +103,10 @@ impl Config {
 /// `cli_args`.
 pub fn from_cli(cli_args: &Anchor) -> Result<Config, String> {
     let eth2_network = if let Some(testnet_dir) = &cli_args.testnet_dir {
+        println!("Loading testnet from {:?}", testnet_dir);
         SsvNetworkConfig::load(testnet_dir.clone())
     } else {
+        println!("Loading default network {:?}", cli_args.network);
```
dknopik: Use tracing.

Also, the network is not necessarily the default network, but can also be the network passed via `--network`.

anchor/common/ssv_network_config/src/lib.rs
```diff
@@ -32,11 +33,12 @@ pub struct SsvNetworkConfig {
     pub ssv_boot_nodes: Option<Vec<Enr<CombinedKey>>>,
     pub ssv_contract: Address,
     pub ssv_contract_block: u64,
+    pub ssv_domain_type: String
```
dknopik: I feel like it would be nicer to convert to our target type here already: a `[u8; 4]`. Then we catch config mistakes early. :)

anchor/common/ssv_network_config/src/lib.rs
```diff
@@ -76,6 +81,7 @@ impl SsvNetworkConfig {
             ssv_boot_nodes,
             ssv_contract: read(&base_dir.join("ssv_contract_address.txt"))?,
             ssv_contract_block: read(&base_dir.join("ssv_contract_block.txt"))?,
+            ssv_domain_type: read(&base_dir.join("domain_type.txt"))?,
```
dknopik: `ssv_domain_type.txt`

anchor/network/src/config.rs
```diff
@@ -66,6 +66,8 @@ pub struct Config {
 
     /// Target number of connected peers.
     pub target_peers: usize,
+
+    pub domain_type: String,
```
dknopik: I think it would be better to pass the `SsvNetworkConfig` into the `network` crate instead of duplicating the field here.

anchor/client/src/config.rs
diegomrsantos: Should the default network be mainnet while in development?

anchor/common/ssv_network_config/src/lib.rs
diegomrsantos: Could you elaborate more on the target type?

anchor/client/src/config.rs
diegomrsantos: I didn't mean to commit it, it was a mistake. I just wanted to see what the default network was. Do we need a log here?

anchor/network/src/network.rs
jxs: after discussing this off band with Diego, I don't think we need a Mutex, we can just pass the current `NodeInfo` (along the subscribed subnets) on every Request and Response

anchor/client/src/config.rs
dknopik: Ah. No, I think we do not necessarily need a log here. But we will need to look at the logging user experience again anyway before release.

anchor/client/src/config.rs
dknopik: Hm. I guess we could change it during development to prevent mistakes. :man_shrugging:

anchor/common/ssv_network_config/src/lib.rs
dknopik: The main purpose of the domain is to be embedded into message ids in order to make messages only valid for one network. For reference, here is my first draft to replicate the message ids: https://github.com/sigp/anchor/blob/unstable/anchor/common/ssv_types/src/msgid.rs#L36. Of course this was a draft only, so feel free to change it as you see fit. For example, the constants will be removed.

We could put `Domain` into the network config and implement `FromStr` for it, attempting to read it from a hex string using the `hex` crate.

EDIT: And to be more precise what I mean by target type: `Domain` or `[u8; 4]` represents more accurately, what a `ssv_domain` actually is: Four arbitrary bytes. A `String` allows way more things, and is therefore unnecessarily loose.

anchor/network/src/handshake/mod.rs
jxs: after off band conversation with Diego I think the cleanest solution is to just implement `Codec` and then reply to the requests and responses at `Swarm` level

anchor/client/src/config.rs
diegomrsantos: Yes, I think it's better. @AgeManning thoughts?

anchor/network/src/config.rs
diegomrsantos: On the other hand, we don't need more info from it in the network crate. We can start passing `DomainType` and change it to `SsvNetworkConfig` if/when we need more. Wdyt?

anchor/client/src/config.rs
diegomrsantos: removed

anchor/network/src/handshake/envelope/codec.rs
diegomrsantos: Yes. In that way, we can use the `?`operator in the codec functions and embed the application errors in the  io::Error.

anchor/network/src/handshake/envelope/codec.rs
diegomrsantos: Nice, thanks. I didn't apply the commit cause I had to remove the alias too.

anchor/common/ssv_network_config/src/lib.rs
diegomrsantos: Done

anchor/common/ssv_network_config/src/lib.rs
diegomrsantos: Let me know what you think about the current version

anchor/network/src/handshake/mod.rs
diegomrsantos: > by directly calling request_response::poll we need to bubble up all it's ToSwarm events, therefore we cannot have the match wildcard arm with empty curlies as it's being done.

Could you please elaborate more on that?

anchor/network/src/handshake/mod.rs
diegomrsantos: One of the core reasons we “respond” in our behaviour code rather than directly in `Codec::write_response` is the separation of concerns between domain logic and low-level I/O logic. Our handshake behaviour—things like deciding which data to send, verifying network IDs, and determining if the handshake is successful—belongs in the domain layer (the behaviour). Meanwhile, the codec is solely responsible for writing and reading raw bytes (the how) once the behaviour has decided what data to send.

Concretely, here’s how it works internally:

1. When we receive an inbound substream, `on_fully_negotiated_inbound` first calls `codec.read_request(...)` to decode the inbound request.
2.  The handler then notifies our behaviour of a new inbound request with a `Message::Request { request, channel, .. }`.
3. 	In our behaviour code—e.g., in `handle_handshake_request`—we run our handshake logic (sealing our node record, verifying the request, etc.) and then call `send_response(channel, response)` to supply the what.
4. 	The handler then takes that response and internally invokes `codec.write_response(...)`, performing the actual low‐level I/O on the substream.

Additionally, consider this snippet from the handler’s poll implementation:

```rust
// Check for inbound requests.
if let Poll::Ready(Some((id, rq, rs_sender))) = self.inbound_receiver.poll_next_unpin(cx) {
    // We received an inbound request.
    return Poll::Ready(ConnectionHandlerEvent::NotifyBehaviour(Event::Request {
        request_id: id,
        request: rq,
        sender: rs_sender,
    }));
}
```

This is how I believe it works in the context of `on_fully_negotiated_inbound`:

1.  In `on_fully_negotiated_inbound`, when a new inbound substream is negotiated, we first clone the codec and generate a new request_id by calling `self.next_inbound_request_id()`. We also clone the `inbound_sender` which is used to forward the request data.

2. We then create an asynchronous task (recv) that:

    - Creates a oneshot channel (rs_send, rs_recv) for the eventual response.
    - Calls `codec.read_request(&protocol, &mut stream)` to decode the inbound request.
    - Upon successful decoding, sends the tuple` (request_id, request, rs_send)` through the inbound_sender channel.
    - Waits for a response on rs_recv. Once the response is received, it calls `codec.write_response(&protocol, &mut stream, response)` to perform the actual write, closes the stream, and returns an event indicating that the response was sent.

3. This task is pushed onto a worker queue (worker_streams). Later, in the handler’s poll method, the asynchronous task’s result becomes available via `self.inbound_receiver.poll_next_unpin(cx)`.
4. When the poll detects that an inbound request is ready (i.e., it receives the tuple `(id, rq, rs_sender)`), it immediately wraps these values in a `Event::Request { request_id: id, request: rq, sender: rs_sender }` and returns it via `ConnectionHandlerEvent::NotifyBehaviour`. This notifies the upper layer (our behaviour) that a new inbound request has been received.
5. At this point, our behaviour (for example, in `handle_handshake_reques`t) can execute its domain-specific handshake logic, decide what the response should be, and then call `send_response(channel, response)`, which, as described, eventually leads to the call to `codec.write_response(...)` in the original asynchronous task.

anchor/network/src/handshake/mod.rs
diegomrsantos: Please refer to [this comment](https://github.com/sigp/anchor/pull/125#discussion_r1952515915) for context. Currently, we can't easily test inbound connections, though outbound connections seem to be working as expected. If you don’t have strong objections to the current design, would you be comfortable proceeding with it for now? We can revisit and refine the approach once we have a devnet or a similar testing environment available.

anchor/network/src/handshake/mod.rs
```diff
@@ -0,0 +1,252 @@
...
+            peer,
+            addr,
+            role_override,
+            port_use,
+        )
+    }
+
+    fn on_swarm_event(&mut self, event: FromSwarm) {
+        // Initiate handshake on new connection
+        if let FromSwarm::ConnectionEstablished(conn_est) = &event {
```
diegomrsantos, line 167: There might be an issue here. We should initiate the handshake only for an outbound connection.

anchor/network/src/network.rs
diegomrsantos: As @jxs and I agreed, I removed the trait and created a manager. We can revisit the lock later, it's not particularly important now imo.

anchor/network/src/network.rs
```diff
@@ -364,23 +362,25 @@ fn build_swarm(
...
-}
-
-impl NodeInfoProvider for DefaultNodeInfoProvider {
-    fn get_node_info(&self) -> NodeInfo {
+    pub fn get_node_info(&self) -> NodeInfo {
         // TODO consider handling lock poisoning.
         self.node_info.lock().unwrap().clone()
     }
+
+    pub fn set_node_info(&self, node_info: NodeInfo) {
```
dknopik, line 380: This is currently unused, but I think in the context where we are going to use this (on updating the subscribed subnets), we'll have a mutable reference available. If so, we won't need a Mutex. Where do you plan on using this where we do not have a mutable reference?

anchor/network/src/handshake/mod.rs
diegomrsantos: > > by directly calling request_response::poll we need to bubble up all it's ToSwarm events, therefore we cannot have the match wildcard arm with empty curlies as it's being done.
> 
> Could you please elaborate more on that?

Thanks for the catch, I addressed it with https://github.com/sigp/anchor/pull/125/commits/4e12f5ed213dc04f60b25cb9111e34c481075d29

anchor/network/src/config.rs
dknopik: Alright!

anchor/common/ssv_network_config/src/lib.rs
dknopik: Very nice, thanks!

anchor/network/src/network.rs
diegomrsantos: Yes, it's unused. I added it to show intent, but we don't need to. The idea is to centralize the global NodeInfo inside the Manager. It can be read by the Handshake and updated by something else, I assumed a Mutex is necessary. Does it make sense?

anchor/network/src/handshake/mod.rs
diegomrsantos: This has been changed

anchor/network/src/handshake/envelope/envelope.proto
diegomrsantos: Done

anchor/network/src/network.rs
```diff
@@ -287,20 +287,18 @@ async fn build_anchor_behaviour(
         discovery
     };
 
-    print!("Domain: {}", network_config.domain_type);
+    let domain_type: String = network_config.clone().domain_type.into();
+    print!("Domain: {}", domain_type);
```
dknopik: This is also to be removed, right?

diegomrsantos: sorry about that

anchor/common/ssv_network_config/src/lib.rs
```diff
@@ -22,21 +23,46 @@ macro_rules! get_hardcoded {
...
+        domain_type.copy_from_slice(&bytes);
+        Ok(Self(domain_type))
+    }
+}
+
+impl From<DomainType> for String {
+    fn from(domain_type: DomainType) -> Self {
+        hex::encode(domain_type.0)
+    }
+}
```
dknopik: Actually, how do you feel about moving this type into `ssv_types`? It is also needed for the `MessageId` there.

diegomrsantos: Done

anchor/network/src/handshake/README.md
diegomrsantos: I created a PR https://github.com/ssvlabs/ssv-spec/pull/520

dknopik:
Hey @diegomrsantos, hey @jxs,

thank you for your discussion in this PR.

You are both making a good case for your proposed variants. To better visualise the changes proposed by João (as I understand them), I implemented them [here](https://github.com/dknopik/anchor/tree/handshake) based on this PR.

To summarize the changes:
- **Move decoding of `NodeInfo` to `Codec`**. This, in my opinion, is a good idea. As you said, Diego, it is good to separate the coding and decoding logic from the rest of the logic. With this change, the conversion of `NodeInfo` to the "Entries"-based JSON format is moved to that separated codec logic. This is desirable if I understood your argument correctly.
- **Do not use `request_response::Behaviour` as a sub behaviour, but as a top-level behaviour**. This has, IMO, advantages and disadvantages. Handling the event on swarm level loosens encapsulation as said by Diego. I think this is a bit mitigated by moving the actual logic into the `handshake` module as I did. The advantage is reduced complexity by not having to impl `NetworkBehaviour`, and no risk of ignored `Waker`s or omitted events when `poll` is called.

Overall, I think I personally prefer this approach. Please take a look at my changes. Of course, I am fine with iterating more on the design. Looking forward to your thoughts.

dknopik (approved):
Thanks for implementing this, especially with the lack of a pre-existing spec! It's also great that we now have a spec.

We'll handle potential improvements separately.

PR #126, opened 2025-02-05
Basic peer manager based on libp2p `peer_store` and `connection_limits`

dknopik:
Draft of basic peer manager using `connection_limits` and a preliminary version of `peer_store`.

This is just a draft, as the exact design of `peer_store` is still TBD and a few things are missing.


dknopik:
I propose we review and merge this, even though upstream has not merged the peer store yet. We need _something_ to test. Very open for other suggestions though

Cargo.toml
```diff
@@ -43,26 +43,26 @@ ssv_types = { path = "anchor/common/ssv_types" }
...
-types = { git = "https://github.com/sigp/lighthouse", rev = "1a77f7a0" }
-unused_port = { git = "https://github.com/sigp/lighthouse", rev = "1a77f7a0" }
-validator_metrics = { git = "https://github.com/sigp/lighthouse", rev = "1a77f7a0" }
-validator_services = { git = "https://github.com/sigp/lighthouse", rev = "1a77f7a0" }
-validator_store = { git = "https://github.com/sigp/lighthouse", rev = "1a77f7a0" }
+types = { git = "https://github.com/sigp/lighthouse", rev = "b71b5f2" }
+unused_port = { git = "https://github.com/sigp/lighthouse", rev = "b71b5f2" }
+validator_metrics = { git = "https://github.com/sigp/lighthouse", rev = "b71b5f2" }
+validator_services = { git = "https://github.com/sigp/lighthouse", rev = "b71b5f2" }
+validator_store = { git = "https://github.com/sigp/lighthouse", rev = "b71b5f2" }
```
diegomrsantos, line 65: Is it possible to define rev only once?

Cargo.toml
```diff
@@ -121,6 +121,12 @@ zeroize = "1.8.1"
 [patch.crates-io]
 # todo: remove when https://github.com/supranational/blst/pull/248 is merged
 blst = { git = "https://github.com/dknopik/blst", branch = "sk-conversion" }
+# todo: remove when https://github.com/libp2p/rust-libp2p/pull/5724 is merged
+# we might need to explicitly list even more crates here if something fails
+libp2p = { git = "https://github.com/dknopik/rust-libp2p.git", rev = "170816f" }
+libp2p-identity = { git = "https://github.com/dknopik/rust-libp2p.git", rev = "170816f" }
+libp2p-mplex = { git = "https://github.com/dknopik/rust-libp2p.git", rev = "170816f" }
```
diegomrsantos: why is mplex necessary?

anchor/network/Cargo.toml
```diff
@@ -11,7 +11,7 @@ ethereum_ssz = "0.8.1"
 ethereum_ssz_derive = "0.8.1"
 futures = { workspace = true }
 hex = "0.4.3"
-libp2p = { version = "0.54", default-features = false, features = [
+libp2p = { git = "https://github.com/dknopik/rust-libp2p.git", rev = "170816f", default-features = false, features = [
```
diegomrsantos: Can we use only the peer store as a standalone dep from this repo?

anchor/network/src/network.rs
```diff
@@ -192,24 +197,39 @@ impl Network {
                 self.swarm
                     .behaviour_mut()
                     .discovery
-                    .start_subnet_query(vec![subnet]);
+                    .set_subscribed(subnet, true);
```
diegomrsantos: Could we do it in a different PR?

anchor/network/src/discovery.rs
```diff
@@ -267,6 +266,22 @@ impl Discovery {
         );
     }
 
+    pub fn set_subscribed(&mut self, subnet: SubnetId, subscribed: bool) {
```
diegomrsantos: Could we do it in a different PR?

anchor/network/src/network.rs
```diff
@@ -178,6 +170,19 @@ impl Network {
...
+    fn on_discovered_peers(&mut self, peers: HashMap<Enr, Option<Instant>>) {
+        debug!(peers =  ?peers, "Peers discovered");
+        let manager = self.peer_manager();
+        // need to collect to avoid double borrow
+        let to_dial = peers
+            .into_iter()
+            .filter_map(|(enr, _)| manager.discovered_peer(enr))
+            .collect::<Vec<_>>();
+        for dial in to_dial {
+            let _ = self.swarm.dial(dial);
```
diegomrsantos, line 229: Lighthouse does something like that https://github.com/sigp/lighthouse/blob/unstable/beacon_node/lighthouse_network/src/peer_manager/network_behaviour.rs#L113

anchor/network/src/peer_manager.rs
```diff
@@ -0,0 +1,329 @@
...
+
+const MIN_PEERS_PER_SUBNET: u16 = 3;
+
+pub struct PeerManager {
+    pub peer_store: peer_store::Behaviour<MemoryStore<Enr>>,
+    pub connection_limits: connection_limits::Behaviour,
+    pub connected: HashSet<PeerId>,
+    pub needed_subnets: HashSet<SubnetId>,
+    pub target_peers: usize,
+    pub max_with_priority_peers: usize,
```
diegomrsantos: Should those be pub?

anchor/network/src/discovery.rs
dknopik: Sure, will do.

diegomrsantos: This info will also come from the handshake https://github.com/sigp/anchor/blob/ca67c05bcd03b10d538c654a15e0ba468e8f269c/anchor/network/src/network.rs#L227

anchor/network/src/peer_manager.rs
```diff
@@ -0,0 +1,329 @@
...
+            connection_limits,
+            connected: HashSet::with_capacity(max_priority_peers),
+            needed_subnets: HashSet::new(),
+            target_peers: config.target_peers,
+            max_with_priority_peers: max_priority_peers,
+        }
+    }
+
+    /// report a discovered peer, and return dial opts if we want to dial it
+    pub fn discovered_peer(&mut self, peer: Enr) -> Option<DialOpts> {
```
diegomrsantos: maybe `enr` or `peer_enr`?

anchor/network/src/network.rs
dknopik: The event only allows us to dial a single peer.

Cargo.toml
dknopik: No.

anchor/network/src/discovery.rs
diegomrsantos: We could potentially handle this too https://github.com/sigp/anchor/blob/ca67c05bcd03b10d538c654a15e0ba468e8f269c/anchor/network/src/network.rs#L228

anchor/network/Cargo.toml
dknopik: Nope. If you do that, you will quickly run into 
> note: there are multiple different versions of crate `libp2p_swarm` in the dependency graph
and similar compiler error.

Rust treats types coming from the same crate, but different versions, as distinct.

anchor/network/src/network.rs
diegomrsantos: Yes, I think it creates an event for each peer. But honestly, I don't know the reason. Maybe decoupling?

Cargo.toml
dknopik: Because else there would be a different mplex somewhere in the dependency graph, causing failure.

But actually, while testing again with a recent version of the libp2p PR, all of these are seemingly not needed anymore. So I'll remove them all.

anchor/network/src/peer_manager.rs
dknopik: probably not. good catch!

anchor/network/src/network.rs
dknopik: @jxs: Is there a relevant difference between these approaches to dial?

anchor/network/src/network.rs
```diff
@@ -13,21 +14,36 @@ use libp2p::swarm::SwarmEvent;
...
+/// A fraction of `PeerManager::target_peers` that we want to be outbound-only connections.
+const TARGET_OUTBOUND_ONLY_FACTOR: f32 = 0.3;
+/// A fraction of `PeerManager::target_peers` that if we get below, we start a discovery query to
+/// reach our target. MIN_OUTBOUND_ONLY_FACTOR must be < TARGET_OUTBOUND_ONLY_FACTOR.
+const MIN_OUTBOUND_ONLY_FACTOR: f32 = 0.2;
+/// The fraction of extra peers beyond the PEER_EXCESS_FACTOR that we allow us to dial for when
+/// requiring subnet peers. More specifically, if our target peer limit is 50, and our excess peer
+/// limit is 55, and we are at 55 peers, the following parameter provisions a few more slots of
+/// dialing priority peers we need for validator duties.
+const PRIORITY_PEER_EXCESS: f32 = 0.2;
```
diegomrsantos: Would this be better encapsulated inside the Peer Manager?

anchor/network/src/network.rs
diegomrsantos: We could also encapsulate this code inside the Peer Manager. Similar to https://github.com/sigp/lighthouse/blob/stable/beacon_node/lighthouse_network/src/service/mod.rs#L1845

anchor/network/src/network.rs
dknopik: yep, will do

anchor/network/src/behaviour.rs
```diff
@@ -13,6 +14,8 @@ pub struct AnchorBehaviour {
     pub gossipsub: gossipsub::Behaviour,
     /// Discv5 Discovery protocol.
     pub discovery: Discovery,
+    /// Anchor peer manager, wrapping libp2p behaviours with minimal added logic for peer selection
```
jxs: nit:
```suggestion
    /// Anchor peer manager, wrapping libp2p behaviours with minimal added logic for peer selection.
```

anchor/network/Cargo.toml
```diff
@@ -12,7 +12,7 @@ ethereum_ssz = { workspace = true }
 ethereum_ssz_derive = { workspace = true }
 futures = { workspace = true }
 hex = "0.4.3"
-libp2p = { version = "0.54", default-features = false, features = [
+libp2p = { git = "https://github.com/libp2p/rust-libp2p.git", rev = "082eb16", default-features = false, features = [
```
jxs, line 15: interesting! Didn't know one could this, I suggest we use our own fork or yours Daniel, as Dr Huang may rebase and I fear that commit may disapear, wdyt?

anchor/network/src/handshake/mod.rs
```diff
@@ -74,6 +74,7 @@ pub fn handle_event(
     match event {
         Event::Message {
             peer,
+            connection_id: _,
```
jxs: nit: use `..` for the non used variables on this and followup cases

anchor/network/src/network.rs
jxs: sorry, missed this!
>@jxs: Is there a relevant difference between these approaches to dial?


no, one calls [the other](https://github.com/libp2p/rust-libp2p/blob/master/swarm/src/lib.rs#L1097) underneath :)

anchor/network/src/peer_manager.rs
```diff
@@ -0,0 +1,368 @@
...
+    type ConnectionHandler = dummy::ConnectionHandler;
+    type ToSwarm = peer_store::Event<memory_store::Event>;
+
+    fn handle_pending_inbound_connection(
+        &mut self,
+        connection_id: ConnectionId,
+        local_addr: &Multiaddr,
+        remote_addr: &Multiaddr,
+    ) -> Result<(), ConnectionDenied> {
+        self.peer_store.handle_pending_inbound_connection(
```
jxs, line 233: for the sake of consistency I suggest we always check the `connection_limits` first as in the `handle_pending_outbound` connection as priority matters in these checks. By checking first `peer_behaviour` first we may introduce a peer to the `peer-store` that then may be reject by `connection-limits` and therefore we aborting the connection to it

jxs (commented):
Overall looks good to me Daniel! Left some minor suggestions

anchor/network/src/peer_manager.rs
dknopik: I put it intentionally this way to add the peer to the peer store regardless (so that we are aware of it and can use it if we need any peers). What do you think?

anchor/network/Cargo.toml
dknopik: I don't think that can happen. The commit can disappear from the branch, but it stays in the repo if it was pushed. For example this commit is still accessible via the link: https://github.com/sigp/lighthouse/commit/7f33d35eb41ef00adfde1d9cebb0092fdfa3cc39, even though it no longer is in the current commit chain of the PR branch.

anchor/network/src/peer_manager.rs
jxs: ah! Interesting, yeah that seems clever Daniel, can we then leave a comment then explaining it?

anchor/network/Cargo.toml
jxs: ok thanks for confirming Daniel :)

jxs (dismissed):
LGTM, thanks Daniel

anchor/network/src/peer_manager.rs
dknopik: done!

anchor/network/src/peer_manager.rs
```diff
@@ -0,0 +1,370 @@
...
+    type ToSwarm = peer_store::Event<memory_store::Event>;
+
+    fn handle_pending_inbound_connection(
+        &mut self,
+        connection_id: ConnectionId,
+        local_addr: &Multiaddr,
+        remote_addr: &Multiaddr,
+    ) -> Result<(), ConnectionDenied> {
+        // we call the peer store here first to remember the peer regardless of whether we accept a
+        // connection with it right now
```
jxs:
```suggestion
        // connection with it right now.
```

jxs (dismissed):

jxs (approved):

diegomrsantos (approved):

dknopik:
Thanks everyone! :tada:

PR #128, opened 2025-02-06
Qbft testing

Zacholme7:
## Issue Addressed

#98 

## Proposed Changes

This PR introduces a framework to provide end to end QBFT Tests. It allows the ability to run multiple instances at once and control the operational status of the nodes along with any byzantine behavior. 

There are also a few bugfixes from issues that were discovered during testing. 

Notes for clarity
- The qbft sender/receiver will be replaced in next pr with signing and network functionality
- `cluster_members` was changed to a `Vec` as we need deterministic leader selection for testing and random ordering with the `HashSet` did not allow for this.


anchor/qbft_manager/src/lib.rs
```diff
@@ -99,13 +105,15 @@ impl<T: SlotClock> QbftManager<T> {
         processor: Senders,
         operator_id: OperatorId,
         slot_clock: T,
+        qbft_out: mpsc::UnboundedSender<Message>,
```
dknopik: we should use a bounded queue here in order to not run into memory issues if we have insufficient bandwidth

anchor/qbft_manager/src/lib.rs
```diff
@@ -307,17 +320,26 @@ async fn qbft_instance<D: QbftData<Hash = Hash256>>(mut rx: UnboundedReceiver<Qb
...
                     // initialize it
                     QbftInstance::Uninitialized { message_buffer } => {
                         // todo: actually send messages somewhere
                         // Create a new instance and receive any buffered messages
-                        let mut instance = Box::new(Qbft::new(config, initial, |_| {}));
+
+                        let mut instance = Box::new(Qbft::new(config, initial, |message| {
+                            if let Err(e) = tx.send(message) {
+                                error!("Failed to send qbft message: {:?}", e);
+                            }
```
dknopik: and here: drop message (with warning) if queue full, or exit if queue closed (as that is irrecoverable and indicates that we have a serious failure or are shutting down)

anchor/common/ssv_types/src/cluster.rs
```diff
@@ -25,7 +24,7 @@ pub struct Cluster {
     /// If the Cluster is liquidated or active
     pub liquidated: bool,
     /// Operators in this cluster
-    pub cluster_members: HashSet<OperatorId>,
+    pub cluster_members: Vec<OperatorId>,
```
dknopik: what prompted this change?

anchor/qbft_manager/Cargo.toml
```diff
@@ -5,13 +5,20 @@ authors = ["Sigma Prime <contact@sigmaprime.io"]
...
 ethereum_ssz_derive = { workspace = true }
+futures = { workspace = true }
 processor = { workspace = true }
 qbft = { workspace = true }
 slot_clock = { workspace = true }
 ssv_types = { workspace = true }
+task_executor = { workspace = true }
 tokio = { workspace = true, features = ["sync"] }
 tracing = { workspace = true }
+tracing-subscriber = { workspace = true }
```
dknopik: All these added deps can be moved to `dev-dependencies`, right?

anchor/qbft_manager/src/tests.rs
```diff
@@ -0,0 +1,757 @@
...
+    #[default]
+    None,
+    // Send conflicting votes for the same round
+    DoubleVote,
+    // Drop all messages of certain types
+    MessageSuppression(QbftMessageType),
+    // Modify the round so that the message is invalid
+    InvalidMessage,
+}
+// Descirbes the behavior of an operator
```
dknopik: Describes

anchor/qbft_manager/src/tests.rs
```diff
@@ -0,0 +1,757 @@
...
+                        for res in self.results.read().unwrap().values().cloned(){
+                            let _ = consensus_tx.send(res);
+                        }
+                        break;
+                    }
+                }
+                // Have to yield here. try_recv is greedy and will starve the runtime
+                else => {
+                    tokio::task::yield_now().await;
+                }
```
dknopik: Use `rx.recv()` in the branches instead of creating an ad-hoc future that just attempts to `try_recv` - this creates a proper future that waits and will not cause a spinlock here. Then you can remove the `else` branch :)

anchor/qbft_manager/src/tests.rs
```diff
@@ -0,0 +1,757 @@
...
+            _shutdown: shutdown,
+            clock,
+        }
+    }
+
+    #[tokio::test]
+    // Test running a single instance and confirm that it reaches consensus
+    async fn test_basic_run() {
+        let setup = setup_test();
+        let mut context = TestContext::<SystemTimeSlotClock, BeaconVote>::new(
```
dknopik: Should probably be a `ManualSlotClock` here

anchor/qbft_manager/src/tests.rs
```diff
@@ -0,0 +1,757 @@
...
+            CommitteeSize::Ten, // Using larger committee for partition testing
+            vec![generate_test_data()],
+        )
+        .await;
+
+        // Initial partition. We have > f offline so we will not be able to reach consensus
+        context.set_operators_offline(&[3, 4, 5, 6]);
+
+        // Wait and change partition
+        tokio::time::sleep(Duration::from_secs(3)).await;
```
dknopik, line 749: Using real sleeps is prone to random test failure and slows testing down. Take a look at https://tokio.rs/tokio/topics/testing#pausing-and-resuming-time-in-tests

anchor/qbft_manager/src/tests.rs
```diff
@@ -0,0 +1,757 @@
...
+use std::sync::{Arc, RwLock, RwLockWriteGuard};
+use std::time::{Duration, SystemTime, UNIX_EPOCH};
+use task_executor::{ShutdownReason, TaskExecutor};
+use tokio::sync::mpsc;
+use tokio::sync::mpsc::{UnboundedReceiver, UnboundedSender};
+use tracing::error;
+use types::{Hash256, Slot};
+
+// Counter for instance height. Allows us to maintain unique data while using deterministic leaders
+static ID_COUNTER: AtomicUsize = AtomicUsize::new(1);
```
dknopik: Do not use globals. This makes tests dependent on the order they are run if the test runner runs all the tests with one invocation of the test binary (which is the case for `cargo test`, but not `cargo nextest`)

anchor/qbft_manager/src/tests.rs
```diff
@@ -0,0 +1,757 @@
...
+            target: types::Checkpoint::default(),
+        };
+
+        (data, id)
+    }
+
+    // Setup env for the test
+    fn setup_test() -> Setup {
+        let env_filter = tracing_subscriber::EnvFilter::new("debug");
+        tracing_subscriber::fmt().with_env_filter(env_filter).init();
```
dknopik: This causes the tests to fail with `cargo test` - so we should only call this once per process. Contrary to my other comment, a global lock for initialization (e.g. std's `LazyLock`) is reasonable here

dknopik (commented):
Great PR! I especially like the support for several kinds of byzantine behaviour. Added some comments on details.

anchor/common/ssv_types/src/cluster.rs
Zacholme7: I needed a way for deterministic leaders for the tests so we can test specific leader based scenarios. With the `HastSet`, the operator order is always random so this made the leader order also always random. The only way I could fix this is just by moving this to a `Vec` so that the order says the same and therefore so does the leader rotation.

anchor/qbft_manager/src/lib.rs
Zacholme7: Was poking around with this a bit. I think this change requires turning a decent amount of the qbft crate async. Do you think that is acceptable or is there another way to handle this?

all of the qbft  `send_*` functions need to then be turned async to send the messages which then trickles up through the rest of the code.

anchor/qbft_manager/src/tests.rs
Zacholme7: This is neat thanks

anchor/qbft_manager/src/tests.rs
Zacholme7: Can change. What is the reason behind this?

anchor/qbft_manager/src/lib.rs
dknopik: [`try_send`](https://docs.rs/tokio/latest/tokio/sync/mpsc/struct.Sender.html#method.try_send) is not async and also what we need, as we want to discard messages on constrained bandwidth.

anchor/common/ssv_types/src/cluster.rs
dknopik: Sounds good. In `qbft::config`, we use `IndexSet`, which ensures both uniqueness and order. What do you think about using it here as well? I am not sure, maybe that is needlessly complex, a `Vec` certainly is simpler. :shrug:

anchor/qbft_manager/src/tests.rs
dknopik: The `SystemTimeSlotClock` uses, as it says, the system time to determine what slot we are in. This is the only reasonable behaviour for production code.

The `ManualSlotClock` allows us to manually chose what time it is. This is very useful for testing, as sometimes tests take way longer than expected, and using `SystemTimeSlotClock` we would run into trouble. For example, the qbft manager cleans up old instances after a while. If the tests take longer than expected (because the running machine is resource constrained or something), the tests may fail because of that. Using the `ManualSlotClock`, we can manually advance time, or not at all, as we do here, preventing these failures. This is the reason why we pass around the `T: SlotClock` everywhere: so that we switch to this for testing :)

Thanks for changing it. But I think we can bubble this up into the `TestContext`, hardcoding the `ManualSlotClock` in the type there instead of using a generic, because I can't think of any scenario where we want to use `SystemTimeSlotClock` instead while testing.

dknopik:
Thought again about the queue we are passing. I am not sure if `Message` is a good type for the output queue. I feel like we should send the signed message there, ready to be sent over the network. What do you think?

Zacholme7:
@dknopik I was actually just planning on scrapping `Message` in the signing pr as it really doesn't provide any use. The message type is already encoded in the `QbftMessage` and I dont see a place where the id is useful. Or am I missing some case where we need this?

I think `UnsignedSSVMessage` is the most appropriate for the output queue. What I have right now in the signing draft is qbft instance outputs the unsigned messages, the unsigned message is passed to a signing task to create a signed message, and then the signed message is serialized and sent to the network.

dknopik:
> I was actually just planning on scrapping `Message` in the signing pr as it really doesn't provide any use. The message type is already encoded in the `QbftMessage` and I dont see a place where the id is useful. Or am I missing some case where we need this?

Well, `Message` was more useful back when the `qbft` crate was more independent from the ssv types/message format, but as we have abandoned that, it is now indeed redundant.

> I think `UnsignedSSVMessage` is the most appropriate for the output queue. What I have right now in the signing draft is qbft instance outputs the unsigned messages, the unsigned message is passed to a signing task to create a signed message, and then the signed message is serialized and sent to the network.

Alright!

anchor/qbft_manager/src/lib.rs
```diff
@@ -323,12 +323,19 @@ async fn qbft_instance<D: QbftData<Hash = Hash256>>(
...
+                            match tx.try_send(message) {
+                                Ok(()) => (),
+                                Err(TrySendError::Full(msg)) => {
+                                    // Queue is full - drop message under constrained bandwidth
+                                    warn!("Dropping QBFT message due to full queue: msg={:?}", msg);
+                                }
+                                Err(TrySendError::Closed(_)) => {
+                                    // Channel closed - critical failure or shutdown
+                                    error!("QBFT message channel closed - initiating shutdown");
+                                    // todo!() need some sort of shutdown
```
dknopik: Sorry, one more nitpick:
```rust
warn!(?msg, "Dropping QBFT message due to full queue");
```

And for the `Err` case: a `return` after the log should suffice here.

Zacholme7: A bit confused on the `Err` case. Dont we need some explicit shutdown? Using `return` here is the same as finishing normally with a implicit `()`. The instance will still be able to send messages and they will all hit the Closed case. The linter also says the return is unneeded.

dknopik: aaah, yeah of course, missed that we are inside a closure.

hm. let's leave it as is for now, the instance also quits if the input queue is closed, so it should be fine

dknopik (approved):
LVGTM

PR #129, opened 2025-02-07
Remove `faulty` field from database and add calculating function instead

dknopik:
We never calculated the `faulty` field of `Cluster` or read it from the DB until now. Remove the field and calculate it instead, it isn't worth caching such a cheap calculation


Zacholme7 (approved):
lgtm!

PR #130, opened 2025-02-10
Ignore case spellcheck

diegomrsantos:
## Issue Addressed

No issue, simple change.

## Proposed Changes

Ignore case in spellcheck and leave only lowercase entries


diegomrsantos:
Closing as we want to catch typos like "BEacon" instead of "Beacon"

Issue #99, opened 2025-01-14
Ignore warn logs during historical sync


Zacholme7:
## Description

During a historical sync, we encounter a number of malformed events and they are just warning logged. These are expected but end users do not know this and getting a screen of warning logs is not very user friendly. These should be ignored during the historical sync but turned back on during the live sync.

PR #132, opened 2025-02-10
Ignore warn log when historical syncing

Zacholme7:
## Issue Addressed

#99 

## Proposed Changes

Only emit malformed event logs during live sync. We know there are a number of malformed events that are encountered during historical sync and these are expected so we dont want to flood the screen with warn logs.


dknopik:
What do you think about emitting a debug log for historical sync?

dknopik (approved):
LGTM

Issue #119, opened 2025-01-31
Sync error handling


dknopik:
I noticed during my local tests that
- sync sometimes crashes instead of retrying
- when sync crashes, the rest of the process lives on

IMO:
- Sync should never crash, but retry indefinitely
- IF it crashes anyways due to a bug, the rest of the process should quit as well (in order for an external restart to trigger)
- Maybe we need some flag if sync is stalled for some reason to stop performing duties during that time, in order to avoid signing stuff for validators that have been removed from our clusters.

Zacholme7:
Few follow ups. 

1) What do you mean here by crash? What specific error is it encountering? Or are there some warn logs emitted? Is there a specific block range it gets hung up on? 
2) It could retry indefinitely but I think there are also arguments that it should not. The chain data does not change so pretty much the only issue would be the rpc being inaccessible/going down. Do we want to keep hitting an rpc that we know does not work or just give it a set number of retrys and exit since we cant make progress anyways?
3) I think a flag could be a good addition. I dont think it is functionally necessary but it is good to have. Say we have 4 operators and 1 has a bad rpc and does not see the validator removed event. The other 3 will stop the duties and the one remaining will not be able to reach consensus anyways. So really no harm, but we are wasting resources there.

dknopik:
1. Here, for example, we never retry.
https://github.com/sigp/anchor/blob/5cc2376c42121a149cb1b03a868bc39207fc501d/anchor/eth/src/sync.rs#L196-L199
This is the very first EL call we make, and therefore very prone to fail because the anchor client might be started simultaneously with EL and CL and the EL might not yet have the endpoint up.

2. >Do we want to keep hitting an rpc that we know does not work or just give it a set number of retrys and exit since we cant make progress anyways?

     IMO, we should keep trying, ideally with capped randomized exponential backoff to avoid DoS.

    If we stay running, we have at least the chance of recovering. So in my opinion, we should stay running and keep trying to connect. In most cases, the users have some kind automatic restart, but we should not rely on that and always try to resume - after all, we want to perform duties.

3. Your example is correct, but I am still worried about cases where multiple operators suffer similar problems - just continuing would then cause issues. So strictly speaking, we should not perform duties if we are not synced, or at least within a certain margin. Yes, this is pedantic - But when having partial responsibility for up to 500 * 32 ETH, safety is rather important.

PR #133, opened 2025-02-11
Sync error handling

Zacholme7:
## Issue Addressed

#119 

## Proposed Changes

This PR introduces a way to handle rpc errors and signal if sync is stalled.

If a websocket goes down or an rpc endpoint is having issues,`OPERATIONAL_STATUS` is set to false. The rest of the application can be conditioned on this value to determine if the execution layer is having sync issues. 

If there is an rpc error, there is nothing we can do until the endpoint is operational again. The simplest way to test this is just to continuously poll for a block number with exponential backoffs. There is no longer a set number of retries. It just keeps retrying until it is valid again.


anchor/eth/src/sync.rs
```diff
@@ -38,14 +38,19 @@ static SSV_EVENTS: LazyLock<Vec<&str>> = LazyLock::new(|| {
     ]
 });
 
+/// Current operational status of sync. If there is an issue with the rpc endpoint or the ws
+/// endpoing, the status is considered down. Otherwise, it is up
+pub static OPERATIONAL_STATUS: AtomicBool = AtomicBool::new(true);
```
dknopik: Why not a field in the event syncer? I am a bit worried about global state. If we proceed with our plans for a simulator, multiple client instances might live in the same process.

Zacholme7: I think both approaches work. 

My thought process was that the status of the sync is a globally important. If sync is not progressing, nothing should progress. Hence why I put a global that can be used to describe the state of sync at any point in time. 

I didn't consider the simulator. Keeping a `Atomicbool` in the syncer could also work. I think the biggest trade off here is having to pass the value every single place we want it but avoiding global state vs using global state and just having to import. 

Wdyt

dknopik: I'm in favor of keeping it in the syncer as an `Arc<AtomicBool>` which we can clone and pass around. I feel like we'll mainly need it in the `validator_store`, so it shouldn't be too bothersome to pass around. :)

As always, we can pull in more opinions if you like :)

dknopik:
I retested it locally. Could not get it to crash now. Nice!

But I noticed that right now, this error condition does not properly apply any backoff: https://github.com/sigp/anchor/blob/b5ffff4901ee45d89cc1523f13221591c30f5788/anchor/eth/src/sync.rs#L413-L414

This got me thinking once more about the approach. How do you feel about, instead of having to apply `troubleshoot_rpc` at multiple appropriate locations throughout the file, just returning an error in the `live` and `historical` functions and doing the reconnect logic at the top level? This also handles cases where we want to fall back to historical sync logic after being offline for a considerable amount of time, and helps us assert that we really do not crash out anymore if something fails. There might be some disadvantages I am not thinking of right now though.

Zacholme7:
@dknopik Threw together a quick POC. I think I that approach works very nice. Take a look when you have a sec and if thats along the lines of what you were thinking ill clean it up and make sure it works.

anchor/eth/execution.rs
```diff
@@ -51,9 +52,10 @@ async fn main() {
     // exist. It will communicate with the rest of the system via processor channels and constantly
     // keep the database up to date with new data for the rest of the system
     let db = Arc::new(NetworkDatabase::new(path, &rsa_pubkey).unwrap());
-    let mut event_syncer = SsvEventSyncer::new(db.clone(), config)
-        .await
-        .expect("Failed to construct event syncer");
+    let mut event_syncer =
+        SsvEventSyncer::new(db.clone(), config, Arc::new(AtomicBool::new(false)))
```
dknopik: I feel like it would be bit cleaner to create the Arc within the `new` method.

dknopik:
> Threw together a quick POC. I think I that approach works very nice. Take a look when you have a sec and if thats along the lines of what you were thinking ill clean it up and make sure it works.

yea, seems good! This is what I meant. :​)

anchor/eth/execution.rs
Zacholme7: Like create it all inside of the constructor and then just pull it out via something like `let status = event_syncer.get_status()`?

dknopik: Yeah. Basically, the caller should not care about how the value is created, regardless how simple it is. The getter should then be a simple  `pub fn operational_status() -> &Arc<AtomicBool>` or so.

dknopik:
Something seems a bit funky still, when disconnecting during live sync, backoff works great, but when reconnecting, we enter a weird state where we loop like this without backoff: 
```text
2025-02-17T08:24:03.006389Z  INFO sync:try_sync{contract_address=0x38a4794cced47d3baf7370ccc43b560d3a1beefa deployment_block=181612}: eth::sync: Starting historical sync
2025-02-17T08:24:03.006411Z DEBUG sync:try_sync{contract_address=0x38a4794cced47d3baf7370ccc43b560d3a1beefa deployment_block=181612}:historical_sync: alloy_rpc_client::call: sending request method=eth_blockNumber id=349
2025-02-17T08:24:03.006536Z DEBUG sync:try_sync{contract_address=0x38a4794cced47d3baf7370ccc43b560d3a1beefa deployment_block=181612}:historical_sync:ReqwestTransport{url=http://127.0.0.1:8545/}: hyper_util::client::legacy::pool: reuse idle connection for ("http", 127.0.0.1:8545)
2025-02-17T08:24:03.026185Z DEBUG sync:try_sync{contract_address=0x38a4794cced47d3baf7370ccc43b560d3a1beefa deployment_block=181612}:historical_sync:ReqwestTransport{url=http://127.0.0.1:8545/}: hyper_util::client::legacy::pool: pooling idle connection for ("http", 127.0.0.1:8545)
2025-02-17T08:24:03.026239Z DEBUG sync:try_sync{contract_address=0x38a4794cced47d3baf7370ccc43b560d3a1beefa deployment_block=181612}:historical_sync:ReqwestTransport{url=http://127.0.0.1:8545/}: alloy_transport_http::reqwest_transport: received response from server status=200 OK
2025-02-17T08:24:03.026283Z DEBUG sync:try_sync{contract_address=0x38a4794cced47d3baf7370ccc43b560d3a1beefa deployment_block=181612}:historical_sync:ReqwestTransport{url=http://127.0.0.1:8545/}: alloy_transport_http::reqwest_transport: retrieved response body. Use `trace` for full body bytes=47
2025-02-17T08:24:03.026347Z  INFO sync:try_sync{contract_address=0x38a4794cced47d3baf7370ccc43b560d3a1beefa deployment_block=181612}:historical_sync: eth::sync: Synced up to the tip of the chain, breaking
2025-02-17T08:24:03.026364Z  INFO sync:try_sync{contract_address=0x38a4794cced47d3baf7370ccc43b560d3a1beefa deployment_block=181612}:historical_sync: eth::sync: Historical sync completed
2025-02-17T08:24:03.026390Z  INFO sync:try_sync{contract_address=0x38a4794cced47d3baf7370ccc43b560d3a1beefa deployment_block=181612}: eth::sync: Starting live sync
2025-02-17T08:24:03.026415Z  INFO sync:try_sync{contract_address=0x38a4794cced47d3baf7370ccc43b560d3a1beefa deployment_block=181612}:live_sync: eth::sync: Network up to sync..
2025-02-17T08:24:03.026430Z  INFO sync:try_sync{contract_address=0x38a4794cced47d3baf7370ccc43b560d3a1beefa deployment_block=181612}:live_sync: eth::sync: Current state
2025-02-17T08:24:03.026444Z  INFO sync:try_sync{contract_address=0x38a4794cced47d3baf7370ccc43b560d3a1beefa deployment_block=181612}:live_sync: eth::sync: Starting live sync contract_address=0x38a4794cced47d3baf7370ccc43b560d3a1beefa
2025-02-17T08:24:03.026489Z ERROR sync: eth::sync: Sync failed, attempting recovery e=WsError("Failed to subscribe to block stream: backend connection task has stopped")
```

Zacholme7:
@dknopik should be fixed now, was matching against wrong error

dknopik (approved):
LGTM. if you consider it ready, feel free to merge

Issue #104, opened 2025-01-15
Qbft signing and verification infra


Zacholme7:
## Description

Currently, the qbft is not signing or verifying signatures on any messages. While the qbft messages should be expanded to contain all relevant signature information, the actual signing should probably happen in the processor or signature collector. Signature verification should probably stay within the qbft crate, but there is also arguments for this also being separated out.

dknopik:
IMO qbft should be strictly responsible for consensus. Signing is done in `sign_and_collect` of the signature collector and triggered by the `validator_store`: e.g. https://github.com/sigp/anchor/blob/2e838d84f11054847e839b854966489d4f518608/anchor/validator_store/src/lib.rs#L623

Zacholme7:
@dknopik yep, qbft can stick to consensus. I believe we are still missing message signing though. The messages get signed with the operators rsa private key at some point in the process which we have not implemented yet.

dknopik:
ah - gotcha. I was thinking about signing the final value, which is of course different from signing the qbft messages.

PR #134, opened 2025-02-11
QBFT Message Signing

Zacholme7:
## Issue Addressed

Part of #104 

## Proposed Changes

This PR introduces the message signing portion qbft. Messages are passed from the qbft instances into a long running signer task, which then sends a blocking task to the processor to do the actual signing. The messages are then sent to a network channel which, as the idea stands, will take the serialized signed message and send it off.


anchor/qbft_manager/src/lib.rs
```diff
@@ -203,6 +221,67 @@ impl<T: SlotClock> QbftManager<T> {
...
+                            Self::sign_and_send_message(pkey, id, unsigned, serialized, network_tx)
+                        {
+                            error!("Signing failed: {}", e);
+                        }
+                    },
+                    QBFT_SIGNER_NAME,
+                )
+                .unwrap_or_else(|e| warn!("Failed to send to processor: {}", e));
+        }
+    }
```
dknopik: what is the advantage of having this long-running task instead of just invoking `send_blocking` directly?

anchor/qbft_manager/src/tests.rs
```diff
@@ -733,6 +733,7 @@ mod manager_tests {
     // Test network partition scenarios
     // This simulates temporary network partitions by taking nodes offline and bringing them back
     async fn test_network_partition() {
+        // todo!() this one is till odd
```
dknopik: What does this mean?

anchor/qbft_manager/src/lib.rs
Zacholme7: Fixed. I had some reason but I have no idea what that reason is anymore. Much cleaner to just put in the closure.

anchor/qbft_manager/src/tests.rs
Zacholme7: Test wasn't working during merge and was a reminder to myself to look at it that I forgot to remove. Fixed.

anchor/qbft_manager/src/lib.rs
```diff
@@ -495,6 +443,34 @@ async fn qbft_instance<D: QbftData<Hash = Hash256>>(
...
+    let signed = SignedSSVMessage::new(
+        vec![sig],
+        vec![*id],
+        unsigned.ssv_message,
+        unsigned.full_data,
+    )?;
+    let serialized_signed = signed.as_ssz_bytes();
+    network_tx
+        .send(serialized_signed)
+        .map_err(|e| format!("Failed to send signed ssv message to network: {}", e))?;
```
dknopik: I just realized that the network will need to know on which subnet to publish this. So unfortunately we need to send the message unserialized to the network or figure that out here. wdyt?

Zacholme7: Changed. That is good I wasnt sure which one to use so I just defaulted to the bytes. This may need to be some signed ssv specific message channel in the future if other things are being sent to to the network as bytes, but just depends how we set that up. Easy to change around.

dknopik (approved):
nice, thx

Issue #103, opened 2025-01-15
Implement and handle QBFT decided message


Zacholme7:
## Description

We are currently missing the Decided state for qbft. Extend the qbft_manager to send this messages when a quorum of commit messages has been reached.

From the spec: 

> We have an extra message named Decided. This message is an aggregation of a quorum of Commit messages plus the proposal data. Its main reason of existence is that, if there is a late replica that didn't receive the quorum of Commits yet, as soon as it receives the Decided message it has all the necessary information that allows it to finish the consensus execution.

dknopik:
This state is reflected by `Qbft::completed` returning `Some`. The `qbft_manager` can send such a message based on that.

PR #136, opened 2025-02-12
Commit aggregation

Zacholme7:
## Issue Addressed

Part of #103 

## Proposed Changes

This PR introduces commit message aggregation, aka decided messages. This is just the aggregation portion of the PR, the next step is to implement handling for this message type.

A aggregated commit message is just a quorum of commit messages where all of the signatures and operators are joined into one message. Instead of creating a whole new message, I just aggregate this all into the first commit message of the quorum.  This also has to be broadcasted into the network so relevant manager functionality is also added.

### DIFF
--- a/anchor/common/qbft/src/lib.rs
+++ b/anchor/common/qbft/src/lib.rs
@@ -165,6 +165,12 @@ where
         self.start_round();
     }
 
+
+    // Get the aggregated commit message, if it exists
+    pub fn get_aggregated_commit(&self) -> Option<SignedSSVMessage> {
+        self.aggregated_commit.clone()
+    }
+
     // Validation and check functions.
     fn check_leader(&self, operator_id: &OperatorId) -> bool {
         self.config.leader_fn().leader_function(
--- a/anchor/common/qbft/src/lib.rs
+++ b/anchor/common/qbft/src/lib.rs
@@ -165,7 +165,6 @@ where
         self.start_round();
     }
 
-
     // Get the aggregated commit message, if it exists
     pub fn get_aggregated_commit(&self) -> Option<SignedSSVMessage> {
         self.aggregated_commit.clone()
--- a/anchor/common/qbft/src/lib.rs
+++ b/anchor/common/qbft/src/lib.rs
@@ -697,9 +697,7 @@ where
             // Sanity check that all of the messages match
             commit_quorum[1..]
                 .iter()
-                .all(|commit_msg| {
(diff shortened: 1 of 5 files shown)


anchor/common/qbft/src/lib.rs
```diff
@@ -659,17 +669,51 @@ where
...
+                // Aggregate all of the commit messages
+                let commit_quorum = self.commit_container.get_quorum_of_messages(round);
+                let aggregated_commit = self.aggregate_commit_messages(commit_quorum);
+                if aggregated_commit.is_some() {
+                    debug!(in = ?self.config.operator_id(), state = ?self.state, "Reached a COMMIT consensus. Success!");
+                    self.state = InstanceState::Complete;
+                    self.completed = Some(Completed::Success(hash));
+                    self.aggregated_commit = aggregated_commit;
+                } else {
+                    warn!("Failed to aggregate commit quorum")
```
dknopik: Should never be the case, right? Then I would propose severity `error`.

anchor/common/qbft/src/lib.rs
```diff
@@ -659,17 +669,51 @@ where
...
 
+    fn aggregate_commit_messages(
+        &self,
+        commit_quorum: Vec<WrappedQbftMessage>,
+    ) -> Option<SignedSSVMessage> {
+        // We know this exists, but in favor of avoiding expect match the first element to Some.
+        // This will be the commit message that we aggregate on top of
+        if let Some(first_commit) = commit_quorum.first() {
+            let mut aggregated_commit = first_commit.signed_message.clone();
+            let aggregated_as_ssz = aggregated_commit.ssv_message().as_ssz_bytes();
```
dknopik: Why do we serialize here?

anchor/common/qbft/src/lib.rs
```diff
@@ -659,17 +669,51 @@ where
...
+                .all(|commit_msg| {
+                    aggregated_as_ssz == commit_msg.signed_message.ssv_message().as_ssz_bytes()
+                })
+                .then_some(())?;
+
+            // Aggregate all of the commits together
+            let signed_commits = commit_quorum[1..]
+                .iter()
+                .map(|msg| msg.signed_message.clone())
+                .collect();
```
dknopik: There's no need to collect here, we can just use the iterator.

dknopik (approved):
lgtm

PR #137, opened 2025-02-14
create validation stub to facilitate parallel work on validation and message passing

dknopik:
Two remaining tasks for the MVP are message passing and validation. This PR enables parallel work on this by adding a stub - and creating a todo location for validation and for message passing, allowing two team members to work simultaneously on one topic each.


anchor/network/src/validation.rs
```diff
@@ -0,0 +1,116 @@
+use crate::types::ssv_message::SignedSSVMessage;
+use crate::Network;
+use libp2p::gossipsub::{Message, MessageAcceptance};
+use ssz::Decode;
+use tracing::debug;
+
+// TODO taken from go-SSV as rough guidance. feel free to adjust as needed. https://github.com/ssvlabs/ssv/blob/e12abf7dfbbd068b99612fa2ebbe7e3372e57280/message/validation/errors.go#L55
+pub enum ValidationFailure {
```
Zacholme7, line 8: I see some errors in here that require qbft specific data. How do we plan on getting access to that data? or should these errors be used within qbft?

dknopik: excellent question :sweat_smile: 

if we really end up needing some of the internal qbft state, then this approach is too naive, and we will need some callback system for parts of the validation. We'll see. This was just a quick copy&paste.

Zacholme7:
Think we should just get this in or keep it as a draft for now?

dknopik:
>Think we should just get this in or keep it as a draft for now? 

marked it as draft as #147 includes it and we can therefore just merge that. As soon as that is done, we can close this one

dknopik:
Superseded by #152.

anchor/network/src/validation.rs
diegomrsantos: There's a potential double decoding cause we need to try to decode to check `UndecodableMessageData`. We are routing messages with the processor and receiving results through queues. It might be controversial, but we could have a queue for something like std::Resut<our::Result, ValidationFailure>. In this way, we could split the validation and do each part where it's more appropriate. Thoughts?

diegomrsantos: @Zacholme7 the validation here https://github.com/sigp/anchor/blob/1648e156af0e8319a12431619a06dfa2f1f7aa10/anchor/common/qbft/src/lib.rs#L436 seems the same as in https://github.com/ssvlabs/ssv/blob/main/message/validation/consensus_validation.go#L116

diegomrsantos: It seems there's a lot of overlap. What kind of validation are we supposed to do in the validator crate?

PR #138, opened 2025-02-14
Handshake followup

dknopik:
follow up on #125 to discuss some changes proposed in that PR


anchor/network/src/handshake/codec.rs
```diff
@@ -15,15 +17,52 @@ impl From<envelope::Error> for io::Error {
...
+    }
+
+    async fn write<T>(&mut self, io: &mut T, node_info: NodeInfo) -> io::Result<()>
+    where
+        T: AsyncWrite + Unpin + Send,
+    {
+        let envelope = node_info.seal(&self.keypair)?;
+        let raw = envelope.encode_to_vec()?;
+        io.write_all(&raw).await?;
+        io.close().await?;
```
jxs, line 57: we probably should also `flush` just in case, as we don't control the underlying implementations, on the worst case it's a noop

anchor/common/ssv_types/src/domain_type.rs
```diff
@@ -7,7 +7,8 @@ impl FromStr for DomainType {
     type Err = String;
 
     fn from_str(hex_str: &str) -> Result<Self, Self::Err> {
-        let bytes = hex::decode(hex_str).map_err(|e| format!("Invalid domain type hex: {}", e))?;
+        let bytes = hex::decode(hex_str.strip_prefix("0x").unwrap_or(hex_str))
```
jxs: are these related to this PR?

anchor/network/src/handshake/mod.rs
```diff
@@ -31,226 +26,148 @@ pub enum Error {
...
+    // NodeInfoProtocol is the protocol.ID used for handshake
+    const NODE_INFO_PROTOCOL: &str = "/ssv/info/0.0.1";
 
-        if let Err(e) = their_info.unmarshal(&envelope.payload) {
-            self.events.push(Event::Failed {
-                peer_id,
-                error: Error::NodeInfo(e),
-            });
-        }
+    let protocol = StreamProtocol::new(NODE_INFO_PROTOCOL);
```
jxs: nit: `NODE_INFO_PROTOCOL` seems to just be used here so I'd  just do:
```suggestion
    let protocol = StreamProtocol::new("/ssv/info/0.0.1");
```

anchor/network/src/network.rs
```diff
@@ -219,16 +243,16 @@ fn subnet_to_topic(subnet: SubnetId) -> IdentTopic {
     IdentTopic::new(format!("ssv.{}", *subnet))
 }
 
-fn handle_handshake_event(ev: Event) {
-    match ev {
-        Event::Completed {
+fn handle_handshake_result(result: Result<handshake::Completed, handshake::Failed>) {
```
jxs: this function is not being used right? But somehow `cargo` doesn't complain

anchor/network/src/handshake/mod.rs
```diff
@@ -31,226 +26,148 @@ pub enum Error {
...
-pub enum Event {
-    Completed {
-        peer_id: PeerId,
-        their_info: NodeInfo,
-    },
-    Failed {
-        peer_id: PeerId,
-        error: Error,
-    },
+pub struct Completed {
```
jxs, line 33: nit: we could probably comment these structures and functions in this module, for better understanding

jxs (dismissed):
This is very good Daniel! Thanks for working on it.
Left some suggestions, feel free to address them as you deem

anchor/common/ssv_types/src/domain_type.rs
dknopik: Ah no, this is a by-product of Diego and me fixing the same thing simultaneously, but a bit differently. I'll roll it back.

anchor/network/src/network.rs
dknopik: huh. yea, weird that it doesn't catch that :thinking: 

This is actually meant to be used - as the comment says, to update some internal records after the handshake.

jxs (approved):
LGTM, thanks Daniel!

PR #139, opened 2025-02-14
Decided msg handling

Zacholme7:
## Issue Addressed

Other part of #103 

## Proposed Changes

A pretty simple implementation. If we recieve a decided message, which is identified by having > 1 signature, we can immediately complete our instance. All verification is to be done upstream. The only remaining part of this is to validate the signatures in #137


anchor/common/qbft/src/lib.rs
```diff
@@ -232,30 +232,52 @@ where
...
-            OperatorId::from(*signer)
-        } else {
+        }
+
+        // The rest of the verification only pertains to messages with one signature
+        if wrapped_msg.signed_message.operator_ids().len() != 1 {
             warn!(
                 num_signers = wrapped_msg.signed_message.operator_ids().len(),
                 "Propose message only allows one signer"
             );
```
dknopik: This log message has to be moved a few lines down, right? And maybe reworded, as we might also end up with other types of messages here?

anchor/common/qbft/src/lib.rs
```diff
@@ -433,7 +461,11 @@ where
             return;
         }
 
-        // Update state
+        // Make sure we have not already accepted another proposal for this round.
+        if self.proposal_accepted_for_current_round {
+            warn!(from = ?operator_id, self=?self.config.operator_id(), "Proposal has already been accepted for this round");
+        }
```
dknopik: Do we have to bail out in here?

Zacholme7: yep thanks

anchor/common/qbft/src/lib.rs
Zacholme7: yep must have been a copy paste, thanks

anchor/common/qbft/src/lib.rs
```diff
@@ -793,6 +824,21 @@ where
         }
     }
 
+    // We have received a decided message
+    fn received_decided(&mut self, wrapped_msg: WrappedQbftMessage) {
+        // Make sure we have a quorum of signatures
+        if wrapped_msg.signed_message.operator_ids().len() >= self.config().quorum_size() {
+            self.received_decided(wrapped_msg);
```
dknopik: Wait, this is a recursive call.

Zacholme7: Im not sure why this is here, will remove and fix the condition.

dknopik (approved):
lgtm

PR #140, opened 2025-02-14
Domain type in discovery

diegomrsantos:
## Issue Addressed

Related to https://github.com/sigp/anchor/issues/35

## Proposed Changes

This adds the domain type predicate when discovering peers. We should connect only to peers in the same domain.


dknopik (approved):
LGTM, thanks!

PR #143, opened 2025-02-17
fix mdbook CI

dknopik:
Fix build of mdbook running on updates to unstable.

rudely stolen from:
https://github.com/sigp/lighthouse/pull/6891

### DIFF
--- a/.github/workflows/book.yml
+++ b/.github/workflows/book.yml
@@ -12,7 +12,7 @@ concurrency:
 jobs:
   build-and-upload-to-s3:
     if: github.repository_owner == 'sigp'
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@v4
 
@@ -20,8 +20,11 @@ jobs:
         uses: jontze/action-mdbook@v3
         with:
           token: ${{secrets.GITHUB_TOKEN}}
+          mdbook-version: "=0.4.44"
           use-mermaid: true
+          mermaid-version: "=0.14.1"
           use-linkcheck: true
+          linkcheck-version: "=0.7.7"
 
       - run: mdbook build
         working-directory: book
--- a/.github/workflows/book.yml
+++ b/.github/workflows/book.yml
@@ -12,7 +12,7 @@ concurrency:
 jobs:
   build-and-upload-to-s3:
     if: github.repository_owner == 'sigp'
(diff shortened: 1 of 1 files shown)


.github/workflows/book.yml
```diff
@@ -12,7 +12,7 @@ concurrency:
 jobs:
   build-and-upload-to-s3:
     if: github.repository_owner == 'sigp'
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
```
diegomrsantos: Do we want to use latest? New releases might break CI.

dknopik: Pinning a version might also break CI, as seen here. I feel like `latest` will break less as we update `mdbook` automatically as well.

So I guess it makes sense to either update all CI images automatically or all manually. The latter creates way more work over time I think. IMO its better to let it do its thing and fix it if it breaks.

diegomrsantos: It shouldn't break if a version is specified, see here https://github.com/jontze/action-mdbook?tab=readme-ov-file#fast-setup

magick93: FYI - https://github.blog/changelog/2024-11-05-notice-of-breaking-changes-for-github-actions/

IMO having `latest` is living dangerously. I mean we all need a bit more excitement in our lives.

dknopik:
@diegomrsantos @magick93 

Switched to pinned versions for the base image, mdbook, mermaid and linkcheck

diegomrsantos:
> @diegomrsantos @magick93
> 
> Switched to pinned versions for the base image, mdbook, mermaid and linkcheck

Thanks. It's also possible to use ~0.4.44 and get patch releases.

diegomrsantos (approved):
LGTM, thanks.

dknopik:
> Thanks. It's also possible to use ~0.4.44 and get patch releases.

Decided against that, as the update that caused this issue was a patch release: https://github.com/rust-lang/mdBook/blob/master/CHANGELOG.md#mdbook-0444

diegomrsantos:
> > Thanks. It's also possible to use ~0.4.44 and get patch releases.
> 
> Decided against that, as the update that caused this issue was a patch release: https://github.com/rust-lang/mdBook/blob/master/CHANGELOG.md#mdbook-0444

True, semantic version allows breaking changes in any version when the major is zero https://semver.org/#spec-item-4

PR #144, opened 2025-02-17
deduplicate types and redundant impls

dknopik:
While implementing network message passing, I noticed that we have a bunch of duplicated code - mainly `network::types::ssv_message` and `ssv_types::message`. 

There were also some remnants of parallel development (and the attempted independence of `qbft`):  `MessageID` vs `MessageId` vs `MsgId` and `OperatorId` vs `OperatorID`.

I now aligned those. Reasoning behind `Id` over `ID` is that it isn't an acronym, but short for `Identifier`.

Last, but not least: We had many locations where we manually implemented `Encode` and `Decode` - I changed this use the derive macro instead.

I do not claim that this PR is complete, there may be a few more places to optimize.


diegomrsantos (approved):
LGTM, thanks.

Zacholme7:
Thanks this was on my todo list. lgtm

Zacholme7 (approved):

PR #145, opened 2025-02-18
Fix handshake network mismatch

diegomrsantos:
## Issue Addressed

The go implementation sends and expects a "0x" prefix in the `networ_id` in `NodeInfo` - this is the same as the `DomainType`.

The current handshake fails with 

```
2025-02-18T14:31:09.753455Z DEBUG network::network: Handshake failed peer_id=16Uiu2HAmHFCLdpkJVbJsRjNzLRsDjRvyDY3mdndmDjduCQjfykSh error=NetworkMismatch { ours: "00000502", theirs: "0x00000502" }
```

## Proposed Changes

Add and remove the "0x" prefix during `NodeInfo` marshal and unmarshal respectively.


dknopik (dismissed):
Nice find! LGTM thanks!

(right after fixed test - was a bit to quick here :D)

dknopik (approved):

PR #146, opened 2025-02-18
Make QBFT and signature collection managers nongeneric

dknopik:
I think this PR will spark joy :)

I realized we do not actually need the stored slot clock in the QBFT and Signature Collection Managers. 

We can therefore get rid of it and only use it as a method parameter for the cleaners, avoiding generic propagation (which would get ugly in my other PR). :)


Zacholme7 (approved):
Nice. Very clean lgtm!

PR #147, opened 2025-02-18
Pass messages from network crate to managers

dknopik:
Passes messages to application by converting into messages understood by the QBFT and signing manager, and directly calling their functions meant for receiving network messages. These functions queue the messages for consumption by the corresponding long running tasks.

There is also some infrastructure created in this PR: 
- CommitteeId
- Partial Signature Message

It includes the changes from #137, and therefore supersedes it.

### DIFF
--- a/anchor/client/src/lib.rs
+++ b/anchor/client/src/lib.rs
@@ -145,11 +145,6 @@ impl Client {
         let subnet_tracker =
             start_subnet_tracker(database.watch(), network::SUBNET_COUNT, &executor);
 
-        // Start the p2p network
-        let network = Network::try_new(&config.network, subnet_tracker, executor.clone()).await?;
-        // Spawn the network listening task
-        executor.spawn(network.run(), "network");
-
         // Initialize slashing protection.
         let slashing_db_path = config.data_dir.join(SLASHING_PROTECTION_FILENAME);
         let slashing_protection =
@@ -361,6 +356,18 @@ impl Client {
             network_tx.clone(),
         )
         .map_err(|e| format!("Unable to initialize qbft manager: {e:?}"))?;
+
+        // Start the p2p network
+        let network = Network::try_new(
+            &config.network,
+            subnet_tracker,
+            qbft_manager.clone(),
+            signature_collector.clone(),
+            executor.clone(),
+        )
+        .await?;
+        // Spawn the network listening task
+        executor.spawn(network.run(), "network");
 
         let validator_store = AnchorValidatorStore::<_, E>::new(
             database.watch(),
--- a/anchor/common/ssv_types/src/cluster.rs
+++ b/anchor/common/ssv_types/src/cluster.rs
@@ -3,6 +3,7 @@ use derive_more::{Deref, From};
 use indexmap::IndexSet;
 use ssz_derive::{Decode, Encode};
 use types::{Address, Graffiti, PublicKeyBytes};
+use crate::committee::CommitteeId;
 
 /// Unique identifier for a cluster
 #[derive(Clone, Copy, Debug, Default, Eq, PartialEq, Hash, From, Deref)]
@@ -35,6 +36,10 @@ impl Cluster {
     /// Exception: Returns 0 if there are no cluster members
     pub fn get_f(&self) -> u64 {
         (self.cluster_members.len().saturating_sub(1) / 3) as u64
+    }
+
+    pub fn committee_id(&self) -> CommitteeId {
+        self.cluster_members.iter().cloned().collect::<Vec<_>>().into()
     }
 }
 
--- a/anchor/common/ssv_types/src/cluster.rs
+++ b/anchor/common/ssv_types/src/cluster.rs
@@ -1,9 +1,9 @@
+use crate::committee::CommitteeId;
 use crate::OperatorId;
 use derive_more::{Deref, From};
 use indexmap::IndexSet;
 use ssz_derive::{Decode, Encode};
 use types::{Address, Graffiti, PublicKeyBytes};
-use crate::committee::CommitteeId;
 
 /// Unique identifier for a cluster
 #[derive(Clone, Copy, Debug, Default, Eq, PartialEq, Hash, From, Deref)]
@@ -39,7 +39,11 @@ impl Cluster {
     }
 
     pub fn committee_id(&self) -> CommitteeId {
-        self.cluster_members.iter().cloned().collect::<Vec<_>>().into()
+        self.cluster_members
+            .iter()
+            .cloned()
+            .collect::<Vec<_>>()
+            .into()
     }
 }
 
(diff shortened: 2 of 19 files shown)


anchor/client/src/lib.rs
```diff
@@ -362,6 +357,18 @@ impl Client {
         )
         .map_err(|e| format!("Unable to initialize qbft manager: {e:?}"))?;
 
+        // Start the p2p network
+        let network = Network::try_new(
+            &config.network,
+            subnet_tracker,
+            qbft_manager.clone(),
```
diegomrsantos, line 364: Do you think the network should be aware of the QBFT Manager? Could we communicate between them using a channel?

anchor/common/ssv_types/src/cluster.rs
```diff
@@ -36,6 +37,14 @@ impl Cluster {
     pub fn get_f(&self) -> u64 {
         (self.cluster_members.len().saturating_sub(1) / 3) as u64
     }
+
+    pub fn committee_id(&self) -> CommitteeId {
```
diegomrsantos, line 41: Would it be better to move this to the file where `CommitteeId` is defined? It'd receive the cluster members and return the Id.

anchor/common/ssv_types/src/committee.rs
```diff
@@ -0,0 +1,33 @@
...
+use alloy::primitives::keccak256;
+use derive_more::{Deref, From};
+
+const COMMITTEE_ID_LEN: usize = 32;
+
+/// Unique identifier for a committee
+#[derive(Clone, Copy, Debug, Default, Eq, PartialEq, Hash, From, Deref)]
+pub struct CommitteeId(pub [u8; COMMITTEE_ID_LEN]);
+
+impl From<Vec<OperatorId>> for CommitteeId {
```
diegomrsantos, line 11: If we'd like to simplify the caller code we could implement sth like:

```rust
impl From<&IndexSet<OperatorId>> for CommitteeId {
    fn from(cluster_members: &IndexSet<OperatorId>) -> Self {
        let mut sorted: Vec<_> = cluster_members.iter().copied().collect();
        sorted.sort();
        let mut data: Vec<u8> = Vec::with_capacity(sorted.len() * 4);
        for id in sorted {
            data.extend_from_slice(&id.to_le_bytes());
        }
        keccak256(data).0.into()
    }
}
```

Then call` let committee_id: CommitteeId = (&self.cluster_members).into();`

anchor/client/src/lib.rs
dknopik: effectively, that's what the QBFT manager struct does. selecting the correct qbft instance and sending a message to it.

anchor/common/ssv_types/src/cluster.rs
dknopik: The one holding a `Cluster` should not need to fiddle with the fields and pass it somewhere - a utility function is clearer at the call site, I think. The actual logic (hashing the operator ids) is already contained in the `committee.rs` file.

anchor/client/src/lib.rs
diegomrsantos: Could this be achieved by decoupling the network and the manager and establishing communication between them using a single channel, through which all messages are routed within the manager?

anchor/common/ssv_types/src/cluster.rs
diegomrsantos: It's not important, but sth like `CommitteeId::from(&self.cluster_members);` could be more natural. But I see, your motivation with this function was to make it even less verbose at the caller.

anchor/client/src/lib.rs
dknopik: one more channel means one more channel that must do one of the following if it can't keep up:
- grow unbounded
- block
- drop messages

not sure if that is worth it

diegomrsantos: Those are good points, but let's pause this for a while. I noticed we're using unbounded channels for both per-instance communication and the network transmission. As in production it could lead to memory issues, I was thinking that switching to bounded channels might be a good idea. What do you think?

dknopik: Agree on that.

diegomrsantos: Regarding the coupling: 

1. Direct method calls between Network → QbftManager create tight coupling
2. Makes it difficult to test network layer without QBFT logic

Central Channel-Based Solution:
```rust
Network → (channel) → QbftManager → (per-instance channels) → QBFT instances
```

Benefits:

1. Enables testing network in isolation by:

    - Mocking the output channel

    - Verifying sent messages without QBFT dependencies

2. Allows testing QBFT manager with:

    -  Mocked network messages

3. Clearer boundaries between layers

diegomrsantos: I need to think deeper about it, but right now it seems to me the core challenge here isn't using a central channel, but what to do when a qbft instance's bounded channel is full.

dknopik: > I need to think deeper about it, but right now it seems to me the core challenge here isn't using a central channel, but what to do when a qbft instance's bounded channel is full.

We should discard the message. And that's fine. If it's happening due to temporary resource constraints, we can recover after (and maybe even operate partially). If we block, we can also recover, but block other incoming messages. If we grow unbounded, we crash.

The central channel makes it worse because we introduce another bottleneck.

dknopik:
Thinking about bottlenecks makes me reconsider validation again. What do you think about moving all validation behind the QBFT manager queues to parallelize them? and only do the most basic validation (e.g. is the message of interest to us) before?

anchor/client/src/lib.rs
diegomrsantos: > The central channel makes it worse because we introduce another bottleneck.

Could you elaborate more on how it's a bottleneck? Isn't it only the qbft instances that process messages? If a specific message can't be delivered it's dropped and the qbft instance won't participate in this consensus round.

anchor/network/src/network.rs
```diff
@@ -211,6 +218,132 @@ impl Network {
...
+            return;
+        };
+
+        match message.ssv_message().msg_type() {
+            MsgType::SSVConsensusMsgType => self.on_consensus_message_received(message),
+            MsgType::SSVPartialSignatureMsgType => self.on_signature_message_received(message),
+        }
+    }
+
+    fn on_consensus_message_received(&mut self, message: SignedSSVMessage) {
```
diegomrsantos, line 255: Could it be moved to the qbft manager?

diegomrsantos:
> Thinking about bottlenecks makes me reconsider validation again. What do you think about moving all validation behind the QBFT manager queues to parallelize them? and only do the most basic validation (e.g. is the message of interest to us) before?

Seems a great idea!

dknopik:
Closing this in favour of a new PR, as we have a new approach

Issue #75, opened 2024-12-18
Improve error handling in network file


diegomrsantos:
## Description

Avoid using `unwrap` and potentially replace `String` with an `Enum` for errors.

PR #148, opened 2025-02-18
Improve network error handling

diegomrsantos:
## Issue Addressed

closes https://github.com/sigp/anchor/issues/75

## Proposed Changes

This PR aims to improve network error handling. The key technical changes proposed include:

  - Enhanced Error Handling: Introduces specific error types for various network-related errors in the Discovery and Network modules using `thiserror`.
  - Detailed Error Messages: Provides more detailed and user-friendly error messages for issues like ENR key parsing, ENR building, and Discv5 initialization and startup errors.
  - Refactoring: Splits complex expressions into multiple lines for better readability and error handling, particularly around await calls and error mappings.
  - Error Propagation: Ensures that errors are properly propagated and handled in asynchronous functions, converting generic errors into specific ones.

### DIFF
--- a/anchor/network/src/discovery.rs
+++ b/anchor/network/src/discovery.rs
@@ -6,7 +6,7 @@ use std::task::{Context, Poll};
 use std::time::Instant;
 use subnet_tracker::SubnetId;
 
-use discv5::enr::{CombinedKey, NodeId};
+use discv5::enr::{CombinedKey, Error, NodeId};
 use discv5::libp2p_identity::{Keypair, PeerId};
 use discv5::multiaddr::Multiaddr;
 use discv5::{Discv5, Enr, ProtocolIdentity};
@@ -51,8 +51,8 @@ pub enum DiscoveryError {
     #[error("Failed to parse keypair into an ENR key: {0}")]
     EnrKey(String),
 
-    #[error("Failed to build ENR: {0}")]
-    EnrBuild(String),
+    #[error("Failed to build local ENR: {0}")]
+    EnrBuild(#[from] Error),
 
     #[error("Discv5 initialization error: {0}")]
     Discv5Init(String),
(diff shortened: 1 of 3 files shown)


anchor/network/src/discovery.rs
```diff
@@ -43,6 +43,24 @@ const TARGET_PEERS_FOR_GROUPED_QUERY: usize = 6;
...
+
+    #[error("Failed to build ENR: {0}")]
+    EnrBuild(String),
+
+    #[error("Discv5 initialization error: {0}")]
+    Discv5Init(String),
+
+    #[error("Discv5 start error: {0}")]
+    Discv5Start(String),
+}
```
dknopik: Do these all need to be `String`? Actual error types should be a bit clearer and should not affect the displayed message as the format string uses the `Display` impl of the inner type, which is the same thing as the `to_string` that you manually call below.

diegomrsantos: Do you mean the original error? If so, that's the type of the original errors, like in https://github.com/sigp/lighthouse/blob/stable/beacon_node/lighthouse_network/src/discovery/enr_ext.rs#L281

diegomrsantos: Good catch, thanks. I changed it for `EnrBuild` and I'll double-check the others.

anchor/network/src/discovery.rs
```diff
@@ -539,14 +566,11 @@ pub fn build_enr(enr_key: &CombinedKey, config: &Config) -> Result<Enr, String>
     }
 
     // set the "subnets" field on our ENR
-    let mut bitfield = BitVector::<U128>::new();
-    bitfield.set(9, true).unwrap();
-
-    builder.add_value::<Bytes>("subnets", &bitfield.as_ssz_bytes().into());
+    builder.add_value::<Bytes>("subnets", &BitVector::<U128>::new().as_ssz_bytes().into());
```
diegomrsantos, line 566: @dknopik This is about what we talked yesterday. Should we create an empty `BitVector` at this point?

dknopik: Yes, the subnets will be updates as the subnet tracker messages come in: https://github.com/sigp/anchor/pull/126/commits/9e9e485c4514b9b2de5a53b8693b5e7b77eb2d96

dknopik: as always, open for other suggestions though

diegomrsantos: From what I understand, it makes sense.

anchor/network/src/discovery.rs
diegomrsantos: Hopefully I covered everything now

dknopik (approved):
lgtm

Issue #32, opened 2024-10-23
BLS Key splitting tooling


AgeManning:
A tool that allows us to split a BLS key into shares to allow for testing on tests and networks.

PR #149, opened 2025-02-18
Keysplitter

Zacholme7:
## Issue Addressed

#32 

## Proposed Changes

This pr introduces our own key splitter tool. This will support 2 modes. Currently only manual mode is implemented
1. Manual mode. This is the same as the ssv keysplitter. All information must be manually input into 
2. Onchain mode. This will perform a quick sync of the chain on relevant events to index all of the owner nonces and the public keys of all of the operators. It will then use this information for keysplitting, making the UX a bit better. 

Edit: Have a PR ready to follow this that includes the onchain split, some logging, and general cleanup. Kept it separate to reduce the size.


Zacholme7:
Example to test with.

[cmd.txt](https://github.com/user-attachments/files/18871713/cmd.txt)
[key.json](https://github.com/user-attachments/files/18871721/key.json)

PR #152, opened 2025-02-20
Create message_validator crate

diegomrsantos:
## Issue Addressed

Extracts validation from https://github.com/sigp/anchor/pull/147.
Related to https://github.com/sigp/anchor/issues/161.

## Proposed Changes

In the end, we decided to go with a dedicated validator. The validation is more involved than expected. There are signature checks, database queries to validate committees, consensus state tracking, etc. They aren't suitable for the network task. 

## Additional Info
This is just putting the validation API in place. The actual validation will be done in future PRs.


anchor/common/message_validator/src/lib.rs
```diff
@@ -0,0 +1 @@
+mod validation;
```
dknopik: why a submodule?

anchor/common/message_validator/src/validation.rs
```diff
@@ -0,0 +1,116 @@
...
+    Accept,
+    Reject,
+    Ignore,
+}
+
+impl Validator {
+    pub(crate) fn validate(
+        &mut self,
+        _message: SignedSSVMessage,
+    ) -> std::result::Result<Result, Error> {
```
dknopik: nit: no need to specify the full type path

anchor/common/message_validator/src/validation.rs
```diff
@@ -0,0 +1,116 @@
...
+            | ValidationFailure::NoDuty
+            | ValidationFailure::EstimatedRoundNotInAllowedSpread => Result::Ignore,
+            _ => Result::Reject,
+        }
+    }
+}
+
+pub struct Validator;
+
+pub enum Error {}
```
dknopik: Why a separate error instead of just returning the `ValidationFailure`?

diegomrsantos: Do we need to return `ValidationFailure`? We return the local `Result` and `Error` is an error in the validation process. Maybe we don't need the Error, I don't know if it can fail yet.

diegomrsantos: Maybe it should be like this?

```rust
pub enum Result {
    Accept,
    Reject(ValidationFailure),
    Ignore,
}

impl Validator {
    pub(crate) fn validate(
        &mut self,
        _message: SignedSSVMessage,
    ) -> Result {

        Accept
    }
}
```

dknopik: Aaah, now I also understand why you specified the path for the std `Result` type.

Well, we will need to report the validation result to the network crate via a queue. So `validate` should not return anything, right?

Nesting a validation failure into the `Result` also feels incorrect: we have a specific validation failure, and this maps to a way we handle the message. And this will be what we report to `network`, right?

diegomrsantos: something like this?
```rust
pub fn start_validator_service(validator: Validator) -> (Sender<SignedSSVMessage>, Receiver<ValidationResult>) {
    // Channel for receiving messages to validate.
    let (msg_tx, msg_rx) = mpsc::channel::<SignedSSVMessage>();
    // Channel for sending back validation results.
    let (result_tx, result_rx) = mpsc::channel::<ValidationResult>();

    spawn(move || {
       
        for signed_msg in msg_rx {
           
            let validation_result = validator.validate(signed_msg);
            
            if result_tx.send(validation_result).is_err() {
                break;
            }
        }
    });

    (msg_tx, result_rx)
}
```

diegomrsantos: But I believe we'd need to adapt it to go through the processor as well

dknopik: I don't think we need a centralized validator task, as there are no central resources to manage. Just create a closure, and send it to a processor queue via `send_blocking`.

diegomrsantos: > Nesting a validation failure into the Result also feels incorrect

We might need it, cause the result can also be `Accept`. There's also a difference between `Reject` and `Ignore`.

dknopik: Yes, we need it, but we do not need to embed the validation failure into it, because we create the validation result (=the way we treat the message) from the validation failure, right? In other words, the failure variant determines whether to `Ignore` or `Reject`.

diegomrsantos: send_blocking doesn't work: 
```
61  |     pub fn send_blocking<F: FnOnce() + Send + 'static>(
    |                             ^^^^^^^^ required by this bound in `Sender::send_blocking`
```

How about (ignore the Result for now)?

```rust
pub fn start_validator_service(
    validator: Validator,
) -> (Sender<SignedSSVMessage>, Receiver<Result>) {
    let (msg_tx, mut msg_rx) = mpsc::channel::<SignedSSVMessage>(100);
    let (result_tx, result_rx) = mpsc::channel::<Result>(100);

    validator
        .processor
        .urgent_consensus
        .send_async(
            async move {
                while let Some(signed_msg) = msg_rx.recv().await {
                    let result = validate(signed_msg);
                    // If the send fails, we can assume the receiver is dropped.
                    if result_tx.send(result).await.is_err() {
                        break;
                    }
                }
            },
            "validator_service",
        )
        .unwrap();

    (msg_tx, result_rx)
}
```

anchor/message_validator/src/lib.rs
```diff
@@ -0,0 +1,208 @@
...
+            message,
+            action,
+        }
+    }
+}
+
+use processor::Senders;
+use tokio::sync::mpsc::error::TrySendError::{Closed, Full};
+use tokio::sync::mpsc::Sender;
+use tracing::{error, trace};
```
dknopik: nit: move to other imports

anchor/message_validator/src/lib.rs
```diff
@@ -0,0 +1,208 @@
...
+        }
+    }
+}
+
+pub struct Result {
+    pub message_id: MessageId,
+    pub propagation_source: PeerId,
+    pub message: SignedSSVMessage,
+    pub action: MessageAcceptance,
+}
```
dknopik: Still not sure about the name. What do you think about `Outcome` instead?

Also, it should provide the deserialised inner message: `QbftMessage` or `PartialSignatureMessages`.

anchor/network/src/network.rs
```diff
@@ -158,10 +166,22 @@ impl Network {
...
+                                                    message_id.clone(),
+                                                    propagation_source,
+                                                    deserialized_message
+                                                ) {
+                                                    Ok(()) => {
+                                                        trace!(?message_id, ?propagation_source, "Message validated");
+                                                    }
+                                                    Err(error) => {
+                                                        error!(?error, ?message_id, ?propagation_source, "Error during message validation");
+                                                    }
```
dknopik: These logs talk about validation, but this is just _triggering_ the validation.

anchor/network/src/network.rs
```diff
@@ -158,10 +166,22 @@ impl Network {
                                         );
                                         match SignedSSVMessage::from_ssz_bytes(&message.data) {
                                             Ok(deserialized_message) => {
-                                                debug!(msg = ?deserialized_message, "SignedSSVMessage deserialized");
+                                                trace!(msg = ?deserialized_message, "SignedSSVMessage deserialized");
```
dknopik: How do you feel about deserialising in the validator? basically passing it the `Vec<u8>`?

dknopik (commented):
Veeery nice. Some minor comments, but otherwise looks like the way to go to me. Thank you!

anchor/message_validator/src/lib.rs
diegomrsantos: thanks

anchor/network/src/network.rs
diegomrsantos: true, thanks

anchor/network/src/network.rs
diegomrsantos: Great catch, thanks

anchor/message_validator/src/lib.rs
dknopik: To be more precise: Both the `SignedSSVMessage` that already in there, and (optionally, if validation got that far), an enum containing either `QbftMessage` or `PartialSignatureMessages`

diegomrsantos: I like `Result` more, but the clash with the std one is annoying. 

Could we handle the inner message later?

diegomrsantos (commented):
Comments addressed

anchor/message_validator/src/lib.rs
dknopik: I would like to fix the API now, as I will be needing that when implementing passing the message to QBFT and the signature collector.

We don't need the full validation right now - just check what kind of message it is, try to deserialize the inner message, and pass it back. That would be enough for me to continue developing in parallel with you.

diegomrsantos: Added it, let me know what you think.

anchor/common/ssv_types/src/message.rs
```diff
@@ -76,7 +76,7 @@ impl Decode for MsgType {
 pub struct SSVMessage {
     msg_type: MsgType,
     msg_id: MessageId, // Fixed-size [u8; 56]
-    data: Vec<u8>,     // Variable-length byte array
+    pub data: Vec<u8>, // Variable-length byte array
```
dknopik: There is a getter `data()` available, adding `pub` here is not necessary.

anchor/common/ssv_types/src/message.rs
```diff
@@ -124,7 +124,7 @@ impl SSVMessage {
 pub struct SignedSSVMessage {
     signatures: Vec<Vec<u8>>, // Vec of Vec<u8>, max 13 elements, each up to 256 bytes
     operator_ids: Vec<OperatorId>, // Vec of OperatorID (u64), max 13 elements
-    ssv_message: SSVMessage,  // SSVMessage: Required field
+    pub ssv_message: SSVMessage, // SSVMessage: Required field
```
dknopik: Same here

anchor/message_validator/src/lib.rs
```diff
@@ -171,19 +212,46 @@ impl ValidatorService for Validator {
...
+                                    Err(failure) => {
+                                        trace!(
+                                            ?failure,
+                                            ?message_id,
+                                            ?propagation_source,
+                                            "Validation failure"
+                                        );
+                                        (
+                                            (&failure).into(),
+                                            Some(ValidatedMessage::new(deserialized_message, None)),
```
dknopik: nitpick: I like the `ValidatedMessage` name, but returning it in case of a validation failure is a bit confusing (same a few lines below). What do you think about returning `None` instead? This should also make the inner `Option` unnecessary.

dknopik (dismissed):
Nice! Two more nitpicks, but otherwise LGTM! Thank you for changing the API.

anchor/common/ssv_types/src/message.rs
diegomrsantos: Fixed! Can you approve it again?

anchor/message_validator/src/lib.rs
diegomrsantos: True. I got confused about the case where we still need to forward the message to other peers, but we don't need the deserialized version, only the id and source.

diegomrsantos: Do we even need the SignedSSVMessage?

anchor/message_validator/src/lib.rs
```diff
@@ -0,0 +1,285 @@
...
+                    .ok()
+                    .map(ValidatedSSVMessage::PartialSignatureMessage)
+                    .ok_or(ValidationFailure::UndecodableMessageData)
+            }
+        }
+    }
+}
+
+impl ValidatorService for Validator {
+    fn validate(
```
diegomrsantos: @dknopik what do you think about naming this `send_for_validation` or just `send`?

anchor/message_validator/src/lib.rs
dknopik: Yes, QBFT requires some information from it. ~We could consider setting an empty `data` vec, to avoid the duplicated memory...~ nope, cant do that either i think

diegomrsantos: What do we need? We can add this to ValidatedMessage instead of the whole signed message.

anchor/message_validator/src/lib.rs
dknopik: sounds good!

anchor/message_validator/src/lib.rs
dknopik: We need the whole message, as we may embed it into round change messages.

anchor/message_validator/src/lib.rs
```diff
@@ -0,0 +1,276 @@
...
+            | ValidationFailure::NoDuty
+            | ValidationFailure::EstimatedRoundNotInAllowedSpread => MessageAcceptance::Ignore,
+            _ => MessageAcceptance::Reject,
+        }
+    }
+}
+
+pub enum ValidatedSSVMessage {
+    QbftMessage(QbftMessage),
+    PartialSignatureMessage(PartialSignatureMessage),
```
dknopik: Sorry, did not notice this before: this needs to be `PartialSignatureMessages` - the plural. They use it to bundle multiple messages in one. It is also used if it only contains one message.

diegomrsantos: done

anchor/client/src/lib.rs
```diff
@@ -354,11 +355,16 @@ impl Client {
             network::SUBNET_COUNT,
         )?;
 
+        let (results_tx, results_rx) = mpsc::channel::<message_validator::Outcome>(9000);
```
jking-aus, line 358: try_send will drop messages if this channel fills but 9000 is likely oversized, assume this is placeholder for now and we can pivot to send and a retry mechanism once we know what demand looks like

anchor/message_validator/src/lib.rs
```diff
@@ -0,0 +1,277 @@
...
+        self: Arc<Self>,
+        message_id: MessageId,
+        propagation_source: PeerId,
+        message_data: Vec<u8>,
+    ) -> Result<(), Error> {
+        let validator = self.clone();
+        Ok(self.processor.urgent_consensus.send_blocking(
+            move || {
+                let (outcome, validated_message) =
+                    match SignedSSVMessage::from_ssz_bytes(&message_data) {
```
jking-aus, line 213: I had originally commented above questioning the use of send_blocking with the concern that this might exhaust the thread pool - I can see you're deserializing here now so that makes more sense given the nested stuff from ssv. 

I couldn't find one but if we aren't already doing so, we should be doing a pre-validation check of message size to avoid passing in large or incorrect data. Should also add gossipsub size limits for max message size.

jking-aus (dismissed):
lgtm - just note comments and would be nice to add a max message size check before deserialization before I forget, may do that myself if I get time

anchor/message_validator/src/lib.rs
diegomrsantos: There's a default max in gossipsub https://github.com/libp2p/rust-libp2p/pull/5906/files

dknopik (approved):

PR #153, opened 2025-02-20
create common component for message serialization and signing

dknopik:
Both the partial signature manager and the QBFT manager need to send messages. This PR introduces a component that schedules such tasks via the processor.

It also provides a mock message sender for use in tests.

### DIFF
--- a/anchor/message_sender/src/lib.rs
+++ b/anchor/message_sender/src/lib.rs
@@ -6,8 +6,15 @@ pub mod testing;
 pub use crate::network::*;
 use ssv_types::consensus::UnsignedSSVMessage;
 use ssv_types::message::SignedSSVMessage;
+use tokio::sync::mpsc::error::TrySendError;
 
 pub trait MessageSender: Send + Sync {
-    fn sign_and_send(&self, message: UnsignedSSVMessage);
-    fn send(&self, message: SignedSSVMessage);
+    fn sign_and_send(&self, message: UnsignedSSVMessage) -> Result<(), Error>;
+    fn send(&self, message: SignedSSVMessage) -> Result<(), Error>;
+}
+
+#[derive(Debug)]
+pub enum Error {
+    Processor(TrySendError<processor::WorkItem>),
+    NetworkQueueClosed,
 }
--- a/anchor/message_sender/src/lib.rs
+++ b/anchor/message_sender/src/lib.rs
@@ -1,0 +1,13 @@
+mod network;
+
+#[cfg(feature = "testing")]
+pub mod testing;
+
+pub use crate::network::*;
+use ssv_types::consensus::UnsignedSSVMessage;
+use ssv_types::message::SignedSSVMessage;
+
+pub trait MessageSender: Send + Sync {
+    fn sign_and_send(&self, message: UnsignedSSVMessage);
+    fn send(&self, message: SignedSSVMessage);
+}
--- a/anchor/message_sender/src/lib.rs
+++ b/anchor/message_sender/src/lib.rs
@@ -6,11 +6,16 @@ pub mod testing;
 pub use crate::network::*;
 use ssv_types::consensus::UnsignedSSVMessage;
 use ssv_types::message::SignedSSVMessage;
+use ssv_types::CommitteeId;
 use tokio::sync::mpsc::error::TrySendError;
 
 pub trait MessageSender: Send + Sync {
-    fn sign_and_send(&self, message: UnsignedSSVMessage) -> Result<(), Error>;
-    fn send(&self, message: SignedSSVMessage) -> Result<(), Error>;
+    fn sign_and_send(
+        &self,
+        message: UnsignedSSVMessage,
+        committee_id: CommitteeId,
+    ) -> Result<(), Error>;
+    fn send(&self, message: SignedSSVMessage, committee_id: CommitteeId) -> Result<(), Error>;
 }
 
 #[derive(Debug)]
--- a/anchor/message_sender/src/network.rs
+++ b/anchor/message_sender/src/network.rs
@@ -1,5 +1,4 @@
 use crate::{Error, MessageSender};
-use database::{NetworkState, UniqueIndex};
 use openssl::error::ErrorStack;
 use openssl::hash::MessageDigest;
 use openssl::pkey::{PKey, Private};
@@ -7,13 +6,12 @@ use openssl::rsa::Rsa;
 use openssl::sign::Signer;
 use ssv_types::consensus::UnsignedSSVMessage;
 use ssv_types::message::SignedSSVMessage;
-use ssv_types::msgid::DutyExecutor;
-use ssv_types::OperatorId;
+use ssv_types::{CommitteeId, OperatorId};
 use ssz::Encode;
 use std::sync::Arc;
 use subnet_tracker::SubnetId;
+use tokio::sync::mpsc;
 use tokio::sync::mpsc::error::TrySendError;
-use tokio::sync::{mpsc, watch};
 use tracing::{debug, error, warn};
 
 const SIGNER_NAME: &str = "message_sign_and_send";
@@ -23,13 +21,16 @@ pub struct NetworkMessageSender {
     processor: processor::Senders,
     network_tx: mpsc::Sender<(SubnetId, Vec<u8>)>,
     private_key: PKey<Private>,
-    network_state_rx: watch::Receiver<NetworkState>,
     operator_id: OperatorId,
     subnet_count: usize,
 }
 
 impl MessageSender for Arc<NetworkMessageSender> {
-    fn sign_and_send(&self, message: UnsignedSSVMessage) -> Result<(), Error> {
+    fn sign_and_send(
+        &self,
+        message: UnsignedSSVMessage,
+        committee_id: CommitteeId,
+    ) -> Result<(), Error> {
         if self.network_tx.is_closed() {
             return Err(Error::NetworkQueueClosed);
         }
@@ -58,14 +59,14 @@ impl MessageSender for Arc<NetworkMessageSender> {
                             return;
                         }
                     };
-                    sender.do_send(message);
+                    sender.do_send(message, committee_id);
                 },
                 SIGNER_NAME,
             )
             .map_err(Error::Processor)
     }
 
-    fn send(&self, message: SignedSSVMessage) -> Result<(), Error> {
+    fn send(&self, message: SignedSSVMessage, committee_id: CommitteeId) -> Result<(), Error> {
         if self.network_tx.is_closed() {
             return Err(Error::NetworkQueueClosed);
         }
@@ -75,7 +76,7 @@ impl MessageSender for Arc<NetworkMessageSender> {
             .urgent_consensus
             .send_blocking(
                 move || {
-                    sender.do_send(message);
+                    sender.do_send(message, committee_id);
                 },
                 SENDER_NAME,
             )
@@ -88,7 +89,6 @@ impl NetworkMessageSender {
         processor: processor::Senders,
         network_tx: mpsc::Sender<(SubnetId, Vec<u8>)>,
         private_key: Rsa<Private>,
-        network_state_rx: watch::Receiver<NetworkState>,
         operator_id: OperatorId,
         subnet_count: usize,
     ) -> Result<Arc<Self>, String> {
@@ -98,20 +98,13 @@ impl NetworkMessageSender {
             processor,
             network_tx,
             private_key,
-            network_state_rx,
             operator_id,
             subnet_count,
         }))
     }
 
-    fn do_send(&self, message: SignedSSVMessage) {
-        let subnet = match self.determine_subnet(&message) {
-            Ok(subnet) => subnet,
-            Err(err) => {
-                error!(?err, "Unable to determine subnet for outgoing message");
-                return;
-            }
-        };
+    fn do_send(&self, message: SignedSSVMessage, committee_id: CommitteeId) {
+        let subnet = SubnetId::from_committee(committee_id, self.subnet_count);
         match self.network_tx.try_send((subnet, message.as_ssz_bytes())) {
             Ok(_) => debug!(?subnet, "Successfully sent message to network"),
             Err(TrySendError::Closed(_)) => warn!("Network queue closed (shutting down?)"),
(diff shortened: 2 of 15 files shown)


anchor/message_sender/src/network.rs
```diff
@@ -0,0 +1,141 @@
...
+impl MessageSender for Arc<NetworkMessageSender> {
+    fn sign_and_send(&self, message: UnsignedSSVMessage) {
+        let sender = self.clone();
+        self.processor
+            .urgent_consensus
+            .send_blocking(
+                move || {
+                    let signature = match sender.sign(&message) {
+                        Ok(signature) => signature,
+                        Err(err) => {
```
diegomrsantos, line 45: Do you know if we need to propagate this error?

anchor/message_sender/src/network.rs
```diff
@@ -0,0 +1,141 @@
...
+                            return;
+                        }
+                    };
+                    let message = match SignedSSVMessage::new(
+                        vec![signature],
+                        vec![sender.operator_id],
+                        message.ssv_message,
+                        message.full_data,
+                    ) {
+                        Ok(signature) => signature,
```
diegomrsantos: very nit: signed_message

anchor/message_sender/src/network.rs
```diff
@@ -0,0 +1,141 @@
...
+                        Err(err) => {
+                            error!(?err, "Creating signed message failed!");
+                            return;
+                        }
+                    };
+                    sender.do_send(message);
+                },
+                SIGNER_NAME,
+            )
+            .unwrap_or_else(|e| warn!("Failed to send to processor: {}", e));
```
diegomrsantos: maybe it's not necessary as I see that we already log in the processor

anchor/message_sender/src/lib.rs
```diff
@@ -0,0 +1,13 @@
+mod network;
+
+#[cfg(feature = "testing")]
+pub mod testing;
+
+pub use crate::network::*;
+use ssv_types::consensus::UnsignedSSVMessage;
+use ssv_types::message::SignedSSVMessage;
+
+pub trait MessageSender: Send + Sync {
```
diegomrsantos, line 12: Very nice. In my PR, I was thinking about a way to split the "interface" and the implementation. First I created two crates, one with the trait and types and another with the implementation. The component C, which needs the functionality, imports only the interface crate, and who create C imports the implementation. But maybe two crates is too much overhead. I think we can achieve the same using one crate with different modules and features. Wdyt?

anchor/message_sender/src/network.rs
dknopik: This is one of those errors that really really should never happen, and where there is not really a way to fall back. So there is no reason to propagate it somewhere.

anchor/message_sender/src/network.rs
dknopik: Good catch, thanks!

anchor/message_sender/src/network.rs
dknopik: True, will change it

anchor/message_sender/src/network.rs
diegomrsantos: According to Murphy's law: Anything that can go wrong will go wrong :) It might be useful to test what happens to the rest of the system when this happens. I'll keep this in mind in my tasks.

anchor/message_sender/src/lib.rs
dknopik: Yeah, I was also tempted to do that. This is for example what we do for the `validator_store`: This only contains the trait and types used in the interface, and there are creates `lighthouse_validator_store` in LH repo and  `anchor_validator_store` in Anchor repo. 

However, as you said, for small stuff like this, it is likely too much noise and too little benefit to split it up.

diegomrsantos: One benefit is enforcing modularity/encapsulation and to make sure that clients of the crate can only use what they are supposed to and nothing else. As a crate that provides a functionality grows, it's tempting for people to just use implementation details directly, bypassing the API when it's more convenient.

diegomrsantos: That's the whole point of the module system, right? Consider the message validator. It's harder to do that with only one crate as the network needs access to the interface and the client module needs access to the implementation, so both need to be public. That's why I mentioned features, I guess that's the only option if there's only one crate.

anchor/message_sender/src/network.rs
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    fn sign(&self, message: &UnsignedSSVMessage) -> Result<Vec<u8>, ErrorStack> {
+        let serialized = message.ssv_message.as_ssz_bytes();
+        let mut signer = Signer::new(MessageDigest::sha256(), &self.private_key)?;
+        signer.update(&serialized)?;
+        signer.sign_to_vec()
+    }
+
+    fn determine_subnet(&self, message: &SignedSSVMessage) -> Result<SubnetId, String> {
```
diegomrsantos: How about a specific error instead of a String?

anchor/message_sender/src/network.rs
```diff
@@ -0,0 +1,141 @@
...
+
+impl NetworkMessageSender {
+    pub fn new(
+        processor: processor::Senders,
+        network_tx: mpsc::Sender<(SubnetId, Vec<u8>)>,
+        private_key: Rsa<Private>,
+        database: watch::Receiver<NetworkState>,
+        operator_id: OperatorId,
+        subnet_count: usize,
+    ) -> Result<Arc<Self>, String> {
```
diegomrsantos, line 94: Would it be better to return only `Self` and let the caller decide if and when to wrap it in an `Arc`?

anchor/message_sender/src/network.rs
```diff
@@ -0,0 +1,141 @@
...
+            .unwrap_or_else(|e| warn!("Failed to send to processor: {}", e));
+    }
+}
+
+impl NetworkMessageSender {
+    pub fn new(
+        processor: processor::Senders,
+        network_tx: mpsc::Sender<(SubnetId, Vec<u8>)>,
+        private_key: Rsa<Private>,
+        database: watch::Receiver<NetworkState>,
```
diegomrsantos: a bit nit: how about `network_state_rx`?

anchor/message_sender/src/network.rs
```diff
@@ -0,0 +1,141 @@
...
+        let subnet = match self.determine_subnet(&message) {
+            Ok(subnet) => subnet,
+            Err(err) => {
+                error!(?err, "Unable to determine subnet for outgoing message");
+                return;
+            }
+        };
+        match self.network_tx.try_send((subnet, message.as_ssz_bytes())) {
+            Ok(_) => debug!(?subnet, "Successfully sent message to network"),
+            Err(TrySendError::Closed(_)) => warn!("Network queue closed (shutting down?)"),
```
diegomrsantos, line 110: This error seems more critical. Should we propagate it and shut down the system?

anchor/message_sender/src/network.rs
```diff
@@ -0,0 +1,141 @@
...
+                let Some(metadata) = database.metadata().get_by(&pubkey) else {
+                    return Err(format!("Unknown validator: {pubkey}"));
+                };
+                let Some(cluster) = database.clusters().get_by(&metadata.cluster_id) else {
+                    return Err(format!(
+                        "Inconsistent database, no cluster for validator: {pubkey}"
+                    ));
+                };
+                cluster.committee_id()
+            }
```
diegomrsantos: Would it make sense to move this to the qbft manager and send the `committee_id` as another parameter in sign_and_send?

anchor/message_sender/src/testing.rs
```diff
@@ -0,0 +1,39 @@
+use crate::MessageSender;
+use ssv_types::consensus::UnsignedSSVMessage;
+use ssv_types::message::SignedSSVMessage;
+use ssv_types::OperatorId;
+use tokio::sync::mpsc;
+
+pub struct TestingMessageSender {
```
diegomrsantos: how about `MockMessageSender` or `NoopSignerMessageSender`?

anchor/message_sender/src/testing.rs
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    fn send(&self, message: SignedSSVMessage) {
+        self.message_tx.send(message).unwrap();
+    }
+}
+
+impl TestingMessageSender {
+    pub fn new(
+        message_tx: mpsc::UnboundedSender<SignedSSVMessage>,
```
diegomrsantos, line 37: nit: this could be created inside the constructor if no assertion is needed in the tests

anchor/message_sender/src/network.rs
```diff
@@ -0,0 +1,141 @@
...
+            Some(DutyExecutor::Validator(pubkey)) => {
+                let database = self.database.borrow();
+                let Some(metadata) = database.metadata().get_by(&pubkey) else {
+                    return Err(format!("Unknown validator: {pubkey}"));
+                };
+                let Some(cluster) = database.clusters().get_by(&metadata.cluster_id) else {
+                    return Err(format!(
+                        "Inconsistent database, no cluster for validator: {pubkey}"
+                    ));
+                };
```
Zacholme7: Im a bit rusty on what I wrote for the db, but cant this just be `database.cluster().get_by(&pubkey)`?

```
/// All of the clusters in the network
/// Primary: cluster id. uniquely identifies a cluster
/// Secondary: public key of the validator. uniquely identifies a cluster
/// Tertiary: owner of the cluster. uniquely identifies a cluster
pub(crate) type ClusterMultiIndexMap =
    MultiIndexMap<ClusterId, PublicKeyBytes, Address, Cluster, UniqueTag, UniqueTag>;
```

dknopik: Ah yes, I think so, thanks!

anchor/message_sender/src/network.rs
dknopik: This might make sense. I am currently implementing the messaging infrastructure for the `signature_collector`, and I need to see if this would work there as well.

anchor/message_sender/src/network.rs
dknopik: To propagate it, we would need a return queue to the sender, and the sender would need to wait for it, defeating the purpose of putting this on another thread in the first place.

dknopik:
Thanks for all your feedback <3

I am currently integrating this into the signature collector, and will likely gather some ways to also improve this in the process of doing so. I will address your feedback as soon as I'm fully confident with the design.

anchor/message_sender/src/network.rs
diegomrsantos: I believe this is fine If we can guarantee that this error happens if and only if the system is already shutting down. Otherwise, the client can keep calling `sign_and_send` and the network never receives a msg as it fails silently, right?

diegomrsantos: We could also add this at the beginning of `sign_and_send`

```rust
 if self.network_tx.is_closed() {
            // return error
        }
```

anchor/message_sender/src/lib.rs
dknopik: I don't get your point. I am using modules. How are features related? They are only for conditional compilation.

anchor/message_sender/src/network.rs
dknopik: Usually I'd agree, but we impl `MessageSender` on `Arc<NetworkMessageSender>`. So it is only useful wrapped.

anchor/message_sender/src/testing.rs
dknopik: But then we have to return `(Self, Receiver)`, which is also kind of ugly.

anchor/message_sender/src/lib.rs
diegomrsantos: something like this

```rust
// src/lib.rs

#[cfg(feature = "api")]
pub mod api;

#[cfg(feature = "impl")]
pub mod message_sender;
```

anchor/qbft_manager/src/lib.rs
```diff
@@ -101,10 +92,8 @@ pub struct QbftManager {
     validator_consensus_data_instances: Map<ValidatorInstanceId, ValidatorConsensusData>,
     // All of the QBFT instances that are voting on beacon data
     beacon_vote_instances: Map<CommitteeInstanceId, BeaconVote>,
-    // Private key used for signing messages
-    pkey: Arc<PKey<Private>>,
-    // Channel to pass signed messages along to the network
-    network_tx: mpsc::UnboundedSender<SignedSSVMessage>,
+    // Utility to sign and serialize network messages
+    message_sender: Arc<dyn MessageSender>,
```
diegomrsantos, line 96: Have you considered using generics?

anchor/message_sender/src/network.rs
```diff
@@ -0,0 +1,121 @@
+use crate::{Error, MessageSender};
```
diegomrsantos, line 1: a bit nit: is network a good name for this file?

anchor/message_sender/src/lib.rs
dknopik: How do the features improve things? If the user decides to use a type with my approach, they import the module. If the user decides to use a type with that approach, they enable the feature. We can't protect the user against ill-advised use.

anchor/qbft_manager/src/lib.rs
dknopik: I expected resistance introducing generics, because they would propagate all over the place, so I didn't bother.

anchor/message_sender/src/network.rs
dknopik: what would you prefer?

anchor/qbft_manager/src/lib.rs
diegomrsantos: It depends. It's annoying when the actual type value needs to be passed deep down in the layers. In this case, the type parameter is defined by the `manager`, and the value is determined by the `client`, there's only one level. Therefore, IMO it's fine.

anchor/message_sender/src/network.rs
diegomrsantos: maybe `msg_sender` or just `sender`?

anchor/subnet_tracker/src/lib.rs
```diff
@@ -87,11 +93,7 @@ async fn subnet_tracker(
         // send a `Leave` event.
         for subnet in previous_subnets.difference(&current_subnets) {
             debug!(?subnet, "send leave");
-            if tx
-                .send(SubnetEvent::Leave(SubnetId(*subnet)))
-                .await
-                .is_err()
-            {
+            if tx.send(SubnetEvent::Leave(*subnet)).await.is_err() {
                 warn!("Network no longer listening for subnets");
```
diegomrsantos, line 97: Maybe it's an error?

anchor/qbft_manager/src/lib.rs
dknopik: the manager would need to be generic as well, so the network & validator_store would need to be generic, et cetera

anchor/message_sender/src/network.rs
dknopik: but that is redundant with the crate name. the module name specifies what kind of sender it is: the one sending it to the network

anchor/network/src/network.rs
```diff
@@ -212,7 +215,19 @@ impl Network {
...
+                            if let Err(err) = self.gossipsub().publish(subnet_to_topic(subnet_id), message) {
+                                error!(?err, "Failed to publish message");
+                            }
+                        }
+                        None => {
+                            error!("message queue was closed");
+                            return;
+                        }
+                    }
+                }
```
diegomrsantos, line 230: This is extremely clean! :)

diegomrsantos (approved):
Looks really good!

PR #158, opened 2025-02-24
Peer manager heartbeat

dknopik:
Implement basic heartbeat for peer manager.

Currently, we only check if we need more peers for some subnets, and try to dial if we find some good peers in the peer store. If we can not dial enough peers, we try to discover more.


dknopik:
NOTE: converted to draft as I found a bug

anchor/network/src/peer_manager.rs
```diff
@@ -127,19 +153,59 @@ impl PeerManager {
...
+        let mut actions = ConnectActions::none();
+        self.determine_actions_for_subnets(
+            &mut actions,
+            &self.needed_subnets.iter().copied().collect::<Vec<_>>(),
+        );
+
+        (!actions.discover.is_empty() || !actions.dial.is_empty()).then_some(actions)
+    }
+
+    fn candidate_peers(&self) -> Vec<(&PeerId, &PeerRecord<Enr>)> {
```
jking-aus, line 198: do we ever prune unreachable ENRs? appreciate a restart would fix this but might be worth removing addresses that fail to connect from peer manager? Apologies if this behaviour is captured in libp2p somewhere and I'm unaware

jking-aus (approved):
lgtm - just note comment to consider pruning unreachable peers

anchor/network/src/peer_manager.rs
dknopik: already metioned this to João. libp2p should handle it. not sure if it made it in in the currently merged version though

PR #159, opened 2025-02-25
Message sending for Signature Collector

dknopik:
Updates signature collector to send messages via the common message sender component. 

There are some bigger changes because I was unaware how the signatures are bundled for the `Committee` role. Let me know if something is unclear.


jking-aus (approved):
lgtm -- nice work daniel

PR #160, opened 2025-02-25
SignedSSVMessage and SSVMessage validation

diegomrsantos:
## Issue Addressed

It's related to https://github.com/sigp/anchor/issues/161

## Proposed Changes

We move basic validation inside the messages

## Additional Info

This is only the first phase of validation, more PRs will come.


anchor/network/src/network.rs
```diff
@@ -57,11 +60,39 @@ pub enum NetworkError {
...
+        // _ => MessageAcceptance::Reject,
+    }
+}
+
 pub struct Network {
     swarm: Swarm<AnchorBehaviour>,
     subnet_event_receiver: mpsc::Receiver<SubnetEvent>,
     peer_id: PeerId,
     node_info: NodeInfo,
+    domain_type: DomainType,
```
dknopik, line 69: There is a domain_type in the node info, right?

diegomrsantos: True, I'll use it. What's the plan to update the subnets in the NodeInfo with this current design?

diegomrsantos: However, since the `network_id` in the `NodeInfo` is in `String` format, we would need to convert it for each message.

dknopik: ah, yeah, forgot about that. nevermind then

Zacholme7 (commented):
Good on my end! Gj. I like the detailed validation errors. These will be good in the qbft crate too. Ill let kingy give this a glad too.

anchor/qbft_manager/src/tests.rs
```diff
@@ -91,19 +91,31 @@ where
...
-                .expect("If consensus was reached, this exists");
+                .expect("If consensus was reached, this must exist");
             assert!(
                 aggregated_commit.signatures().len() as u64
                     >= (self.tester.size as u64 - self.tester.size.get_f())
             );
         }
+
+        // At this point the channel has closed, so if we never received anything, fail the test.
+        assert!(
```
Zacholme7, line 115: I dont think we want to assert here. There are tests we can write where it wont reach consensus such as not having a quorum online. Maybe just passing a flag in to `verify_consensus` stating if it should reach consensus or not is a better way to do this?

diegomrsantos: Would it be enough not to call this function in these tests?

diegomrsantos: When it's called and doesn't fail, do we expect that consensus has been reached? According to the way it was written, it doesn't fail if the channel is dropped and no message is received.

Zacholme7: After thinking about this for a bit, if there is a test where we expect consensus not to be reached it is probably better to have an explicit function named something like `verify_consensus_not_reached`. So I think these changes look good. 

Consensus should always eventually be reached when it is called.

diegomrsantos: We could rename it to `has_consensus_been_reached` and return a bool. But not sure about the second assertion.

Zacholme7: I think that sounds good

diegomrsantos: The problem is that it makes two assertions. How about we leave it like that and make the change if/when the need arises?

Zacholme7: Yea lets go with that

diegomrsantos:
This is waiting for https://github.com/sigp/anchor/pull/152 to be merged and then I'll change it to ready.

jking-aus (approved):
lgtm

PR #164, opened 2025-02-28
Add domain to ENR and correct checking

dknopik:
1. Put our domaintype into the ENR so that we can be discovered
2. The domaintype is stored as four bytes, not as string. Fixes parsing

### DIFF
--- a/anchor/network/src/discovery.rs
+++ b/anchor/network/src/discovery.rs
@@ -327,16 +327,8 @@ impl Discovery {
         let local_domain_type = self.domain_type.clone();
 
         let domain_type_predicate = move |enr: &Enr| {
-            if let Some(Ok(domain_type_str)) = enr.get_decodable::<String>("domaintype") {
-                if let Ok(domain_type_bytes) = <[u8; 4]>::try_from(domain_type_str.as_bytes()) {
-                    local_domain_type == DomainType::from(domain_type_bytes)
-                } else {
-                    trace!(
-                        domain_type = domain_type_str.as_bytes(),
-                        "ENR domain type is not 4 bytes",
-                    );
-                    false
-                }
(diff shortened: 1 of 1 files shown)


anchor/network/src/discovery.rs
```diff
@@ -327,16 +327,8 @@ impl Discovery {
...
-                if let Ok(domain_type_bytes) = <[u8; 4]>::try_from(domain_type_str.as_bytes()) {
-                    local_domain_type == DomainType::from(domain_type_bytes)
-                } else {
-                    trace!(
-                        domain_type = domain_type_str.as_bytes(),
-                        "ENR domain type is not 4 bytes",
-                    );
-                    false
-                }
+            if let Some(Ok(domain_type)) = enr.get_decodable::<[u8; 4]>("domaintype") {
```
diegomrsantos, line 330: Could we handle the else case?

dknopik: What do you mean? There is an else below. If the field can not decoded, the enr is not acceptable.

diegomrsantos: Sorry, I should have been clearer. We don't log if there's an error when decoding. 

It was like that before but we also don't log if the field doesn't exist. Maybe we should if it's expected to always exist at this point.

jking-aus (approved):

anchor/network/src/discovery.rs
dknopik: It's not expected to always exist. For example, non-SSV ENRs might arrive at this point, and they won't have this field. I will add a trace log for debugging anyway. Thanks!

diegomrsantos: I think this should avoid non-ssv nodes https://github.com/sigp/anchor/blob/unstable/anchor/network/src/discovery.rs#L100

dknopik: Oh right!

PR #166, opened 2025-03-04
Properly handle messages sent by ourselves

dknopik:
The QBFT instance needs to handle messages that it has sent. For this, it needs the signed message - for example, so that it can embed signed quorums in the round change messages.

This PR adds an optional callback mechanism to the common message sender. This is used in the QBFT instance to enqueue signed message for handling by the same instance.

While we're at it, we get rid of `qbft::Message`, which is no longer adequate as outgoing message type. The different variants are not treated differently at all in the current design, where we directly use SSV-specific messages.


jking-aus (approved):
nice -- lgtm

PR #167, opened 2025-03-04
Embed message id into outgoing qbft messages

dknopik:
We need to set the correct message ID for outgoing QBFT messages - previously, an all zero ID was used. This PR adds the infrastructure to set this to the message ID as expected by the protocol.


jking-aus (dismissed):
lgtm

jking-aus (approved):

PR #168, opened 2025-03-04
Handle validated messages

dknopik:
This PR implements passing the validated messages to the QBFT manager or signature collector.


anchor/network/src/network.rs
```diff
@@ -64,18 +66,23 @@ pub struct Network<V: ValidatorService> {
...
     // p2p network.
+    #[allow(clippy::too_many_arguments)]
     pub async fn try_new(
         config: &Config,
         subnet_event_receiver: mpsc::Receiver<SubnetEvent>,
         message_rx: mpsc::Receiver<(SubnetId, Vec<u8>)>,
         message_validator: V,
-        results_rx: mpsc::Receiver<message_validator::Outcome>,
+        qbft_manager: Option<Arc<QbftManager>>,
+        signature_collector: Option<Arc<SignatureCollectorManager>>,
```
diegomrsantos: We could create a new crate `message_receiver` and pass it to the network. Then we'd move the validation and message routing there. It'd make the network less aware of specifics, especially now that it handles only bytes, and hopefully help with the `clippy::too_many_arguments` - it's probably a sign it's doing too much.

dknopik: ooh, sounds good! 

this new crate could also then handle filtering of relevant messages - which, as i have realized just now, is not implemented for the signature collector I think.

Thanks!

diegomrsantos: What do you think about moving the validation to the receiver? Then, we could have only a channel reporting Reject/Ignore to the Network. For Acceptance, we'd forward the message to the manager, and the Network would receive the result from the Sender.

diegomrsantos: It'd reduce copying and moving messages around through channels.

dknopik: Sounds good, will do

dknopik: done, what do you think?

dknopik: @diegomrsantos can you please take a look if I implemented it as you meant?

jking-aus (dismissed):
lgtm - just fix merge conf

dknopik:
@jking-aus thanks - before you merge this, lets wait for diego's feedback

anchor/message_receiver/src/manager.rs
```diff
@@ -0,0 +1,105 @@
...
+use signature_collector::SignatureCollectorManager;
+use ssv_types::msgid::DutyExecutor;
+use std::sync::Arc;
+use tokio::sync::watch;
+use tracing::error;
+
+const RECEIVER_NAME: &str = "message_receiver";
+
+/// A message receiver that passes messages to responsible managers.
+pub struct ManagerMessageReceiver<V: ValidatorService + 'static> {
```
diegomrsantos, line 24: nit: should manager be at the end?

anchor/message_validator/src/lib.rs
```diff
@@ -200,78 +164,18 @@ impl Validator {
...
-                    }
-                }
-            },
-            "validator",
-        )?)
+    fn validate(&self, message_data: Vec<u8>) -> Result<ValidatedMessage, ValidationFailure> {
+        match SignedSSVMessage::from_ssz_bytes(&message_data) {
+            Ok(deserialized_message) => {
+                trace!(msg = ?deserialized_message, "SignedSSVMessage deserialized");
+                self.do_validate(&deserialized_message)?;
```
diegomrsantos, line 171: I deleted this fn as it's doing nothing

diegomrsantos (dismissed):
Looks great! Thanks for that. Only two minor comments.

anchor/message_receiver/src/manager.rs
dknopik: The reason why I put it in front is that it is a receiver, that forwards messages to the managers, not a receiver that _is_ a manager.

But I am also not happy with the name. I just wanted to avoid `MessageReceiverImpl` at all costs. :D

anchor/message_validator/src/lib.rs
dknopik: Yeah, as noted there will be merge conflicts anyway. So you can remove it then.

anchor/message_receiver/src/manager.rs
diegomrsantos: Maybe NetworkMessageReceiver or SSVMessageReceiver?

jking-aus (approved):

PR #163, opened 2025-02-28
Properly set subscribed subnets in ENR and NodeInfo

dknopik:

PR #169, opened 2025-03-04
Add ENR-related logging

dknopik:
Follow-up to #164 and #163 to address suggestions for additional logging by @diegomrsantos, which I meant to add but was too slow before @jking-aus merged everything


diegomrsantos (approved):
LGTM, thanks!

Issue #161, opened 2025-02-26
Message validation


diegomrsantos:
Messages arriving from the network must be validated and the result reported to Gossipsub.

PR #171, opened 2025-03-04
Stateless QBFT consensus message semantic validation

diegomrsantos:
## Issue Addressed

## Proposed Changes

Add an initial validation for QBFT consensus messages that doesn't depend on the consensus state.

## Additional Info

Validation using consensus state is coming in the next PR.


diegomrsantos:
This is based on https://github.com/sigp/anchor/pull/160, as soon as it gets merged, I'll remove the common code and make it ready for review. But feel free to take a look in the mean time.

anchor/common/qbft/src/lib.rs
```diff
@@ -252,21 +246,9 @@ where
...
-                let valid_data = Some(ValidData::new(None, wrapped_msg.qbft_message.root));
-                return Some((valid_data, OperatorId::from(0)));
-            }
-            // Otherwise, this is invalid data
-            warn!(
-                num_signers = wrapped_msg.signed_message.operator_ids().len(),
-                "Message only allows one signer"
-            );
-            return None;
+            // Do not care about data here, just that we had a success
```
diegomrsantos, line 249: The previous validation is now being done in https://github.com/sigp/anchor/pull/171/files#diff-e78a617e82f41db85b09de526249f65676c24a328fe1ac1c2475f33cd8449aafR254. @Zacholme7, could you confirm if it makes sense?

diegomrsantos:
@jking-aus @dknopik @Zacholme7 Sorry for the big PR. I can move the tests to a new one if necessary.

anchor/message_validator/src/lib.rs
```diff
@@ -169,33 +174,159 @@ pub trait ValidatorService {
...
+        &self,
+        signed_ssv_message: &SignedSSVMessage,
+        consensus_message: &QbftMessage,
+    ) -> Result<(), ValidationFailure> {
+        let signers = signed_ssv_message.operator_ids().len();
+
+        let committee_id = match signed_ssv_message.ssv_message().msg_id().duty_executor() {
+            Some(DutyExecutor::Committee(id)) => id,
+            _ => return Err(ValidationFailure::NonExistentCommitteeID),
+        };
```
diegomrsantos, line 191: This is incomplete, I'm not handling the validator case. But this PR is already quite big, I will handle it in my next one which is coming soon.

anchor/message_validator/src/lib.rs
```diff
@@ -275,3 +402,490 @@ impl ValidatorService for Validator {
...
+}
+
+fn hash_data_root(full_data: &[u8]) -> [u8; 32] {
+    let mut hasher = Sha256::new();
+    hasher.update(full_data);
+    let hash: [u8; 32] = hasher.finalize().into();
+    hash
+}
+
+#[cfg(test)]
```
diegomrsantos, line 335: The tests could be greatly simplified if it's possible to call the validation functions without creating a validator. It's an interesting idea that I'll explore in a new PR.

anchor/common/qbft/src/lib.rs
Zacholme7: Changes look correct. For clarity, I would also adjust this comment to directly state that in this flow path the message is decided message. This can be deciphered by looking at the validation code but helps to have it here too.

anchor/database/src/state.rs
```diff
@@ -281,3 +298,25 @@ impl NetworkState {
...
+    fn get_cluster_members(&self, committee_id: &CommitteeId) -> Option<IndexSet<OperatorId>> {
+        let db_state = self.state_rx.borrow();
+        db_state
+            .multi_state
+            .clusters
+            .get_all_by(committee_id)
+            .and_then(|clusters| clusters.first().cloned())
+            .map(|cluster| cluster.cluster_members)
+    }
+}
```
Zacholme7, line 317: Why is this needed? 

Can `get_cluster_members` just be defined on the network state and then use `state()` and  `watch()` on the database to get access to `NetworkState`?

anchor/database/src/state.rs
```diff
@@ -43,6 +51,9 @@ impl NetworkState {
         // 5) Owner -> Nonce (u16)
         let nonces = Self::fetch_nonces(&conn)?;
 
+        //CommitteeId -> ClusterId. It's populated in the loop that populates the multi-index maps
+        let mut clusters_by_committee_id = HashMap::new();
```
Zacholme7: Is this being used anywhere?

anchor/message_validator/src/lib.rs
```diff
@@ -169,34 +174,168 @@ pub trait ValidatorService {
...
+                || (msg_type == QbftMessageType::Commit && signers == 1)
+            {
+                return Err(ValidationFailure::PrepareOrCommitWithFullData);
+            }
+
+            let hashed_full_data = hash_data_root(signed_ssv_message.full_data());
+            // Rule: Full data hash must match root
+            if hashed_full_data != consensus_message.root {
+                return Err(ValidationFailure::InvalidHash);
+            }
```
Zacholme7: This is also being checking in qbft. Can add a new issue to remove the checks there

anchor/database/src/state.rs
```diff
@@ -8,12 +10,18 @@ use openssl::rsa::Rsa;
...
 };
 use std::collections::{HashMap, HashSet};
 use std::str::FromStr;
+use tokio::sync::watch;
 use types::Address;
 
+pub trait NetworkStateService: Send + Sync {
+    fn get_cluster_members(&self, committee_id: &CommitteeId) -> Option<IndexSet<OperatorId>>;
+}
+
```
Zacholme7, line 24: +1 to comment below, can this just be added to the `NetworkState`?

Zacholme7 (commented):
This is awesome, thanks for doing this. Have a few comments.

anchor/database/src/state.rs
diegomrsantos: You mean why the trait is needed? If so, it's much easier to use it in the tests then recreating the whole NetworkState. But after talking to @dknopik, there might be an easier solution if we extract the validation to functions that can be called without a validator instance. I'll explore this design in a new PR.

anchor/database/src/state.rs
diegomrsantos: it's not used anymore, thanks!

anchor/common/qbft/src/lib.rs
diegomrsantos: What is a decided message? The previous code checks the type is commit.

Zacholme7: Decided message is a commit message with > 1 signers. It is also called an aggregated commit message. 

Relevant check is here. 
https://github.com/sigp/anchor/blob/878c0dead46c4e00766d4f645da881b2581820c6/anchor/message_validator/src/lib.rs#L255-L261

diegomrsantos: Should I update the check to` > 1` instead of `!= 1`?

Zacholme7: Yes probably. Messages will either have `>= quorum size` signatures (which is already checked) or 1 signature. 

Is there a check to make sure that all other message have `1` signature? Im not seeing anything.

diegomrsantos: There's this check for partial signatures. I'll create the PR soon.

diegomrsantos: Pushed, let me know what you think

Zacholme7: Perfect

anchor/message_validator/src/lib.rs
diegomrsantos: [Here](https://github.com/sigp/anchor/blob/unstable/anchor/common/qbft/src/lib.rs#L435)?

Zacholme7: Yes

diegomrsantos: Created an issue https://github.com/sigp/anchor/issues/182

diegomrsantos: Removed

anchor/common/ssv_types/src/consensus.rs
```diff
@@ -60,12 +60,12 @@ pub struct QbftMessage {
...
-    pub fn validate(&self) -> bool {
-        if self.qbft_message_type > QbftMessageType::RoundChange {
-            return false;
-        }
-        true
+    pub fn max_round(&self) -> Option<u64> {
+        self.identifier.role().and_then(|role| match role {
+            Role::Committee | Role::Aggregator => Some(12),
+            Role::Proposer | Role::SyncCommittee => Some(6),
+            _ => None,
```
jking-aus, line 67: this may be obvious and I'm not getting it but why are we hardcoding 12 and 6 as max rounds? is that ssv spec? 2secs per round and 2 slot window?

anchor/database/src/state.rs
```diff
@@ -281,3 +293,25 @@ impl NetworkState {
...
+impl WatchableNetworkState {
+    pub fn new(state_rx: watch::Receiver<NetworkState>) -> Self {
+        Self { state_rx }
+    }
+}
+
+impl NetworkStateService for WatchableNetworkState {
+    fn get_cluster_members(&self, committee_id: &CommitteeId) -> Option<IndexSet<OperatorId>> {
+        let db_state = self.state_rx.borrow();
```
jking-aus, line 309: not sure I like this borrow -  I think it will block on write lock and wont err if it gets stuck. Maybe not a problem depending on how often we're using this, but would prefer to just design around it with something like try_borrow with a retry mechanism. 

If you've got something else cooking as per comment with zac below happy to merge and revisit.

jking-aus (changes requested):
great work as usual diego -- just a couple of comments and resolve the conflicts please

anchor/common/ssv_types/src/consensus.rs
diegomrsantos: I copied it from here https://github.com/ssvlabs/ssv/blob/main/message/validation/consensus_validation.go#L370

anchor/database/src/state.rs
diegomrsantos: I only read it here. I don't know exactly how it works, I'm using it how it is used in other parts of the codebase. Maybe @dknopik can elaborate more.

dknopik: I think the borrow is fine.

We write only rarely and only little data during normal operation. Yes, during that time, we will block, but as we are in a dedicated (non-async) task, it will not block the whole network thread or the tokio runtime. So the block will delay only the handling of the message.

A retry mechanism would not be beneficial IMO, as it effectively also delays the message - so it might be even worse because we get admitted immediately after the writing borrow ends.

jking-aus: alright - wasn't sure how often this gets called, if it's to verify a large portion of network traffic there's a good chance the block hurts but trust you've got it handled

anchor/common/ssv_types/src/consensus.rs
jking-aus: all good we need to match them on this or we won't timeout at the same time. I would like to ask them where they're pulling this number from because I would assume it's just a guess like follow distance. I will just add an inline comment saying where it's from.

jking-aus (approved):
lgtm

anchor/common/ssv_types/src/consensus.rs
diegomrsantos: This is used only for validation, maybe I shouldn't have added it here.

diegomrsantos: I can move it to the validator if you think it's better.

PR #172, opened 2025-03-05
Add pre-commit hook for formatting, linting, and sorting

diegomrsantos:
## Proposed Changes

Create a pre-commit file in `.githooks`. You only need to run `git config core.hooksPath .githooks` and the old days of CI failures are over.


dknopik:
I think invoking the appropriate commands in the Makefile would be more appropriate, in order to use the flags specified there. i.e. `make lint`

diegomrsantos:
> I think invoking the appropriate commands in the Makefile would be more appropriate, in order to use the flags specified there. i.e. `make lint`

I agree. Are they `make cargo-fmt`, `make lint`, and `make sort`?

diegomrsantos:
@dknopik should we get it merged like this?

dknopik:
@diegomrsantos how about the switch to invoking the `make` targets?

diegomrsantos:
> @diegomrsantos how about the switch to invoking the `make` targets?

The make targets have an error that causes the script to fail. Additionally, the fmt and lint ones only check and don't apply fixes.

dknopik:
The make targets work fine on my end, and are also what's used in CI. But I will likely not use this anyway :man_shrugging:

jking-aus (changes requested):
lgtm - I have my own local pre-commit hooks but happy to merge if this is easier

jking-aus (approved):

PR #177, opened 2025-03-07
Set "ssv" field in ENR

dknopik:
There is a check for a "ssv" field in the ENR: https://github.com/ssvlabs/ssv/blob/dc0c830e8f0718cc17d61d20fc802887df252569/network/discovery/dv5_filters.go#L39-L49 

This PR sets the field in our own ENR. We do not check for it, as that is unneccessary


diegomrsantos (approved):
thanks for fixing this

PR #178, opened 2025-03-10
Implement version endpoint

ThreeHrSleep:
## Issue Addressed

part of https://github.com/sigp/anchor/issues/26


jking-aus (dismissed):
good stuff man - I think the structure here is useful to build on

jking-aus (approved):

PR #181, opened 2025-03-11
Add `Role::{ValidatorRegistration, VoluntaryExit}`

dknopik:
Add two Roles that have not been added yet and have been causing validation failures on my end.

`ValidatorRegistration` is for registering the validator at MEV relays and has been previously erroneously treated as `Proposer`.
`VoluntaryExit` is for chain triggered exits, which is currently unsupported by Anchor.

### DIFF
--- a/anchor/common/ssv_types/src/msgid.rs
+++ b/anchor/common/ssv_types/src/msgid.rs
@@ -12,6 +12,8 @@ pub enum Role {
     Aggregator,
     Proposer,
     SyncCommittee,
+    ValidatorRegistration,
+    VoluntaryExit,
 }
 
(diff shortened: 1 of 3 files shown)


jking-aus (approved):

diegomrsantos:
Why do they say unknown role [here](https://github.com/ssvlabs/ssv/blob/main/message/validation/consensus_validation.go#L370)

diegomrsantos:
> Why do they say unknown role [here](https://github.com/ssvlabs/ssv/blob/main/message/validation/consensus_validation.go#L370)

Got it, it's not expected for consensus messages.

Issue #154, opened 2025-02-20
Operator Key generation


Zacholme7:
Re-write their [operator key generator](https://github.com/ssvlabs/ssv/blob/main/cli/generate_operator_keys.go) as there is the header issue

PR #184, opened 2025-03-12
Rsa keygen

Zacholme7:
## Issue Addressed

#154

## Proposed Changes

This pr introduces a new cli tool to generate rsa keys. It will generate a `key.pem` file to be used with the anchor nodes and a `keys.json` file that contains the base64 encoded keys to be used with the ssv ui webapp. 

Note: this replaces the base64 headers in the key so that we can interop with the ssv-client for ssv-mini


anchor/keygen/src/lib.rs
```diff
@@ -0,0 +1,93 @@
...
+    } else {
+        PathBuf::from(".") // Current working directory
+    };
+
+    // Create output paths for both files
+    let pem_file = output_dir.join("key.pem");
+    let json_file = output_dir.join("keys.json");
+
+    // Write the PEM file
+    fs::write(&pem_file, &private_pem).map_err(|e| {
```
dknopik: I think we should avoid to accidentally overwrite existing files

dknopik (commented):
Looks good!

What do you think about changing the client to implicitly invoke this if it finds no key? Replacing its current key gen mechanism.

anchor/keygen/src/lib.rs
```diff
@@ -83,11 +76,25 @@ pub fn run_keygen(keygen: Keygen) -> Result<(), KeygenError> {
...
+        // Write the JSON file
+        fs::write(&json_file, pretty_json)
+            .map_err(|e| KeygenError::Output(format!("Failed to write keys to JSON file: {e}")))?;
+
+        info!("JSON keys written to: {}", json_file.display());
+    } else {
+        warn!(
+            "PEM file or JSON file already exist in {:?}, skipping write",
+            output_dir
+        );
```
dknopik: I think we should return an error here to abort client startup. It makes no sense to operate on a key which has not been saved.

dknopik (commented):
Nice, thanks! Just a nitpick:

anchor/keygen/src/lib.rs
diegomrsantos: Might be relevant https://github.com/sigp/anchor/pull/43#discussion_r1876112496

dknopik: An important difference here is that we need the operator key to perform our duties. If we lose it or overwrite it, we can not act as that operator again, especially as (opposed to validator keys) the operator key is not derived from a mnemonic. So we should be very careful with it.

dknopik:
Looks good! Another thing that came to my mind is zeroing the memory after use with `zeroize`. It's a bit complicated to do correctly (as moves might leave unzeroed memory, but clones don't) and it's difficult to reason about invoked library functions, but even doing it on a "best effort" basis would be better than nothing. Do you want to try it? If so, I recommend reading through the main docs page: https://docs.rs/zeroize/latest/zeroize/

Zacholme7:
Update: Need the ability to be able to force overwrites in ssv-mini so added cli flag to accomplish this

Zacholme7:
We are writing passwords out in plaintext right now, which kinda defeats the purpose of zeroing out the memory. 

Adding an issue to enable password protection for the private key file.

anchor/keygen/src/lib.rs
```diff
@@ -0,0 +1,112 @@
...
+use clap::Parser;
+use openssl::{pkey::Private, rsa::Rsa};
+use serde::Serialize;
+use std::{fs, path::PathBuf};
+use tracing::info;
+use zeroize::{Zeroize, ZeroizeOnDrop, Zeroizing};
+
+#[derive(Debug)]
+pub enum KeygenError {
+    Generate(String),
```
diegomrsantos: Could we keep the original Error instead of a String?

Zacholme7: We could, but the errors here are very simple, should really never fail, and we are printing the original errors in the format call anyways. 

What is the benefit of just the original errors?

anchor/keygen/src/lib.rs
```diff
@@ -0,0 +1,112 @@
...
+    Pem(String),
+    Output(String),
+    Utf8(String),
+}
+
+#[derive(Parser, Clone, Debug)]
+#[clap(name = "keygen", about = "RSA key generation tool")]
+pub struct Keygen {
+    #[clap(long, help = "Path to output keys to", value_name = "OUTPUT_PATH")]
+    pub output_path: Option<String>,
```
diegomrsantos, line 35: Is this used as the param name? If so, how about `keys_output_path`?

anchor/keygen/src/lib.rs
diegomrsantos: It might give more detailed and structured information useful for logging. It'd allow matching on specific error variants as well. It's quite straightforward to do that using thiserror.

anchor/keygen/src/lib.rs
diegomrsantos: But if the context is clearly the key generation, the current name might be good enough.

anchor/keygen/src/lib.rs
Zacholme7: Fixed, something like that?

anchor/keygen/src/lib.rs
Zacholme7: Yeah this is run as `anchor keygen --output-path ...` so its self describing

anchor/keygen/src/lib.rs
```diff
@@ -39,22 +53,14 @@ struct PrettyOutput {
...
-            .map_err(|e| KeygenError::Pem(format!("Failed to convert private key to PEM: {e}")))?,
-    );
-
-    let public_pem = private_key
-        .public_key_to_pem()
-        .map_err(|e| KeygenError::Pem(format!("Failed to convert public key to PEM: {e}")))?;
-
-    let public_pem_string = String::from_utf8(public_pem)
-        .map_err(|e| KeygenError::Utf8(format!("Failed to convert public key to UTF8: {e}")))?;
+    let private_pem = Zeroizing::new(private_key.private_key_to_pem().map_err(KeygenError::Pem)?);
```
diegomrsantos, line 61: Does only `let private_pem = Zeroizing::new(private_key.private_key_to_pem()?);` work?

Zacholme7: No, compile error since `From` is not implemented on Keygen error

anchor/keygen/src/lib.rs
diegomrsantos: yes, thanks!

jking-aus (approved):
nice -- just left a few todo comments so we remember to encrypt

PR #185, opened 2025-03-13
Remove traits and refactor tests

diegomrsantos:
## Issue Addressed

This is mostly a refactoring related to https://github.com/sigp/anchor/issues/161

## Proposed Changes

Removes traits (NetworkStateService, ValidatorService, and MessageReceiver) and simplifies the code. 
Fixes the committee info computation for the validator role (sorry for the misleading title).


anchor/message_validator/src/lib.rs
```diff
@@ -133,186 +134,201 @@ pub enum Error {
...
-            Ok(msgs) => msgs,
-            Err(_) => return Err(ValidationFailure::UndecodableMessageData),
-        };
-
-        Ok(ValidatedSSVMessage::PartialSignatureMessages(messages))
+pub fn validate_ssv_message(
+    signed_ssv_message: &SignedSSVMessage,
+    committee_info: &CommitteeInfo,
+    role: Role,
+) -> Result<ValidatedSSVMessage, ValidationFailure> {
```
dknopik: should this be `pub`?

anchor/message_receiver/src/lib.rs
```diff
@@ -1,17 +1,21 @@
...
 
 pub trait MessageReceiver: Send + Sync {
     fn receive(
         &self,
         propagation_source: PeerId,
         message_id: MessageId,
         message: Message,
-    ) -> Result<(), processor::Error>;
+    ) -> Result<(), crate::Error>;
 }
```
dknopik: IMO, we should consider removing this at some point.

Currently, it is only needed for the "create network" test, which IMHO is a test of questionable value.

But maybe, we'll find more use cases, so let's keep it for now :)

anchor/network/src/network.rs
```diff
@@ -431,24 +431,22 @@ fn build_swarm(
...
     use crate::network::Network;
     use crate::Config;
-    use message_receiver::testing::MessageReceiverMock;
-    use message_validator::{ValidatedMessage, ValidationFailure};
     use std::time::Duration;
     use subnet_tracker::test_tracker;
     use task_executor::TaskExecutor;
     use tokio::sync::mpsc;
-
     pub struct ValidatorServiceMock;
```
dknopik: remove

dknopik (commented):
Very good. I am very happy about this solution! :)

LGTM, minor notes below.

anchor/message_validator/src/lib.rs
diegomrsantos: pub(crate)?

anchor/message_receiver/src/lib.rs
diegomrsantos: If we aren't testing anything for now, I can remove it in this PR and we can add it again if/when it is needed.

anchor/message_validator/src/lib.rs
dknopik: if it's necessary, yes, but if it compiles with no `pub` at all, even better :)

anchor/network/src/network.rs
diegomrsantos: Thanks. Weird that clippy doesn't flag it.

anchor/message_receiver/src/lib.rs
dknopik: As you prefer. :)

anchor/message_validator/src/lib.rs
diegomrsantos: it works with private. Probably cause the tests are in the same file.

anchor/message_receiver/src/lib.rs
diegomrsantos: removed

anchor/message_validator/src/lib.rs
dknopik: It would even work if they were in a different file, as long as they are in the same module or a submodule.

jking-aus (dismissed):
lgtm -- nice work diego, just hit those conflicts and I'll merge

jking-aus:
nice lgtm

jking-aus (approved):

PR #186, opened 2025-03-13
Update deps

dknopik:
This PR updates our dependencies. This is grouped into commits for easier review:

1. "Update Lighthouse": This updates to the newest commit in the `anchor` branch of Lighthouse :partying_face:. This includes goodies like the tracing PR and fixes for the CVEs that have been breaking CI.
2. "Align libp2p versions": This cleans up the libp2p versions, by pulling everything from crates.io, except `peer-store` (which is not on crates.io yet) and `gossipsub` (which we get from the SigP fork).
3. "`cargo update`": Rans `cargo update` to get all the trivially updatable versions.
4. "update `alloy`, `derive_more`, `strum`": These were easily updatable without code changes being necessary.
5. "update `rand` where possible": Updates `rand` to 0.9.0, except in `bls_lagrange`, because our dependencies there are still on `0.8.5`. I think it is worth having two versions in the dependency graph here, because then our move to 2024 edition of Rust will be cleaner (as 0.9 avoids collision with the new reserved keyword `gen`).
6. "remove audit exceptions": We are now updated, and the exceptions are no longer necessary. :partying_face:


jking-aus (approved):
excellent! great staying on top of things

PR #187, opened 2025-03-14
Add comments for easier audit of crypto code

dknopik:
Our crypto code is to be audited soon :partying_face: 

This code adds some comments that explain what is happening.

The attentive reader notices that I talk about passing a reference where we actually don't. The corresponding locations will appropriately change through #186.


jking-aus:
I assume you mean at line 40 in bls_lagrange/lib.rs? 186 went in so adding ref to secret key result. Just double check that's what you wanted because going deep on blst stuff is above my head right now

dknopik:
@jking-aus yep looks good :)

jking-aus (approved):

PR #189, opened 2025-03-14
Rust Dependency Caching

diegomrsantos:
## Issue Addressed

Speeds up CI builds by caching Rust dependencies.

## Proposed Changes

1. Added cache: true to all moonrepo/setup-rust actions that were previously missing it or had it set to false
2. Set appropriate cache-target values:
   - release for jobs that perform release builds
   - debug for jobs that work with debug builds
3. Fixed the check-msrv job structure by combining the action with proper caching configuration

These changes will speed up CI by reusing dependencies between workflow runs, significantly reducing build times after the first run.

### DIFF
--- a/.github/workflows/test-suite.yml
+++ b/.github/workflows/test-suite.yml
@@ -147,19 +147,13 @@ jobs:
     runs-on: ubuntu-22.04
     steps:
     - uses: actions/checkout@v4
-    - name: Get latest version of stable Rust
-      uses: moonrepo/setup-rust@v1
-      with:
-          cache: true
-          cache-target: debug
-    - name: Install Rust at MSRV
+    - name: Install Rust at Minimum Supported Rust Version (MSRV)
       run: |
         metadata=$(cargo metadata --no-deps --format-version 1)
         msrv=$(echo $metadata | jq -r '.packages | map(select(.name == "anchor")) | .[0].rust_version')
         rustup override set $msrv
     - name: Run cargo check
       run: cargo check --workspace
-
   cargo-udeps:
     name: cargo-udeps
     needs: [check-labels]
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -123,7 +123,7 @@ test-full: cargo-fmt test-release test-debug
 # Lints the code for bad style and potentially unsafe arithmetic using Clippy.
 # Clippy lints are opt-in per-crate for now. By default, everything is allowed except for performance and correctness lints.
 lint:
-	cargo clippy --workspace --tests $(EXTRA_CLIPPY_OPTS) --features "$(TEST_FEATURES)" -- \
+	cargo clippy --workspace --tests --no-deps $(EXTRA_CLIPPY_OPTS) --features "$(TEST_FEATURES)" -- \
```
diegomrsantos, line 126: Even with this flag, the dependencies are still being checked.

.github/workflows/test-suite.yml
```diff
@@ -77,8 +78,8 @@ jobs:
       if: env.SELF_HOSTED_RUNNERS == 'false'
       uses: moonrepo/setup-rust@v1
       with:
-          cache: false
           channel: stable
+          cache: true
```
diegomrsantos, line 82: Was it intentional to set the cache to false?

dknopik:
https://github.com/sigp/anchor/issues/123

jking-aus (approved):

dknopik:
Did anyone have the chance to review the issue I flagged above? @diegomrsantos

jking-aus:
> Did anyone have the chance to review the issue I flagged above? @diegomrsantos

this only occurs when the code is built multiple times on the same runner? I will try and force this issue to happen -- if it does we'll roll back the cache change.

Issue #182, opened 2025-03-11
Double check redundant validations were removed from qbft


diegomrsantos:
Stateless validation (that doesn't depend on the current consensus state) is being moved to the message validator. This issue is to double-check we didn't leave redundant checks in qbft.

diegomrsantos:
Qbft logic also needs to be checked, like mentioned here https://github.com/sigp/anchor/pull/191/files#r2014993628

PR #191, opened 2025-03-14
Validate qbft logic

diegomrsantos:
## Issue Addressed

Implements https://github.com/ssvlabs/ssv/blob/main/message/validation/consensus_validation.go#L172

## Proposed Changes

See https://github.com/sigp/anchor/pull/191#pullrequestreview-2714959369


anchor/message_validator/src/consensus_message.rs
```diff
@@ -112,12 +150,208 @@ pub(crate) fn validate_justifications(
...
+/// Validate that the message round is within the allowed spread
+fn validate_round_in_allowed_spread(
+    consensus_message: &QbftMessage,
+    received_at: SystemTime,
+    slot_clock: impl SlotClock,
+) -> Result<(), ValidationFailure> {
+    // Get the slot
+    let slot = Slot::new(consensus_message.height);
+    let slot_start_time: SystemTime = UNIX_EPOCH + slot_clock.start_of(slot).unwrap();
+
```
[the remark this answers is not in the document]
diegomrsantos: fixed

anchor/message_validator/src/consensus_state.rs
```diff
@@ -0,0 +1,190 @@
...
+    state: Vec<Option<SignerState>>,
+    max_slot: Slot,
+    max_epoch: Epoch,
+    last_epoch_duties: u64,
+    prev_epoch_duties: u64,
+}
+
+impl OperatorState {
+    fn new(stored_slot_count: usize) -> Self {
+        println!("stored_slot_count: {}", stored_slot_count);
```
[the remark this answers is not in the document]
diegomrsantos: fixed

anchor/common/ssv_types/src/round.rs
```diff
@@ -0,0 +1,54 @@
...
+use std::num::NonZeroUsize;
+use std::ops::Add;
+
+/// This represents an individual round, these change on regular time intervals
+#[derive(Clone, Copy, Debug, Deref, PartialEq, Eq, Hash, PartialOrd, Ord)]
+pub struct Round(NonZeroUsize);
+
+impl From<u64> for Round {
+    fn from(round: u64) -> Round {
+        Round(NonZeroUsize::new(round as usize).expect("round == 0"))
```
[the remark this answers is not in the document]
diegomrsantos: @dknopik what do you think?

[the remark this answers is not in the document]
diegomrsantos: There isn't a way to recover from this error.

anchor/common/ssv_types/src/round.rs
```diff
@@ -0,0 +1,54 @@
...
+use std::fmt::{Display, Formatter};
+use std::num::NonZeroUsize;
+use std::ops::Add;
+
+/// This represents an individual round, these change on regular time intervals
+#[derive(Clone, Copy, Debug, Deref, PartialEq, Eq, Hash, PartialOrd, Ord)]
+pub struct Round(NonZeroUsize);
+
+impl From<u64> for Round {
+    fn from(round: u64) -> Round {
```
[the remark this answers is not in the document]
diegomrsantos: We do not support targets with 32-bit pointer size

anchor/message_sender/src/network.rs
```diff
@@ -18,16 +19,16 @@ use tracing::{debug, error, warn};
 const SIGNER_NAME: &str = "message_sign_and_send";
 const SENDER_NAME: &str = "message_send";
 
-pub struct NetworkMessageSender {
+pub struct NetworkMessageSender<S: SlotClock> {
```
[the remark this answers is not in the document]
diegomrsantos: SlotClock is defined like `trait SlotClock: Send + Sync + Sized + Clone`

anchor/common/ssv_types/src/round.rs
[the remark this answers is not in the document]
dknopik: IMO, the suggested variant is a bit cleaner - but I guess it only moves the location where we panic? Still might be better because we can then more easily see that only nonzero values will be passed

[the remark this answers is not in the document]
diegomrsantos: Why cleaner? Won't it make the conversion more cumbersome?

anchor/message_validator/src/consensus_state.rs
```diff
@@ -0,0 +1,189 @@
...
+                .insert(hash_operators(signed_ssv_message.operator_ids()));
+        }
+
+        self.message_counts
+            .record_consensus_message(consensus_message.qbft_message_type);
+    }
+}
+
+/// Hash a list of operator IDs to create a unique identifier, using SHA-256
+fn hash_operators(operators: &Vec<OperatorId>) -> [u8; 32] {
```
Zacholme7: nit: `&Vec<OperatorId>` to `&[OperatorId]`

anchor/message_validator/src/consensus_state.rs
```diff
@@ -0,0 +1,189 @@
...
+use ssv_types::consensus::{QbftMessage, QbftMessageType};
+use ssv_types::message::SignedSSVMessage;
+use ssv_types::OperatorId;
+use ssv_types::{Epoch, Slot};
+use std::cell::RefCell;
+use std::collections::{HashMap, HashSet};
+
+/// ConsensusState manages the state for consensus validation across operators and slots
+pub(crate) struct ConsensusState {
+    operators: HashMap<OperatorId, RefCell<OperatorState>>,
```
Zacholme7: It looks like it all works good here, but this `RefCell` is very hard to reason about with all the clones and could cause some unexpected runtime panics. 

I recommend we either really reason about this, also use a DashMap here, or use the `HashMap Entry` API more directly

anchor/message_validator/src/lib.rs
```diff
@@ -140,14 +169,22 @@ pub enum Error {
     Processor(#[from] ::processor::Error),
 }
 
+const SLOTS_PER_EPOCH: u64 = 32;
+
 #[derive(Clone)]
-pub struct Validator {
+pub struct Validator<S: SlotClock> {
```
Zacholme7, line 173: nit: Maybe this name should be changed. Can be confusing with an actual Validator.

anchor/message_validator/src/lib.rs
```diff
@@ -185,37 +222,58 @@ impl Validator {
...
     }
+
+    /// Gets the consensus state for a message ID, creating a new one if it doesn't exist
+    fn get_consensus_state(&self, message_id: &MessageId) -> Arc<Mutex<ConsensusState>> {
+        self.consensus_state_map
+            .entry(message_id.clone())
+            .or_insert_with(|| {
+                // Create a new consensus state with storage for two epochs worth of slots
+                let slots_per_epoch = SLOTS_PER_EPOCH;
+                let stored_slot_count = slots_per_epoch * 2; // Store last two epochs
```
Zacholme7, line 258: nit: why not just `SLOTS_PER_EPOCH * 2`

anchor/message_validator/src/lib.rs
```diff
@@ -185,37 +222,58 @@ impl Validator {
...
             }
         }
     }
+
+    /// Gets the consensus state for a message ID, creating a new one if it doesn't exist
+    fn get_consensus_state(&self, message_id: &MessageId) -> Arc<Mutex<ConsensusState>> {
+        self.consensus_state_map
+            .entry(message_id.clone())
+            .or_insert_with(|| {
+                // Create a new consensus state with storage for two epochs worth of slots
```
Zacholme7: Is there anything that is cleaning up `ConsensusState` after two epochs have passed?

anchor/message_validator/src/consensus_message.rs
```diff
@@ -112,12 +149,207 @@ pub(crate) fn validate_justifications(
...
+                return Err(ValidationFailure::DecidedWithSameSigners);
+            }
+        }
+    }
+
+    // Rule: Round must be within allowed spread from current time
+    if signers.len() == 1 {
+        validate_round_in_allowed_spread(consensus_message, received_at, slot_clock)?;
+    }
+
```
Zacholme7, line 238: I think we can now remove some of these checks from qbft. separate issue

anchor/message_validator/src/lib.rs
```diff
@@ -140,14 +169,22 @@ pub enum Error {
     Processor(#[from] ::processor::Error),
 }
 
+const SLOTS_PER_EPOCH: u64 = 32;
+
 #[derive(Clone)]
-pub struct Validator {
+pub struct Validator<S: SlotClock> {
     network_state_rx: Receiver<NetworkState>,
+    consensus_state_map: DashMap<MessageId, Arc<Mutex<ConsensusState>>>,
```
Zacholme7, line 175: nit: In terms of readability it might help to document the structure of the map. It goes very deep and has a lot of moving parts

Here is my understanding. Top level we have `MessageID` -> `ConsensusState` where `ConsensusState` holds all the state for Operators pertaining to `MessageID`.

`OperatorState` contains specific operator data for each slot via a circular buffer. For each slot, there is `SignerState` which represents signer info if this operator was a signer.

anchor/message_validator/src/consensus_state.rs
```diff
@@ -0,0 +1,189 @@
...
+
+    // Write each operator ID to the buffer in little endian order
+    for &operator in operators {
+        buffer.extend_from_slice(&operator.0.to_le_bytes());
+    }
+
+    // Hash the buffer with SHA-256
+    let mut hasher = Sha256::new();
+    hasher.update(&buffer);
+    hasher.finalize().into()
```
Zacholme7: Is there any reason behind not sorting? Otherwise this is the same computation for `CommitteeId`

anchor/message_validator/src/message_counts.rs
```diff
@@ -0,0 +1,71 @@
...
+            QbftMessageType::Commit
+                if signed_message.operator_ids().len() == 1
+                    && self.commit >= MAX_MESSAGES_PER_ROUND =>
+            {
+                Err(ValidationFailure::DuplicatedMessage {
+                    got: format!("commit, having {}", self),
+                })
+            }
+            QbftMessageType::RoundChange if self.round_change >= MAX_MESSAGES_PER_ROUND => {
+                Err(ValidationFailure::DuplicatedMessage {
```
Zacholme7, line 54: Is `DuplicatedMessage` the right error here? or would something like `MessageLimitExceeded` be better.

Zacholme7 (commented):
Phew this was dense! Awesome job though this is great and super useful. Definitely should get @dknopik and @jking-aus eyes on this too.

anchor/message_validator/src/consensus_message.rs
diegomrsantos: Do you remember where it is done?

anchor/message_validator/src/consensus_state.rs
diegomrsantos: Good catch, thanks. I removed the RefCell.

anchor/message_validator/src/lib.rs
diegomrsantos: Possibly, but it's inside the `message_validator` crate.

anchor/message_validator/src/lib.rs
diegomrsantos: This has been changed.

anchor/message_validator/src/lib.rs
diegomrsantos: I'm not sure if this comment is correct. When does the consensus instance associated with the message IDs stop being active?

anchor/message_validator/src/message_counts.rs
diegomrsantos: Aren't they equivalent? An exceeded limit means a duplicated msg cause the limit is 1.

anchor/message_validator/src/consensus_state.rs
diegomrsantos: There's a validation checking this list is sorted, so I'm not sure sorting is necessary.

anchor/message_validator/src/lib.rs
Zacholme7: Im probably just confused. It doesn't stop being active, but my understanding from the comment was that a new fresh instance would be created every two epochs. What happens to the consensus state after two epochs pass and the storage is taken up?

anchor/message_validator/src/consensus_message.rs
Zacholme7: I will have to take a look, would be good to group this into issue #182. Just a passover the whole pipeline to make sure we are not performing redundant work.

anchor/message_validator/src/lib.rs
Zacholme7: Fair point.

anchor/message_validator/src/consensus_state.rs
Zacholme7: If the operators are sorted, this is a duplicate of computing `CommitteeId`. Can that just be used here? 
https://github.com/sigp/anchor/blob/045cb2c99934ac7d0043c364696e9953144ddf22/anchor/common/ssv_types/src/committee.rs#L26-L40

anchor/message_validator/src/message_counts.rs
Zacholme7: Yes my bad, I missed that max messages is 1

anchor/message_validator/src/lib.rs
```diff
@@ -98,7 +125,7 @@ impl From<&ValidationFailure> for MessageAcceptance {
             | ValidationFailure::EarlySlotMessage
             | ValidationFailure::LateSlotMessage
             | ValidationFailure::SlotAlreadyAdvanced
-            | ValidationFailure::RoundAlreadyAdvanced
+            | ValidationFailure::RoundAlreadyAdvanced { got: _, want: _ }
```
dknopik: Nitpick: use `..` instead of listing the fields, is a bit cleaner. Same below.

anchor/message_receiver/src/manager.rs
```diff
@@ -20,18 +22,38 @@ pub struct Outcome {
...
 /// A message receiver that passes messages to responsible managers.
-pub struct MessageReceiver {
+pub struct NetworkMessageReceiver<S: SlotClock> {
     processor: processor::Senders,
     qbft_manager: Arc<QbftManager>,
     signature_collector: Arc<SignatureCollectorManager>,
     network_state_rx: watch::Receiver<NetworkState>,
     outcome_tx: mpsc::Sender<Outcome>,
-    validator: Validator,
+    validator: Validator<S>,
```
dknopik: Now that the `Validator` actually has state, this needs to be an `Arc`, right? Or else the state of the receiver validator and sender validator diverges.

anchor/message_validator/src/consensus_message.rs
```diff
@@ -112,12 +153,204 @@ pub(crate) fn validate_justifications(
...
+    slot_clock: impl SlotClock,
+) -> Result<(), ValidationFailure> {
+    // Rule: For proposals, signer must be the leader
+    let signers = signed_ssv_message.operator_ids();
+    if consensus_message.qbft_message_type == QbftMessageType::Proposal {
+        if signers.is_empty() {
+            return Err(ValidationFailure::NoSigners);
+        }
+
+        let signer = signers[0];
```
dknopik: very very nitpick, but a pattern I like in these cases is
```rust
let Some(signer) = signers.first() else {
    return Err(ValidationFailure::NoSigners);
}
```
to combine the check and retrieval.

anchor/message_validator/src/consensus_message.rs
```diff
@@ -112,12 +153,204 @@ pub(crate) fn validate_justifications(
...
+        return Err(ValidationFailure::NonExistentCommitteeID);
+    }
+
+    // Calculate the first round index
+    let first_round_index = if height != FIRST_HEIGHT {
+        // If not the first height, add height % len(committee) to the index
+        height % committee.len() as u64
+    } else {
+        0
+    };
```
dknopik: I don't get this. `0 % anything` is `0`, so why the differentiation?

anchor/message_validator/src/consensus_message.rs
```diff
@@ -112,12 +153,204 @@ pub(crate) fn validate_justifications(
...
+    role: Role,
+    received_at: SystemTime,
+    slot_clock: impl SlotClock,
+) -> Result<(), ValidationFailure> {
+    // Get the slot
+    let slot = Slot::new(consensus_message.height);
+    let slot_start_time = match slot_clock.start_of(slot) {
+        Some(time) => UNIX_EPOCH + time,
+        None => return Err(ValidationFailure::SlotStartTimeNotFound),
+    };
```
dknopik, line 277: The attestation voting only starts one third into the round - so always starting to measure at the start of the slot seems incorrect

dknopik (commented):
Nice! Some remarks, but mostly nitpicks.

Anyway, before we merge this I'd like to test this on interop when convenient.

anchor/message_validator/src/lib.rs
dknopik: The message id does not change between instance heights. The only case where we do not clean up is when we leave a committee or a validator is removed

dknopik: This kind of inherits the confusing "message id" from go-ssv, which I hope to get rid of in a future SIP ;)

anchor/message_validator/src/consensus_state.rs
dknopik: Sorry, unresolving: Now that we have the DashMap, is an `Arc<Mutex<_>>` really necessary?

anchor/message_receiver/src/manager.rs
diegomrsantos: Good question. Do we want to share the consensus state? I think so.

anchor/message_validator/src/consensus_message.rs
diegomrsantos: Nice suggestion, thanks. Had to use:
```rust
let Some(&signer) = signers.first() else {
     return Err(ValidationFailure::NoSigners);
};
 ```

anchor/message_validator/src/consensus_message.rs
diegomrsantos: Done

anchor/message_validator/src/consensus_message.rs
diegomrsantos: You're right, copied their approach and didn't catch this.

anchor/message_validator/src/consensus_state.rs
diegomrsantos: Nice catch

anchor/message_validator/src/lib.rs
diegomrsantos: The comment is incorrect, I removed it. It's one `ConsensusState` per message id

anchor/message_validator/src/lib.rs
diegomrsantos: Done

anchor/message_validator/src/consensus_state.rs
diegomrsantos: Pushed a new version, let me know what you think

diegomrsantos (commented):
Thanks a lot, replied to most of the comments. I'll finish tomorrow

anchor/message_validator/src/consensus_state.rs
diegomrsantos: This is a different map and there I need the Mutex cause I can't change the consensus state concurrently

anchor/message_validator/src/consensus_state.rs
Zacholme7: lgtm

dknopik (approved):
LGTM, thanks!

anchor/message_validator/src/lib.rs
diegomrsantos: That's right. I'll add more docs in the next PR.

PR #192, opened 2025-03-17
Fix decoding QbftMessage

dknopik:
# Problem

The inner `QbftMessage` has a field `identifier`, which contains the message id. The message id is also contained in the outer `SsvMessage`.

As this is repeated, we used our type `MessageId` in both places, but the types actually differ, even though they contain the *same data* in every message: 

https://github.com/ssvlabs/ssv-spec/blob/36c660085fd74464dcfd301a0084e1112ed264ea/qbft/messages.go#L51 vs https://github.com/ssvlabs/ssv-spec/blob/4066ef02de3c159ea9852bc5615358ca03384fba/types/messages.go#L96

We fail to decode the inner `QbftMessage`, as the field is tagged `ssz-max`, which indicates that this is a variable length field. This is encoded differently in SSZ vs a fixed sized field.

# Solution

As the inner identifier is redundant anyway, don't bother with our custom type, and treat it as an opaque variable sized byte list with max length of 56. We still ensure that it is identical to the outer identifier.

### DIFF
--- a/anchor/common/qbft/src/lib.rs
+++ b/anchor/common/qbft/src/lib.rs
@@ -892,7 +892,7 @@ where
             qbft_message_type: msg_type,
             height: *self.instance_height as u64,
             round: data.round,
-            identifier: self.identifier.clone(),
+            identifier: self.identifier.clone().into(),
             root: data.root,
             data_round: data.data_round,
             round_change_justification,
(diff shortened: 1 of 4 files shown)


jking-aus (approved):
good job mate -- I just added a todo for when we come back and fix this

PR #195, opened 2025-03-18
Anchor simulator

Zacholme7:
## Issue Addressed

N/A

## Proposed Changes

This PR extends on the simulator introduced in lighthouse to add support for anchor. 

## Note
The tests wont currently pass since some of the node functionality still does not work,  but once everything is working all the tests should be drop in. There is also an issue where we will stack overflow a tokio worker after running for a bit. This also occurs in the lighthouse simulator. The culprit is the `LocalBeaconClient` simply growing too large over the course of execution mixed with some deep async calls. This would require a significant refactor to address, so simply configuring the stack size before running is the way to go. 

`RUST_MIN_STACK=8388608 cargo run --bin integration basic-sim`


dknopik:
I suppose the `http_api` was moved to `anchor_http_api` to avoid a conflict with LH's `http_api`. I propose that we leave the files in their previous location and just rename the crates in the `Cargo.toml`. (if really necessary)

anchor/client/src/lib.rs
```diff
@@ -316,9 +316,6 @@ impl Client {
         let proposer_nodes = Arc::new(proposer_nodes);
         start_fallback_updater_service::<_, E>(executor.clone(), proposer_nodes.clone())?;
 
-        // Wait until genesis has occurred.
-        wait_for_genesis(&beacon_nodes, genesis_time).await?;
-
```
dknopik, line 344: Why did you move it?

anchor/integration/src/basic_sim.rs
```diff
@@ -0,0 +1,206 @@
...
+        let genesis_delay = GENESIS_DELAY;
+        spec.seconds_per_slot /= sim_config.speed_up_factor;
+        spec.seconds_per_slot = max(1, spec.seconds_per_slot);
+        spec.genesis_delay = genesis_delay;
+        spec.min_genesis_time = 0;
+        spec.min_genesis_active_validator_count = total_validator_count as u64;
+        spec.altair_fork_epoch = Some(Epoch::new(ALTAIR_FORK_EPOCH));
+        spec.bellatrix_fork_epoch = Some(Epoch::new(BELLATRIX_FORK_EPOCH));
+        spec.capella_fork_epoch = Some(Epoch::new(CAPELLA_FORK_EPOCH));
+        spec.deneb_fork_epoch = Some(Epoch::new(DENEB_FORK_EPOCH));
```
dknopik, line 77: Add electra

anchor/integration/src/basic_sim.rs
```diff
@@ -0,0 +1,206 @@
...
+pub const ALTAIR_FORK_EPOCH: u64 = 0;
+pub const BELLATRIX_FORK_EPOCH: u64 = 0;
+pub const CAPELLA_FORK_EPOCH: u64 = 1;
+pub const DENEB_FORK_EPOCH: u64 = 2;
+
+pub struct SimConfig {
+    pub node_count: usize,
+    pub proposer_nodes: usize,
+    pub validators_per_node: usize,
+    pub speed_up_factor: u64,
```
dknopik: Do we also speed up round timers?

anchor/integration/src/basic_sim.rs
```diff
@@ -0,0 +1,206 @@
...
+                compression: false,
+                is_restricted: true,
+                sse_logging: false,
+                extra_info: false,
+            },
+            matches,
+            EnvironmentBuilder::mainnet(),
+        );
+
+        let mut env = env_builder.multi_threaded_tokio_runtime()?.build()?;
```
dknopik, line 65: Have you tried enabling the tokio pause time thingy that basically uses a virtual time? But I guess that does not work well with the slot clock out of the box?

anchor/integration/src/cli.rs
```diff
@@ -0,0 +1,59 @@
...
+                )
+                .arg(
+                    Arg::new("continue-after-checks")
+                        .short('c')
+                        .long("continue_after_checks")
+                        .action(ArgAction::SetTrue)
+                        .help("Continue after checks (default false)"),
+                ),
+        )
+}
```
dknopik: Why not the declarative style as used by Anchor?

anchor/integration/src/mock_websocket.rs
```diff
@@ -0,0 +1,42 @@
...
+}
+
+impl MockServer {
+    pub async fn start() -> Result<Self, String> {
+        // Create a simple WebSocket route that just accepts connections
+        let ws_route = warp::ws().map(|ws: warp::ws::Ws| {
+            ws.on_upgrade(|_websocket| async {
+                // Connection established, but we don't need to do anything with it
+            })
+        });
```
dknopik: We should use `axum` instead of `warp`.

dknopik (commented):
Very nice!

I have some question.

We should also immediately add this to CI IMO

anchor/client/src/lib.rs
Zacholme7: There was an issue where the beacon and execution nodes would fully start up and wait for genesis while the anchor node would wait for genesis and then perform a bunch more startup before duties could occur.

This is the same spot that lighthouse has it and I think it makes sense. We should fully start up the anchor node before waiting for genesis to perform duties.

anchor/integration/src/mock_websocket.rs
Zacholme7: Will change. Why axum over wrap though?

anchor/integration/src/cli.rs
Zacholme7: No reason, this is what the lighthouse simulator does. Will change

anchor/client/src/lib.rs
dknopik: But effectively - we might as well remove it fully, because it is after the step waiting for operator id, right? Or do you pre-set the operator id before genesis in the simulator?

Zacholme7: DB is already populated with all relevant data, so nodes will start up and begin communicating right away while waiting for genesis whereas previously they were blocked waiting

We still need it because there is a configurable period of time to wait before genesis occurs.

Does that make sense?

diegomrsantos:
> The tests wont currently pass since some of the node functionality still does not work

Could you please mention what ones? Could we create an issue to enable them when the pre-conditions are met?

anchor/integration/src/basic_sim.rs
Zacholme7: No ill take a look

anchor/integration/src/basic_sim.rs
Zacholme7: ah no good catch. not sure it really matters since we are over localhost anyways but should still do.

Zacholme7:
@diegomrsantos no tests pass right now, they all hinge on block production and finalization but should be drop it when that is working. They run but just error.

anchor/integration/src/basic_sim.rs
Zacholme7: Round timers are hardcoded to 2 seconds and im not really sure its worth the effort here to nest some round timer config deep into the code just consistency with the sped up slot time.

anchor/integration/src/basic_sim.rs
Zacholme7: Looked into this and it is important to have actual time pass for work to be done so the virtual time is not applicable here.

diegomrsantos:
Is there copied code from LH in this PR?

diegomrsantos:
> @diegomrsantos no tests pass right now, they all hinge on block production and finalization but should be drop it when that is working. They run but just error.

Are you still working on block production and finalization in this PR?

diegomrsantos:
> I suppose the `http_api` was moved to `anchor_http_api` to avoid a conflict with LH's `http_api`. I propose that we leave the files in their previous location and just rename the crates in the `Cargo.toml`. (if really necessary)

Agree, seems redundant to use `anchor` prefix inside the project. Shouldn't the namespace be enough to distinguish them?

anchor/client/src/config.rs
```diff
@@ -54,6 +54,8 @@ pub struct Config {
     pub execution_nodes_tls_certs: Option<Vec<PathBuf>>,
     /// Configuration for the processor
     pub processor: processor::Config,
+    /// If database sync should be skipped
+    pub skip_sync: bool,
```
diegomrsantos, line 61: Why is this necessary?

anchor/client/src/lib.rs
```diff
@@ -109,7 +113,11 @@ impl Client {
             "Starting the Anchor client"
         );
 
-        let spec = Arc::new(config.ssv_network.eth2_network.chain_spec::<E>()?);
+        let spec = if let Some(spec) = spec {
+            Arc::new(spec)
+        } else {
+            Arc::new(config.ssv_network.eth2_network.chain_spec::<E>()?)
+        };
```
diegomrsantos, line 122: why do we need this?

anchor/client/src/lib.rs
```diff
@@ -733,18 +741,32 @@ async fn wait_for_operator_id_and_sync(
...
+                    result.ok()?;
+                    if let Some(id) = state.borrow().get_own_id() {
+                        break id;
+                    }
                 }
+                _ = sleep(sleep_duration) => info!("Waiting for operator id"),
             }
-            _ = sleep(sleep_duration) => info!("Waiting for operator id"),
         }
     };
```
diegomrsantos:
```rust
// Extract the value while the borrow is active.
let maybe_id = {
    let current_state = state.borrow();
    current_state.get_own_id()
};

// Now the borrow is dropped; match on the extracted value.
let id = match maybe_id {
    Some(id) => id,
    None => loop {
        select! {
            result = state.changed() => {
                result.ok()?;
                if let Some(id) = state.borrow().get_own_id() {
                    break id;
                }
            }
            _ = sleep(sleep_duration) => info!("Waiting for operator id"),
        }
    },
};
```

diegomrsantos: But maybe your version is better

anchor/eth/src/sync.rs
```diff
@@ -214,7 +214,7 @@ impl SsvEventSyncer {
 
     #[instrument(skip(self))]
     /// Try to perform both a historical and live sync from the chain
-    pub async fn sync(&mut self) -> Result<(), ExecutionError> {
+    pub async fn sync(&mut self, skip: bool) -> Result<(), ExecutionError> {
```
diegomrsantos, line 226: Can we split the `sync` function into `historical_sync` and `live_sync`, and decide to have `historical_sync` or not at the caller instead of passing a param?

diegomrsantos: Maybe it's better to keep everything unified in the sync function, but I'd rename the param to `skip_historical_sync`.

diegomrsantos: Thinking more about it, do we want to skip it, or should we do the historical_sync from a mock implementation or something like that?

anchor/client/src/lib.rs
Zacholme7: This is the hacky workaround that I mentioned in the internal chat related to not being able to set the slot timers. Its a band-aide over a bigger problem but I put it in in favor of being able to make some progress.

anchor/eth/src/sync.rs
Zacholme7: Will rename the param and I think we should keep it as a unified function as it was setup like this to handle errors and retry smoothly. 

Syncing from an mock contract was something I mentioned earlier but we decided against it as this is more of the purpose for ssv-mini in comparison to the simulator. It would be nice to have a fully simulated environment with all of the contracts to sync from  but it would probably be another 1.5k lines of infra to setup.

Zacholme7:
@diegomrsantos 

> Is there copied code from LH in this PR?

Yes I had to move the beacon node and execution node setup logic over

> Are you still working on block production and finalization in this PR?

Not planned for this pr

diegomrsantos:
> @diegomrsantos
> 
> > Is there copied code from LH in this PR?
> 
> Yes I had to move the beacon node and execution node setup logic over

Do we have a plan to address this duplication?

> > Are you still working on block production and finalization in this PR?
> 
> Not planned for this pr

How are we gonna merge it if the tests are failing?

anchor/eth/src/sync.rs
diegomrsantos: I don't know how valuable this would be, but having the simulator on CI testing this for each PR is different from having it on ssv-mini, isn't it?

anchor/client/src/lib.rs
diegomrsantos: Is there any major blocker for creating the simulator spec?

dknopik:
>How are we gonna merge it if the tests are failing?

We can merge it without integrating it into CI

anchor/eth/src/sync.rs
Zacholme7: Yes this is just an end to end integration test to make sure we dont have any unexpected behavior regressions while the simulator is a full devnet that tests interop with ssv nodes

diegomrsantos:
> > How are we gonna merge it if the tests are failing?
> 
> We can merge it without integrating it into CI

Is there any advantage in merging broken tests instead of adding them later when they'll work?

Zacholme7:
> Do we have a plan to address this duplication?

There is no way to address it without upstreaming anchor support into the lighthouse simulator which we dont want to do.

diegomrsantos:
> > Do we have a plan to address this duplication?
> 
> There is no way to address it without upstreaming anchor support into the lighthouse simulator which we dont want to do.

IMO, it's not a good idea to add a massive duplication without a plan to address it.

anchor/client/src/lib.rs
Zacholme7: Havent looked into that integration too much, but working block productions is the biggest blocker.

anchor/eth/src/sync.rs
diegomrsantos: I mean, how valuable would it be to have `historical_sync` in the regression tests?

anchor/client/src/lib.rs
diegomrsantos: I'm particularly not a fan of adding hacky workarounds unless there's no other sane option.

diegomrsantos:
> > Do we have a plan to address this duplication?
> 
> There is no way to address it without upstreaming anchor support into the lighthouse simulator which we dont want to do.

Isn't it possible to modularize and generalize the LH Simulator and reuse the common parts in Anchor? Similar to what I planned to do with the Peer Manager? @AgeManning

anchor/eth/src/sync.rs
Zacholme7: I personally dont think it is important to include here. I think the core idea here is to test core behavior like attestations & block production. The syncing section of the codebase is pretty finalized imo and if there are modifications this should be caught by ssv-mini or testnet deployment

Zacholme7:
> > > Do we have a plan to address this duplication?
> > 
> > 
> > There is no way to address it without upstreaming anchor support into the lighthouse simulator which we dont want to do.
> 
> Isn't it possible to modularize and generalize the LH Simulator and reuse the common parts in Anchor? Similar to what I planned to do with the Peer Manager? @AgeManning

I will look into it and see if there is anything that can be done. The main issue is that the LH simulator revolves around a `LocalNetwork` that does not include anchor nodes. This is pretty much used everywhere in the simulator so either I add some wrapper around it here or we just duplicate some code in favor of readability.

diegomrsantos:
> > > > Do we have a plan to address this duplication?
> > > 
> > > 
>>> There is no way to address it without upstreaming anchor support into the lighthouse simulator which we dont want to do.
> > 
> > 
> > Isn't it possible to modularize and generalize the LH Simulator and reuse the common parts in Anchor? Similar to what I planned to do with the Peer Manager? @AgeManning
> 
> I will look into it and see if there is anything that can be done. The main issue is that the LH simulator revolves around a `LocalNetwork` that does not include anchor nodes. This is pretty much used everywhere in the simulator so either I add some wrapper around it here or we just duplicate some code in favor of readability.



> > > > Do we have a plan to address this duplication?
> > > 
> > > 
> > > There is no way to address it without upstreaming anchor support into the lighthouse simulator which we dont want to do.
> > 
> > 
> > Isn't it possible to modularize and generalize the LH Simulator and reuse the common parts in Anchor? Similar to what I planned to do with the Peer Manager? @AgeManning
> 
> I will look into it and see if there is anything that can be done. The main issue is that the LH simulator revolves around a `LocalNetwork` that does not include anchor nodes. This is pretty much used everywhere in the simulator so either I add some wrapper around it here or we just duplicate some code in favor of readability.

Got it. If you have time, please point me to the more relevant parts in the LH code, and I can try to think about something similar to what I had in mind for the Peer Manager.

diegomrsantos:
> > > > Do we have a plan to address this duplication?
> > > 
> > > 
> > > There is no way to address it without upstreaming anchor support into the lighthouse simulator which we dont want to do.
> > 
> > 
> > Isn't it possible to modularize and generalize the LH Simulator and reuse the common parts in Anchor? Similar to what I planned to do with the Peer Manager? @AgeManning
> 
> I will look into it and see if there is anything that can be done. The main issue is that the LH simulator revolves around a `LocalNetwork` that does not include anchor nodes. This is pretty much used everywhere in the simulator so either I add some wrapper around it here or we just duplicate some code in favor of readability.

Got it. If you have time, please point me to the more relevant parts in the LH code, and I can try to think about something similar to what I had in mind for the Peer Manager.

anchor/integration/Cargo.toml
```diff
@@ -0,0 +1,30 @@
...
+environment = { git = "https://github.com/sigp/lighthouse", rev = "22faccb" }
+ethereum_serde_utils = { workspace = true }
+execution_layer = { git = "https://github.com/sigp/lighthouse", rev = "22faccb" }
+futures = { workspace = true }
+futures-util = "0.3.31"
+kzg = { git = "https://github.com/sigp/lighthouse", rev = "22faccb" }
+lighthouse_network = { workspace = true }
+logging = { git = "https://github.com/sigp/lighthouse", rev = "22faccb" }
+network = { workspace = true }
+node_test_rig = { git = "https://github.com/sigp/lighthouse", rev = "22faccb" }
```
dknopik: even though this is only used here - let's use the root Cargo.toml to specify the lighthouse revision. this makes updating it way easier

AgeManning:
Yeah, i think in general we want to avoid code duplication. However the simulator, I think can be quite dependent on the thing its simulating. 

As its mainly used for CI and testing, I think its fine to just have duplicate code, I'd personally be more concerned with core components. I.e I still think we merge the new Anchor peer manager into Lighthouse at some point. (Maybe that would help with the simulator also).

Zacholme7:
~~Thought the stack overflow was resolved but just hit it again, looking into it.~~

Mentioned in summary

Zacholme7:
Closing for now

PR #201, opened 2025-03-20
Anchor to respect log-level flag

AgeManning:
## Issue Addressed

Currently the --debug-level CLI flag in the Anchor subcommand doesn't do anything. 

In fact we set up the logging prior to running the subcommand. 

I've shifted the CLI flag to exist globally so we can adjust the loglevel for all subcommands and made the code respect the CLI choice. 

I think this might be useful for interop testing to optionally remove some noise if we want.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -730,6 +730,7 @@ dependencies = [
  "regex",
  "sensitive_url",
  "serde",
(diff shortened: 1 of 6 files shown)


jking-aus (approved):
lgtm

PR #205, opened 2025-03-21
Validate own messages

dknopik:
Validate outgoing messages. Useful for debugging. We can make this toggleable later.


jking-aus (approved):

anchor/message_sender/src/network.rs
```diff
@@ -22,6 +23,7 @@ pub struct NetworkMessageSender {
     network_tx: mpsc::Sender<(SubnetId, Vec<u8>)>,
     private_key: PKey<Private>,
     operator_id: OperatorId,
+    validator: Option<Validator>,
```
diegomrsantos, line 26: why an Option?

Issue #208, opened 2025-03-23
Improve Logging


AgeManning:
Currently logging in Anchor allows us to set the log level, however we don't have the ability to log libp2p or discovery logs. 

Also we don't have logging to file. 

Some of these are already templated in Lighthouse and it might be nice to move some of these feature across.

PR #209, opened 2025-03-25
File Logging

ThreeHrSleep:
## Issue Addressed

https://github.com/sigp/anchor/issues/208
file logging should work now
~still need to work on getting libp2p & discv5 logs though~ 
~edit1: (removed the stuff from here,will handle them in separate PR)~
edit2: added libp2p & discv5 stuff back in here as the fix was small enough


dknopik:
Few notes from my side:

- We should use a default log directory if none were set, e.g. <datadir>/logs
- I added ` --debug-level debug` but see only INFO for some reason
- Differing levels for console and file would be nice if possible, so that the user (console) only sees info, but we have debug in the file for troubleshooting. I think it's done this way in LH as well

anchor/src/main.rs
```diff
@@ -2,21 +2,24 @@ use clap::Parser;
...
+use logging::{filter_dependency_log, init_file_logging, LoggerConfig};
 use task_executor::ShutdownReason;
+use tracing::Level;
+use tracing_appender::non_blocking::WorkerGuard;
+use tracing_subscriber::{filter::FilterFn, fmt, prelude::*, EnvFilter};
 use types::EthSpecId;
 
 #[derive(Parser, Clone, Debug)]
 struct Cli {
+    #[clap(flatten)]
```
macladson: Ooh I like this pattern, I'm definitely stealing this for Lighthouse

anchor/src/main.rs
```diff
@@ -134,3 +136,39 @@ fn start_anchor(anchor_config: Node, mut environment: Environment) {
...
+    let (file_appender, guard) = init_file_logging(logger_config.clone());
+
+    let file_layer = fmt::layer().with_writer(file_appender);
+
+    if let Err(e) = tracing_subscriber::registry()
+        .with(env_filter)
+        .with(fmt::layer())
+        .with(dependency_log_filter)
+        .with(file_layer)
+        .try_init()
```
macladson: For the `file_layer`, it might be worth disabling the layer entirely when it's disabled (instead of setting to layer to output as `std::io::sink`). You can use `with()` with a `Vec<Layer>` so you can actually conditionally push the layers you want. I recently implemented that in Lighthouse so it might be useful for you here too.

anchor/logging/src/logging.rs
```diff
@@ -0,0 +1,84 @@
...
+    pub debug_level: Level,
+    pub max_log_size: u64,
+    pub max_log_number: usize,
+    pub compression: bool,
+}
+impl Default for LoggerConfig {
+    fn default() -> Self {
+        LoggerConfig {
+            path: None,
+            debug_level: Level::TRACE,
```
macladson: This doesn't match the `Level` set in `default_debug_level()`. I assume these should be the same?

anchor/logging/src/logging.rs
```diff
@@ -0,0 +1,84 @@
...
+    pub guard: WorkerGuard,
+}
+
+pub fn init_file_logging(config: LoggerConfig) -> (NonBlocking, WorkerGuard) {
+    let filename = PathBuf::from("anchor.log");
+
+    let file_appender = match config.path {
+        None => {
+            eprintln!("No logfile path provided, logging to file is disabled");
+            return tracing_appender::non_blocking(std::io::sink());
```
macladson: See my comment in `main.rs` about this

anchor/src/main.rs
```diff
@@ -134,3 +136,39 @@ fn start_anchor(anchor_config: Node, mut environment: Environment) {
...
     };
 }
+
+fn enable_logging(logging_flags: &LoggingFlags) -> WorkerGuard {
+    let cli = logging_flags.clone();
+    let filter_level: Level = cli.debug_level.into();
+
+    let logger_config = LoggerConfig {
+        path: cli.logfile_dir.clone(),
+        debug_level: filter_level,
```
macladson, line 188: I don't think this value gets used anywhere, which is probably why `--debug-level` doesn't work.

macladson (commented):
Just giving this a quick review, but I didn't test anything so was just looking over the code. Hopefully this is helpful

anchor/src/main.rs
ThreeHrSleep: lol 😆 ,, happy to hear that

anchor/src/main.rs
ThreeHrSleep: Nice idea! Thanks.. omw to ~steal~ get inspired from Lighthouse's code  again

anchor/logging/src/logging.rs
ThreeHrSleep: Yup, my bad .. will change

anchor/src/main.rs
ThreeHrSleep: so i poked around a bit & i think that the major issue lied in [`--debug-level`](https://github.com/ThreeHrSleep/anchor/blob/85c6e1b8621ea823de01b4da63239a2fae6817d6/anchor/client/src/cli.rs#L36-L40) not being marked with `global = true`

but i changed a bunch of other stuff before actually fixing the issue so there might be other problematic parts aswell which could've contributed to the issue, so i could be wrong here

ThreeHrSleep:
Thanks for the review(s)!
i think all the suggestions should be covered now : )

anchor/logging/src/logging.rs
```diff
@@ -0,0 +1,88 @@
...
+use logroller::{Compression, LogRollerBuilder, Rotation, RotationSize};
+use serde::{Deserialize, Serialize};
+use tracing::Level;
+use tracing_appender::non_blocking::{NonBlocking, WorkerGuard};
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct LoggerConfig {
+    pub path: Option<PathBuf>,
+    #[serde(skip_serializing, skip_deserializing, default = "default_debug_level")]
+    pub debug_level: Level,
```
macladson: The `LoggerConfig` doesn't really _need_ to implement `Serialize` or `Deserialize`, however, if in the future we want to do something like reading in a config file instead of cli flags, we need a way to parse the file into the `LoggerConfig` struct, hence deserializing would be helpful. I believe the current code (and it's like this in Lighthouse too), would just ignore whatever field is in the config file for `debug_level` and set it as `INFO` regardless.

Since reading in from a config file isn't supported yet, it's a non-issue, so you could just remove the derived impls entirely if you wanted to. 

(Lighthouse impls `Serialize` and `Deserialize` on its config structs since it has some tests/tooling to dump out the config files and check certain values, even though we don't yet support reading in config files)

anchor/logging/src/logging.rs
```diff
@@ -0,0 +1,88 @@
...
+            path: None,
+            debug_level: Level::INFO,
+            max_log_size: 20,
+            max_log_number: 5,
+            compression: false,
+        }
+    }
+}
+
+fn default_debug_level() -> Level {
```
macladson: Removing serde would also allow you to remove this function.

anchor/logging/src/logging.rs
```diff
@@ -0,0 +1,88 @@
...
+    pub max_log_number: usize,
+    pub compression: bool,
+}
+impl Default for LoggerConfig {
+    fn default() -> Self {
+        LoggerConfig {
+            path: None,
+            debug_level: Level::INFO,
+            max_log_size: 20,
+            max_log_number: 5,
```
macladson: Personally, would love to see some `const`s here instead

macladson (commented):
I still have some more things to review but here's some things I noticed for now

anchor/logging/src/logging.rs
ThreeHrSleep: Ohh! i just blindly stole the stuff without thinking much ...
Thanks! cleaned it up

anchor/client/src/cli.rs
```diff
@@ -30,6 +31,83 @@ pub static LONG_VERSION: LazyLock<String> = LazyLock::new(|| {
 
 pub const FLAG_HEADER: &str = "Flags";
 
+#[derive(Parser, Debug, Clone, Deserialize, Serialize)]
+pub struct LoggingFlags {
```
macladson: This is more of a stylistic thing, and it will depend on how the rest of the anchor project is structured, but personally I'd try to keep all logging code inside the `logging` crate.

Then only import `LoggingFlags` (either here in `Node` or in `main.rs` inside `Cli`: see my other comment) combined with `#[clap(flatten)]` would make it all very tidy imo

anchor/client/src/cli.rs
```diff
@@ -444,6 +522,9 @@ pub struct Node {
         display_order = 0
     )]
     pub disable_slashing_protection: bool,
+
+    #[clap(flatten)]
+    pub logging_flags: LoggingFlags,
```
macladson, line 519: I don't think this is required since the same code is already in `main.rs`.

anchor/src/main.rs
```diff
@@ -1,22 +1,34 @@
...
+use tracing::Level;
+use tracing_appender::non_blocking::WorkerGuard;
+use tracing_subscriber::{filter::FilterFn, fmt, prelude::*, EnvFilter};
 use types::EthSpecId;
 
 #[derive(Parser, Clone, Debug)]
 struct Cli {
+    #[clap(flatten)]
+    pub logging_flags: LoggingFlags,
+
```
macladson: Having `LoggingFlags` inside `Cli` here will apply these flags to all anchor subcommands including `Node`, `Keygen` and `Keysplit`. If you want `anchor keygen` and `anchor keysplit` to accept logging flags and they are relevant, then this is completely fine (I haven't looked at the code enough to know)

But if we only initialize logging for `Node` then we should remove this from `Cli` and instead put it directly inside `Node`. That way it won't show up for every subcommand.

macladson (commented):
Didn't get a chance to do a very thorough review but just had some more code structure comments for you

anchor/src/main.rs
ThreeHrSleep: Nice idea! yeahh i don't think `keygen` & `keysplit` will be needing these flags, 
so keeping this stuff to only `Node` should be fine  
(just putting a one-liner `tracing_subscriber::fmt().init()` in keygen & keysplit separately and we should be good ig)
Thanks!

anchor/client/src/cli.rs
ThreeHrSleep: removed `logging_flags` from `main.rs`(to address the other comment), 
so i guess we can keep it here now?

ThreeHrSleep:
Thanks again @macladson !! : )
Could you please take one last look to confirm if we're good to go(or if anything needs fixing😅)  ?

dknopik:
Tried this out on interop, there seem to be a few issues. Will dig into it more later.

anchor/logging/src/logging.rs
```diff
@@ -0,0 +1,170 @@
...
+        default_value_t = DebugLevel::Info,
+        help = "Specifies the verbosity level used when emitting logs to the terminal")]
+    pub debug_level: DebugLevel,
+
+    #[arg(
+        long,
+        global = true,
+        default_value_t = DebugLevel::Info,
+        help = "Specifies the verbosity level used when emitting logs to the log file")]
+    pub logfile_debug_level: DebugLevel,
```
dknopik: For the file logging, the default should be `Debug` IMO

dknopik:
There seems to be a bug as `libp2p_gossipsub` and `discv5` logs show up in the `anchor.log` file.

EDIT: actually, all three log files seem to contain all logs.

dknopik:
Without this branch, `cargo run -- node -t not_existing_dir` returns

```
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 12.92s
     Running `target/debug/anchor node -t not_existing_dir`
2025-04-22T13:18:37.668948Z ERROR anchor: Unable to initialize configuration e="Unable to read \"not_existing_dir/ssv_contract_address.txt\": No such file or directory (os error 2)"
```

With this branch, it returns

```
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.62s
     Running `target/debug/anchor node -t not_existing_dir`
```
so apparently, the error is suppressed.

anchor/keysplit/src/lib.rs
```diff
@@ -43,6 +43,7 @@ struct ValidatorKeys {
 }
 
 pub fn run_keysplitter(keysplit: Keysplit) -> Result<(), KeysplitError> {
+    tracing_subscriber::fmt().init();
```
dknopik: Instead of putting this here and in `run_keygen`, what do you think about putting it into the `_` match arm in `main.rs`?

ThreeHrSleep: Nice idea! ..Moved it : )

ThreeHrSleep:
Thanks for catching all the issues ! (& apologies for not testing the changes properly myself)
all of these should fixed now(hopefully), 

please lmk if i broke anything else this time : )

anchor/src/main.rs
```diff
@@ -134,3 +149,118 @@ fn start_anchor(anchor_config: Node, mut environment: Environment) {
...
+        .join(
+            config
+                .ssv_network
+                .eth2_network
+                .config
+                .config_name
+                .as_deref()
+                .unwrap_or("custom"),
+        )
+        .join("logs");
```
dknopik: Sorry, found another small issue. This should be `datadir/logs` if a datadir is set via CLI.

ThreeHrSleep: Ah,Thanks for pointing it out!
Done :)

anchor/src/main.rs
```diff
@@ -134,3 +149,126 @@ fn start_anchor(anchor_config: Node, mut environment: Environment) {
...
+                    .unwrap_or("custom"),
+            )
+            .join("logs")
+    } else {
+        anchor_config
+            .datadir
+            .clone()
+            .unwrap_or_else(|| PathBuf::from("."))
+            .join("logs")
+    };
```
dknopik: We can simplify this a bit and avoid some clones :)

```suggestion
    let default_logs_dir = if let Some(datadir) = &anchor_config.datadir {
        datadir.join("logs")
    } else {
        dirs::home_dir()
            .unwrap_or_else(|| PathBuf::from("."))
            .join(DEFAULT_ROOT_DIR)
            .join(
                config
                    .ssv_network
                    .eth2_network
                    .config
                    .config_name
                    .as_deref()
                    .unwrap_or("custom"),
            )
            .join("logs")
    };
```

dknopik (commented):
small nitpick

Otherwise, LGTM, seems to work well on interop. Can someone else take an additional look before merging?

jking-aus (approved):

PR #211, opened 2025-03-25
Keygen bugfix

Zacholme7:
## Issue Addressed

N/A

## Proposed Changes

Passwords were being defaulted to "" if it was not supplied. This would probably confuse a lot of users as to what their password was unless they read the code.


jking-aus (approved):
lol good catch

PR #212, opened 2025-03-26
Introduce validator index syncer

dknopik:
- Switch index to Option
- Clean up event syncer Config
- Remove obsolete testing binary


anchor/client/src/cli.rs
```diff
@@ -111,6 +111,15 @@ pub struct Node {
...
     pub execution_nodes: Option<Vec<String>>,
 
+    #[clap(
+        long,
+        value_name = "NETWORK_ADDRESSES",
+        help = "Comma-separated addresses to one or more beacon node WS APIs. \
+                Default is http://localhost:8546.",
+        display_order = 0
+    )]
+    pub execution_nodes_websocket: Option<Vec<String>>,
```
Zacholme7: nits on these. 
1) help is calling execution endpoints beacon endpoints
2) Maybe simplify these to execution_http & execution_ws?

anchor/database/src/validator_operations.rs
```diff
@@ -61,4 +63,38 @@ impl NetworkDatabase {
...
+                if let Some(mut validator) =
+                    state.multi_state.validator_metadata.get_by(&public_key)
+                {
+                    // Update in memory
+                    validator.index = Some(index);
+                    state
+                        .multi_state
+                        .validator_metadata
+                        .update(&public_key, validator);
+                }
```
Zacholme7, line 100: Maybe warn/error log missing validators? Should the entries always exist?

anchor/eth/src/index_sync.rs
```diff
@@ -0,0 +1,144 @@
...
+use tokio::time::sleep;
+use tracing::{debug, error, info, warn};
+use types::PublicKeyBytes;
+
+pub type Tx = UnboundedSender<PublicKeyBytes>;
+
+const INDEX_SYNCER_NAME: &str = "validator_index_syncer";
+
+const MAX_BATCH_SIZE: usize = 512;
+const BATCHING_DELAY: Duration = Duration::from_secs(1);
```
Zacholme7, line 23: Think 1second is large enough? sig verify is the bottle neck in processing and this coincides with validators added. On a heavy block with lots of batched validator added events it might be useful to have more of a buffer on this. Fetching from the bn is the quick part.

Zacholme7 (commented):
This is slick, nice job. Left tiny nits

anchor/eth/src/index_sync.rs
```diff
@@ -0,0 +1,144 @@
...
+
+            // sort and skip to current position
+            from_database.sort_unstable_by_key(|x| x.serialize());
+            batch.extend(
+                from_database
+                    .into_iter()
+                    .skip(db_sweep)
+                    .take(space)
+                    .map(ValidatorId::PublicKey),
+            );
```
Zacholme7: I think this could all be turned into a quick db query. Something like 
`
SELECT public_key 
FROM validators 
WHERE index IS NULL;
`
not sure if this would result in a speedup or not though

anchor/database/src/validator_operations.rs
dknopik: There is the edge case of a validator being removed while the request is in flight. That is not really an error though - so I decided against logging. But given the rarity of that edge case a warn log might be worth it, yeah

anchor/eth/src/index_sync.rs
dknopik: The delay resets if a validator is fetched from the queue - so it's second per validator, which should be fine. my tests on interop showed that we regularly hit the max batch size during sync from the queue alone

I also do not want to increase the delay too much as we want to get the index ASAP during live operation.

Zacholme7: Ah okay, nice.

anchor/eth/src/index_sync.rs
dknopik: pushed a test, lets try it out :)

Will report back

anchor/client/src/cli.rs
dknopik: Decided on `execution_rpc` instead of `execution_http` to emphasise the difference, is that fine?

anchor/eth/src/index_sync.rs
dknopik: It's slower by a factor of ~50 (which surprised me), therefore I propose leaving it like this

anchor/client/src/cli.rs
Zacholme7: yep, thanks

anchor/eth/src/index_sync.rs
Zacholme7: Oh interesting, I suppose its still fetching from db at the end of the day. thanks for giving it a shot

anchor/client/src/config.rs
```diff
@@ -36,6 +36,8 @@ pub struct Config {
     pub proposer_nodes: Vec<SensitiveUrl>,
     /// The http endpoints of the execution node APIs.
     pub execution_nodes: Vec<SensitiveUrl>,
+    /// The websocket endpoints of the execution node APIs.
+    pub execution_nodes_websocket: Vec<SensitiveUrl>,
```
diegomrsantos, line 40: nit: maybe group them in a struct?

anchor/client/src/config.rs
```diff
@@ -132,14 +133,22 @@ pub fn from_cli(cli_args: &Node) -> Result<Config, String> {
...
     }
 
+    if let Some(execution_nodes_websocket) = &cli_args.execution_ws {
+        config.execution_nodes_websocket = execution_nodes_websocket
+            .iter()
+            .map(|s| SensitiveUrl::parse(s))
+            .collect::<Result<_, _>>()
+            .map_err(|e| format!("Unable to parse execution node URL: {:?}", e))?;
+    }
+
```
diegomrsantos: It'd be nice to remove the duplication with the code above

anchor/common/ssv_types/src/cluster.rs
```diff
@@ -77,7 +77,7 @@ pub struct ValidatorMetadata {
     /// The cluster that is responsible for this validator
     pub cluster_id: ClusterId,
     /// Index of the validator
-    pub index: ValidatorIndex,
+    pub index: Option<ValidatorIndex>,
```
diegomrsantos, line 81: Could you elaborate on the motivation behind this?

dknopik: We do not always have a validator index, as it is not supplied via SSV contract event. There, the validator public key is supplied instead.

The validator only gets a validator index once it receives a beacon chain deposit. This means someone might enable a validator in SSV, but the validator is not active in the beacon chain. So we have no validator index for that validator (and might never have, actually!).

Before the change, we represented this scenario by setting a validator index of 0. This is not idiomatic vs just representing the absence of a value via an `Option`, which is why I changed it.

anchor/client/src/config.rs
dknopik: Hmm, I think personally I prefer it flat here. Do you have specific advantages in mind that I might not see right now?

diegomrsantos: something like that seems nicer and easier to pass around:

```rust
pub struct ExecutionNodes {
    pub http: Vec<SensitiveUrl>,
    pub websocket: Vec<SensitiveUrl>,
}

pub struct NodesConfig {
    pub beacon_nodes: Vec<SensitiveUrl>,
    pub proposer_nodes: Vec<SensitiveUrl>,
    pub execution_nodes: ExecutionNodes,
}
```

diegomrsantos: But maybe a matter of taste

Zacholme7: I also prefer it flat

anchor/client/src/config.rs
```diff
@@ -236,6 +237,22 @@ pub fn from_cli(cli_args: &Node) -> Result<Config, String> {
     Ok(config)
 }
 
+/// Read SensitiveUrls from given CLI Strings
+fn read_urls(
```
diegomrsantos: maybe `sanitize_urls`?

anchor/client/src/config.rs
```diff
@@ -127,21 +128,21 @@ pub fn from_cli(cli_args: &Node) -> Result<Config, String> {
...
-            .iter()
-            .map(|s| SensitiveUrl::parse(s))
-            .collect::<Result<_, _>>()
-            .map_err(|e| format!("Unable to parse execution node URL: {:?}", e))?;
-    }
+    read_urls(
+        &mut config.beacon_nodes,
+        &cli_args.beacon_nodes,
+        "beacon node",
+    )?;
```
diegomrsantos:
```suggestion
if let Some(ref beacon_nodes) = cli_args.beacon_nodes {
    read_urls(&mut config.beacon_nodes, beacon_nodes, "beacon node")?;
}
```

anchor/client/src/config.rs
```diff
@@ -236,6 +237,22 @@ pub fn from_cli(cli_args: &Node) -> Result<Config, String> {
...
+    if let Some(beacon_nodes) = src {
+        *dest = beacon_nodes
+            .iter()
+            .map(|s| SensitiveUrl::parse(s))
+            .collect::<Result<_, _>>()
+            .map_err(|e| format!("Unable to parse {kind} URL: {:?}", e))?;
+    }
+    Ok(())
+}
+
```
diegomrsantos:
```suggestion
fn read_urls(
    dest: &mut Vec<SensitiveUrl>,
    src: &[String],
    kind: &str,
) -> Result<(), String> {
    *dest = src
        .iter()
        .map(|s| SensitiveUrl::parse(s))
        .collect::<Result<_, _>>()
        .map_err(|e| format!("Unable to parse {kind} URL: {:?}", e))?;
    Ok(())
}

```

anchor/client/src/lib.rs
```diff
@@ -322,25 +324,26 @@ impl Client {
         // Wait until genesis has occurred.
         wait_for_genesis(&beacon_nodes, genesis_time).await?;
 
+        // Start validator index syncer
+        let index_sync_tx =
```
diegomrsantos, line 328: It's unrelated to this PR, but could we split the `run` function into smaller ones? It's almost 500 lines long, way too big, IMO.

anchor/common/ssv_types/src/cluster.rs
diegomrsantos: > This means someone might enable a validator in SSV, but the validator is not active in the beacon chain

Is there a way to allow validator registration only if it's active in the beacon chain?

dknopik: I think it might be possible if you pass a giant merkle proof over the beacon state to the SSV contract :D It's definitely not worth it.

There is also the use-case of registering a validator at the same time as depositing it - the validator is not immediately active then, but will become active soon.

anchor/client/src/config.rs
dknopik: I wanted to minimise repetition, which was the purpose of this change

diegomrsantos: Imo readability of the code is also important. It's more intuitive to not call the function if the input is none.

diegomrsantos: Another option is grouping the configs in something like this:

```rust
pub enum NodeType {
    Beacon,
    Proposer,
    Execution(ExecutionType),
}

pub enum ExecutionType {
    Http,
    WebSocket,
}

pub struct NodeEndpoint {
    pub node_type: NodeType,
    pub url: SensitiveUrl,
}

pub struct NodesConfig {
    pub endpoints: Vec<NodeEndpoint>,
}
```

But I don't know if it'd make other things more difficult

diegomrsantos: Or even simpler

```rust
pub enum Node {
    Beacon(SensitiveUrl),
    Proposer(SensitiveUrl),
    ExecutionHttp(SensitiveUrl),
    ExecutionWebSocket(SensitiveUrl),
}

pub struct NodesConfig {
    pub nodes: Vec<Node>,
}
```

dknopik: Then we have to filter out the appropriate URLs later on, again creating Vecs of the filtered values :sweat_smile: really not a fan of this

diegomrsantos: It's not that bad to add a `get_beacon_nodes` function to NodesConfig, this `vec` is pretty small. But I see you don't like it, and I don't have a strong opinion at all.

anchor/database/src/validator_operations.rs
```diff
@@ -62,4 +66,40 @@ impl NetworkDatabase {
...
+        &self,
+        map: HashMap<PublicKeyBytes, ValidatorIndex>,
+    ) -> Result<(), DatabaseError> {
+        // Update the database
+        let mut conn = self.connection()?;
+        let transaction = conn.transaction()?;
+        for (public_key, index) in map.iter() {
+            transaction
+                .prepare_cached(SQL[&SqlStatement::SetIndex])?
+                .execute(params![
```
diegomrsantos, line 80: how big can the map get?

anchor/eth/src/event_processor.rs
```diff
@@ -8,13 +8,21 @@ use tracing::{debug, error, info, instrument, trace, warn};
...
-    error::ExecutionError, event_parser::EventDecoder, gen::SSVContract,
+    error::ExecutionError, event_parser::EventDecoder, gen::SSVContract, index_sync,
     network_actions::NetworkAction, util::*,
 };
 
 // Specific Handler for a log type
 type EventHandler = fn(&EventProcessor, &Log) -> Result<(), ExecutionError>;
 
+pub enum Mode {
+    Node {
```
diegomrsantos, line 23: Could we add a comment describing what this mode means?

anchor/eth/src/event_processor.rs
```diff
@@ -23,12 +31,12 @@ pub struct EventProcessor {
     /// Reference to the database
     pub db: Arc<NetworkDatabase>,
     /// Signal if we should only do relevant keysplitting processing
-    keysplit: bool,
+    mode: Mode,
```
diegomrsantos, line 41: Should we update the comment?

anchor/eth/src/event_processor.rs
```diff
@@ -235,9 +239,12 @@ impl EventProcessor {
         })?;
 
         // During keysplitting, we only care about the nonce
-        if self.keysplit {
+        let Mode::Node {
```
diegomrsantos, line 249: Should we update the comment?

anchor/eth/src/index_sync.rs
```diff
@@ -0,0 +1,145 @@
...
+) -> Tx {
+    let (tx, rx) = unbounded_channel();
+    executor.spawn(validator_index_syncer(nodes, db, rx), INDEX_SYNCER_NAME);
+    tx
+}
+
+async fn validator_index_syncer(
+    nodes: Arc<BeaconNodeFallback<impl SlotClock>>,
+    db: Arc<NetworkDatabase>,
+    mut validator_queue_rx: UnboundedReceiver<PublicKeyBytes>,
```
diegomrsantos: Should we make it bounded?

anchor/eth/src/index_sync.rs
```diff
@@ -0,0 +1,145 @@
...
+                .values()
+                .filter_map(|v| {
+                    (v.index.is_none()
+                        && !batch.contains(&ValidatorId::PublicKey(v.public_key))
+                        && clusters
+                            .get_by(&v.cluster_id)
+                            .is_some_and(|c| !c.liquidated))
+                    .then_some(v.public_key)
+                })
+                .collect::<Vec<_>>();
```
diegomrsantos: This is a bit convoluted for a declarative approach imo

diegomrsantos: But if you'd like to keep it, I'd extract the logic inside the filter to a function

anchor/eth/src/index_sync.rs
```diff
@@ -0,0 +1,145 @@
...
+            if db_sweep >= count {
+                db_sweep = 0;
+            }
+        }
+
+        if !batch.is_empty() {
+            debug!(len = batch.len(), "Sending request");
+            let validators = nodes
+                .first_success(move |client| {
+                    let batch = batch.clone();
```
diegomrsantos: don't know how often this will be called, but can this clone get expensive?

anchor/database/src/validator_operations.rs
dknopik: Up to `MAX_BATCH_SIZE`, which is 512.

anchor/eth/src/event_processor.rs
dknopik: No, the purpose is still the same

anchor/eth/src/event_processor.rs
dknopik: No, the purpose is still the same

anchor/eth/src/index_sync.rs
dknopik: Unfortunately, I do not see a way to avoid a clone, as `post_beacon_states_validators` wants a `Vec`. This is called once per connected beacon node.

dknopik: I'll shuffle the code around a bit to also handle mapping from PublicKeyBytes to ValidatorId here, this way it looks a bit more useful ;D

diegomrsantos (approved):
Thank you for this PR and for addressing the comments. Sorry they were too many ;)

dknopik:
Thank you for having a good look at my code!

PR #214, opened 2025-03-28
Address BLST audit

dknopik:
## Issue Addressed

After an internal audit, one major and several minor issues were found in our lagrange interpolation implementation using `blst`. Recall that this implementation was not in active use until now.

## Proposed Changes

I will go over the issues found and the solution implemented for it.

#### Misuse of `blst_scalar_from_uint64` :hot_pepper: 

`blst_scalar_from_uint64` reads four `u64` instead of one (as assumed by the author). This caused the `From` and `TryFrom` implementations to read additional memory after the passed `u64`. This was not caught before as the crate was never used productively, and the tests triggered this in a way where the IDs ended up the same regardless, causing no apparent issue.

We now instead use `blst_scalar_from_le_bytes`, which allows us to explicitly choose the amount of bytes to read. Fixed in commit https://github.com/sigp/anchor/commit/25bb849d281b660bb7f82e26ca3bb5348b4df448.

#### Rename `keys` and `msk`

These variable names are unclear: while "keys" corresponds to the type, they are not actually used as keys, but coefficients in our polynomial. `msk` is an unclear acronym.

In commit https://github.com/sigp/anchor/commit/ab74ae02e35c1f2a1d075e81bc7612bbadef11a7, `keys` has been renamed to `random_coefficients`, and `msk` has been renamed to `coefficients`.

#### Check for zero input key

The coefficients must not be zero. This includes the secret as constant coefficient, and the random coefficients.

The random coefficients can not be zero, as `blst`s `key_gen` [ensures that it does not generate a zero key](https://github.com/supranational/blst/blob/35844fecb2fedf3612adb161bbbdc0da439b9a2e/src/keygen.c#L194-L195). While it difficult for a user to even create a zero key to be split, a check was added in commit https://github.com/sigp/anchor/commit/3b72be2b7b6eb3701dd362ef3b6a99e1d9400ed0 to be safe.

#### Initialise scratch memory

We create some scratch space memory for use in `blst`s implementation of Pippenger's algorithm. While the memory is allowed to be uninitialised as a raw pointer is passed, we switch to initialising the `Vec` instead of just allocating it in commit https://github.com/sigp/anchor/commit/42ba3025ef39ad170d47d6f3ed623a21755bdad6 for clarity.

## Additional Info

Now that we are audited, this PR switches to use the single thread `blst` implementation by default for now, and remove the warning. Some additional error case tests have been added.


anchor/common/bls_lagrange/src/blst.rs
```diff
@@ -36,7 +23,7 @@ impl TryFrom<u64> for KeyId {
         if value != 0 {
             unsafe {
                 let mut id = blst_scalar::default();
-                blst_scalar_from_uint64(&mut id, &value);
+                blst_scalar_from_le_bytes(&mut id, &value.to_le_bytes()[0], 8);
```
kirk-baird: This works as is, though I'd consider caching the u64 -> bytes and using the length of bytes rather than hard coding. e.g.

```rust
                let value_le_bytes = value.to_le_bytes();
                blst_scalar_from_le_bytes(&mut id, key_ley_bytes.as_ptr(), value_le_bytes.len());
```

anchor/common/bls_lagrange/src/blst.rs
```diff
@@ -51,7 +38,7 @@ impl From<NonZeroU64> for KeyId {
     fn from(value: NonZeroU64) -> Self {
         unsafe {
             let mut id = blst_scalar::default();
-            blst_scalar_from_uint64(&mut id, &value.get());
+            blst_scalar_from_le_bytes(&mut id, &value.get().to_le_bytes()[0], 8);
```
kirk-baird: Same as [above](https://github.com/sigp/anchor/pull/214/files#r2020307966)

kirk-baird (dismissed):
Looks good to me.

Just 2 nit picks which you can take or leave, just a bit of a coding style.

anchor/common/bls_lagrange/src/blsful.rs
```diff
@@ -55,23 +56,24 @@ pub fn split_with_rng(
...
     let key = if result.is_some().into() {
-        Zeroizing::new(IdentifierPrimeField(result.unwrap()))
+        let scalar = result.unwrap();
+        if scalar.is_zero().unwrap_u8() != 0 {
+            return Err(Error::ZeroKey);
+        }
+        Zeroizing::new(IdentifierPrimeField(scalar))
     } else {
         return Err(Error::InternalError);
     };
```
diegomrsantos:
```rust
    let scalar = result.ok_or(Error::InternalError)?;
    if bool::from(scalar.is_zero()) {
        return Err(Error::ZeroKey);
    }
    let key = Zeroizing::new(IdentifierPrimeField(scalar));
 ```

dknopik: Result is a `CtOption`, which does not have `ok_or`

diegomrsantos: How about `let scalar = result.into_option().ok_or(Error::InternalError)?;`

dknopik:
@jking-aus I think you should take a look at this, especially as it sets `blst` as the active implementiation

jking-aus (approved):
Thanks for the audit kirk! And Daniel, must feel good to get that warning out of codebase.

PR #215, opened 2025-03-28
Signature verification

diegomrsantos:
## Issue Addressed

## Proposed Changes

This pull request adds signature verification and refactors the validation logic by introducing a centralized ValidationContext and updates several message validation functions and tests to use this new context. Key changes include:

- Replacing multiple function parameters with a single ValidationContext struct in partial signature and consensus message verification.
- Integrating OpenSSL-based signature verification into the validation flows.
- Updating tests to reflect the new ValidationContext usage and enhanced signature verification.


anchor/message_validator/src/lib.rs
```diff
@@ -173,6 +185,14 @@ pub enum Error {
     Processor(#[from] ::processor::Error),
 }
 
+struct ValidationContext<'a> {
```
Zacholme7, line 188: Are the borrows here just to reduce some clone overheads?

anchor/message_validator/src/lib.rs
```diff
@@ -250,6 +280,23 @@ impl<S: SlotClock> Validator<S> {
         }
     }
 
+    fn get_operator_pks(
```
Zacholme7, line 283: nit: I would probably extract this into a function in `state.rs` instead of keeping it here.

Zacholme7 (commented):
small comments, otherwise looking good!

anchor/message_validator/src/lib.rs
diegomrsantos: Exactly, wdyt?

anchor/message_validator/src/lib.rs
diegomrsantos: why? It's not related to the consensus state

anchor/message_validator/src/lib.rs
```diff
@@ -268,36 +315,81 @@ impl<S: SlotClock> Validator<S> {
 }
 
 fn validate_ssv_message(
-    signed_ssv_message: &SignedSSVMessage,
-    committee_info: &CommitteeInfo,
-    role: Role,
+    validation_context: &ValidationContext,
```
dknopik: I like that the ValidationContext borrows its contents, but borrowing again here basically creates a double indirection. Wdyt about passing the context by value?

dknopik: Or, alternatively, having the Context own the values and passing around a reference, which is maybe even better.

diegomrsantos: Good catch! But if it owns the values, we'll have to copy it when creating the struct. Is it necessary?

dknopik: Moving the values once might have better performance than having one additional level of pointer indirection during the whole of validation. But I don't know. We can also go ahead with my first suggestion if you prefer that.

diegomrsantos: Now I remember that the reason I passed it as a ref is to avoid it being moved. Doesn't the compiler optimize away the indirections?

dknopik: How would it?

diegomrsantos: If I move it to a function, it can't be used anymore

anchor/message_validator/src/lib.rs
Zacholme7: I meant `state.rs` in the database as this function deals with fetching state. But if this is the only place this is needed then there is not much point in moving it I suppose

anchor/message_validator/src/lib.rs
diegomrsantos: Ah, you mean the indirections. The contents of the struct aren’t mutated anywhere, so there’s no reason to keep loading them from memory many times.

anchor/message_validator/src/lib.rs
diegomrsantos: I see, yeah it's used only here

diegomrsantos:
@dknopik @Zacholme7 is there anything pending?

dknopik:
Gonna test it on interop tomorrow, after that I think we are good to go imo

dknopik (approved):
lgtm

PR #216, opened 2025-03-28
Disable slashing protection

Zacholme7:
## Issue Addressed

N/A

## Proposed Changes

Add cli option to disable slashing protection


dknopik:
We handle metrics within `handle_slashing_check_result`, so I suggest passing `Safe::Valid` to it if slashing protection is disabled instead of not calling it.

dknopik:
General question: What do you think about making this a hidden option? The user should not be tempted - even if there is an all caps warning...

diegomrsantos:
Is it for ssv-mini? Isn't it possible to clean the db before running?

Zacholme7:
@dknopik I think that is a good idea, will update

Zacholme7:
@diegomrsantos This was prompted by the simulator, not ssv-mini. I did go the route of cleaning the dbs before running, but this is still an option that we should include regardless.

dknopik (approved):
LGTM

PR #218, opened 2025-03-31
Add support for Electra data

dknopik:
Add `DATA_VERSION_ELECTRA` and properly set it in our qbft starting values for blocks and aggregates.

This is useful for post-Electra networks :)


diegomrsantos (approved):
LGTM

PR #220, opened 2025-04-01
No error on sending duplicates

dknopik:
Decided messages are sent on successful completion of a QBFT instance, by all participants of that instance. 

This causes exact copies of these messages to be sent by multiple parties. These messages have the same message ID, as they have the same content. When calling Gossipsub's `publish`, a "Duplicate" error is returned if the message has been seen before.

We do not really want to check if we have seen the message before, as the infrastructure and overhead is not worth it. Instead, ignore "Duplicate" errors. There is still a warning logged by libp2p, which is IMO sufficient.

### DIFF
--- a/anchor/network/src/network.rs
+++ b/anchor/network/src/network.rs
@@ -5,7 +5,9 @@ use std::sync::Arc;
 use std::time::{Duration, Instant};
 
 use futures::StreamExt;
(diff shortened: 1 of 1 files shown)


diegomrsantos (approved):
LGTM

PR #221, opened 2025-04-01
Change qbft invalid state warnings to debug level

dknopik:
Due to latency, we will from regularly receive messages relevant for an older state, e.g. the fourth PREPARE message when three others have already been received.

This is a normal thing to happen and not worth a warning, so this PR changes the log level to debug.

### DIFF
--- a/anchor/common/qbft/src/lib.rs
+++ b/anchor/common/qbft/src/lib.rs
@@ -397,7 +397,7 @@ where
     ) {
         // Make sure that we are actually waiting for a proposal
         if !matches!(self.state, InstanceState::AwaitingProposal) {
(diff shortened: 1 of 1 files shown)


diegomrsantos:
Does it happen very often? If so, it might be better to use `trace`.

dknopik:
It happened 78 times in the last 30min on interop. I think that is fair enough for debug. It is good to know why the qbft instance discarded a certain message if we end up debugging using some user's logs.

Kind of off topic, but this is IMO the differentiation we should use:

INFO - the user should see this and it won't confuse them too much
DEBUG - we want to see this if the user sends us their log file
TRACE - details, with many additional values, e.g. full messages and so on

Zacholme7 (approved):
lgtm!

PR #222, opened 2025-04-02
logging: span in message receiver, more details around qbft decisions

dknopik:
add some debug logging that helped me figure out some issues. might be also helpful in future


dknopik:
@diegomrsantos 

Ah there is the CI issue that occurs with the cache enabled. Do you want to try troubleshooting it with this PR?

diegomrsantos:
> @diegomrsantos
> 
> Ah there is the CI issue that occurs with the cache enabled. Do you want to try troubleshooting it with this PR?

I'll take a look

jking-aus (approved):

PR #224, opened 2025-04-02
Deserialization fuzzing

Zacholme7:
## Issue Addressed

N/A

## Proposed Changes

Fuzzing for first key entry-point into the program - network deserialization


diegomrsantos:
Do you think we should create an issue for the fuzzing tasks?

PR #225, opened 2025-04-03
Remove unnecessary mutex in validation

dknopik:
As noted [here](https://github.com/sigp/anchor/pull/191#discussion_r2016971752), the mutex is unnecessary. The `DashMap` provides synchronization.


diegomrsantos (approved):
Thank you!

PR #226, opened 2025-04-03
Remove unused NetworkActions

Zacholme7:
## Issue Addressed

N/A

## Proposed Changes

Removed network actions that we never ended up needing


diegomrsantos (approved):
LGTM

dknopik:
Are we not going to need some form of this? We have not implemented exiting a validator if requested on-chain yet

jking-aus (changes requested):
how are we handling the events removed? we still need to be able to evitvals, process liquidations etc. Are these being handled elsewhere?

Zacholme7:
My thought process was that we would have a channel to watch for these other changes or have some approach like we do with new validators. 

The original idea is that we would pipe all of these events somewhere and then have some common handling trait that is implemented to process each one but we didnt end up going that route.

Closing until we figure out for sure, can easily delete later.

PR #228, opened 2025-04-03
Duty logic validation for consensus messages

diegomrsantos:
## Issue Addressed

## Proposed Changes

This PR adds duty validation logic for SSV messages and integrates a new duties management module.

- Introduced a new duties module and duties_tracker for validating sync committee and proposer duties messages.

## Additional Info

This is pending https://github.com/sigp/anchor/issues/277 and will be implemented in a future PR.


anchor/message_validator/src/beacon_network.rs
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    /// Returns the last slot of the given sync committee period
+    pub fn last_slot_of_sync_period(&self, period: u64) -> Slot {
+        let last_epoch = self.first_epoch_of_sync_period(period + 1) - 1;
+        // If we are in the sync committee that ends at slot x we do not generate a message
+        // during slot x-1 as it will never be included, hence -2.
+        self.get_epoch_first_slot(last_epoch + 1) - 2
+    }
+}
```
dknopik: As the functionality of each function is minimal, I am not sure if this wrapper is worth it.

anchor/message_validator/src/beacon_network.rs
```diff
@@ -0,0 +1,132 @@
...
+
+    /// Estimates the current epoch
+    pub fn estimated_current_epoch(&self) -> Epoch {
+        self.estimated_epoch_at_slot(self.estimated_current_slot())
+    }
+
+    /// Estimates the epoch at the given slot
+    pub fn estimated_epoch_at_slot(&self, slot: Slot) -> Epoch {
+        Epoch::new(slot.as_u64() / self.slots_per_epoch)
+    }
```
dknopik: There is no estimation happening in the function, why is this named `estimated`?

anchor/message_validator/src/beacon_network.rs
```diff
@@ -0,0 +1,132 @@
...
+use slot_clock::SlotClock;
+use types::{Epoch, Slot};
+
+/// Wrapper around SlotClock to provide beacon chain network functionality
+#[derive(Clone)]
+pub struct BeaconNetwork<S: SlotClock> {
+    slot_clock: S,
+    slots_per_epoch: u64,
+    epochs_per_sync_committee_period: u64,
+}
```
dknopik: The name is misleading. This does not represent a Beacon Network.

anchor/message_validator/src/beacon_network.rs
```diff
@@ -0,0 +1,132 @@
...
+
+    /// Estimates the epoch at the given slot
+    pub fn estimated_epoch_at_slot(&self, slot: Slot) -> Epoch {
+        Epoch::new(slot.as_u64() / self.slots_per_epoch)
+    }
+
+    /// Returns the first slot at the given epoch
+    pub fn first_slot_at_epoch(&self, epoch: u64) -> Slot {
+        Slot::new(epoch * self.slots_per_epoch)
+    }
```
dknopik: There is `Epoch::start_slot`

anchor/message_validator/src/beacon_network.rs
```diff
@@ -0,0 +1,132 @@
...
+
+    /// Returns the number of epochs per sync committee period
+    pub fn epochs_per_sync_committee_period(&self) -> u64 {
+        self.epochs_per_sync_committee_period
+    }
+
+    /// Estimates the sync committee period at the given epoch
+    pub fn estimated_sync_committee_period_at_epoch(&self, epoch: Epoch) -> Epoch {
+        epoch / self.epochs_per_sync_committee_period
+    }
```
dknopik: There is `Epoch::sync_committee_epoch`

anchor/message_validator/src/beacon_network.rs
```diff
@@ -0,0 +1,132 @@
...
+    pub fn first_epoch_of_sync_period(&self, period: u64) -> u64 {
+        period * self.epochs_per_sync_committee_period
+    }
+
+    /// Returns the last slot of the given sync committee period
+    pub fn last_slot_of_sync_period(&self, period: u64) -> Slot {
+        let last_epoch = self.first_epoch_of_sync_period(period + 1) - 1;
+        // If we are in the sync committee that ends at slot x we do not generate a message
+        // during slot x-1 as it will never be included, hence -2.
+        self.get_epoch_first_slot(last_epoch + 1) - 2
```
dknopik: you subtract 1, just to re-add 1 in the next statement.

Besides that, this seems to be unused.

anchor/message_validator/src/consensus_message.rs
```diff
@@ -355,6 +365,267 @@ fn current_estimated_round(since_slot_start: Duration) -> Round {
...
+                return Err(ValidationFailure::SlotAlreadyAdvanced {
+                    got: consensus_message.height,
+                    want: max_slot.as_u64(),
+                });
+            }
+        }
+    }
+
+    let msg_slot = Slot::new(consensus_message.height);
+    let randao_msg = false; // Default to false as in the Go code
```
dknopik, line 385: What is the purpose of this?

anchor/message_validator/src/duties/duties_tracker.rs
```diff
@@ -0,0 +1,376 @@
...
+        // Prune old duties.
+        self.duties
+            .proposers
+            .write()
+            .retain(|&epoch, _| epoch + HISTORICAL_DUTIES_EPOCHS >= current_epoch);
+
+        Ok(())
+    }
+
+    pub fn start(self: Arc<Self>, executor: TaskExecutor) {
```
diegomrsantos, line 239: @dknopik do we need to handle cancellation or an orderly shutdown hook when starting tasks like that?

dknopik: nope :)

dknopik (commented):
Nice, overall I like it! There are some notes, some might be outdated as I wrote them over some time.

anchor/message_validator/src/duties/duties_tracker.rs
dknopik: no, looks good

anchor/message_validator/src/beacon_network.rs
diegomrsantos: removed this one

anchor/message_validator/src/beacon_network.rs
diegomrsantos: removed

anchor/message_validator/src/beacon_network.rs
diegomrsantos: removed

anchor/message_validator/src/consensus_message.rs
diegomrsantos: Good question https://github.com/ssvlabs/ssv/blob/main/message/validation/consensus_validation.go#L269

anchor/message_validator/src/consensus_message.rs
```diff
@@ -355,11 +351,260 @@ fn current_estimated_round(since_slot_start: Duration) -> Round {
...
+    validate_beacon_duty(
+        validation_context,
+        msg_slot,
+        randao_msg,
+        duty_provider.clone(),
+    )?;
+
+    // Rule: current slot(height) must be between duty's starting slot and:
+    // - duty's starting slot + 34 (committee and aggregation)
+    // - duty's starting slot + 3 (other types)
```
diegomrsantos, line 396: @dknopik, any idea why these numbers?

anchor/message_validator/src/consensus_message.rs
```diff
@@ -355,11 +351,260 @@ fn current_estimated_round(since_slot_start: Duration) -> Round {
...
+            // TODO For voluntary exit, check the stored duties https://github.com/sigp/anchor/issues/277
+            // This would need to be adapted to use the actual duty store
+            Ok((2, true))
+        }
+        Role::Aggregator | Role::ValidatorRegistration => Ok((2, true)),
+        Role::Committee => {
+            let validator_index_count = validator_indices.len() as u64;
+            let slots_per_epoch_val = validation_context.slots_per_epoch;
+
+            // Skip duty search if validators * 2 exceeds slots per epoch
```
diegomrsantos, line 579: @Zacholme7 @dknopik, does this make sense? If the number of validators in the committee is >= slots per epoch, it doesn't mean one of them is in the sync committee. Additionally, do they use the Committee role for the sync committee? What's the difference to SyncCommittee?

anchor/message_validator/src/consensus_message.rs
dknopik: The committee duty is needed for attestation, which can be submitted on chain until the end of the next epoch. So a large window for message validity is a good idea here. Same reasoning for attestation aggregation.

anchor/message_validator/src/consensus_message.rs
```diff
@@ -355,11 +351,260 @@ fn current_estimated_round(since_slot_start: Duration) -> Round {
...
+    Ok(())
+}
+
+/// Determines duty limit based on role and validator indices
+fn duty_limit(
+    validation_context: &ValidationContext<impl SlotClock>,
+    slot: Slot,
+    validator_indices: &[ValidatorIndex],
+    duty_provider: Arc<impl DutiesProvider>,
+) -> Result<(u64, bool), ValidationFailure> {
```
dknopik: Why do we use a `(u64, bool)` instead of a `Option<u64>`?

anchor/message_validator/src/consensus_message.rs
dknopik: The Committee Role is used for `BeaconVote` data, which includes the current head. This is needed for attestations AND sync committee contributions.

The word "Committee" in the "Committee role" refers to the SSV committee, not the sync committee, as this role is executed once per SSV committee, not once per validator.

As for the check:

The limit is at most `slots_per_epoch` because we decide a `BeaconVote` at most once per slot. 
We need to vote for a `BeaconVote` if
- at least one validator is in the sync committee, or
- at least one validator attests in this slot

Furthermore, validators attest once per epoch. So, if we have 32 validators, it is possible that we have to perform the Committee role every slot - regardless of whether one of them is in the sync committee.

That being said, I am not sure why we multiply/divide by two.

anchor/message_validator/src/consensus_message.rs
```diff
@@ -355,11 +355,260 @@ fn current_estimated_round(since_slot_start: Duration) -> Round {
...
+        // No lateness check for these roles
+        Role::ValidatorRegistration | Role::VoluntaryExit => return Ok(Duration::from_secs(0)),
+    };
+
+    let deadline = slot_start_time(slot + ttl, validation_context.slot_clock.clone())
+        .map_err(|_| ValidationFailure::SlotStartTimeNotFound { slot })?
+        .checked_add(LATE_MESSAGE_MARGIN)
+        .unwrap_or_else(|| {
+            SystemTime::now() // Fallback if overflow occurs
+        });
```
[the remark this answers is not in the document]
diegomrsantos: fixed

anchor/message_validator/src/consensus_message.rs
diegomrsantos: fixed

anchor/message_validator/src/consensus_message.rs
diegomrsantos: @nkryuchkov could you please advise here?

anchor/message_validator/src/consensus_message.rs
```diff
@@ -355,11 +351,261 @@ fn current_estimated_round(since_slot_start: Duration) -> Round {
...
+                )?;
+
+                // Check if at least one validator is in the sync committee
+                for &index in validator_indices {
+                    if duty_provider.is_validator_in_sync_committee(period, index) {
+                        return Ok(Some(slots_per_epoch_val));
+                    }
+                }
+            }
+            Ok(Some(std::cmp::min(
```
diegomrsantos, line 594: @@nkryuchkov given the if at line 580, it seems when we get here, `2 * validator_index_count >= slots_per_epoch_val`, so we could simply return `slots_per_epoch_val`?

anchor/message_validator/src/consensus_message.rs
nkryuchkov: @diegomrsantos @dknopik 

https://github.com/ssvlabs/ssv/pull/1892 added some extra logic for randao messages in `validateBeaconDuty` that tolerates missing duties for randao signatures during the first slot of an epoch, while duties are still being fetched from the Beacon node: https://github.com/ssvlabs/ssv/blob/7f928b898e674f479a5ad9042faac78cb10ed92e/message/validation/common_checks.go#L126

Randao is a partial signature message type used in post-consensus messages in proposer role, so it's checked only in partial signature message validation: https://github.com/ssvlabs/ssv/blob/7f928b898e674f479a5ad9042faac78cb10ed92e/message/validation/partial_validation.go#L158

Consensus messages are unrelated to partial signature message type, so it's always false in consensus message validation

anchor/message_receiver/src/manager.rs
```diff
@@ -92,12 +94,12 @@ impl<S: SlotClock + 'static> MessageReceiver for Arc<NetworkMessageReceiver<S>>
                 } = match result {
                     Ok(message) => message,
                     Err(failure) => {
-                        debug!(?failure, "Validation failure");
+                        debug!(gosspisub_message_id = ?message_id, ?failure, "Validation failure");
```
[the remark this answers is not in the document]
diegomrsantos: fixed

anchor/message_validator/src/consensus_message.rs
nkryuchkov: When `validator_index_count < slots_per_epoch_val / 2`, the code inside the `if` returns only if one of the indices satisfies the `duty_provider.is_validator_in_sync_committee(period, index)` condition, so it can also go out of the `if` block and reach the last line. To get rid of `min`, we'd need to add an extra check:
```go
		// Skip duty search if validators * 2 exceeds slots per epoch,
		// as the maximum duties per epoch is capped at the number of slots.
		// This avoids unnecessary checks.
		if validatorIndexCount*2 >= slotsPerEpoch {
			return slotsPerEpoch, true
		}
		// Check if there is at least one validator in the sync committee.
		// If so, the duty limit is equal to the number of slots per epoch.
		period := mv.netCfg.Beacon.EstimatedSyncCommitteePeriodAtEpoch(mv.netCfg.Beacon.EstimatedEpochAtSlot(slot))
		for _, i := range validatorIndices {
			if mv.dutyStore.SyncCommittee.Duty(period, i) != nil {
				return slotsPerEpoch, true
			}
		}

		return 2 * validatorIndexCount, true
```

diegomrsantos: Good catch, I like your solution more. But do we need to skip `duty_provider.is_validator_in_sync_committee(period, index)` tho? Isn't it just a map get?

anchor/message_validator/src/consensus_message.rs
diegomrsantos: Makes sense, thank you!

anchor/message_validator/src/consensus_message.rs
nkryuchkov: @diegomrsantos I'm not sure I understand, what do you mean by skipping? Yes, it's just a map get

diegomrsantos: is the code below necessary?
```go
if validatorIndexCount*2 >= slotsPerEpoch {
	return slotsPerEpoch, true
}
```

nkryuchkov: @diegomrsantos as I understand, yes if we want to get rid of `min`, or we need to return `min(slotsPerEpoch, 2*validatorIndexCount)` otherwise

diegomrsantos: can it be only?
```go
		period := mv.netCfg.Beacon.EstimatedSyncCommitteePeriodAtEpoch(mv.netCfg.Beacon.EstimatedEpochAtSlot(slot))
		for _, i := range validatorIndices {
			if mv.dutyStore.SyncCommittee.Duty(period, i) != nil {
				return slotsPerEpoch, true
			}
		}

		return validatorIndexCount, true
```

diegomrsantos: Why the `*2` in the condition `validator_index_count * 2 >= slots_per_epoch_val`? If we want to skip the check, shouldn't it be `validator_index_count >= slots_per_epoch_val`?

anchor/message_validator/src/consensus_message.rs
```diff
@@ -355,11 +351,261 @@ fn current_estimated_round(since_slot_start: Duration) -> Round {
...
+        // Get current duty count for this signer
+        let epoch = slot.epoch(validation_context.slots_per_epoch);
+        let duty_count = signer_state.get_duty_count(epoch);
+
+        if duty_count >= limit {
+            return Err(ValidationFailure::ExcessiveDutyCount {
+                got: duty_count,
+                limit,
+            });
+        }
```
dknopik: Is `>=` correct? We hit this frequently on interop, and I feel like it should be `>`.

diegomrsantos: Fixed, thanks

dknopik (approved):
great work, LGTM! sorry for the delay

anchor/message_validator/src/consensus_message.rs
nkryuchkov: @diegomrsantos sorry for the delay. The `*2` is because there might be 2 duties per validator. Do I understand correctly that your question is only about the `*2` part?

dknopik: @nkryuchkov I assume you mean sync committee and attestation duty. But isn't the `BeaconData` only voted on once per committee regardless?

nkryuchkov: @dknopik IIRC the main reason was reorgs. @MatheusFranco99 can you please confirm it? If one of validators participates in sync committee, the duty limit changes to `slotsPerEpoch`

MatheusFranco99: @nkryuchkov @dknopik 
Yes, the reason is reorgs. We allow up to 2 attestation duties per epoch, i.e. `2*validatorIndexCount`. For example, a committee with a single validator is allowed to send messages for up to 2 duties.
However, for the case `validatorIndexCount*2 >= slotsPerEpoch`, it's meaningless to set the maximum number of committee duties to more than 32, so we cap it to 32.

The only exception is when one of the committee's validators is in the sync committee period. In this case, the committee can send msgs for up to 32 duties, i.e. all slots.

@diegomrsantos Answering your question, yes, the following code is equivalent:

```go
period := mv.netCfg.Beacon.EstimatedSyncCommitteePeriodAtEpoch(mv.netCfg.Beacon.EstimatedEpochAtSlot(slot))
for _, i := range validatorIndices {
    if mv.dutyStore.SyncCommittee.Duty(period, i) != nil {
        return slotsPerEpoch, true
    }
}
return min(validatorIndexCount*2, slotsPerEpoch), true
```

We added the IF check before `mv.dutyStore.SyncCommittee.Duty` is called because:
- it's very **likely** that `validatorIndexCount*2 >= slotsPerEpoch`, immediately returning.
- and it's very **unlikely** that any `mv.dutyStore.SyncCommittee.Duty` will return true, so the FOR loop complexity is expected to be  `len(validatorIndices)`

PR #232, opened 2025-04-04
Fuzzing via libfuzzer 

Zacholme7:
## Issue Addressed

N/A

## Proposed Changes

Another approach to fuzzing to prompt discussion about which path we should follow. This is the same approach that is taken in lighthouse via beacon-fuzz. The fuzzing here corresponds to #224  & #227

cd in fuzz dir and run `cargo fuzz run {target}`

have not hit any panics yet, good code or bad fuzzing??


anchor/fuzz/fuzz_targets/setup.rs
```diff
@@ -0,0 +1,155 @@
...
+    let config: Config<DefaultLeaderFunction> = ConfigBuilder::new(
+        1.into(),
+        InstanceHeight::default(),
+        (1..=4).map(OperatorId::from).collect(),
+    )
+    .build()
+    .unwrap();
+
+    let data = BeaconVote {
+        block_root: Hash256::random(),
```
gnattishness: Prefer deterministic to random. E.g. hard-coded or generating based on a seed from fuzz input (if you want to mutate the setup)

(Unless you have some funky feature flags so that `Hash256::random()` is no longer random)
Another option looks to be using `random_with()` and a constant seed.

anchor/fuzz/fuzz_targets/setup.rs
```diff
@@ -0,0 +1,155 @@
...
+}
+
+// Sets up a real Validator for fuzzing
+pub fn setup_test_message_validator() -> Arc<Validator<ManualSlotClock>> {
+    let slot_clock = ManualSlotClock::new(
+        Slot::new(0),
+        Duration::from_secs(0),
+        Duration::from_secs(12),
+    );
+    let rsa = Rsa::generate(2048).expect("Keygen will not fail");
```
gnattishness: Randomness. Each setup is different

anchor/fuzz/fuzz_targets/setup.rs
```diff
@@ -0,0 +1,155 @@
...
+
+    let processor_config = processor::Config { max_workers: 2 };
+    let processor_senders = processor::spawn(processor_config, executor);
+
+    let slot_clock = ManualSlotClock::new(
+        types::Slot::new(0),
+        Duration::from_secs(0),
+        Duration::from_secs(12),
+    );
+    let rsa = Rsa::generate(2048).expect("Keygen will not fail");
```
gnattishness: Random. Prefer fixed key, or deterministic

anchor/fuzz/fuzz_targets/setup.rs
```diff
@@ -0,0 +1,155 @@
...
+        source: types::Checkpoint::default(),
+        target: types::Checkpoint::default(),
+    };
+
+    Qbft::new(config, data, MessageId::from([0; 56]), send_message)
+}
+
+// Sets up a real Validator for fuzzing
+pub fn setup_test_message_validator() -> Arc<Validator<ManualSlotClock>> {
+    let slot_clock = ManualSlotClock::new(
```
gnattishness, line 97: Likely want to reset the slot clock (and any other state) between fuzz invocations.
Want each execution to be isolated from the others

gnattishness (changes requested):
Some rough thoughts around improving determinism.

Zacholme7:
Quick thought, do we want fuzzing in the main repo for everyone to have access to? Or make it private in some fuzzing fork

Cargo.toml
```diff
@@ -1,6 +1,33 @@
...
+    "anchor/client",
+    "anchor/common/api_types",
+    "anchor/common/bls_lagrange",
+    "anchor/common/qbft",
+    "anchor/common/ssv_network_config",
+    "anchor/common/ssv_types",
+    "anchor/common/version",
+    "anchor/database",
+    "anchor/eth",
+    "anchor/fuzz",
```
diegomrsantos, line 14: Is there a way to achieve the same without the duplication?

Zacholme7: Did some looking just to double check and this appears to be the most idiomatic solution unless we want to build fuzz alongside the rest of the workspace every time.

dknopik: If nothing depends on fuzz, running `cargo run` should only build the relevant binary, so I do not think the duplication is actually needed. Invocations to Clippy can be done like `cargo clippy -p anchor` to only check the dependency tree of the main binary. Finally, because build results are cached, I don't think it is a huge hassle to let it build once and then it's fine until it is changed.

jking-aus (approved):

PR #237, opened 2025-04-07
Remove validation expect

Zacholme7:
## Issue Addressed

N/

## Proposed Changes

Qbft fuzzing hits the expect. Just propagate option via `?` instead of calling `expect`


anchor/common/qbft/src/lib.rs
```diff
@@ -254,11 +254,7 @@ where
         }
 
         // Message is not a decide message, we know there is only one signer
-        let signer = wrapped_msg
-            .signed_message
-            .operator_ids()
-            .first()
-            .expect("Confirmed to exist");
+        let signer = wrapped_msg.signed_message.operator_ids().first()?;
```
diegomrsantos, line 257: Does it happen when `operator_ids` is empty? If so, should execution get to this point?

Zacholme7: This should never happen hence the expect. The fuzzing is very targeted. 

My thought was that although this should never happen in practice, if the fuzzer can hit it then we should probably just handle it. Behavior is exactly the same. 

Maybe there is also an argument that if this happens then we should panic and stop the node.

jking-aus (approved):

PR #238, opened 2025-04-08
Remove unused deps

diegomrsantos:
Remove unused deps


Zacholme7 (approved):
Weird I thought we had something in CI to catch unused deps.

Lgtm thanks

diegomrsantos:
> Weird I thought we had something in CI to catch unused deps.
> 
> Lgtm thanks

What could we use?

dknopik:
@Zacholme7 We use `cargo udeps`, but it tends to have false negatives. So it is good to manually check from time to time. Thanks @diegomrsantos !

PR #239, opened 2025-04-08
Fix CI issues

Zacholme7:
## Issue Addressed

N/A

## Proposed Changes

Hitting same CI issue with all PRs. Running `cargo update` + settings gossipsub to point to same rev as [lighthouse-anchor](https://github.com/sigp/lighthouse/blob/22faccb4575792ffe815ed02a92ea86f05f9366b/Cargo.toml#L147) fixes the issue. 

Need to change branch because we pull in `sigp-gossipsub` while `lighthouse_network` pulls in the specific `rev` which causes an audit issue.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -94,9 +94,9 @@ checksum = "683d7910e743518b0e34f1186f92494becacb047c7b6bf616c96772180fef923"
 
 [[package]]
 name = "alloy"
-version = "0.12.5"
+version = "0.12.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


Zacholme7:
Is there any reason we need to be on latest `sigp-gossipsub`?

diegomrsantos:
The previous PRs (https://github.com/sigp/anchor/pull/238) also passed, but when we update the current open ones, they fail.

dknopik:
While it is good to pin the `gossipsub` version - what CI issue is this supposed to fix? This does not fix the caching issue.

Zacholme7:
@dknopik was a bandaid over the caching issue. Closing

PR #240, opened 2025-04-09
Disable cache

diegomrsantos:
## Issue Addressed

CI build fails due to missing LH genesis file https://github.com/sigp/anchor/actions/runs/14338027416/job/40190094805?pr=237

## Proposed Changes

Disable cache for now while we find a solution.


dknopik:
Please add a link to https://github.com/sigp/anchor/issues/123 as comment above the `cache: false` line to prevent someone from re-enabling it in the future.

dknopik (approved):
LGTM, thank you!

PR #243, opened 2025-04-10
Bump deps

dknopik:
There is a vulnerability in a dependency: https://rustsec.org/advisories/RUSTSEC-2025-0024

So we bump all the dependencies. I tried updating the dependencies that are non-semver compatible, but no are non-conflicting right now. 

Furthermore, we pin Gossipsub to the same version as used in Lighthouse to avoid the duplicated dependency.


Zacholme7 (approved):
lgtm

Issue #39, opened 2024-10-23
Spec tests


jking-aus:
Implement Spec tests from SSV repo located [here](https://github.com/ssvlabs/ssv-spec/tree/main/ssv/spectest)

PR #247, opened 2025-04-11
Spec tests Framework

Zacholme7:
## Issue Addressed

#39 

## Proposed Changes

Introductory rough draft to spec tests. This is extremely WIP and any feedback is welcome.

I have attempted to setup a general structure that provides a clean interface for parsing, setting up, and running the spec tests. A key observation here is that spec tests are divided into categories (can discern which category based on .json file name) and each category has a different json structure. We need to be able to read these into an internal format, setup the test, and then run the test. 

Test are identified via the `SpecTestType` enum, which is further subdivided into specifics areas where each inner enum has a variant for each test category. 

The core behind this is the `SpecTest` trait. Each different test category must implement this trait on its respective struct. This provides functionality for setting up the test, running the test, and identifying which test category it is. 

All test categories must register itself with the `register_test_loaders` macro. This allows for a very clean way of parsing and running all of the test categories in a generic way. Without this, we would have to match against all variants and optimistically decode each one in succession. 

The whole flow is roughly the following
1) call `run_tests` with the variant of test that you would like to run
2) Walk through the test directory and read in/parse all of the file
3) Perform any setup, if necessary
4) Run all all of the tests 

We are still missing implementations for the type spec and the ssv spec. It is pretty easy to create the scaffolding for these by following the qbft example. 

There is still a lot of work to be done for setting up the tests and mapping the types into our `ssv_types` representation, but I think this provides a very extensible and clean framework for implmenting all of the spec tests.


anchor/spec_tests/src/lib.rs
```diff
@@ -0,0 +1,124 @@
...
+}
+
+// Impl display for path construction. Do not change
+impl fmt::Display for SpecTestType {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            SpecTestType::Qbft(_) => write!(f, "src/ssv-spec/qbft/spectest/generate/tests"),
+        }
+    }
+}
```
dknopik: why a display impl instead of a `fn path(&self) -> String/&'static str`?

Zacholme7: no reason, fixed.

anchor/spec_tests/src/lib.rs
```diff
@@ -0,0 +1,121 @@
...
+#[derive(Eq, PartialEq, Hash)]
+enum SpecTestType {
+    Qbft(QbftSpecTestType),
+}
+
+// Impl display for path construction. Do not change
+impl SpecTestType {
+    fn path(&self) -> &'static Path {
+        match self {
+            SpecTestType::Qbft(_) => Path::new("src/ssv-spec/qbft/spectest/generate/tests"),
```
diegomrsantos, line 22: Could this be a property in the variant?

anchor/spec_tests/src/lib.rs
```diff
@@ -0,0 +1,121 @@
...
+trait SpecTest {
+    // Retrieve the name of the test
+    fn name(&self) -> &str;
+
+    // Setup a runner for the test. This will configure and construct eveything required to
+    // execute the test
+    fn setup(&mut self);
+
+    // Run the test and verify that the output is what we were expecting.
+    fn run(&self) -> bool;
```
diegomrsantos, line 39: Maybe we could split it into `run` and `assert`

anchor/spec_tests/src/lib.rs
Zacholme7: Not sure what you mean, do you mean on the trait? or the inner enum variant?

anchor/spec_tests/src/lib.rs
Zacholme7: Yea I want to head in this direction but im holding off for now because I am not sure how the structure of this would work. Need to dig deeper into the other tests first. 

Might be a case of setting up tests and then refactoring to an assertion func too

anchor/spec_tests/src/lib.rs
diegomrsantos: In the enum variant

Zacholme7:
Closing for now

PR #271, opened 2025-04-16
Trim whitespace from testnet directory files

dknopik:
It is easy to accidentally put white space (e.g. a trailing newline) into testnet directory files. This may cause errors when parsing the contained values, so let's trim all trailing or leading white space.


Zacholme7 (approved):
This was bugging me but I forgot about fixing it. Thanks!

PR #274, opened 2025-04-16
Fix deadlock in message validator

dknopik:
I noticed that sometimes, during a burst of incoming messages, Anchor would deadlock. More specifically, threads holding a `processor` permit would deadlock and the processor queue would become full shortly after.

The issue was doubly locking the network state. As `RwLock` (and thus `tokio::sync::watch`) avoids writer starvation by not allowing new readers while a writer is waiting, so trying to (read) borrow the lock while 
a) holding a read lock already, and
b) a writer is waiting
is a deadlock, as we are now waiting for the writer, while the writer is waiting for us to release the already held read lock.

This PR fixes the issue by using the existing borrowed value instead of borrowing again. Furthermore, we explicitly drop the read guard to release it as early as possible.

### DIFF
--- a/anchor/message_validator/src/lib.rs
+++ b/anchor/message_validator/src/lib.rs
@@ -248,8 +248,9 @@ impl<S: SlotClock> Validator<S> {
                             .ok_or(ValidationFailure::UnknownValidator)?
                     }
                 };
-
-                let operators_pks = self.get_operator_pks(signed_ssv_message.operator_ids())?;
+                let operators_pks =
+                    get_operator_pks(&network_state, signed_ssv_message.operator_ids())?;
+                drop(network_state);
 
                 let mut consensus_state =
                     self.get_consensus_state(ssv_message.msg_id(), self.slots_per_epoch);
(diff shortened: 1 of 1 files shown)


anchor/message_validator/src/lib.rs
```diff
@@ -248,8 +248,9 @@ impl<S: SlotClock> Validator<S> {
                             .ok_or(ValidationFailure::UnknownValidator)?
                     }
                 };
-
-                let operators_pks = self.get_operator_pks(signed_ssv_message.operator_ids())?;
+                let operators_pks =
+                    get_operator_pks(&network_state, signed_ssv_message.operator_ids())?;
+                drop(network_state);
```
diegomrsantos, line 253: That's a good catch, thanks. Is there a way we can avoid repeating this in the future? It still seems error-prone.

dknopik: Well, we have`lockbud` in CI, which is supposed to catch stuff like this, but I guess it has its limits. 

Otherwise we just have to stay extra vigilant during review, and adopt patterns like either dropping locks ASAP, or even better, acquiring them in a limited scope such as a function.

jking-aus (approved):

PR #279, opened 2025-04-22
Rust Edition 2024

dknopik:
Switch to Rust Edition 2024 and update dependencies.

In a nutshell,
- the formatter seems to want to do things a bit differently,
- `gen` is now a keyword, so I renamed a module, and
- `set_var` is now `unsafe`, so I added a comment explaining why we are fine.


Zacholme7 (approved):
lgtm

Issue #264, opened 2025-04-14
Stop handling liquidated clusters and update fee recipient in validator store


dknopik:
Currently, we only check if any validator has been added or removed when polling from the database. We do not properly check for clusters that have been liquidated or for updated fee recipients.

PR #280, opened 2025-04-22
Handle liquidated/reactivated clusters and update fee recipient

Zacholme7:
## Issue Addressed

#264 

## Proposed Changes

Handles the addition/removal of validators based on cluster liquidation status and updated fee recipients.

For liquidations,we can filter based on the liquidation status. If the cluster is now liquidated, we will not mark those validators as seen and they will be removed. If a cluster has been reactivated, they will pass the filter and get added back in.


anchor/validator_store/src/lib.rs
```diff
@@ -165,8 +166,38 @@ impl<T: SlotClock, E: EthSpec> AnchorValidatorStore<T, E> {
...
+                        // recipient
+
+                        // Create a new cluster with updated fee recipient
+                        let new_cluster = Arc::new(Cluster {
+                            cluster_id: old_cluster.cluster_id,
+                            owner: old_cluster.owner,
+                            fee_recipient: cluster.fee_recipient,
+                            liquidated: old_cluster.liquidated,
+                            cluster_members: old_cluster.cluster_members.clone(),
+                        });
```
dknopik: what do you think about keeping this general and just comparing whether the old cluster and the new cluster are the same? That way we also catch e.g. someone rapidly switching over a validator to another cluster with the same fee recipient.

Zacholme7: Yep, cleaner too. fixed.

anchor/common/ssv_types/src/cluster.rs
```diff
@@ -21,7 +21,7 @@ impl Debug for ClusterId {
 ///
 /// Each cluster is owned by a unqiue EOA and only that Address may perform operators on the
 /// Cluster.
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq)]
```
dknopik: Nitpick: Also derive `Eq`

dknopik (commented):
nitpick, otherwise LGTM!

dknopik:
taking this for a drive on interop along with some other stuff, will merge after

dknopik (approved):
LGTM

PR #281, opened 2025-04-22
setup mev-boost support

Zacholme7:
## Issue Addressed

#261 

## Proposed Changes

Just have to plumb MEV options into the validator store

### DIFF
--- a/anchor/client/src/cli.rs
+++ b/anchor/client/src/cli.rs
@@ -474,7 +474,7 @@ pub struct Node {
     #[clap(
         long,
         alias = "private-tx-proposals",
-        help = "If this flag is set, Lighthouse will query the Beacon Node for only block \
+        help = "If this flag is set, Anchor will query the Beacon Node for only block \
                 headers during proposals and will sign over headers. Useful for outsourcing \
                 execution payload construction during proposals.",
         display_order = 0,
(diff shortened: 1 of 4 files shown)


anchor/client/src/cli.rs
```diff
@@ -470,6 +470,47 @@ pub struct Node {
         display_order = 0
     )]
     pub work_queue_size: Vec<String>,
+
+    #[clap(
+        long,
+        alias = "private-tx-proposals",
+        help = "If this flag is set, Lighthouse will query the Beacon Node for only block \
```
dknopik: Anchor instead of Lighthouse

anchor/client/src/cli.rs
```diff
@@ -470,6 +470,47 @@ pub struct Node {
...
+                when choosing between a builder payload header and payload from \
+                the local execution node.",
+        conflicts_with = "prefer_builder_proposals",
+        display_order = 0
+    )]
+    pub builder_boost_factor: Option<u64>,
+
+    #[clap(
+        long,
+        help = "If this flag is set, Lighthouse will always prefer blocks \
```
dknopik: Anchor instead of Lighthouse

anchor/client/src/cli.rs
```diff
@@ -470,6 +470,47 @@ pub struct Node {
...
+    pub prefer_builder_proposals: bool,
+
+    #[clap(
+        long,
+        value_name = "UNIX-TIMESTAMP",
+        help = "This flag takes a unix timestamp value that will be used to override the \
+                timestamp used in the builder api registration.",
+        display_order = 0
+    )]
+    pub builder_registration_timestamp_override: Option<u64>,
```
dknopik: I don't think we can offer this option.

We need to sign with the start of the epoch as timestamp, as that is what go-ssv does. Deviating from that would cause us to fail to sign.

dknopik (commented):
nice!

dknopik (approved):
LGTM

PR #282, opened 2025-04-23
Health endpoint

ThreeHrSleep:
## Issue Addressed

part of https://github.com/sigp/anchor/issues/26


dknopik:
You seem to have updated the dependencies in this PR with `cargo update` or by deleting the `Cargo.lock` file, as the `Cargo.lock` file is updated with many version changes. I assume this was by accident. Can you please revert that, and make sure only adding the dependencies to the `http_api` is reflected there?

ThreeHrSleep:
ahh,,,sorry, my bad
cleaned it..I think it should be better now

dknopik (approved):
Thanks, LGTM!

PR #283, opened 2025-04-23
Enable aarch64 builds

dknopik:
Enable `aarch64` builds.

Long running branches that require quicker builds (such as `interop1`) can simply remove this again, as GitHub will pick the workflow from the branch.


antondlr (approved):
lgtm!

PR #284, opened 2025-04-23
Fix flipped condition for committee relevancy check

dknopik:
I introduced a bug when fixing a TODO in #265. 

When checking whether a committee message is relevant for us, we correctly check whether we are a member of it, but then discard the message if so, instead of if not. This PR flips the condition to fix this.


Zacholme7 (approved):
lgtm

Issue #252, opened 2025-04-14
Extend Syncer to actually handle multiple URLs


dknopik:
While support for multiple Beacon Nodes is implemented via `FallbackBeaconNodes`, there is currently no support for multiple EL RPC and WS endpoints - we just take the first supplied via the CLI for each.

We should implement some fallback logic that switches over to another URL if we encounter faults.

It is unclear how complex this needs to be (do we need some kind of scoring regarding node quality, or is it sufficient to just fail over to the next one on failure), and whether there is some existing mechanism for this available, e.g. in Alloy.

PR #285, opened 2025-04-23
Node fallback endpoints

Zacholme7:
## Issue Addressed

#252

## Proposed Changes

Support fallback url for rpc provider. 

Unfortunately there is no easy way to also support this for websockets. Another potential option is to use [polling](https://docs.rs/alloy/latest/alloy/providers/trait.Provider.html#method.watch_blocks) instead of websocket connections to provider fallbacks.


dknopik:
As we will not accept fallback WS URLs for now, we should adjust the CLI to only accept one WS URL.

dknopik (approved):
LGTM, ty!

Issue #268, opened 2025-04-15
Report accurate Node Metadata


dknopik:
Currently, we initialize the Node Metadata used in the handshake as follows: 

```rust
            Some(NodeMetadata {
                node_version: "1.0.0".to_string(),
                execution_node: "geth/v1.10.8".to_string(),
                consensus_node: "lighthouse/v1.5.0".to_string(),
                subnets: "00000000000000000000000000000000".to_string(),
            }),
```

Only the `subnets` field is adjusted dynamically.

IMO, we should at least correctly fill the `node_version`, as that might be useful for troubleshooting. 

`execution_node` and `consensus_node` are less important - especially considering that the handshake mechanism can be considered deprecated and due for replacement or removal via an SIP. So I'd be fine with dummy values for those. Looking forward to your opinion.

diegomrsantos:
> the handshake mechanism can be considered deprecated

Why is it?

dknopik:
It has flaws in its design, especially with regard to data encoding. I think it is bound to be replaced with some SIP.

PR #286, opened 2025-04-23
Node version in metadata

Zacholme7:
## Issue Addressed

#268

## Proposed Changes

Put the node version into the metadata. Will have to update version prefix alongside releases.

Question: Do we also want platform? Or just version


dknopik:
> Question: Do we also want platform? Or just version

Personally, I do not care. Go-SSV does not report platform (but they also do not report that they are Go-SSV)...

dknopik (approved):
LGTM, but feel free to wait for the other's feedback if you want

diegomrsantos:
Adding platform is fine imo

PR #288, opened 2025-04-24
Fix index uniqueness for cluster owners

dknopik:
Owners can own multiple clusters. Accurately reflect this by in our `MultiIndexMap` by setting the `NonUniqueTag`.


Zacholme7 (approved):
Ah yea this makes sense. lgtm

Issue #289, opened 2025-04-24
Track fee recipient accurately


dknopik:
Currently, we initialise each cluster's fee recipient to the owner, and update all clusters owned by the owner if they issue a "setFeeRecipientAddress" transaction.

This is not correct, as future clusters created by the owner will have the owner address as fee recipient, even though it should be the last address set via "setFeeRecipientAddress".

We should adjust our database to accurately track the fee recipient address for each owner address. This likely involves a new table, `owners`, with two columns, `owner_address` and `fee_recipient_address`.

PR #291, opened 2025-04-24
Correctly set fee recipient

Zacholme7:
## Issue Addressed

#289

## Proposed Changes

Makes it so that when a fee recipient address is updated, we store that mapping and reflect it in all current and future clusters.


dknopik:
Thanks! 

While I think it is nice to still store the recipient address with the cluster for the in-memory representation, I feel like we should remove the `fee_recipient` column in the `clusters` table, as it will always be the same as the `fee_recipient` in the `owners` table. What do you think?

Zacholme7:
@dknopik agreed. Fixed

dknopik (approved):
LGTM, thanks!

PR #298, opened 2025-04-28
Networking metrics

### LINKED ISSUE #256: Networking metrics
Add metrics about networking.

An initial list of interesting metrics to implement might be https://github.com/sigp/lighthouse/blob/e98209d1186419264868a28591c2b14af3f6abff/beacon_node/lighthouse_network/src/metrics.rs

ThreeHrSleep:
## Issue Addressed

https://github.com/sigp/anchor/issues/256


anchor/network/src/peer_manager.rs
```diff
@@ -409,15 +405,24 @@ impl NetworkBehaviour for PeerManager {
...
         }
+        if updated_connected {
+            lighthouse_network::metrics::set_gauge(
+                &lighthouse_network::metrics::PEERS_CONNECTED,
+                self.connected.len().try_into().unwrap_or(0),
+            );
+        }
         self.connection_limits.on_swarm_event(event);
         self.peer_store.on_swarm_event(event);
     }
```
diegomrsantos: How about something like

```rust
fn on_swarm_event(&mut self, event: FromSwarm) {
    use FromSwarm::*;

    // `changed` is `true` only when the set actually grew or shrank.
    let changed = match event {
        ConnectionEstablished(ConnectionEstablished { peer_id, .. }) =>
            self.connected.insert(peer_id),
        ConnectionClosed(ConnectionClosed { peer_id, .. }) =>
            self.connected.remove(&peer_id),
        _ => false,
    };

    if changed {
            lighthouse_network::metrics::set_gauge(
                &lighthouse_network::metrics::PEERS_CONNECTED,
                self.connected.len().try_into().unwrap_or(0),
            );
    }

    // forward the same `FromSwarm` to the other helpers
    self.connection_limits.on_swarm_event(event);
    self.peer_store.on_swarm_event(event);
}
```

dknopik: aaah, good catch, thx!

dknopik (approved):
LGTM, thank you Sayan!

Issue #259, opened 2025-04-14
Support chain-supplied validator exits


dknopik:
Validators can not only removed from the cluster, but the cluster can also be ordered to sign a voluntary exit and publish it. This needs to be implemented by having the event processor trigger this process, for example by spawning a task or via a queue.

PR #300, opened 2025-04-30
Support chain-supplied validator voluntary exits

diegomrsantos:
## Issue Addressed

https://github.com/sigp/anchor/issues/259

## Proposed Changes

This PR adds support for chain-supplied voluntary exits by introducing dedicated tracking and processing components for validator exit requests. Key changes include the creation of a voluntary exit tracker and processor, integration of signature collection within the validator store, and the subsequent wiring of these components into the client and event sync flows.

## Additional Info

In the next PR, I'll use the tracker for duty validation.


anchor/eth/src/event_processor.rs
```diff
@@ -516,29 +521,247 @@ impl EventProcessor {
...
+                // If the channel is closed, we can't send the exit request
+                // This is a fatal error and should be handled by the caller
+                error!(
+                    validator_pubkey = %validator_pubkey,
+                    ?err,
+                    "Failed to send validator exit request to processor"
+                );
+                return Err(ExecutionError::InvalidEvent(
+                    "Failed to send validator exit request to processor".to_string(),
+                ));
```
dknopik: `InvalidEvent` is not really accurate here

anchor/eth/src/event_processor.rs
```diff
@@ -516,29 +521,247 @@ impl EventProcessor {
...
+                    "No committee info found for validator".to_string(),
+                ));
+            }
+        };
+        drop(state);
+
+        // Check if our operator is part of this validator's committee
+        let is_our_validator = committee_info.committee_members.contains(&own_operator_id);
+        Ok(Some(is_our_validator))
+    }
```
dknopik: We store shares in the multi index map only if they are relevant for the local operator, so you can just check if that contains an entry for the pubkey

anchor/eth/src/event_processor.rs
```diff
@@ -516,29 +521,247 @@ impl EventProcessor {
...
+        }
+
+        if cluster.liquidated {
+            warn!(
+                validator_pubkey = %validator_pubkey,
+                computed_cluster_id = ?computed_cluster_id,
+                "Cluster is already liquidated, skipping exit processing"
+            );
+            return Ok(());
+        }
```
dknopik: The word "already" is confusing me. Also, why is `Ok(())` returned?

anchor/eth/src/event_processor.rs
```diff
@@ -516,29 +521,247 @@ impl EventProcessor {
...
+            error!(
+                validator_pubkey = %validator_pubkey,
+                registered_owner = ?cluster.owner,
+                contract_event_owner = ?owner,
+                "Owner mismatch: the address in the contract event is not the validator's registered owner"
+            );
+            return Err(ExecutionError::InvalidEvent(
+                "Contract event owner does not match the validator's registered owner".to_string(),
+            ));
+        }
```
dknopik, line 752: This is enforced by the contract.

anchor/validator_store/src/lib.rs
```diff
@@ -557,6 +554,35 @@ impl<T: SlotClock, E: EthSpec> AnchorValidatorStore<T, E> {
             }
         }
     }
+
+    pub async fn collect_voluntary_exit_signatures(
+        &self,
+        validator_pubkey: PublicKeyBytes,
+        voluntary_exit: VoluntaryExit,
+        slot: Slot,
+    ) -> Result<SignedVoluntaryExit, Error> {
```
dknopik: `collect_voluntary_exit_signature` in singular, right?

anchor/eth/src/voluntary_exit_processor.rs
```diff
@@ -0,0 +1,141 @@
...
+        block_timestamp,
+    } = request;
+
+    info!(
+        validator_pubkey = %validator_pubkey,
+        validator_index = ?validator_index,
+        "Processing voluntary exit request"
+    );
+
+    let block_time = Duration::from_millis(block_timestamp);
```
dknopik: I think the block timestamp is in seconds.

dknopik (commented):
Nice, looks good overall! Some first comments below, will test it out tomorrow.

anchor/eth/src/event_processor.rs
diegomrsantos: I believe we should also do our validation.

anchor/validator_store/src/lib.rs
diegomrsantos: What do you mean, aren't we collecting signatures?

anchor/eth/src/voluntary_exit_processor.rs
diegomrsantos: You might be right, good catch, thanks!

anchor/eth/src/event_processor.rs
diegomrsantos: No particular reason, what would you recommend?

anchor/eth/src/event_processor.rs
diegomrsantos: But also risky to make validation based on implicit assumptions that might change and we could not notice.

anchor/eth/src/event_processor.rs
diegomrsantos: True, how about `Misc`?

anchor/client/src/lib.rs
```diff
@@ -346,11 +349,17 @@ impl Client {
         let index_sync_tx =
             start_validator_index_syncer(beacon_nodes.clone(), database.clone(), executor.clone());
 
+        // We create the channel here so that we can pass the receiver to the syncer. But we need to
+        // delay starting the voluntary exit processor until we have created the validator store.
+        let (exit_tx, exit_rx) = unbounded_channel();
```
diegomrsantos, line 358: @dknopik maybe we should bound this?

anchor/voluntary_exit/src/voluntary_exit_processor.rs
```diff
@@ -0,0 +1,298 @@
...
+                    }
+                })
+                .await
+            {
+                Ok(_) => {
+                    info!(
+                        validator_pubkey = %validator_pubkey,
+                        "Successfully submitted voluntary exit to beacon node"
+                    );
+                    // metrics::inc_counter_vec(
```
diegomrsantos, line 233: I'm not sure what's the best way to handle this

anchor/eth/src/event_processor.rs
```diff
@@ -558,14 +568,45 @@ impl EventProcessor {
...
             Some(ts) => ts,
             None => {
                 debug!("Block timestamp not available");
-                return Err(ExecutionError::InvalidEvent(
-                    "Block timestamp not available".to_string(),
-                ));
+
+                // Get the block_number for epoch calculation
+                let block_number = match log.block_number {
+                    Some(ts) => ts,
```
dknopik: `Some(block_number) => block_number` -  or even better, use `ok_or_else` and `?`

anchor/voluntary_exit/src/voluntary_exit_processor.rs
diegomrsantos: This has been fixed in the next PR

anchor/client/src/lib.rs
dknopik: You know my liberal policy about `unbounded_channel` that have a natural boundary due to on-chain limitations ;D

Still, if you want you can estimate how many exits can be burst in a single block, add some safety margin on top and apply that as limit.

anchor/eth/src/event_processor.rs
```diff
@@ -32,72 +39,83 @@ pub enum Mode {
...
+                }
+
+                hash if *hash == SSVContract::ValidatorExited::SIGNATURE_HASH => {
+                    self.process_validator_exited(log).await
+                }
+                _ => {
+                    debug!(?topic0, "Unknown event signature, skipping");
+                    continue;
+                }
+            };
```
dknopik:
```rust
match *topic0 {
    SSVContract::OperatorAdded::SIGNATURE_HASH => {
    ....
```
shouuuld work, right?

anchor/eth/src/event_processor.rs
```diff
@@ -32,72 +39,83 @@ pub enum Mode {
...
+            rpc_client,
+        }
     }
 
     /// Process a new set of logs
     #[instrument(skip(self, logs), fields(logs_count = logs.len()))]
-    pub fn process_logs(&self, logs: Vec<Log>, live: bool) {
+    pub async fn process_logs(&self, logs: Vec<Log>, live: bool) {
         info!(logs_count = logs.len(), "Starting log processing");
         let timer = metrics::start_timer(&metrics::EXECUTION_LOG_PROCESSING_TIME);
```
dknopik, line 64: GH does not let me comment it directly, but at the bottom of this fn is the `NetworkAction` machinery, which I think is now unneeded and unplanned as the approach in this PR is better. IMO think its fine to remove it now?

anchor/eth/src/event_processor.rs
dknopik: "already" implies that liquidation is a result of exit processing, which it is not.

returning Ok here does not skip the exit processing as claimed by the log message

anchor/validator_store/src/lib.rs
dknopik: we collect partial singatures to create a single voluntary exit signature, so I guess it depends on what words you remove from that sentence ;D

anchor/voluntary_exit/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "voluntary_exit"
+version = "0.1.0"
+edition = "2024"
```
dknopik: use workspace edition

anchor/voluntary_exit/src/voluntary_exit_processor.rs
```diff
@@ -0,0 +1,298 @@
...
+                sleep(Duration::from_secs(1)).await;
+                continue;
+            }
+        };
+
+        // Don't process if we are still in the same slot
+        if last_processed_slot == Some(current_slot) {
+            sleep(Duration::from_secs(1)).await;
+            continue;
+        }
```
dknopik: use `duration_to_next_slot` instead

anchor/voluntary_exit/src/voluntary_exit_processor.rs
```diff
@@ -0,0 +1,298 @@
...
+        .unwrap_or_default();
+
+    let next_slot_start = slot_clock
+        .start_of(current_slot + 1)
+        .unwrap_or_else(|| now + slot_duration); // Fallback to now + slot_duration if start_of fails
+
+    // Sleep until just before the next slot (50ms before)
+    let time_until_next_slot = next_slot_start
+        .saturating_sub(now)
+        .saturating_sub(Duration::from_millis(50));
```
dknopik: why?

anchor/voluntary_exit/src/voluntary_exit_processor.rs
```diff
@@ -0,0 +1,298 @@
...
+                    info!(
+                        validator_index = ?exit_duty.validator_index,
+                        "Successfully processed and removed voluntary exit"
+                    );
+                } else {
+                    // Exit remains in tracker for retry
+                    debug!(
+                        validator_index = ?exit_duty.validator_index,
+                        "Failed to process voluntary exit, will retry later"
+                    );
```
dknopik: I do not think a retry is appropriate, I found no retry mechanism in Go-ssv, but maybe I missed it

anchor/eth/src/event_processor.rs
dknopik: It's not implicit: https://github.com/sigp/anchor/blob/b724c9829b18e07db3710797c52b19d88e5e88e4/anchor/database/src/lib.rs#L44-L47

diegomrsantos: It's implicit because it's an implementation choice. If we change this map to have all shares and forget about this dependency, a bug is introduced. Our operator being part of the committee for a validator is a domain property.

anchor/eth/src/event_processor.rs
diegomrsantos: True, good catch. How about?
```rust
        if cluster.liquidated {
            warn!(
                validator_pubkey = %validator_pubkey,
                computed_cluster_id = ?computed_cluster_id,
                "Cluster is liquidated, skipping exit processing"
            );
            return Err(ExecutionError::Misc(
                "Cluster is liquidated, skipping exit processing".to_string(),
            ));
        }
```

anchor/validator_store/src/lib.rs
diegomrsantos: True. How about `collect_voluntary_exit_partial_signatures`?

anchor/eth/src/event_processor.rs
diegomrsantos: done

anchor/voluntary_exit/Cargo.toml
diegomrsantos: done

anchor/voluntary_exit/src/voluntary_exit_processor.rs
```diff
@@ -0,0 +1,298 @@
...
+        exit_tracker.prune(current_slot, slots_per_epoch);
+
+        let sleep_duration = calculate_sleep_duration(slot_clock.clone(), current_slot);
+
+        sleep(sleep_duration).await;
+        continue;
+    }
+}
+
+fn calculate_sleep_duration(slot_clock: impl SlotClock, current_slot: Slot) -> Duration {
```
diegomrsantos: I didn't know `duration_to_next_slot`. Maybe I could use it instead of this function?

anchor/voluntary_exit/src/voluntary_exit_processor.rs
diegomrsantos: What's the issue with the retry?

anchor/voluntary_exit/src/voluntary_exit_processor.rs
diegomrsantos: This is just a check in case we woke up still in the same slot, and it shouldn't happen. So we sleep a bit more, and hopefully it'll be the next one.

anchor/eth/src/event_processor.rs
```diff
@@ -32,72 +39,83 @@ pub enum Mode {
...
+            let topic0 = match log.topic0() {
+                Some(topic) => topic,
+                None => {
+                    warn!("Log missing topic0, skipping");
+                    continue;
+                }
+            };
+
+            // Use a match statement instead of a HashMap lookup
+            let result = match *topic0 {
```
Zacholme7, line 79: Any reason to move from the lookup to the match? I just did the lookup because I just tend to prefer it over verbose matching when there is conditional dispatching. Just personal style preference though. 

Fine to stick with the match, but the comment should probably be adjusted to something like `Process log based on signature hash`

anchor/eth/src/event_processor.rs
```diff
@@ -525,20 +543,269 @@ impl EventProcessor {
 
     // A validator has exited the beacon chain
     #[instrument(skip(self, log), fields(validator_pubkey, owner))]
-    fn process_validator_exited(&self, log: &Log) -> Result<(), ExecutionError> {
+    async fn process_validator_exited(&self, log: &Log) -> Result<(), ExecutionError> {
+        let exit_tx = match &self.mode {
```
Zacholme7: Bit more idiomatic imo
`
let Mode::Node { exit_tx, .. } = &self.mode else {
    return Ok(());
};
`

anchor/eth/src/event_processor.rs
```diff
@@ -525,20 +543,269 @@ impl EventProcessor {
...
+                };
+
+                // Calculate the slot at which to process this exit
+                block.header.timestamp
+            }
+        };
+
+        let validator_index = match self.get_validator_index(&validator_pubkey) {
+            Ok(Some(value)) => value,
+            Ok(None) => return Ok(()),
```
Zacholme7, line 592: Should `Ok(None)` also be considered an error here since the validator should exist?

Zacholme7 (commented):
Small nits, really good job here!

anchor/eth/src/event_processor.rs
diegomrsantos: Nice, `ok_or_else` is much better!

anchor/eth/src/event_processor.rs
```diff
@@ -32,72 +39,83 @@ pub enum Mode {
...
+                }
+                _ => {
+                    debug!(?topic0, "Unknown event signature, skipping");
+                    continue;
+                }
+            };
 
-            if let Err(e) = handler(self, log) {
+            // Handle any errors from the event processing
+            if let Err(e) = result {
```
diegomrsantos, line 118: @Zacholme7 @dknopik That's what I mentioned some time ago, we are logging twice.

diegomrsantos: I created https://github.com/sigp/anchor/issues/317

anchor/eth/src/event_processor.rs
diegomrsantos: The `process_logs` and `process_validator_exited` functions are async now. Could the previous code handle it?

anchor/eth/src/event_processor.rs
diegomrsantos: done

anchor/eth/src/event_processor.rs
dknopik: > If we change this map to have all shares and forget about this dependency, a bug is introduced.

but this argument can be applied to all code ever written. if one decides to change the behaviour of something without considering other parts of the code dependent on it, it breaks. 

If you want, we can rename the accessor to the map to `own_shares` to make it more visible.

anchor/validator_store/src/lib.rs
dknopik: sounds good!

anchor/eth/src/event_processor.rs
diegomrsantos: I think we can have a validator without an index for a while. @dknopik

anchor/voluntary_exit/src/voluntary_exit_processor.rs
dknopik: Still, if you need to sleep again, use `duration_to_next_slot` again. 1 sec of sleep is a long time, especially since there is ongoing research in shortening Ethereums slot times.

anchor/eth/src/event_processor.rs
diegomrsantos: I don't think it's true. The fact that our operator is part of the committee means it's "our" validator isn't an implementation detail; it's a property of the domain, and we can't just change that.

anchor/voluntary_exit/src/voluntary_exit_processor.rs
dknopik: We need to follow the spec (= the behaviour of go-ssv). If that mandates no retry, then we do not retry. In a mixed committee, if we retry but go-ssv does not, we send additional messages that will never result in a successfully reconstructed exit.

anchor/voluntary_exit/src/voluntary_exit_processor.rs
dknopik: sounds good!

anchor/voluntary_exit/src/voluntary_exit_processor.rs
diegomrsantos: True. I'll double-check if they do that, and if not, I'll remove it.

anchor/eth/src/event_processor.rs
Zacholme7: ah missed that, handler is `fn` so no.

anchor/voluntary_exit/src/voluntary_exit_processor.rs
diegomrsantos: Removed. Good catch, thanks.

anchor/voluntary_exit/src/voluntary_exit_processor.rs
diegomrsantos: 6d792c4eaacd1c42742f5303c92cd088456316b8

anchor/voluntary_exit/src/voluntary_exit_processor.rs
diegomrsantos: Not relevant anymore

anchor/voluntary_exit/src/voluntary_exit_processor.rs
diegomrsantos: I removed it, not sure if this is an issue.

anchor/eth/src/event_processor.rs
diegomrsantos: you mean?
```rust
            // If live is true, then we are currently in a live sync and want to take some action in
            // response to the log. Parse the log into a network action and send to be processed;
            if live {
                let action = log.try_into().unwrap_or_else(|e| {
                    error!("Failed to convert log into NetworkAction {e}");
                    NetworkAction::NoOp
                });
                if action != NetworkAction::NoOp && live {
                    debug!(action = ?action, "Network action ready for processing");
                    // TODO: handle the ExitValidator event and remove the other events.
                    // https://github.com/sigp/anchor/issues/259
                }
            }
 ```
 
 removed

File: anchor/eth/src/sync.rs [hunk 21]
```diff
@@ -604,32 +604,7 @@ impl SsvEventSyncer {
...
-                        "Processing events from block {}", relevant_block
-                    );
-
-                    // process the logs and update the last block we have recorded
-                    self.event_processor.process_logs(logs, true).await;
-                    self.event_processor
-                        .db
-                        .processed_block(relevant_block)
-                        .expect("Failed to update last processed block number");
+                    continue;
```
dknopik: moving this out of the loop makes us resubscribe to the block stream after every processed block

anchor/eth/src/event_processor.rs
dknopik: Yeah, while the validator _should_ exist, it is not guaranteed. The owner might (for some reason) submit the exit request before actually depositing the validator. The correct thing to do then is to ignore it.

anchor/eth/src/event_processor.rs
dknopik: Yes, and we fill the share map by that exact same domain property. We built the client that way and also rely on it elsewhere. So why not rely on it here?

diegomrsantos: Is there a reason to change the current approach?

dknopik: Reduced complexity and, while not the main motivator, better performance is also a bonus.

diegomrsantos: It seems a very premature optimization for a feature that won't be used often

diegomrsantos: I changed it, you're right. I don't think that performance will make much difference, but your approach is MUCH simpler. But how does the shares handle the absence of the operator ID? Now I don't check it.

dknopik: Thank you. If we have no operator id, we also have no shares belonging to us. Therefore the shares map will be empty and your changed function will return `false`.

dknopik (commented):
LGTM, thank you! Will defer to @jking-aus for final approval due to architectural significance.

anchor/common/ssv_types/src/share.rs
```diff
@@ -6,7 +6,7 @@ use crate::{ClusterId, OperatorId};
 pub const ENCRYPTED_KEY_LENGTH: usize = 256;
 
 /// One of N shares of a split validator key.
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Eq, PartialEq)]
```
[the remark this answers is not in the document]
diegomrsantos: @dknopik ?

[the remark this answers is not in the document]
dknopik: cant hurt :shrug:

anchor/voluntary_exit/src/voluntary_exit_tracker.rs
```diff
@@ -0,0 +1,123 @@
...
+                for exit in entry.value() {
+                    ready_exits.push(exit.clone());
+                }
+            }
+        }
+        ready_exits
+    }
+
+    /// Remove a specific exit that has been successfully processed
+    pub fn remove_processed_exit(&self, exit_duty: &ExitDuty) {
```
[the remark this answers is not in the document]
diegomrsantos: That's a very good point. They don't decrement the counter in the go implementation. Any idea why? @dknopik @nkryuchkov

jking-aus (approved):

anchor/voluntary_exit/src/voluntary_exit_tracker.rs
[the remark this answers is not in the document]
nkryuchkov: @diegomrsantos We remove the entire slot data after one epoch: https://github.com/ssvlabs/ssv/blob/v2.3.1/operator/duties/voluntary_exit.go#L117. Or did you mean anything else?

Issue #302, opened 2025-05-05
Make `common/qbft` general purpose


dknopik:
There currently is no general purpose QBFT implementation in the Rust ecosystem.

We should use the opportunity and make our QBFT implementation usable by the general public. 

This means:
- It should not depend on ANY Lighthouse or Anchor crates
- It should not depend on Ethereum-specifc crates such as `ethereum_ssz`.

While we have taken some initial steps (e.g. through the `QbftData` trait), we have since deviated quite a bit from this path.

A rough and incomplete list of steps to do:
1. Move `QbftData` trait back into `qbft`.
2. Add further generics to avoid overfitting to the SSV case, such as `ParticipantId` instead of the hardcoded `OperatorId` type.
3. Do not rely on a specific hashing method.
4. Move SSV specific logic such as the construction of SSV messages into the QBFT manager.
5. Rethink message sender closure - maybe we should just return outgoing messages instead.
6. Offer better return types for every function to allow the user to better react to certain events.
7. `no_std` support would be awesome.
8. Move `common/qbft` into it's own repo.
 
CC @jxs

PR #303, opened 2025-05-06
Refactor and fix QBFT Manager 

dknopik:
I found several bugs related to the QBFT Manager and used the opportunity to refactor it a bit.

- Start the round interval at precisely specified instants
- Clean up validator-specific instances correctly
- Refactor by splitting up `qbft_instance` logic
- Properly detect timeout when going over `max_rounds`
- Correctly set `max_rounds`


anchor/common/qbft/src/lib.rs
```diff
@@ -59,6 +59,16 @@ impl<D: QbftData<Hash = Hash256>> ValidData<D> {
     }
 }
 
+pub trait MessageSender {
```
diegomrsantos, line 62: What's the difference between this and our current `MessageSender`?

anchor/qbft_manager/src/instance.rs
```diff
@@ -0,0 +1,300 @@
...
+    Uninitialized(Uninitialized),
+    // The instance is initialized
+    Initialized(Initialized<D>),
+    // The instance has been decided
+    Decided(Decided<D>),
+}
+
+#[derive(Default)]
+struct Uninitialized {
+    // A buffer of message that are being sent into the system before the instance has been
```
diegomrsantos:
```suggestion
    // A buffer of messages being sent into the system before the instance is
```

anchor/qbft_manager/src/instance.rs
```diff
@@ -0,0 +1,300 @@
...
+    // initialized. The maximum size of this is effectively capped by duty limits for messages
+    // and maximum instance lifetime enforced by the `cleaner`.
+    message_buffer: Vec<WrappedQbftMessage>,
```
diegomrsantos, line 43: Maybe it's a good idea to set an explicit limit here.

anchor/qbft_manager/src/instance.rs
```diff
@@ -0,0 +1,300 @@
...
+}
+
+struct Initialized<D: QbftData<Hash = Hash256>> {
+    qbft: Box<Qbft<D>>,
+    round_end: Interval,
+    sent_by_us: UnboundedReceiver<WrappedQbftMessage>,
```
diegomrsantos:
```suggestion
    msgs_sent_by_us: UnboundedReceiver<WrappedQbftMessage>,
```

anchor/validator_store/src/lib.rs
```diff
@@ -1246,6 +1280,10 @@ impl<T: SlotClock, E: EthSpec> ValidatorStore for AnchorValidatorStore<T, E> {
                     version: DATA_VERSION_PHASE0,
                     data_ssz: DataSsz::Contributions(data).as_ssz_bytes(),
                 },
+                self.get_instant_in_slot(
+                    slot,
+                    Duration::from_secs(self.spec.seconds_per_slot) * 2 / 3,
+                )?,
```
diegomrsantos: Could we extract this and the similar cases above into variables to improve readability?

anchor/qbft_manager/src/lib.rs
```diff
@@ -63,30 +65,33 @@ pub enum ValidatorDutyKind {
...
     // deserialziation we dermine the message is for the qbft instance and decode it into a
     // wrapped qbft messsage consisting of the signed message and the qbft message
     NetworkMessage(WrappedQbftMessage),
 }
 
-type Qbft<D, S> = qbft::Qbft<DefaultLeaderFunction, D, S>;
+#[derive(Debug)]
+pub struct QbftInitialization<D: QbftData> {
+    initial: D,
+    // The message id to be embedded into outgoing messages
```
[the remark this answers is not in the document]
dknopik: I consider adding field documentation to the struct instead of to the fields an antipattern. But I'll add some docs written by myself.

anchor/validator_store/src/lib.rs
```diff
@@ -557,6 +558,21 @@ impl<T: SlotClock, E: EthSpec> AnchorValidatorStore<T, E> {
...
+            .slot_clock
+            .start_of(slot)
+            .map(|start| start + delay)
+            .ok_or(SpecificError::SlotClock)?;
+        let now_duration = self
+            .slot_clock
+            .now_duration()
+            .ok_or(SpecificError::SlotClock)?;
+        let difference = target_duration - now_duration;
+        let instant = Instant::now() + difference;
```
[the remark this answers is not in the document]
dknopik: Reasonable idea, will add better comments though

anchor/common/qbft/src/lib.rs
```diff
@@ -114,17 +124,17 @@ where
...
 }
 
 impl<F, D, S> Qbft<F, D, S>
 where
     F: LeaderFunction + Clone,
     D: QbftData<Hash = Hash256>,
-    S: FnMut(UnsignedWrappedQbftMessage),
+    S: MessageSender,
 {
     // Construct a new QBFT Instance and start the first round
```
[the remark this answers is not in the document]
dknopik: Will add some docs

anchor/common/qbft/src/lib.rs
dknopik: Previously, we had a `FnMut(UnsignedWrappedQbftMessage)`, which is basically the same signature as this trait. During refactoring, it was hard to pass the closure across multiple types as closures are unnamable. Switching to a custom trait instead of using `FnMut` allowed for easier splitting of the logic in this PR.

anchor/qbft_manager/src/instance.rs
dknopik: I think the complexity to calculate a proper adequate limit is not worth it.

dknopik: Or do you think about a hardcoded limit based on some napkin estimation of the upper bound?

anchor/common/qbft/src/lib.rs
diegomrsantos: My question was more related to why we need it if we already have a `MessageSender` trait. What is the purpose of this new trait?

dknopik: In our implementation of the trait, you can see that some special handling is needed for the QBFT manager. 

Furthermore, the two traits represent different abstraction levels: Ideally, the QBFT crate does not know about how we handle e.g. subnets. The `message_sender` crate however does care about subnets, which is why it needs the committee id, which is completely irrelevant for QBFT itself. As soon as we generalize the QBFT crate (see #302), the differences will become even more apparent.

anchor/qbft_manager/src/instance.rs
diegomrsantos: I mean the last case, I reason that relying on implicit limits is risky. There's always the possibility of new code being introduced in the future, some hidden bug or attack vectors not foreseen. We can set a big enough limit to accommodate normal use, which will prevent unbound growth.

anchor/common/qbft/src/lib.rs
diegomrsantos: Could we use a channel instead?

diegomrsantos: Here it says Network Sender https://github.com/sigp/anchor/pull/303/files#diff-1480a846af2b359779733974f2f345dadfe4e3380369d7cc67218cf8494286d8R126

dknopik: >Could we use a channel instead?

The `qbft` crate should be not bound to any async runtime implementation.

anchor/qbft_manager/src/instance.rs
dknopik: done

diegomrsantos: Thanks a lot, looks great!

jking-aus (approved):

PR #305, opened 2025-05-06
Logging improvements

dknopik:
Collection of logging improvements all over the code base. 

These changes especially aim to allow debuggability using logfiles sent in by a user, while maintaining a nice overview on the console output without unnecessary detail.


diegomrsantos:
I was considering tackling areas of the codebase where we log and return an error, like in https://github.com/sigp/anchor/blob/0f74f321be2f54ecc535230165918beaec6e42e2/anchor/eth/src/event_processor.rs#L193. In cases like this, we should only return an error with the proper message. This error should be kept as the source, in full, if the caller wishes to wrap it with another error for context. In this way, we avoid log duplication. Perhaps, we could do this in this PR.

dknopik:
>In cases like this, we should only return an error with the proper message.

Well, ideally we return errors with structured data as long as possible and only turn into a `String` when we want to output it somewhere. But maybe you mean that. 

On the other hand, logging where the error originates gives us more information due to active spans that might not be active where they are caught. :thinking:

diegomrsantos:
> > In cases like this, we should only return an error with the proper message.
> 
> Well, ideally we return errors with structured data as long as possible and only turn into a `String` when we want to output it somewhere. But maybe you mean that.
> 
> On the other hand, logging where the error originates gives us more information due to active spans that might not be active where they are caught. 🤔

We need to capture the backtrace https://docs.rs/anyhow/latest/anyhow/struct.Error.html

anchor/common/qbft/src/lib.rs
```diff
@@ -424,7 +424,7 @@ where
         };
         self.data.insert(valid_data.hash, data);
 
-        debug!(from = ?operator_id, in = ?self.config.operator_id(), state = ?self.state, "PROPOSE received");
+        debug!(from = ?operator_id, state = ?self.state, "PROPOSE received");
```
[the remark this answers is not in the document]
dknopik: Generally, the local operator id is well known. Furthermore, info like that is more a candidate for spans covering the qbft instance if it is desired in e.g. testing contexts.

anchor/common/qbft/src/lib.rs
```diff
@@ -274,13 +274,20 @@ where
...
             _ => {
-                warn!(in = ?self.config.operator_id(), "Invalid data");
+                error!(
+                    msg = %wrapped_msg,
+                    "Invalid full data received",
+                );
+                debug!(
+                    full_data = hex::encode(wrapped_msg.signed_message.full_data()),
+                    "Raw invalid full data",
+                );
```
diegomrsantos, line 294: Do you think this is necessary?

dknopik: I added this because I had some trouble with mismatching data being sent - especially "Decode DataSsz variants explicitly based on duty and fork" of #310. I hope to have caught all cases of different data formats being sent by go-ssv and anchor now - but if there is still an issue, it is important that we can debug it with logs sent in by users. This is why we print an error with the message's display implementation (which omits some data for brevity to not nuke a user's terminal), and additionally log the full data in debug severity (which ends up in the log file, which we can then use to figure out why the data mismatches the expected format.

In practice, I think (and hope) that we do not hit this anymore, but if we do, we need to be able to debug it.

diegomrsantos: thanks for explaining

anchor/validator_store/src/lib.rs
```diff
@@ -469,7 +469,7 @@ impl<T: SlotClock, E: EthSpec> AnchorValidatorStore<T, E> {
         validator_pubkey: PublicKeyBytes,
         block: BeaconBlock<E, P>,
     ) -> Result<SignedBeaconBlock<E, P>, Error> {
-        debug!(?block, "Decided on BeaconBlock to sign");
+        trace!(?block, "Decided on BeaconBlock to sign");
```
diegomrsantos: Not totally sure about this and the next ones. What's your reasoning?

anchor/qbft_manager/src/instance.rs
```diff
@@ -260,13 +260,14 @@ pub async fn qbft_instance<D: QbftData<Hash = Hash256>>(
...
             RecvResult::Message(msg) => {
-                debug!(msg = ?msg.kind, "Handling message in qbft_instance");
                 match msg.kind {
                     QbftMessageKind::Initialize(initialization) => {
+                        debug!(msg_id = ?initialization.message_id, "Received initialization message");
                         instance = instance.initialize(initialization, &message_sender).await;
                     }
                     // We got a new network message, this should be passed onto the instance
                     QbftMessageKind::NetworkMessage(message) => {
+                        debug!(msg = %message, "Received message in qbft_instance");
```
diegomrsantos, line 270: Also not sure about debug here

anchor/message_validator/src/message_counts.rs
```diff
@@ -4,7 +4,7 @@ use ssv_types::{
     partial_sig::{PartialSignatureKind, PartialSignatureMessages},
 };
 
-use crate::ValidationFailure;
+use crate::ValidationFailureKind;
```
diegomrsantos: Why has this been changed?

anchor/message_receiver/src/manager.rs
```diff
@@ -94,7 +94,16 @@ impl<S: SlotClock + 'static, D: DutiesProvider> MessageReceiver
...
+                        // If we could decode the message, display that, else display the bytes.
+                        if let Some(msg) = failure.decoded_message {
+                            debug!(%msg, failure = ?failure.kind, "Validation failure");
+                        } else {
+                            debug!(
+                                msg = hex::encode(&message.data),
+                                failure = ?failure.kind,
+                                "Validation failure"
+                            );
+                        }
```
diegomrsantos: Could we do this inside the validator?

anchor/message_validator/src/lib.rs
```diff
@@ -102,7 +107,7 @@ pub enum ValidationFailure {
     InconsistentSigners,
     InvalidHash,
     FullDataHash,
-    UndecodableMessageData,
+    UndecodableMessageData(DecodeError),
```
diegomrsantos, line 137: Or we could put the msg here

anchor/validator_store/src/lib.rs
dknopik: Blocks can be rather large, especially if encoded via `Debug`. Mega-bytes worth data. Especially in conjunction with #310, where we include blobs in consensus. Throwing these into the logs is a waste of disk space IMO

`AggregateAndProof` and `Contribution` are smaller, but also mostly binary data (signatures and bitfields) that is mostly uninteresting for debugging.

For the block, we could instead log the block header only. For the other two, I do not care too strongly.

diegomrsantos: In this case It seems better to keep debug and log info necessary to identify the block

anchor/message_validator/src/message_counts.rs
dknopik: To be able to figure out why a message failed validation, I want to include failed messages in the logs. To enable that, I introduced moved `ValidationFailure` to `ValidationFailureKind` and added a struct called `ValidationFailure` that includes the decoded message for logging.

anchor/validator_store/src/lib.rs
diegomrsantos: Imo trace is for high frequency events that could help identify an issue, but we would use them only if debug logs weren't enough

anchor/message_receiver/src/manager.rs
dknopik: The advantage of having it outside is having different logging for outgoing/incoming message validation:

||Outgoing|Incoming|
|-|-|-|
|MessageAcceptance::Ignore| Not a big problem, happens from time to time if we send a message too late (e.g. due to late initialization of a qbft instance) | Would have ideally been caught by outgoing validation by the sender, but also unavoidable due to latency |
|MessageAcceptance::Reject| Wow, we could have been punished by peers for that, serious bug, log loudly |  Hm, weird, but might just be a bad peer, not worth an `error!` |

anchor/message_validator/src/lib.rs
dknopik: We want to put the message if it is decodable, so that we can log the decoded version instead of the raw bytes

diegomrsantos: Don't we return the decoded message already? Maybe we can use a custom type instead of a Result - containing the msg and an optional failure. Wdyt?

anchor/qbft_manager/src/instance.rs
dknopik: yeeeah, understandable. this is definitely something we should tone down later, and is kind of paranoid right now. however, when troubleshooting the qbft issues for proposals, this really helped to figure out who sent what and when, allowing me to find differences in behaviour between anchor and go-ssv. I already reduced impact as much as possible by creating the `Display` implementation and not logging large binary data such as signatures and `full_data`. 

I am kind of on the fence here, but knowing we'll likely have a lot to troubleshooting ahead of us (people do not like missed proposals and they are bound to happen for some reason), I think I would like to keep this in.

anchor/message_validator/src/lib.rs
diegomrsantos: Or both optional, as the msg might not be decodable.

anchor/qbft_manager/src/instance.rs
diegomrsantos: Sure, makes sense. Maybe just a comment explaining this reasoning.

anchor/message_validator/src/lib.rs
dknopik: Yeeeah, that sounds like a better approach. I'll change it on friday

anchor/message_validator/src/message_counts.rs
dknopik: See below, will change back

anchor/validator_store/src/lib.rs
dknopik: Okay, will check how we can change this back to `debug!` with low impact on logfile size

anchor/message_validator/src/lib.rs
dknopik: That was also kind of awkward as we have two types of messages because of the inner ssv message. I tried another approach, let me know what you think :sweat_smile:

anchor/validator_store/src/lib.rs
dknopik: done, wdyt?

anchor/message_validator/src/lib.rs
```diff
@@ -34,22 +34,49 @@ use tracing::{error, trace};
...
 
 pub(crate) const FIRST_ROUND: u64 = 1;
 
-#[derive(Debug, PartialEq)]
-pub struct ValidationFailure {
-    pub decoded_message: Option<Box<SignedSSVMessage>>,
-    pub kind: ValidationFailureKind,
+#[derive(Debug)]
+pub enum ValidationResult {
+    Success(ValidatedMessage),
```
diegomrsantos, line 46: Looks good. How about we rename `ValidatedMessage` to `DecodedMessage` and use it here and in `PostDecodeFailure`?

dknopik: The `ValidatedMessage` also has a `ValidatedSSVMessage` enum for the inner message variant, which we may not have in the `PostDecodeFailure` case, as we check some properties of the `SignedSSVMessage` first.

We could introduce a fourth case for errors that occur before the inner message is decoded.

diegomrsantos: I see, I don't have a strong opinion.

anchor/validator_store/src/lib.rs
diegomrsantos: Looks good, thanks

anchor/message_validator/src/lib.rs
dknopik: Then let's leave it as-is for now. :+1:

diegomrsantos (dismissed):
LGTM, thanks for the work on this

diegomrsantos (approved):

Issue #277, opened 2025-04-17
Validate messages related to voluntary exit


diegomrsantos:
Check if the validator is exising after https://github.com/sigp/anchor/issues/259 is done

dknopik:
@diegomrsantos this is fully done, right?

diegomrsantos:
> [@diegomrsantos](https://github.com/diegomrsantos) this is fully done, right?

The partial signature might be pending. Do both need to be validated related to exits?

dknopik:
>Do both need to be validated related to exits?

both what?

diegomrsantos:
> > Do both need to be validated related to exits?
> 
> both what?

Consensus and partial sig.

dknopik:
exited validators might still perform duties (the weird sync committee edge case), so filtering out all messages for exited validators is dangerous - I think we are fine like it is.

And if you mean messages related to the actual exit duty, AFAICT you implemented it completely

So I am closing this, if some specific thing we missed pops up we can open another

PR #306, opened 2025-05-07
Voluntary exit msg validation

diegomrsantos:
## Issue Addressed

https://github.com/sigp/anchor/issues/277

## Proposed Changes

This PR removes the old voluntary_exit module and introduces a standalone duties_tracker crate to fetch voluntary exit duty counts, wiring it into the message validator and ETH clients.

- Move exit duty logic into duties_tracker and update all crates to depend on it
- Integrate get_voluntary_exit_duty_count into consensus message validation for Role::VoluntaryExit


anchor/validator_store/src/lib.rs
```diff
@@ -584,6 +584,35 @@ impl<T: SlotClock, E: EthSpec> AnchorValidatorStore<T, E> {
...
         Ok(instant)
     }
+
+    pub async fn collect_voluntary_exit_signatures(
+        &self,
+        validator_pubkey: PublicKeyBytes,
+        voluntary_exit: VoluntaryExit,
+        slot: Slot,
+    ) -> Result<SignedVoluntaryExit, Error> {
+        let domain_hash = self.spec.get_builder_domain();
```
dknopik: You need to use `spec.get_domain` to get the correct domain hash.

Parameters:
`epoch` should be the epoch of `slot`,
`domain` should be `Domain::VoluntaryExit`,
`fork` should be `spec.fork_version_for_name(ForkName::Capella)` if we are at Capella or later, or the corresponding fork if earlier (see https://eips.ethereum.org/EIPS/eip-7044 and https://github.com/sigp/lighthouse/blob/82faf975b346e20fd82ee22dba3a845b99bd7c22/consensus/types/src/voluntary_exit.rs#L44-L49),
`genesis_validators_root` should be the genesis validators root.

dknopik (approved):
nice, LGTM! Thanks!

PR #307, opened 2025-05-07
add `--enable-high-validator-count-metrics`

dknopik:
add CLI flag to specify if we should enable per validator metrics if we have a high validator count


diegomrsantos (approved):
Thanks a lot for handling this!

PR #308, opened 2025-05-12
Run ARM build on ARM instance

dknopik:
Instead of running the ARM build inside QEMU on an x86 instance (which takes long), use Githubs ARM instances. Furthermore, mark `SELF_HOSTED_RUNNERS` as false while we do not use them.


.github/workflows/docker.yml
```diff
@@ -14,8 +14,7 @@ concurrency:
   cancel-in-progress: true
 
 env:
-    # Enable self-hosted runners for the sigp repo only.
-    SELF_HOSTED_RUNNERS: ${{ github.repository == 'sigp/anchor' }}
+    SELF_HOSTED_RUNNERS: false
```
diegomrsantos: Is this necessary?

.github/workflows/docker.yml
```diff
@@ -32,12 +31,14 @@ jobs:
...
-        runs-on: ubuntu-24.04
+        runs-on: ${{ matrix.arch.os }}
         strategy:
             matrix:
-                cpu_arch: [aarch64, x86_64]
+                arch:
+                - os: ubuntu-24.04
+                  short: amd64
+                - os: ubuntu-24.04-arm
+                  short: arm64
```
diegomrsantos: How about `cpu` instead of `short`?

diegomrsantos:
This workflow doesn't run on PR's. I suggest enabling it to run temporarily here to check if everything is working. Or even better, adding `workflow_dispatch:` allows us to trigger it manually. But I believe it is gonna work only after it's merged, not sure.

.github/workflows/docker.yml
dknopik: removed and opened #318

.github/workflows/docker.yml
dknopik: i changed it to `docker`, as it is only used in the docker step. my concern is that in rust `amd64` is not valid, as `x86_64` is used.

dknopik:
@diegomrsantos for reference, it has served me well on interop: https://github.com/sigp/anchor/actions/workflows/docker.yml?query=branch%3Ainterop1

can still enable it if you want to make sure

diegomrsantos:
Up to you, but the changes made here can't be tested before we merge, can they?

diegomrsantos (approved):
looks great, thanks!

PR #309, opened 2025-05-12
Avoid double processing of HEAD-8 if HEAD reorgs

dknopik:
## Issue Addressed

We have a follow distance of 8 to avoid reorgs. However, if the head block reorgs, our block subscription returns the new block - which reuses the previous block number, and cause us to again process the block at the follow distance - causing issues such as doubly incremented nonces.

## Proposed Changes

Before syncing `relevant_block`, check if we already have done so by comparing with `get_last_processed_block`.

## Additional Info

I could not hesitate to make some of the surrounding code a bit more idiomatic and avoid one level of indentation.


Zacholme7:
Thinking through this, dont we face the issue that we arent properly processing the new block which will still lead to incorrect state?

dknopik:
We are processing the new block as soon as it becomes the relevant block - which is 8 blocks in the future.

Zacholme7 (dismissed):
lgtm, thx

diegomrsantos:
> We have a follow distance of 8 to avoid reorgs

Could you please elaborate more on that?

Zacholme7:
@diegomrsantos I took the 8 block follow distance from the SSV code. For log fetching we are always 8 blocks behind the head of the chain so the head may reorg but we are following at a distance anyways. We would face an issue if there was an 8 block reorg, but for all intents and purposes I think this isnt something we have to worry about.

anchor/eth/src/sync.rs
```diff
@@ -612,54 +612,60 @@ impl SsvEventSyncer {
...
+                        relevant_block, "Already synced block - likely reorg"
                     );
+                    continue;
+                }
 
-                    metrics::set_gauge(
-                        &metrics::EXECUTION_CURRENT_BLOCK,
-                        block_header.number as i64,
-                    );
+                debug!(
```
diegomrsantos, line 641: `trace` might be better to avoid too many logs

dknopik: this is done once per slot, and signals the block numbers we are treating. though this log (which existed before this pr, see diff above) I actually figured out this isse, so I definitely think it has value

diegomrsantos (approved):
lgtm, good catch!

PR #310, opened 2025-05-12
Proposal fixes

dknopik:
This PR fixes proposals! The commits fix several issues, I am not sure which are strictly necessary to fix proposals, but each is an improvement anyway.

[Decode `DataSsz` variants explicitly based on duty and fork](https://github.com/sigp/anchor/pull/310/commits/cf573b6e72744a8bc0bcda4f81347af43d1d7fc9): Currently, Fulu and Electra blocks are identical when serialized in SSZ. Therefore, when we deserialize/deserialise the block, `from_ssz_bytes` accidentally decodes the block into the `Fulu` variant, causing us to send a Fulu block to the BN, which in turn refuses to publish it. The commit explicitly decodes the block using the `version` field from the decided data and similarly decodes other objects directly as well in other duties (to be safe).

[Set `max_transmit_size`](https://github.com/sigp/anchor/pull/310/commits/a353880036d9baee590250b7775d8cda2373b710): Some messages with large blocks failed to be sent as they are quite large - we set `max_transmit_size` to a large enough value of 5 million bytes. 

**Outdated, see "Improve fix for round timeouts" below** - [Avoid conflict between timer-based and message-based round increments](https://github.com/sigp/anchor/pull/310/commits/159126ed89ff1097c29df2a4da674399377743cd): If we receive a quorum of round change messages, we set our own round to the round mentioned in these messages. However, if a timer tick arrives after, it increased the round regardless - even though actually it meant to increase the round to the round found in those messages. We fix this by tracking the round requested by ticks separately.

[Remember PREPARE even if we have not received the PROPOSE yet](https://github.com/sigp/anchor/pull/310/commits/f6d89986384f0d0c13683fb1708b63dbd79a6f4d): We discarded PREPARE messages if we have not received a PROPOSE message yet - which in some cases resulted in missed quorums. We record unknown prepare messages, but still only check for quorum if we have the PROPOSE message.

[Use current slot for RANDAO Reveal](https://github.com/sigp/anchor/pull/310/commits/460a288ce3f3359a3eb3c8d8d9e8131d80c65634): We need to use the current slot instead of the last slot of the epoch for go-ssv to pick up the messages correctly.

[Address review and fix proposal data](https://github.com/sigp/anchor/pull/310/commits/a309453dee9a04373e117778b7a407e4669b7661): While the logic was correct for blinded block and pre-Deneb full blocks, it was not correct for Deneb and Electra blocks as blobs need to decided along with the block. This commit sets the data appropriately - but this requires a small change in LH which is pending approval: https://github.com/sigp/lighthouse/pull/7497.

[Improve fix for round timeouts](https://github.com/sigp/anchor/pull/310/commits/73c5ce77a2b228ba841f533bc2aa223421401ae7): Round changes might be triggered by messages instead of timer ticks. In order to avoid incorrect timeouts after a round has been changed by a message (e.g. after replaying buffered messages), we calculate the currently adequate timeout every time before we wait for a message or round timeout.


anchor/common/ssv_types/src/consensus.rs
```diff
@@ -255,6 +254,22 @@ impl TreeHash for DataVersion {
     }
 }
 
+impl TryFrom<DataVersion> for ForkName {
+    type Error = ();
+
+    fn try_from(value: DataVersion) -> Result<ForkName, Self::Error> {
```
diegomrsantos: is this related to https://github.com/sigp/anchor/pull/310/commits/cf573b6e72744a8bc0bcda4f81347af43d1d7fc9#diff-25a340f5184292e9132e4b012313bfd9e056682eb85555b96830b1909801b2f1R441?

anchor/common/qbft/src/lib.rs
```diff
@@ -603,6 +601,12 @@ where
             warn!(from = ?operator_id, "PREPARE message is a duplicate")
         }
 
+        // Make sure that we have accepted a proposal for this round
+        if !self.proposal_accepted_for_current_round {
+            warn!(from=?operator_id, ?self.state, self=?self.config.operator_id(), "Have not accepted Proposal for current round yet");
+            return;
+        }
+
```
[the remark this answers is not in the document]
dknopik: what? no. i removed the other one

anchor/common/ssv_types/src/consensus.rs
dknopik: yes: `ForkName::try_from(completed_data.version)`

anchor/common/ssv_types/src/consensus.rs
```diff
@@ -255,6 +254,22 @@ impl TreeHash for DataVersion {
     }
 }
```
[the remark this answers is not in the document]
dknopik: This is self-explanatory. a 12 line comment explaining this very simple functionality is noise.

anchor/validator_store/src/lib.rs
```diff
@@ -459,10 +460,21 @@ impl<T: SlotClock, E: EthSpec> AnchorValidatorStore<T, E> {
...
+            BlindedBeaconBlock::from_ssz_bytes_for_fork(&completed_data.data_ssz, fork)
+        {
+            Ok(DataSsz::BlindedBeaconBlock(blinded_block))
+        } else if let Ok(block) =
+            BeaconBlock::from_ssz_bytes_for_fork(&completed_data.data_ssz, fork)
+        {
+            Ok(DataSsz::BeaconBlock(block))
+        } else {
+            error!(%fork, "Failed to deserialize decided block");
+            Err(Error::SpecificError(SpecificError::InvalidQbftData))
```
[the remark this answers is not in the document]
dknopik: fair enough, but I think one error is enough information here

anchor/common/ssv_types/src/consensus.rs
```diff
@@ -225,7 +224,7 @@ impl TreeHash for BeaconRole {
...
 
-#[derive(Clone, Debug, PartialEq, Decode, Encode)]
+#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Decode, Encode)]
 #[ssz(struct_behaviour = "transparent")]
 pub struct DataVersion(u64);
```
diegomrsantos: Could this be an enum?

dknopik: yes, and it probably should be. do you want me to change it in this PR or in a different one?

diegomrsantos: I don't have any preference, whatever works better for you

anchor/common/ssv_types/src/consensus.rs
```diff
@@ -255,6 +254,22 @@ impl TreeHash for DataVersion {
...
+
+    fn try_from(value: DataVersion) -> Result<ForkName, Self::Error> {
+        match value {
+            DATA_VERSION_PHASE0 => Ok(ForkName::Base),
+            DATA_VERSION_ALTAIR => Ok(ForkName::Altair),
+            DATA_VERSION_BELLATRIX => Ok(ForkName::Bellatrix),
+            DATA_VERSION_CAPELLA => Ok(ForkName::Capella),
+            DATA_VERSION_DENEB => Ok(ForkName::Deneb),
+            DATA_VERSION_ELECTRA => Ok(ForkName::Electra),
+            _ => Err(()),
```
diegomrsantos: How about a more descriptive error?

anchor/common/ssv_types/src/consensus.rs
dknopik: let's do it separately, I am eager to refactor some types anyway :)

anchor/common/ssv_types/src/consensus.rs
dknopik: there is only one error case. I am usually against returning strings in `TryFrom` implementations, as the caller might not even need an error message as it can fall back to some other logic. A single variant enum is also noise IMO.

diegomrsantos: how about 

```rust
#[derive(Debug)]
pub struct UnknownDataVersion(pub DataVersion);

impl TryFrom<DataVersion> for ForkName {
    type Error = UnknownDataVersion;

    fn try_from(v: DataVersion) -> Result<Self, Self::Error> {
        match v {
            DATA_VERSION_PHASE0   => Ok(ForkName::Base),
            DATA_VERSION_ALTAIR   => Ok(ForkName::Altair),
            DATA_VERSION_BELLATRIX=> Ok(ForkName::Bellatrix),
            DATA_VERSION_CAPELLA  => Ok(ForkName::Capella),
            DATA_VERSION_DENEB    => Ok(ForkName::Deneb),
            DATA_VERSION_ELECTRA  => Ok(ForkName::Electra),
            _ => Err(UnknownDataVersion(v)),
        }
    }
}
```

anchor/common/ssv_types/src/consensus.rs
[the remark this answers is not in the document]
diegomrsantos: Unfortunately, it got quite verbose with this kind of suggestion. At least, it says [nitpick].

anchor/validator_store/src/lib.rs
```diff
@@ -413,43 +414,55 @@ impl<T: SlotClock, E: EthSpec> AnchorValidatorStore<T, E> {
...
+        };
+
+        // Get the validator index, ensuring it exists
+        let validator_index = validator
+            .metadata
+            .index
+            .ok_or(SpecificError::MissingIndex)?;
+
+        // Determine the appropriate version based on block type
+        let block_version = match &block {
```
diegomrsantos: how about sth like 

```rust
impl<E: EthSpec, P: AbstractExecPayload<E>> TryFrom<BeaconBlock<E, P>> for DataVersion {
    type Error = UnknownDataVersion;

    fn try_from(value: BeaconBlock<E, P>) -> Result<DataVersion, Self::Error> {
        match value {
            BeaconBlock::Base(_) => Ok(DATA_VERSION_PHASE0),
            BeaconBlock::Altair(_) => Ok(DATA_VERSION_ALTAIR),
            BeaconBlock::Bellatrix(_) => Ok(DATA_VERSION_BELLATRIX),
            BeaconBlock::Capella(_) => Ok(DATA_VERSION_CAPELLA),
            BeaconBlock::Deneb(_) => Ok(DATA_VERSION_DENEB),
            BeaconBlock::Electra(_) => DATA_VERSION_ELECTRA,
            _ => Err(UnknownDataVersion),
        }
    }
}
```

anchor/validator_store/src/lib.rs
diegomrsantos: Is trial and error the only way to do that? The previous code also checked for more cases, wasn't it necessary?

anchor/validator_store/src/lib.rs
```diff
@@ -1038,11 +1050,16 @@ impl<T: SlotClock, E: EthSpec> ValidatorStore for AnchorValidatorStore<T, E> {
             Completed::Success(data) => data,
         };
 
-        let data_ssz = DataSsz::from_ssz_bytes(&data.data_ssz);
-
-        let message = match data_ssz {
-            Ok(DataSsz::AggregateAndProof(message)) => message,
-            _ => return Err(Error::SpecificError(SpecificError::InvalidQbftData)),
+        let message = if data.version < DATA_VERSION_ELECTRA {
```
diegomrsantos: We could use a match here

anchor/common/qbft/src/lib.rs
```diff
@@ -98,8 +98,11 @@ where
     valid_start_data: ValidData<D>,
     /// All of the data that we have seen
     data: HashMap<D::Hash, Arc<D>>,
-    /// The current round this instance state is in.a
+    /// The current round this instance state is in.
     current_round: Round,
+    /// The round requested by the user. Used to not accidentally start a round that was already
```
diegomrsantos: > The round requested by the user

Could you elaborate on what this means?

anchor/common/qbft/src/lib.rs
```diff
@@ -159,6 +162,7 @@ where
             valid_start_data,
             data: HashMap::new(),
             current_round,
+            requested_round: current_round,
```
diegomrsantos: Should this be `None` at this point?

anchor/common/qbft/src/lib.rs
```diff
@@ -832,14 +836,17 @@ where
...
 
-        // Bump the current round
-        self.current_round = next_round;
-
-        // Set the state so SendRoundChange so we include Round + 1 in message
-        self.state = InstanceState::SentRoundChange;
-
-        self.send_round_change(Hash256::default());
-        self.start_round();
+        self.requested_round = next_round;
```
diegomrsantos: `requested_round` is initially equal to `current_round` and this seems to be the only place where `requested_round` is updated. How is the value updated by this "request"?

anchor/common/qbft/src/lib.rs
```diff
@@ -832,14 +836,17 @@ where
...
+        if self.current_round < next_round {
```
diegomrsantos: Why is the `current_round` updated only conditionally now?

anchor/common/qbft/src/lib.rs
Zacholme7: Second this

Zacholme7 (commented):
Diego covered pretty much everything. Also just seconded some small terminology clarification on my end.

anchor/validator_store/src/lib.rs
dknopik: Yes, trial and error is the only way as the surrounding structure does not indicate whether the block is blinded or not.

dknopik: The previous code tried all `DataSsz` variants, uselessly, as we would fail anyway in the calling function if an invalid variant would be encountered.

anchor/validator_store/src/lib.rs
dknopik: personally, for decisions with 2 options, i prefer a classic if-else.

as noted elsewhere, the data version should be a enum anyway, so i will need to touch this again. as soon as this is an enum, I might need to change it anyway :)

anchor/common/qbft/src/lib.rs
dknopik: No.

By initializing the instance, the user sets the instance into the first round as requested in the config, therefore the first round is the requested one

anchor/common/qbft/src/lib.rs
dknopik: There is an explanation of the issue in the PR description - I am not sure which part of it is unclear, please let me know which aspect I should elaborate more.

anchor/common/qbft/src/lib.rs
dknopik: It is incremented by one, see `next` above

anchor/common/qbft/src/lib.rs
dknopik: User, as in user of the crate. The user calls `end_round` if they want the round they requested previously to end, usually timer-based. They previously requested a round via initialization config or another `end_round` call.

anchor/validator_store/src/lib.rs
dknopik: I've reworked the `DataVersion` a bit, using `ForkName` internally. Looking forwards to your thoughts about that approach, I am a bit on the fence about it.

anchor/common/ssv_types/src/consensus.rs
dknopik: @diegomrsantos I've reworked the DataVersion a bit, using ForkName internally. Looking forwards to your thoughts about that approach, I am a bit on the fence about it.

dknopik:
Hey @diegomrsantos and @Zacholme7,

I found and solved more issues (see description), and made the code around versions a bit cleaner (IMO). Let me know if you have any questions or suggestions!

As one fix requires a (maybe slightly questionable) change in LH, I'd like to wait until that is approved upstream, but feel free to review already. :) Thank you!

anchor/common/ssv_types/src/consensus.rs
```diff
@@ -226,48 +226,81 @@ impl TreeHash for BeaconRole {
...
-    fn tree_hash_root(&self) -> tree_hash::Hash256 {
-        self.0.tree_hash_root()
+    fn ssz_fixed_len() -> usize {
+        <u64 as Decode>::ssz_fixed_len()
+    }
+
+    fn from_ssz_bytes(bytes: &[u8]) -> Result<Self, DecodeError> {
+        let num = u64::from_ssz_bytes(bytes)?;
+        match num {
+            0 => Ok(DataVersion::Unknown),
```
diegomrsantos: How does the code proceed if we hit this?

anchor/common/ssv_types/src/consensus.rs
```diff
@@ -226,48 +226,81 @@ impl TreeHash for BeaconRole {
...
         match value {
-            DATA_VERSION_PHASE0 => Ok(ForkName::Base),
-            DATA_VERSION_ALTAIR => Ok(ForkName::Altair),
-            DATA_VERSION_BELLATRIX => Ok(ForkName::Bellatrix),
-            DATA_VERSION_CAPELLA => Ok(ForkName::Capella),
-            DATA_VERSION_DENEB => Ok(ForkName::Deneb),
-            DATA_VERSION_ELECTRA => Ok(ForkName::Electra),
-            _ => Err(()),
+            DataVersion::Fork(fork) => Ok(fork),
+            _ => Err(UnknownDataVersion),
```
diegomrsantos: If that's the answer to my question above, can we get rid of the unknown variant and return an error like this during decoding?

anchor/common/ssv_types/src/consensus.rs
```diff
@@ -227,33 +226,82 @@ impl TreeHash for BeaconRole {
     }
 }
 
-#[derive(Clone, Debug, PartialEq, Decode, Encode)]
-#[ssz(struct_behaviour = "transparent")]
-pub struct DataVersion(u64);
+#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
+pub enum DataVersion {
```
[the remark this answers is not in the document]
dknopik: It would not compile if we truly needed it m8

anchor/common/ssv_types/src/consensus.rs
dknopik: Hmm, a "Unknown" version is still well defined with a value of "0" and is different case from a value that is not well defined at all (such as "17"). This might be a bit pedantic though as it seems to be unused in SSV.

Eh, lets get rid of it.

diegomrsantos: Thank you. Perhaps one last improvement would be adding a comment explaining why we can't use `ForkName` directly. If I saw the code for the first time, that would be the question in my mind.

Zacholme7 (commented):
lgtm, ill give @jking-aus final review

dknopik:
pls do not merge yet, I want to properly update the LH deps in a separate PR

anchor/qbft_manager/src/instance.rs
```diff
@@ -177,6 +171,12 @@ impl<D: QbftData> From<Option<QbftMessage<D>>> for RecvResult<D> {
 
 impl<D: QbftData<Hash = Hash256>> Initialized<D> {
     async fn recv(&mut self, rx: &mut UnboundedReceiver<QbftMessage<D>>) -> RecvResult<D> {
+        // We calculate the sleep dynamically, as both messages and the local timer might cause the
+        // round to advance
+        let round_end = calculate_round_timeout(self.qbft.get_round().into(), &self.start_time);
```
diegomrsantos, line 176: Does it mean `round_end_time` or `round_timeout`?

anchor/qbft_manager/src/instance.rs
```diff
@@ -185,11 +185,7 @@ impl<D: QbftData<Hash = Hash256>> Initialized<D> {
                     drop_on_finish: None
                 }).into()
             },
-            _ = &mut self.round_end => {
-                // Compute timeout for next round
-                let next_round = self.qbft.get_round().get() as u64 + 1_u64;
-                let sleep = calculate_round_timeout(next_round, &self.start_time);
-                self.round_end = Box::pin(tokio::time::sleep_until(sleep));
+            _ = &mut round_timeout_sleep => {
```
diegomrsantos, line 188: Is there any behaviour change in extracting the logic, or is it only to improve readability? Why was the `+ 1` removed?

anchor/common/qbft/src/lib.rs
```diff
@@ -842,17 +837,14 @@ where
...
-            // Set the state so SendRoundChange so we include Round + 1 in message
-            self.state = InstanceState::SentRoundChange;
+        // Set the state so SendRoundChange so we include Round + 1 in message
+        self.state = InstanceState::SentRoundChange;
 
-            self.send_round_change(Hash256::default());
-            self.start_round();
-        }
+        self.send_round_change(Hash256::default());
+        self.start_round();
```
diegomrsantos: I wouldn't expect this to be called inside this function.

diegomrsantos:
@dknopik, can we add tests to be sure (and avoid regressions) that there's no race related to the round timeout?

anchor/validator_store/src/lib.rs
```diff
@@ -382,72 +382,60 @@ impl<T: SlotClock, E: EthSpec> AnchorValidatorStore<T, E> {
...
-                current_slot,
-            });
-        }
-
-        let wrapped = wrapper(block.clone());
+        signable_block: impl SignableBlock<E>,
+    ) -> Result<UnsignedBlock<E>, Error> {
         let validator = self.validator(validator_pubkey)?;
 
+        let block = signable_block.as_block();
```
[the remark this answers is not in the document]
dknopik: Nah, this sanity check is only needed before signing

anchor/validator_store/src/lib.rs
```diff
@@ -1006,8 +1016,8 @@ impl<T: SlotClock, E: EthSpec> ValidatorStore for AnchorValidatorStore<T, E> {
         let validator = self.validator(validator_pubkey)?;
 
         let version = match &aggregate {
-            Attestation::Base(_) => DATA_VERSION_PHASE0,
-            Attestation::Electra(_) => DATA_VERSION_ELECTRA,
+            Attestation::Base(_) => ForkName::Base.into(),
+            Attestation::Electra(_) => ForkName::Electra.into(),
```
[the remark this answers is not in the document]
dknopik: There are only these two types of attestations

anchor/common/qbft/src/lib.rs
dknopik: this function is not modified anymore by this PR in total (only by the latest commit), so it is not really in scope for this review

anchor/qbft_manager/src/instance.rs
dknopik: yes, there is a change: by extracting the logic we make sure to catch any round changes caused by messages and time out at the end of the new round if so

The `+ 1` is removed because we are now calculating the timeout for the current round, and not for the "next" round as done by the previous code

dknopik:
> @dknopik, can we add tests to be sure (and avoid regressions) that there's no race related to the round timeout?

Sure. Do you want me to do so? If you want, you can take a stab at it instead - that might prevent issues like me "overfitting" the test on my mental model of the issue.

anchor/qbft_manager/src/instance.rs
dknopik: seeing that the type is `Instant`, there is not really a difference IMO. But always happy to considers suggestions for better naming

anchor/common/ssv_types/src/consensus.rs
diegomrsantos: @dknopik has this been done?

diegomrsantos:
> > @dknopik, can we add tests to be sure (and avoid regressions) that there's no race related to the round timeout?
> 
> Sure. Do you want me to do so? If you want, you can take a stab at it instead - that might prevent issues like me "overfitting" the test on my mental model of the issue.

Feel free to do it.

diegomrsantos (dismissed):
Thanks for the work on this one, it's really great.

anchor/common/ssv_types/src/consensus.rs
dknopik: Sorry - forgot this. Thanks for pinging me. I've added a comment, please take a look if it is sufficient

diegomrsantos: It's great, thank you.

dknopik:
@diegomrsantos thanks for the idea about a test - I found another minor data race thanks to it.

We removed the message round check in the qbft crate as we now check for this in message validation. However, as the message might be handled by the qbft crate after the round has advanced, this caused an interesting issue where a message belonging to a previous round change arrived and caused an instance to revert to a previous round.

I reintroduced the round check.

anchor/common/qbft/src/lib.rs
```diff
@@ -235,6 +235,19 @@ where
         &self,
         wrapped_msg: &WrappedQbftMessage,
     ) -> Option<(Option<ValidData<D>>, OperatorId)> {
+        // Ensure that this message is for the correct round
+        let current_round = self.current_round.get();
+        if (wrapped_msg.qbft_message.round < current_round as u64)
+            || (wrapped_msg.qbft_message.round > self.config.max_rounds() as u64)
+        {
+            warn!(
```
diegomrsantos: Do you think it should be a `warn`?

anchor/common/qbft/src/lib.rs
```diff
@@ -235,6 +235,19 @@ where
...
+                propose_round = wrapped_msg.qbft_message.round,
+                current_round = *self.current_round,
+                "Message received for a invalid round"
```
diegomrsantos: How about "for a previous round"?

anchor/qbft_manager/src/tests.rs
```diff
@@ -110,21 +112,37 @@ where
...
+                    let aggregated_commit = result
+                        .aggregated_commit
+                        .expect("If consensus was reached, this must exist");
+                    assert!(
+                        aggregated_commit.signatures().len() as u64
+                            >= (self.tester.size as u64 - self.tester.size.get_f())
+                    );
+                },
+                _ = &mut timeout => {
+                    panic!("test timed out")
```
diegomrsantos, line 156: How about defining the timeout here?

dknopik: There is a difference between defining it here vs above. If we define it here, it is refreshed every time a message comes in from the queue. If we define it above, it is only created once, so we wait at most 5 minutes. Personally, I prefer the latter as it is more predictable, but I don't feel strongly.

anchor/common/qbft/src/lib.rs
dknopik: changed message and removed check for `max_round`, as we check that in validation and is not subject to a data race

anchor/common/qbft/src/lib.rs
dknopik: nope, fixed

anchor/qbft_manager/src/tests.rs
```diff
@@ -653,6 +680,7 @@ mod manager_tests {
             setup.executor,
             CommitteeSize::Four,
             setup.all_data,
+            HashMap::new(),
```
diegomrsantos: I've been thinking about this. Unfortunately, Rust doesn't have default parameters. Could you tell me what's the idiomatic way to handle this? Creating an additional function `new_with_delay`?

anchor/qbft_manager/src/tests.rs
diegomrsantos: Or even better, a const in a more visible place.

anchor/qbft_manager/src/tests.rs
```diff
@@ -828,6 +865,31 @@ mod manager_tests {
...
+    async fn test_late_initialization() {
+        let setup = setup_test(1);
+        let mut context = TestContext::<BeaconVote>::new(
+            setup.clock,
+            setup.executor,
+            CommitteeSize::Four,
+            setup.all_data,
+            HashMap::from([
+                (OperatorId(2), Duration::from_secs(3)),
+                (OperatorId(3), Duration::from_secs(5)),
```
diegomrsantos: Very nice docs at the beginning. How about extracting it to a new variable with a descriptive name? It might be useful to state the round time too.

anchor/qbft_manager/src/tests.rs
dknopik: if there is only only one "optional" parameter, `new_with_x` is fine. However, once you have more optional parameters, a builder struct is probably the way to go, because `new_with_x_and_y` gets ugly fast unless you have a nice way of phrasing the different `new_*` functions in the context of the application. I'll add `new_with_delay`.

anchor/qbft_manager/src/tests.rs
diegomrsantos: Looks good, thanks

anchor/common/ssv_types/src/consensus.rs
```diff
@@ -227,75 +227,61 @@ impl TreeHash for BeaconRole {
...
+    fn ssz_append(&self, buf: &mut Vec<u8>) {
+        let num: u64 = match self.0 {
+            ForkName::Base => 1,
+            ForkName::Altair => 2,
+            ForkName::Bellatrix => 3,
+            ForkName::Capella => 4,
+            ForkName::Deneb => 5,
+            ForkName::Electra => 6,
+            ForkName::Fulu => 7,
+        };
```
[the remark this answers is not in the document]
dknopik: no

anchor/common/qbft/src/lib.rs
```diff
@@ -235,6 +235,17 @@ where
         &self,
         wrapped_msg: &WrappedQbftMessage,
     ) -> Option<(Option<ValidData<D>>, OperatorId)> {
+        // Ensure that this message is for the correct round
+        let current_round = self.current_round.get();
+        if wrapped_msg.qbft_message.round < current_round as u64 {
```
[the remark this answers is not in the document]
diegomrsantos: I think we can use into() instead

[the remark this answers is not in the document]
dknopik: No, there is no `From<usize> for u64`

This kind of discrepancy between our message types and our custom types such as `Round` is something I have been meaning to fix for a while... When that PR lands, this will be nicer ;)

diegomrsantos (approved):

PR #327, opened 2025-05-16
Start curr_epoch_duties at 0 for a new epoch

diegomrsantos:
## Issue Addressed

https://github.com/sigp/anchor/pull/301/files starts `curr_epoch_duties`at 1 for a new epoch, but it should be zero as the state is updated only after checking. In the current code, the check will fail if our limit is 1.

## Proposed Changes

Start curr_epoch_duties at 0 for a new epoch


diegomrsantos:
cc @nkryuchkov

diegomrsantos:
Sorry guys, ignore it, the current code is correct. The epoch is changed after we check the message.

Issue #322, opened 2025-05-15
Correct round time logic


dknopik:
We seem to not have correctly implemented the round timeout logic:

https://github.com/ssvlabs/ssv/blob/main/protocol/v2/qbft/roundtimer/timer.go#L105

See also:
https://github.com/ssvlabs/ssv/pull/1120

PR #330, opened 2025-05-16
Timeout fix

Zacholme7:
## Issue Addressed

#322 

## Proposed Changes

Fix the round change timeouts to be in line with ssv.


anchor/qbft_manager/src/instance.rs
```diff
@@ -107,13 +109,19 @@ impl<D: QbftData<Hash = Hash256>> QbftInstance<D> {
...
+    ) -> Initialized<D, T> {
         // Create the interval and tick it right away to wait until the start time if necessary.
-        let mut interval = tokio::time::interval_at(init.start_time, init.config.round_time());
+        let timeout = calculate_round_timeout(
+            init.message_id.role(),
+            init.config.instance_height(),
+            &init.config.round(),
+            &init.slot_clock,
+        );
+        let mut interval = tokio::time::interval_at(init.start_time, timeout);
```
dknopik: Intervals are not really the appropriate tool here anymore, as we might change the round timeout dynamically as determined by `calculate_round_timeout`. Further down you replace the interval on every tick, so I think `sleep_until` might be a more appropriate tool.

Furthermore, here it is sufficient to sleep until `start_time`.

anchor/qbft_manager/src/timeout.rs
```diff
@@ -0,0 +1,65 @@
...
+        Role::Aggregator | Role::SyncCommittee => {
+            // two-thirds of the slot time
+            slot_duration / 3 * 2
+        }
+        _ => {
+            if round.get() <= QUICK_TIMEOUT_THRESHOLD.get() {
+                return Duration::from_secs(QUICK_TIMEOUT);
+            } else {
+                return Duration::from_secs(SLOW_TIMEOUT);
+            }
```
dknopik: With these `return`s you calculate the time of a single round, while you return the total time to sleep down below.

anchor/qbft_manager/src/timeout.rs
```diff
@@ -0,0 +1,65 @@
+use std::{ops::Add, sync::LazyLock, time::Duration};
+
+use slot_clock::SlotClock;
+use ssv_types::{Round, msgid::Role};
+use types::Slot;
+
+use crate::InstanceHeight;
+
+pub static QUICK_TIMEOUT_THRESHOLD: LazyLock<Round> = LazyLock::new(|| Round::from(8));
```
dknopik: Use a `const` here.

anchor/qbft_manager/src/timeout.rs
```diff
@@ -0,0 +1,65 @@
...
+        let quick_portion = Duration::from_secs(QUICK_TIMEOUT_THRESHOLD.get() as u64 * 2);
+        let slow_portion = Duration::from_secs(
+            (round.get() as u64 - QUICK_TIMEOUT_THRESHOLD.get() as u64) * SLOW_TIMEOUT,
+        );
+        quick_portion + slow_portion
+    };
+
+    let total_timeout = base_duration.add(additional_timeout);
+
+    time_to_slot_start + total_timeout
```
dknopik: As `Durations` can not be negative (and `time_to_slot_start` is therefore `0` at best), this logic is kind of flawed, as we will therefore wait for `total_timeout` for a single round. 

Instead, add `additional_timeout` to `start_time` as passed to the `qbft_instance`

anchor/qbft_manager/src/timeout.rs
```diff
@@ -0,0 +1,65 @@
...
+pub static QUICK_TIMEOUT_THRESHOLD: LazyLock<Round> = LazyLock::new(|| Round::from(8));
+
+const QUICK_TIMEOUT: u64 = 2; // 2 Seconds
+const SLOW_TIMEOUT: u64 = 120; // 2 Minutes
+
+pub fn calculate_round_timeout<T: SlotClock + 'static>(
+    role: Option<Role>,
+    height: &InstanceHeight,
+    round: &Round,
+    slot_clock: &T,
```
dknopik: This should just be `start_time` and `round`.

anchor/qbft_manager/src/instance.rs
```diff
@@ -181,7 +190,18 @@ impl<D: QbftData<Hash = Hash256>> Initialized<D> {
...
+            _ = self.round_end.tick() => {
+                // Update the timeout
+                let timeout = calculate_round_timeout(
+                    self.qbft.get_message_id().role(),
+                    self.qbft.get_instance_height(),
+                    self.qbft.get_round(),
+                    &self.slot_clock,
+                );
+                self.round_end = tokio::time::interval(timeout);
+                self.round_end.tick().await;
```
dknopik: Calculate and store a new timeout sleep here. Do not immediately tick it, the `select` variant does that.

dknopik (commented):
Thanks! 

I think we can avoid the slot clock, I left some remarks below. Let me know if you have any questions!

anchor/qbft_manager/src/timeout.rs
Zacholme7: We cant here. 

`error[E0015]: cannot call non-const associated function `<Round as From<u64>>::from` in constants
 --> anchor/qbft_manager/src/timeout.rs:9:44
  |
9 | pub const QUICK_TIMEOUT_THRESHOLD: Round = Round::from(8);
  |                                            ^^^^^^^^^^^^^^
  |
  = note: calls in constants are limited to constant functions, tuple structs and tuple variants`

anchor/qbft_manager/src/timeout.rs
Zacholme7: not really sure if I understand this. this is what the go reference code does. https://github.com/ssvlabs/ssv/blob/e38319a33d5e4bcbb79a81863de78f76122ffa97/protocol/v2/qbft/roundtimer/timer.go#L119

anchor/qbft_manager/src/timeout.rs
Zacholme7: Sorry, not sure I understand again. 

So youre saying  calculating timeouts as UNIX timestamps instead of having specific round timeouts?

For example, sleep until 11:00 vs sleep for 2 seconds?

> Instead, add additional_timeout to start_time as passed to the qbft_instance`

Also not sure I understand this. While anchor nodes will be in sync, wont this cause them to go out of sync with go-ssv nodes?

anchor/qbft_manager/src/instance.rs
Zacholme7: > Further down you replace the interval on every tick

Isnt this the point? when the tick future resolves, we know the timeout is up and the round has elapsed so we want to create a new interval dynamically

anchor/qbft_manager/src/timeout.rs
Zacholme7: actually, this seems like a bug in the go impl https://github.com/ssvlabs/ssv/blob/e38319a33d5e4bcbb79a81863de78f76122ffa97/protocol/v2/qbft/roundtimer/timer.go#L89

anchor/qbft_manager/src/instance.rs
dknopik: Citing the Tokio documentation: "[`Interval`](https://docs.rs/tokio/latest/tokio/time/fn.interval.html) is a stream yielding a value at a fixed period. It is initialized with a [`Duration`](https://doc.rust-lang.org/nightly/core/time/struct.Duration.html) and repeatedly yields each time the duration elapses."

It seems to be a tool for repeating ticks at a fixed interval. However, due to the now varying round times, we do not have a fixed interval. Therefore I think it is not appropriate to use anymore. An interval where we wait until the interval has elapsed once is just a sleep.

anchor/qbft_manager/src/timeout.rs
dknopik: The go reference code returns the time of a single round. This is also what you do here. However it is inconsistent with the logic below, as I'll explain in more detail there.

anchor/qbft_manager/src/timeout.rs
dknopik: I think it is preferable to create a `const` constructor for `Round` then.

anchor/qbft_manager/src/timeout.rs
dknopik: Let me give an example.

Let's say we are at the beginning of slot `X`, so we say `t = X`. We are starting a QBFT instance with `Role::Committee`. The timeout of the first round is `slot_duration / 3 + 2s` (which results in `6s` for usual Ethereum slot times). We are right at the slot start, so `time_to_slot_start` is `0`.

In `Uninitialized::initialized`, we create an interval starting at `init.start_time` (which is `X + 4s`) and having (as per the result of the calculation above) a tick time of `6s`. Therefore, the initial tick directly below the interval initialization lets us sleep until `X + 4s`. This is fine as that is the intended start time of the instance. However, as the interval is `6s`, the round will time out `4s` too late, as we are supposed to time out `6s` into the slot, not `6s` after the start time.

I realize I am not really explaining this well :( I took the liberty of pushing a test into your branch that should illustrate the issue better - feel free to rebase it out again. I am also happy to hop into a call as that might be easier. :)

> Also not sure I understand this. While anchor nodes will be in sync, wont this cause them to go out of sync with go-ssv nodes?

Of course, the goal is to align the two client. My point was that we already calculate the offset for the first round timeout with `start_time`, and so there is no need to calculate it again in the new function.

diegomrsantos:
@Zacholme7, can you pass the `slot_clock` only to the function where it is needed?

Zacholme7:
@dknopik the test helped a lot, thx

@diegomrsantos was able to remove slot clock

dknopik (approved):
LGTM, thank you!

anchor/qbft_manager/src/timeout.rs
```diff
@@ -0,0 +1,26 @@
...
+        // If we are below the quick timeout threshold the additonal timeout is round *
+        // QUICK_TIMEOUT
+        Duration::from_secs(round * QUICK_TIMEOUT)
+    } else {
+        // For higher rounds, use a combination of quick and slow timeouts
+
+        // The quick portion is the timeout threshold * QUICK_TIMEOUT
+        let quick_portion = Duration::from_secs(QUICK_TIMEOUT_THRESHOLD * QUICK_TIMEOUT);
+
+        // The slow poritin is (round - threshold) * SLOW_TIMEOUT
```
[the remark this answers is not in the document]
dknopik: We should do a typo-sweep through all the comments anyway at some point

anchor/qbft_manager/src/instance.rs
```diff
@@ -137,6 +135,11 @@ impl Uninitialized {
                 sender,
             },
         ));
+
+        // Calculate inital round sleep time at round 1
```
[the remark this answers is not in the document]
dknopik: We should do a typo-sweep through all the comments anyway at some point

Issue #319, opened 2025-05-14
Better database transaction handling


dknopik:
Currently, we use many small database transactions during sync. This is slow and also risky, as we may end up inserting data but exit before we can update the last updated block. If we switch to a single transaction per block/batch, we benefit from better performance and atomicity of changes.

PR #331, opened 2025-05-19
Adjust DB transactions to be one per log batch

Zacholme7:
## Issue Addressed

#319

## Proposed Changes

Setup database transactions so that we only open/use one transaction per log batch. This makes it so that if processing falls mid batch, our last check pointed block is the previously fully processed batch so we are not failing with an inconsistent view of the database.


anchor/eth/src/event_processor.rs
```diff
@@ -78,35 +93,36 @@ impl EventProcessor {
...
 
                 hash if hash == SSVContract::FeeRecipientAddressUpdated::SIGNATURE_HASH => {
-                    self.process_fee_recipient_updated(log)
+                    self.process_fee_recipient_updated(log, &tx)
                 }
 
                 hash if hash == SSVContract::ValidatorExited::SIGNATURE_HASH => {
-                    self.process_validator_exited(log).await
+                    // Block so we can keep this fn sync
+                    tokio::runtime::Handle::current().block_on(self.process_validator_exited(log))
```
dknopik, line 128: What is the reason for this change?

dknopik (commented):
Very nice, thanks! Just one question, will test this tomorrow.

anchor/eth/src/event_processor.rs
Zacholme7: `rustqlite::Connection` is not `Send` so we need it to be non-async to open a single tx

dknopik (approved):
LGTM, thank you!

PR #337, opened 2025-05-21
Fix signer state update

diegomrsantos:
## Issue Addressed

The `update` function incorrectly passed a `SignerState` reference to the `set` function that cloned it (creating a new instance) and stored it in the `OperatorState`, which incorrectly updated its instance, assuming they were the same.

## Proposed Changes

We now pass ownership to the `set` functions, and they return mutable references, allowing for in-place updates. Key changes include:

- Updating the getter/setter functions in OperatorState to use mutable references.
- Refactoring the test helpers to create and verify consensus messages without duplicating code.

### DIFF
--- a/anchor/message_validator/src/consensus_state.rs
+++ b/anchor/message_validator/src/consensus_state.rs
@@ -265,8 +265,10 @@ impl SignerState {
                 .insert(signed_ssv_message.operator_ids().as_slice().into());
         }
 
-        self.message_counts
-            .record_consensus_message(consensus_message.qbft_message_type);
+        self.message_counts.record_consensus_message(
+            consensus_message.qbft_message_type,
+            signed_ssv_message.operator_ids().len(),
+        );
     }
 }
 
--- a/anchor/message_validator/src/consensus_state.rs
+++ b/anchor/message_validator/src/consensus_state.rs
@@ -131,8 +131,12 @@ impl OperatorState {
             .filter(|s| s.slot == *slot)
     }
 
-    /// Sets the signer state in the circular buffer at the computed index.
-    fn set_signer_state(&mut self, slot: &Slot, signer_state: SignerState) -> &mut SignerState {
+    /// Sets the signer state for a round change in the circular buffer at the computed index.
+    fn set_signer_state_for_round_change(
+        &mut self,
+        slot: &Slot,
+        signer_state: SignerState,
+    ) -> &mut SignerState {
         let index = slot.as_usize() % self.state.len();
         self.state[index] = Some(signer_state);
         self.state[index].as_mut().unwrap()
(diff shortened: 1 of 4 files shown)


anchor/message_validator/src/consensus_state.rs
```diff
@@ -118,21 +118,23 @@ impl OperatorState {
...
-    pub(crate) fn get_signer_state(&self, slot: &Slot) -> Option<SignerState> {
-        match &self.state[slot.as_usize() % self.state.len()] {
-            Some(s) if s.slot == *slot => Some(s.clone()),
+    pub(crate) fn get_signer_state(&mut self, slot: &Slot) -> Option<&mut SignerState> {
+        let len = self.state.len();
+        match self.state[slot.as_usize() % len].as_mut() {
+            Some(s) if s.slot == *slot => Some(s),
             _ => None,
         }
     }
```
dknopik: Use `Option::filter` instead.

anchor/message_validator/src/consensus_state.rs
```diff
@@ -118,21 +118,23 @@ impl OperatorState {
...
-        self.state[index] = Some(signer_state.clone());
+        self.state[index] = Some(signer_state);
+        self.state[index].as_mut().unwrap()
     }
 
     /// Returns true if we have not seen a message for a duty in `slot` yet.
-    pub(crate) fn is_first_message_for_duty(&self, slot: Slot) -> bool {
+    pub(crate) fn is_first_message_for_duty(&mut self, slot: Slot) -> bool {
         self.get_signer_state(&slot).is_none()
     }
```
dknopik: This should not need a `&mut self`. I think it would be nicer to have both `get_signer_state` and `get_signer_state_mut`.

dknopik (commented):
Great find! Two nitpicks, otherwise LGTM. Will put this on interop overnight to make sure that the now working validation is not too eager (e.g. due to another bug).

anchor/message_validator/src/consensus_state.rs
diegomrsantos: Done.

anchor/message_validator/src/consensus_state.rs
diegomrsantos: Done. Very good suggestions, thanks a lot

dknopik:
@diegomrsantos Fixing this uncovered an issue with the underlying validation.

The validation only allows one `Commit` message per QBFT round. However, `Commit` messages with more than one signer (also known as `Decided` messages) are not counted: https://github.com/ssvlabs/ssv/blob/4ab561d97c008ca33b57070c7f8de661c5038c15/message/validation/message_counts.go#L103-L105

We incorrectly count those: https://github.com/sigp/anchor/blob/f1ac1e11ac2eea29a71c9be0ef7174647570406a/anchor/message_validator/src/message_counts.rs#L63-L71

We should fix that before merging this. I have no preference between doing it in this PR or a separate PR.

diegomrsantos:
@dknopik Applied a fix and a test. Great finding too!

dknopik (approved):
LGTM, thanks!

Feel free to hit merge if this is good to go.

PR #339, opened 2025-05-23
Avoid async in event processor

dknopik:
## Issue Addressed

The approach of calling the `process_validator_exited` function with `block_on` does not work, as that panics:

```
thread 'tokio-runtime-worker' panicked at anchor/eth/src/event_processor.rs:126:55:
Cannot start a runtime from within a runtime. This happens because a function (like `block_on`) attempted to block the current thread while the thread is being used to drive asynchronous tasks.
```

## Proposed Changes

Instead, during live sync, check whether the logs need any block timestamp from the EL **before** calling `process_logs`.

Also, process exits only during live sync, as they are already expired when history syncing.

### DIFF
--- a/anchor/eth/src/event_processor.rs
+++ b/anchor/eth/src/event_processor.rs
@@ -1,12 +1,6 @@
 use std::sync::Arc;
 
-use alloy::{
-    eips::BlockNumberOrTag,
-    primitives::Address,
-    providers::{Provider, RootProvider},
-    rpc::types::Log,
-    sol_types::SolEvent,
-};
+use alloy::{primitives::Address, rpc::types::Log, sol_types::SolEvent};
 use database::{NetworkDatabase, UniqueIndex};
 use eth2::types::PublicKeyBytes;
 use indexmap::IndexSet;
@@ -47,18 +41,12 @@ pub struct EventProcessor {
     pub db: Arc<NetworkDatabase>,
     /// Signal if we should only do relevant keysplitting processing
     mode: Mode,
-    /// RPC client for fetching data from the network
-    rpc_client: Arc<RootProvider>,
 }
 
 impl EventProcessor {
     /// Construct a new EventProcessor
-    pub fn new(db: Arc<NetworkDatabase>, mode: Mode, rpc_client: Arc<RootProvider>) -> Self {
-        Self {
-            db,
-            mode,
-            rpc_client,
-        }
+    pub fn new(db: Arc<NetworkDatabase>, mode: Mode) -> Self {
+        Self { db, mode }
     }
 
     /// Process a new set of logs
@@ -95,37 +83,34 @@ impl EventProcessor {
 
             // Process log based on signature hash
             let result = match *topic0 {
-                hash if hash == SSVContract::OperatorAdded::SIGNATURE_HASH => {
-                    self.process_operator_added(log, &tx)
-                }
+                SSVContract::OperatorAdded::SIGNATURE_HASH => self.process_operator_added(log, &tx),
 
-                hash if hash == SSVContract::OperatorRemoved::SIGNATURE_HASH => {
+                SSVContract::OperatorRemoved::SIGNATURE_HASH => {
                     self.process_operator_removed(log, &tx)
                 }
 
-                hash if hash == SSVContract::ValidatorAdded::SIGNATURE_HASH => {
+                SSVContract::ValidatorAdded::SIGNATURE_HASH => {
                     self.process_validator_added(log, &tx)
                 }
 
-                hash if hash == SSVContract::ValidatorRemoved::SIGNATURE_HASH => {
+                SSVContract::ValidatorRemoved::SIGNATURE_HASH => {
                     self.process_validator_removed(log, &tx)
                 }
 
-                hash if hash == SSVContract::ClusterLiquidated::SIGNATURE_HASH => {
+                SSVContract::ClusterLiquidated::SIGNATURE_HASH => {
                     self.process_cluster_liquidated(log, &tx)
                 }
 
-                hash if hash == SSVContract::ClusterReactivated::SIGNATURE_HASH => {
+                SSVContract::ClusterReactivated::SIGNATURE_HASH => {
                     self.process_cluster_reactivated(log, &tx)
                 }
 
-                hash if hash == SSVContract::FeeRecipientAddressUpdated::SIGNATURE_HASH => {
+                SSVContract::FeeRecipientAddressUpdated::SIGNATURE_HASH => {
                     self.process_fee_recipient_updated(log, &tx)
                 }
 
-                hash if hash == SSVContract::ValidatorExited::SIGNATURE_HASH => {
-                    // Block so we can keep this fn sync
-                    tokio::runtime::Handle::current().block_on(self.process_validator_exited(log))
+                SSVContract::ValidatorExited::SIGNATURE_HASH if live => {
+                    self.process_validator_exited(log)
                 }
                 _ => {
                     debug!(?topic0, "Unknown event signature, skipping");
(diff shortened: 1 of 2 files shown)


anchor/eth/src/event_processor.rs
```diff
@@ -95,37 +83,34 @@ impl EventProcessor {
...
-                hash if hash == SSVContract::FeeRecipientAddressUpdated::SIGNATURE_HASH => {
+                SSVContract::FeeRecipientAddressUpdated::SIGNATURE_HASH => {
                     self.process_fee_recipient_updated(log, &tx)
                 }
 
-                hash if hash == SSVContract::ValidatorExited::SIGNATURE_HASH => {
-                    // Block so we can keep this fn sync
-                    tokio::runtime::Handle::current().block_on(self.process_validator_exited(log))
+                SSVContract::ValidatorExited::SIGNATURE_HASH if live => {
+                    self.process_validator_exited(log)
```
diegomrsantos, line 113: What was the reason for adding the previous code?

anchor/eth/src/sync.rs
```diff
@@ -562,6 +562,53 @@ impl SsvEventSyncer {
...
+            }
+
+            let block_number = log.block_number.ok_or_else(|| {
+                ExecutionError::InvalidEvent("Block number not available".to_string())
+            })?;
+
+            if let Some(timestamp) = timestamps.get(&block_number) {
+                log.block_timestamp = Some(*timestamp);
+            } else {
+                debug!("Block timestamp not available");
```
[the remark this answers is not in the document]
diegomrsantos: I believe this can be trace

anchor/eth/src/sync.rs
```diff
@@ -562,6 +562,53 @@ impl SsvEventSyncer {
...
+
+                let block = match self
+                    .rpc_client
+                    .get_block_by_number(BlockNumberOrTag::from(block_number))
+                    .await
+                {
+                    Ok(Some(block)) => {
+                        debug!(?block, "Fetched block");
```
diegomrsantos: also can be trace

anchor/eth/src/sync.rs
```diff
@@ -620,13 +667,17 @@ impl SsvEventSyncer {
                     )
                     .await?;
 
-                info!(
-                    log_count = logs.len(),
-                    "Processing events from block {}", relevant_block
-                );
+                self.preprocess_exits(&mut logs).await?;
```
diegomrsantos: I believe `preprocess_exits` isn't an accurate name for this function. Also, calling this to mutate the log feels a bit hacky.

anchor/eth/src/event_processor.rs
```diff
@@ -95,37 +83,34 @@ impl EventProcessor {
 
             // Process log based on signature hash
             let result = match *topic0 {
```
Zacholme7, line 85: can we go back to the lookup table now? https://github.com/sigp/anchor/blob/80ed2e01186b4840c4b8638aa171f6c23e5a282f/anchor/eth/src/event_processor.rs#L48

anchor/eth/src/event_processor.rs
Zacholme7: Getting the db transaction uses a `RefCell` internally which is not `Send`. The only way around this was to make sure this function was was not async. There is also a lot of logic that goes on here so I think it is nice to be able to keep it sync instead of going to async for just one rpc call.

anchor/eth/src/sync.rs
```diff
@@ -562,6 +562,53 @@ impl SsvEventSyncer {
...
+                || log.block_timestamp.is_some()
+            {
+                continue;
+            }
+
+            let block_number = log.block_number.ok_or_else(|| {
+                ExecutionError::InvalidEvent("Block number not available".to_string())
+            })?;
+
+            if let Some(timestamp) = timestamps.get(&block_number) {
```
Zacholme7: we are in live sync so shouldnt all of these logs have the same block number? Cant we just fetch the timestamp once and apply it to all logs missing it

Zacholme7 (commented):
This is my fault, sorry! Thx for the fix

anchor/eth/src/event_processor.rs
Zacholme7: Getting the db tx above uses a `RefCell` internally which is not `Send`. The only way around this was to make the fuction sync. There is also a lot of logic that goes on in this function and I think it it nice to be able to keep it async free as there is just one rpc call that does on.

anchor/eth/src/sync.rs
dknopik: I considered that also, but then we have the (implicit, documentation-only) prerequisite that all logs need to be from the same block. If we change something in the future (e.g. not falling back to historical sync unless we have been offline for a while), the approach you suggested will cause trouble, so I implemented it this way to catch this case.

anchor/eth/src/event_processor.rs
dknopik: Personally, I still prefer the match, especially as one variant has the special case of `if live`. @diegomrsantos what is your opinion?

anchor/eth/src/sync.rs
dknopik: I like mutating the logs as it is basically a step to account for EL clients not replying with the timestamp. After this step, the logs are in a correct state regardless of client support of this feature.

Yeah, I'll think about the name again. any suggestions?

diegomrsantos: My reasoning is that I wouldn't expect this type to be modified by our code as we aren't the one managing it.

diegomrsantos: I'm not even sure it should be mutable.

anchor/eth/src/event_processor.rs
diegomrsantos: Imo the match is more idiomatic and simpler

anchor/eth/src/sync.rs
diegomrsantos: We could move the if that checks if it's an exit outside the function, get the timestamp and call the process exit function directly passing the timestamp as a param. Wdyt?

diegomrsantos: But tbh I don't have a strong opinion. I think we could also rename it to sth like get_block_timestamp, set it to the log and call the process_logs if you like it more.

dknopik: Sorry, I do not get what exactly you now mean. Please push your suggestion to this branch

diegomrsantos: I'm fine with the current implementation. I'd only rename this function to sth like `set_block_timestamp`

anchor/eth/src/event_processor.rs
diegomrsantos: What about a Mutex or RwLock?

diegomrsantos: Or maybe sth like https://docs.rs/crossbeam/latest/crossbeam/atomic/struct.AtomicCell.html

anchor/eth/src/sync.rs
```diff
@@ -562,6 +562,53 @@ impl SsvEventSyncer {
         Ok(result)
     }
 
+    /// Exit logs need the block timestamps set. Ensure every exit in a batch of logs has a block
+    /// timestamp set, fetching it from the EL if needed.
+    async fn preprocess_exits(&mut self, logs: &mut [Log]) -> Result<(), ExecutionError> {
+        let mut timestamps = HashMap::new();
```
diegomrsantos: How about `block_timestamp_cache`?

dknopik: done

anchor/eth/src/sync.rs
dknopik: done, but used plural: `set_block_timestamps`

anchor/eth/src/event_processor.rs
diegomrsantos: @dknopik wdyt?

dknopik: > What about a Mutex or RwLock?

It kind of feels bad to use that in a context where we are not actually accessing the value concurrently :sweat_smile: 

> AtomicCell

Is kind of awkward as well as we have to take the inner value (can't get a reference) but might work with `into_inner`.

I also looked into primitives such as https://doc.rust-lang.org/std/sync/struct.Exclusive.html but it might not be the correct tool for the job (and it is unstable).

Maybe we can explore this later on.

diegomrsantos: Yeah makes sense, thanks for looking into it

diegomrsantos (approved):

PR #341, opened 2025-05-23
Fix cargo sort

dknopik:
- Pin version
- Apply new table order
- Remove taplo
- Add `--grouped` arg

## Additional Info

It now properly handles the workspace toml, so we can remove taplo! :partying_face: 

Personally, I am fine with the new order of `[[bin]]`, `features`, and `dependencies`, wdyt?

Make sure to update your local `cargo sort` with `cargo install cargo-sort`!


Zacholme7 (approved):
Nice!

PR #343, opened 2025-05-26
Properly record `validator_store` errors in metrics

### LINKED ISSUE #242: Better result reporting for metrics used in the validator store
Currently, we only record successful signatures in the validator store metrics. E.g.:

```rust
        let signature = self
            .timeout_within_slot(
                slot,
                delay,
                self.collect_signature(
                    PartialSignatureKind::SelectionProofPartialSig,
                    Role::Aggregator,
                    None,
                    self.validator(validator_pubkey)?,
                    signing_root,
                    slot,
                ),
            )
            .await?;

        validator_metrics::inc_counter_vec(
            &validator_metrics::SIGNED_SELECTION_PROOFS_TOTAL,
            &[validator_metrics::SUCCESS],
        );

        Ok(signature.into())
```

If we return early via `?` due to an error (e.g. timeout), we do not record that. 

Add some function that we can pass a metric name and result to that adds an accurate metric value for that result.

dknopik:
## Issue Addressed

- #242 

## Proposed Changes

Wrap all major logic in futures whose errors are then centrally treated in a new function

### DIFF
--- a/anchor/validator_store/src/lib.rs
+++ b/anchor/validator_store/src/lib.rs
@@ -78,6 +78,16 @@ use validator_store::{
 /// This acts as a maximum safe-guard against clock drift.
 const SLASHING_PROTECTION_HISTORY_EPOCHS: u64 = 512;
 
+const RANDAO_REVEAL_LOG_NAME: &str = "RANDAO reveal";
+const BLOCK_LOG_NAME: &str = "block";
+const ATTESTATION_LOG_NAME: &str = "attestation";
+const VALIDATOR_REGISTRATION_LOG_NAME: &str = "validator registration";
+const AGGREGATE_LOG_NAME: &str = "aggregate";
+const SELECTION_PROOF_LOG_NAME: &str = "selection proof";
+const SYNC_SELECTION_PROOF_LOG_NAME: &str = "sync selection proof";
+const SYNC_COMMITTEE_SIGNATURE_LOG_NAME: &str = "sync committee signature";
+const SYNC_COMMITTEE_CONTRIBUTION_LOG_NAME: &str = "sync committee contribution";
+
 #[derive(Clone)]
 struct InitializedValidator {
     cluster: Arc<Cluster>,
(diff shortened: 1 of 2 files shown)


anchor/validator_store/src/lib.rs
```diff
@@ -802,13 +809,18 @@ impl<T: SlotClock, E: EthSpec> ValidatorStore for AnchorValidatorStore<T, E> {
...
-        self.collect_signature(
-            PartialSignatureKind::RandaoPartialSig,
-            Role::Proposer,
-            None,
-            self.validator(validator_pubkey)?,
-            signing_root,
-            self.slot_clock.now().ok_or(SpecificError::SlotClock)?,
+
+        handle_result(
+            "RANDAO reveal",
```
diegomrsantos: We could extract those strings to a const

anchor/validator_store/src/lib.rs
```diff
@@ -615,37 +610,41 @@ impl<T: SlotClock, E: EthSpec> AnchorValidatorStore<T, E> {
     }
 }
 
-fn handle_slashing_check_result(
-    slashing_status: Result<Safe, NotSafe>,
-    object: impl Debug,
+async fn handle_result<T>(
```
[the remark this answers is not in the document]
diegomrsantos: maybe `run_and_update_metrics`

anchor/validator_store/src/lib.rs
```diff
@@ -722,6 +721,14 @@ impl From<QbftError> for SpecificError {
     }
 }
 
+fn handle_slashing_check_result(value: Result<Safe, NotSafe>) -> Result<(), Error> {
```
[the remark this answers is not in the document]
dknopik: ok

anchor/validator_store/src/lib.rs
dknopik: personally, I do not tend to like consts that are both used only in a single place and just like `const THING_NAME: &str = "thing";` - they do not really add value IMO. Pushed a suggestion regardless

diegomrsantos: Yeah, I agree.

diegomrsantos (approved):
LGTM, thanks

PR #344, opened 2025-05-26
Extend notifier with Anchor-specific logic

### LINKED ISSUE #251: Add basic notification routine
Lighthouse's VC has a basic user notification routine: https://github.com/sigp/lighthouse/blob/61a932b280a41651e34b54468e5f42be2cc44ecf/validator_client/src/notifier.rs#L36

This logs some basic info and updates some metrics once per slot. Almost everything there applies to us as well, so we should consider to reuse it (possibly slightly generalising it), and maybe add some Anchor specific logging/metrics.

We need to move it to use it, as we should not import the `validator_client` crate.

See also: #250

- [x] reuse existing
- [ ] extend

dknopik:
## Issue Addressed

- closes #251 

## Proposed Changes

Instead of calling the notifier service directly, we create our own service to add a bit of own logic to it - this avoids a LH specific message and logs some Anchor specific data such as the operator ID and amount of clusters we're in.

### DIFF
--- a/anchor/client/src/notifier.rs
+++ b/anchor/client/src/notifier.rs
@@ -1,0 +1,67 @@
+use std::sync::Arc;
+
+use anchor_validator_store::AnchorValidatorStore;
+use database::NetworkState;
+use slot_clock::SlotClock;
+use task_executor::TaskExecutor;
+use tokio::{
+    sync::watch,
+    time::{Duration, sleep},
+};
+use tracing::{error, info};
+use types::{ChainSpec, EthSpec};
+
+use crate::duties_service::DutiesService;
+
+/// Spawns a notifier service which periodically logs information about the node. It reuses LH's
+/// `notifier_service` for most functionality but adds some Anchor-specific information and logic.
+pub fn spawn_notifier<E: EthSpec, T: SlotClock + 'static>(
+    duties_service: Arc<DutiesService<AnchorValidatorStore<T, E>, T>>,
+    network_state: watch::Receiver<NetworkState>,
+    executor: TaskExecutor,
+    spec: &ChainSpec,
+) -> Result<(), String> {
+    let slot_duration = Duration::from_secs(spec.seconds_per_slot);
+
+    let interval_fut = async move {
+        loop {
+            if let Some(duration_to_next_slot) = duties_service.slot_clock.duration_to_next_slot() {
+                sleep(duration_to_next_slot + slot_duration / 2).await;
+                notify(&duties_service, &network_state).await;
+            } else {
+                error!("Failed to read slot clock");
+                // If we can't read the slot clock, just wait another slot.
+                sleep(slot_duration).await;
+                continue;
+            }
+        }
+    };
+
+    executor.spawn(interval_fut, "validator_notifier");
+    Ok(())
+}
+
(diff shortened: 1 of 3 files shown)


anchor/client/src/notifier.rs
```diff
@@ -0,0 +1,67 @@
...
+            }
+        }
+    };
+
+    executor.spawn(interval_fut, "validator_notifier");
+    Ok(())
+}
+
+/// Performs a single notification routine.
+async fn notify<E: EthSpec, T: SlotClock + 'static>(
```
diegomrsantos, line 51: Could you explain more about the objective of this function? Is it updating metrics? If so, I'm not sure if `notify` is the best name.

anchor/client/src/notifier.rs
```diff
@@ -0,0 +1,67 @@
...
+    network_state: watch::Receiver<NetworkState>,
+    executor: TaskExecutor,
+    spec: &ChainSpec,
+) -> Result<(), String> {
+    let slot_duration = Duration::from_secs(spec.seconds_per_slot);
+
+    let interval_fut = async move {
+        loop {
+            if let Some(duration_to_next_slot) = duties_service.slot_clock.duration_to_next_slot() {
+                sleep(duration_to_next_slot + slot_duration / 2).await;
```
diegomrsantos, line 33: Could you add a comment explaining the reasoning for this value?

anchor/client/src/notifier.rs
```diff
@@ -0,0 +1,67 @@
...
+    let (operator_id, clusters) = {
+        let state = network_state.borrow();
+        let operator_id = state.get_own_id();
+        let clusters = state.get_own_clusters().len();
+        (operator_id, clusters)
+    };
+
+    if let Some(operator_id) = operator_id {
+        if duties_service.total_validator_count() > 0 {
+            info!(%operator_id, clusters, "Operator active");
```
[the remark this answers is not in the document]
diegomrsantos: Additionally, do we want to log this as `info` that often?

anchor/client/src/notifier.rs
dknopik: Added a comment

anchor/client/src/notifier.rs
dknopik: done

anchor/client/src/notifier.rs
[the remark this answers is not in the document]
dknopik: Yeah. Compare e.g. with the logging emitted by the beacon node or validator client. Both have a regular logging routine to show some key data to the user (and, I guess, reassure them that we are in fact doing something). There is also the added benefit of logging the operator id, which also ends up in the logfile, allowing us to cross-reference logs when analysing some issue using logs from multiple nodes.

jking-aus (approved):

anchor/client/src/notifier.rs
diegomrsantos: It might be a language thing, but I believe "user" is usually used for an actual person. A "client" might be more idiomatic for this domain. Wdyt?

dknopik: In this case, I also mean an actual person :)

Issue #348, opened 2025-05-27
Disable mainnet network config


dknopik:
Prevent the user from running Anchor on Mainnet. This includes disabling `--network mainnet` and disallowing pointing `--testnet-dir` to a directory actually describing mainnet.

PR #350, opened 2025-05-29
Disable mainnet support

diegomrsantos:
## Issue Addressed

https://github.com/sigp/anchor/issues/348

## Proposed Changes

This PR disables mainnet support across the codebase by removing mainnet-specific branches from network matching and updating associated CLI options, configuration checks, and documentation.

- Removed mainnet handling in the onchain split logic.
- Updated CLI help messages and enum definitions to only list supported networks (Holesky and Hoodi).
- Added a configuration check in the client to error out if mainnet is attempted and revised documentation accordingly.

### DIFF
--- a/anchor/client/src/config.rs
+++ b/anchor/client/src/config.rs
@@ -140,14 +140,6 @@ pub fn from_cli(cli_args: &Node) -> Result<Config, String> {
             .and_then(|net| net.ok_or_else(|| format!("Unknown network {}", cli_args.network)))
     }?;
 
-    if let Some(network) = eth2_network.clone().eth2_network.config.config_name {
-        if network == "mainnet" {
-            return Err(
-                "Mainnet is not supported. Please use a testnet configuration.".to_string(),
-            );
-        }
-    }
-
     let mut config = Config::new(eth2_network);
 
     if let Some(datadir) = cli_args.datadir.clone() {
--- a/anchor/client/src/config.rs
+++ b/anchor/client/src/config.rs
@@ -140,6 +140,14 @@ pub fn from_cli(cli_args: &Node) -> Result<Config, String> {
             .and_then(|net| net.ok_or_else(|| format!("Unknown network {}", cli_args.network)))
     }?;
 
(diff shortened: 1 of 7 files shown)


anchor/client/src/config.rs
```diff
@@ -140,6 +140,14 @@ pub fn from_cli(cli_args: &Node) -> Result<Config, String> {
             .and_then(|net| net.ok_or_else(|| format!("Unknown network {}", cli_args.network)))
     }?;
 
+    if let Some(network) = eth2_network.clone().eth2_network.config.config_name {
+        if network == "mainnet" {
+            return Err(
+                "Mainnet is not supported. Please use a testnet configuration.".to_string(),
+            );
+        }
+    }
```
dknopik: This can be circumvented by naming the network e.g. `definitely_not_mainnet`. The name is afaik not used in any checks and is for human reference only.

I recommend checking the `genesis_fork_version`, which is `[0, 0, 0, 0]` for mainnet. This is e.g. also used by `beacon_node_fallback` to check if the connected beacon node is on the correct network: https://github.com/sigp/lighthouse/blob/999b04517e35f8fce78525132d0b8e0a4816f31e/validator_client/beacon_node_fallback/src/lib.rs#L317-L325

dknopik:
Other than that, LGTM! thanks for handling this :)

anchor/client/src/config.rs
diegomrsantos: Makes sense. Does the user supply this? Do you know how I can check it here?

dknopik: At this point the `genesis_fork_version` might be user supplied if we hit `load` above. 

I recommend something like 

```rust
const MAINNET_GENESIS_FORK_VERSION: [u8; 4] = [0, 0, 0, 0];
if eth2_network.eth2_network.config.genesis_fork_version == MAINNET_GENESIS_FORK_VERSION {
...
```

Zacholme7 (approved):
lgtm

PR #351, opened 2025-05-29
fix(eth): abort batches on internal event processing failures

### LINKED ISSUE #317: Remove duplicate logging
In several places in the code, we both log and return an error. This will likely cause duplicated logs. We should either log or return an error. The returned error should preserve any existing error chain and capture the current context. I believe the latter can be done with the anyhow library.

diegomrsantos:

- Internal failures in EL event handling were being logged like malformed events and then skipped, which could let Anchor commit progress even when the failure was in Anchor itself rather than in the chain data.
- Duplicate logging: several places both logged and returned an error, causing duplicated log entries.
- Evidence: [#317](https://github.com/sigp/anchor/issues/317) identified duplicate logging and weak error propagation.
- Closes #317

## Change Overview

- Switched `ExecutionError` to `thiserror` and preserved typed decode/operator conversion errors instead of flattening them to strings.
- Introduced `LogErrorDisposition` to categorize event-processing errors into three buckets:
  - **SkipMalformed**: decode failures, invalid events, duplicates — logged and skipped.
  - **SkipExpected**: known no-ops — logged at lower level and skipped.
  - **AbortBatch**: internal/dependency failures (database, sync, RPC) — block processing aborted.
- Removed duplicate lower-level logging so errors are emitted once at the boundary that decides whether to continue or abort.
- Returned typed `operator_key::ConversionError` from `Operator::new` instead of `String`.
- Intentionally did not change: historical `ValidatorExited` behavior, the external API of the `eth` crate, or the broader stringly-typed error surface outside the touched paths.

- Main risk: internal failures that were previously logged-and-skipped will now abort the current block and force retry/recovery behavior.
- This favors correctness and transactional consistency over best-effort forward progress.
- Mitigations: only internal/dependency failures are block-fatal; malformed, duplicate, and expected-skip events preserve the old continue-processing behavior.

## Validation

- `cargo test -p eth --tests` — all 11 integration tests pass
- `cargo clippy -p eth --all-targets -- -D warnings` — clean
- `cargo clippy -p ssv_types -- -D warnings` — clean

## Rollback

- Safe to revert as a normal code change.
- No schema, config, or data migration impact.
- Reverting restores the previous behavior where internal failures could be logged and skipped, and restores duplicate lower-level logging in the touched paths.

- Converting more `ExecutionError` variants from string payloads to typed source errors is a natural follow-up.
- The `ValidatorExited` replay gap (historical sync silently drops exit side effects after a live-sync failure) is a known pre-existing issue and should be tracked in a separate issue.


anchor/eth/src/error.rs
```diff
@@ -1,20 +1,121 @@
...
 pub enum ExecutionError {
+    #[error("Sync error: {0}")]
     SyncError(String),
-    InvalidEvent(String),
+
+    #[error("Invalid event for operator {operator_id:?} (owner: {owner:?}): {message}")]
+    InvalidOperatorEvent {
+        operator_id: ssv_types::OperatorId,
+        owner: alloy::primitives::Address,
+        message: String,
```
dknopik: Having a `message: String` field in here feels like an anti pattern. I'd rather have a non `Option`al sub error for context

anchor/eth/src/error.rs
```diff
@@ -1,20 +1,121 @@
...
+        #[source]
+        source: Option<Box<dyn std::error::Error + Send + Sync>>,
+    },
+
+    #[error("Invalid event: {message}")]
+    InvalidEvent {
+        message: String,
+        #[source]
+        source: Option<Box<dyn std::error::Error + Send + Sync>>,
+    },
```
dknopik: Instead of a variant per event type, maybe embed the event into this variant?

anchor/eth/src/error.rs
```diff
@@ -1,20 +1,121 @@
...
     SyncError(String),
-    InvalidEvent(String),
+
+    #[error("Invalid event for operator {operator_id:?} (owner: {owner:?}): {message}")]
+    InvalidOperatorEvent {
+        operator_id: ssv_types::OperatorId,
+        owner: alloy::primitives::Address,
+        message: String,
+        #[source]
+        source: Option<Box<dyn std::error::Error + Send + Sync>>,
```
dknopik: This is kind of emulating `anyhow`'s `context`, right?

dknopik (commented):
Thanks for looking into this! I added some early thoughts

anchor/eth/src/error.rs
diegomrsantos: I think anyhow's context is a message describing what we were trying to do when the function called failed. The `source` is the error that the function called returned.

Zacholme7:
We still planning to keep this or is it closable?

diegomrsantos:
> We still planning to keep this or is it closable?

I wanna do it at some point

anchor/eth/src/error.rs
diegomrsantos: Addressed in the rewrite — the new approach keeps simple `(String)` tuple variants without `message: String` struct fields. The only typed source error uses `#[from]` on `DecodeError` for `alloy::sol_types::Error`, which preserves the error chain without the anti-pattern.

anchor/eth/src/error.rs
diegomrsantos: Addressed — the rewrite keeps a single `InvalidEvent(String)` variant instead of splitting into per-event-type variants (`InvalidOperatorEvent`, `InvalidValidatorEvent`, etc.).

anchor/eth/src/error.rs
diegomrsantos: Addressed — the `#[source] source: Option<Box<dyn Error>>` pattern is gone. We now only use `#[from]` where there's a real typed source error (`DecodeError` wrapping `alloy::sol_types::Error`). No anyhow-style context emulation.

anchor/eth/src/event_processor.rs
```diff
@@ -659,16 +596,9 @@ impl EventProcessor {
                 );
             }
             Err(err) => {
-                // If the channel is closed, we can't send the exit request
```
shane-moore: consider resurrecting this `error!` since just returning `Err` propogates up to a `warn` path, and this failure case is a bit more severe

shane-moore:
maybe create gh issue for this?
`Converting more variants (e.g. Database) from String to typed source errors is a natural follow-up but left out to keep this PR focused`

anchor/eth/src/event_processor.rs
diegomrsantos: Now it's gonna be logged as an error

shane-moore (dismissed):

diegomrsantos:
@dknopik @shane-moore

This change correctly fixes one problem: before this PR, `process_logs()` treated all per-log failures as "malformed event, log and continue," which meant even internal failures on valid chain events could be skipped while the batch still committed and `last_processed_block` still advanced. The new `LogErrorDisposition` split is a real improvement because internal failures now abort the batch instead of being silently swallowed.

However, there is still an architectural edge case around `ValidatorExited` that is worth calling out explicitly. `ValidatorExited` is not handled like ordinary replayable DB state. During live sync, Anchor processes logs with `live = true` in `sync.rs`, and `process_validator_exited()` turns the event into a live side effect by enqueueing work to the exit processor. During historical sync, the same code path is called with `live = false`, and `process_validator_exited()` explicitly ignores the event in `event_processor.rs`.

That distinction matters because Anchor can move from live back to historical after a failure. `sync()` loops forever; if live processing returns an error, it restarts by calling `try_sync()` again. `try_sync()` always begins with historical sync from `last_processed_block + 1`. So if a live batch aborts for any reason and that batch included a valid `ValidatorExited`, the retry path can re-read the same block range historically, where the exit is now treated as a no-op instead of an actionable event.

I’m not saying this PR is wrong for preserving the old historical-exit behavior, and I’m not claiming the new `ExitProcessorUnavailable` test proves a production bug. The concern is narrower: exit handling is mode-dependent, so replay after a live failure is not semantically equivalent to the original live pass. Regular DB-backed events are reconstructed by historical replay; `ValidatorExited` is not. If that is intentional, I think it should be documented or tested explicitly, because otherwise it is very easy to assume that retrying the same blocks fully reconstructs behavior when it actually only reconstructs durable state.

diegomrsantos:
@dknopik @shane-moore Tracked the ValidatorExited replay gap in #911. The rest of this PR is now rebased on unstable and ready for re-review.

diegomrsantos:
@claude review this PR

diegomrsantos:
@shane-moore I pushed a narrow follow-up here.

Current state:
- `#351` still aborts the batch for clearly internal failures (`Database`, sync/RPC/WS errors).
- It now preserves the old skip-and-continue behavior for the ambiguous missing-state paths that Claude called out:
  - missing operator in `validate_operators()`
  - missing validator metadata / cluster in `process_validator_removed()`
- I handled that with a separate `MissingCommittedState -> SkipAmbiguous` path so this PR does not force the `InvalidEvent` vs `Database` decision.

`#930` stays open to decide the semantics of those ambiguous cases. I intentionally left `get_validator_index()` alone in this PR because it already remains non-fatal under the new disposition split.

diegomrsantos:
@claude review it again

diegomrsantos:
On current head `01ad2baa`, I do not think the remaining two findings are blockers.

- `get_validator_index()` is only reached after `verify_validator_owner()` succeeds. `verify_validator_owner()` loads the cluster via `get_cluster_by_validator_tx()`, and that query joins through `validators`, so if it succeeds the validator row already exists. In the current flow, the `None` branch in `get_validator_index()` is therefore defensive rather than a real ambiguous-state path.
- `verify_validator_owner()` returning `InvalidEvent` for a missing cluster is a consistency nit at most. Under `#351` that path is already non-fatal either way, so changing it to `MissingCommittedState` would only affect classification/log wording, not behavior.

So I do not see a remaining merge blocker here.

@claude review it again

shane-moore (approved):
great job working through the last few findings today. agree with your comments here as well https://github.com/sigp/anchor/pull/351#issuecomment-4201582535. No nits to add at this point, lgtm!

PR #361, opened 2025-06-05
Avoid unnecessary allocations and conversions in database

dknopik:
There are a bunch of allocations and conversion in the database that can be avoided.

### DIFF
--- a/anchor/eth/src/event_processor.rs
+++ b/anchor/eth/src/event_processor.rs
@@ -167,33 +167,25 @@ impl EventProcessor {
             )));
         }
 
-        // Parse ABI encoded public key string and trim off 0x prefix for hex decoding
-        let public_key_str = publicKey.to_string();
-        let public_key_str = public_key_str.trim_start_matches("0x");
-        let data = hex::decode(public_key_str).map_err(|e| {
-            debug!(operator_id = ?operator_id, error = %e, "Failed to decode public key data from hex");
-            ExecutionError::InvalidEvent(
-                format!("Failed to decode public key data from hex: {e}")
-            )
-        })?;
+        let data = publicKey.as_ref();
 
         // If the data is 704 bytes, remove the ssv encoding. Else, just parse the key
         let data = if data.len() == 704 {
             let data = &data[64..];
-            let data = String::from_utf8(data.to_vec()).map_err(|e| {
+            let data = str::from_utf8(data).map_err(|e| {
                 debug!(operator_id = ?operator_id, error = %e, "Failed to convert to UTF8 String");
                 ExecutionError::InvalidEvent(format!("Failed to convert to UTF8 String: {e}"))
             })?;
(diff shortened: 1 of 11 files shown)


dknopik:
Regarding copilot's low confidence comment: It turns out the hex encoding was added by "to_string", so we added the encoding and immediately removed it again. Using `as_ref` uses the raw data directly.

anchor/eth/src/event_processor.rs
```diff
@@ -167,33 +167,22 @@ impl EventProcessor {
...
+            while let [rest @ .., 0] = data {
+                data = rest;
+            }
+            data
         } else {
-            String::from_utf8(data.to_vec()).map_err(|e| {
-                debug!(operator_id = ?operator_id, error = %e, "Failed to convert to UTF8 String");
-                ExecutionError::InvalidEvent(format!("Failed to convert to UTF8 String: {e}"))
-            })?
+            data
```
diegomrsantos, line 181: Could you please explain this change?

anchor/eth/src/event_processor.rs
```diff
@@ -167,33 +167,22 @@ impl EventProcessor {
...
         let data = if data.len() == 704 {
-            let data = &data[64..];
-            let data = String::from_utf8(data.to_vec()).map_err(|e| {
-                debug!(operator_id = ?operator_id, error = %e, "Failed to convert to UTF8 String");
-                ExecutionError::InvalidEvent(format!("Failed to convert to UTF8 String: {e}"))
-            })?;
-            data.trim_matches(char::from(0)).to_string()
+            let mut data = &data[64..];
+            // while there is a 0 at the end of the data, remove it
+            while let [rest @ .., 0] = data {
```
diegomrsantos, line 176: [trim_matches](https://doc.rust-lang.org/std/primitive.str.html#method.trim_matches) also removes from the beginning

dknopik: `trim_matches` is not available for byte slices, only for string slices.

anchor/eth/src/event_processor.rs
dknopik: There is no need to convert into a string here, as `BASE64_STANDARD::decode` (called in Operator::new) also accepts a `&[u8]`

anchor/eth/src/event_processor.rs
diegomrsantos: The point is that the code isn't equivalent

dknopik: Ah! sorry, I misunderstood you.

In practice, the code is equivalient, as `let mut data = &data[64..];` cuts all zero bytes. Only if the data is invalid, `0x00` bytes remain at the beginning, and this is then later caught by `parse_rsa`.

diegomrsantos (approved):

PR #363, opened 2025-06-05
Avoid accidentally skipping blocks when switching to live sync

dknopik:
If a block happens just as we switch from historic to live sync, we may miss a block. This is highly unlikely, but possible.

This PR gets the logs from the block after the last processed block to the relevant block, instead of only the relevant block. In most cases, this should be equivalent, but it can't hurt to make sure.


anchor/eth/src/sync.rs
```diff
@@ -641,7 +641,9 @@ impl SsvEventSyncer {
 
                 // If the relevant block was already processed, do not process it again. This can
```
diegomrsantos, line 632: Should this be updated? What does a relevant block mean?

anchor/eth/src/sync.rs
```diff
@@ -661,7 +663,7 @@ impl SsvEventSyncer {
 
                 let mut logs = self
                     .fetch_logs(
-                        relevant_block,
+                        last_processed_block + 1,
```
diegomrsantos, line 656: Was passing `relevant_block` twice a bug?

anchor/eth/src/sync.rs
dknopik: No, this comment still applies.

The relevant block is the newest block we want to get. It is 8 blocks before the head of the chain.

anchor/eth/src/sync.rs
dknopik: Yes. Before, we were not guaranteed to fetch all logs. For example, when switching from historic to live sync, we may miss a block, and passing `relevant_block` twice returns only logs for that block. By requesting the range of blocks from the last processed block to the relevant block, we avoid this.

anchor/eth/src/sync.rs
```diff
@@ -631,7 +631,9 @@ impl SsvEventSyncer {
 
                 // If the relevant block was already processed, do not process it again. This can
                 // happen if `block_header.number` was seen before due to a reorg.
-                if relevant_block <= self.event_processor.db.state().get_last_processed_block() {
+                let last_processed_block =
+                    self.event_processor.db.state().get_last_processed_block();
+                if relevant_block <= last_processed_block {
```
Zacholme7, line 636: What is the switch here? just code quality?

Zacholme7 (approved):
makes sense

anchor/eth/src/sync.rs
dknopik: yep - we use `last_processed_block` again later.

PR #364, opened 2025-06-05
Reduce logging during sync

dknopik:
Reduce some logging during sync, by demoting some events from `debug` to `trace` and by removing several spans. Note that this PR focuses on logging during the "happy path", as @diegomrsantos has ongoing efforts for error logging.


anchor/eth/src/event_processor.rs
```diff
@@ -143,7 +143,6 @@ impl EventProcessor {
     }
 
     // A new Operator has been registered in the network.
-    #[instrument(skip(self, log), fields(operator_id, owner), level = "debug")]
```
diegomrsantos, line 146: Why does it need to be removed?

dknopik: It caused internal fields of `tx` to be logged, and redundancies with `operator_id` and `owner` which are also explicitly given in log statements below.

Zacholme7 (approved):
Good with me

anchor/eth/src/event_processor.rs
diegomrsantos: `tx` can be added to skip. I'm not sure yet, but this might be necessary in my open PR about logs

dknopik: feel free to readd it - in its current form it added no value.

PR #365, opened 2025-06-05
Fetch batches while processing

dknopik:
Instead of 

```
fetch 50 batches -> process -> fetch 50 batches -> process -> ...
```

continually fetch batches while processing the already arrived batches. This allows better parallelism and reduces load spikes on the EL client.


anchor/eth/src/sync.rs
```diff
@@ -69,8 +66,11 @@ const KEYSPLIT_EVENTS: &[&str] = &[
 /// Batch size for log fetching
 const BATCH_SIZE: u64 = 10000;
 
-/// Batch size for task groups
-const GROUP_SIZE: usize = 50;
+/// Log after this many batches have been processed
+const LOG_AFTER_BATCHES: u64 = 50;
+
+/// Number of batches to fetch concurrently
+const FETCH_CONCURRENT: usize = 50;
```
Zacholme7, line 73: It this just a finger to the wind or a number you narrowed in on? I think this could be bumped up.

anchor/eth/src/sync.rs
```diff
@@ -390,70 +395,101 @@ impl SsvEventSyncer {
...
 
-            // Further chunk the block ranges into groups where each group covers 500k blocks, so
-            // there are 50 tasks per group. BATCH_SIZE * 50 = 500k
-            let mut task_groups = Vec::new();
-            while !tasks.is_empty() {
-                // Drain takes elements from the original vector, moving them to a new vector
-                // take up to chunk_size elements (or whatever is left if less than chunk_size)
-                let chunk: Vec<_> = tasks.drain(..tasks.len().min(GROUP_SIZE)).collect();
-                task_groups.push(chunk);
+            let mut fetching_batches = FuturesOrdered::new();
```
Zacholme7, line 412: Have you taken a look at `futures::stream` and `buffer_unordered`? I this whole process could be  simplified a lot using those instead.

dknopik: We need the batches in order. That is very important, as we e.g. need to process `OperatorAdded` before processing `ValidatorAdded` referring to that operator.

Zacholme7: That was the purpose of this. What is still keeping them ordered?
```rust
                let event_logs: Vec<Log> = event_logs.into_iter().flatten().collect();

                // The futures may join out of order block wise. The individual events within the
                // block retain their tx ordering. Due to this, we can reassemble
                // back into blocks and be confident the order is correct
                let mut ordered_event_logs: BTreeMap<u64, Vec<Log>> = BTreeMap::new();
                for log in event_logs {
                    let block_num = log
                        .block_number
                        .ok_or("Log is missing block number")
                        .map_err(|e| {
                            ExecutionError::RpcError(format!("Failed to fetch block number: {e}"))
                        })?;
                    ordered_event_logs.entry(block_num).or_default().push(log);
                }
                let ordered_event_logs: Vec<Log> =
                    ordered_event_logs.into_values().flatten().collect();
```

Edit: I guess this really only works if we have the batch of logs beforehand though and not in a constantly concurrent setting

anchor/eth/src/sync.rs
dknopik: This is just keeping the previous value :) 

I played around with other values, but as batches begin to get larger during the later stages of sync, processing them is far slower than fetching them anyway, so larger values are not beneficial there. During the beginning of sync, increasing concurrency adds stress to the EL client, and actually "distracts" the EL from getting us the first batch ASAP.

I have not done proper timing tests with different values yet, but 50 felt good when doing some comparisons.

anchor/eth/src/sync.rs
dknopik: >What is still keeping them ordered?

`FuturesOrdered` guarantees that `next` returns the results in the order they were passed into it. We also take care that `pending_batches` and `fetched_batches` are not shuffled around and used as fifo queues.

Zacholme7 (approved):
nice, lgtm

PR #366, opened 2025-06-09
Add role to ExcessiveDutyCount error

diegomrsantos:
## Issue Addressed

On an `ExcessiveDutyCount` error, we currently can't see the role tha caused it.

## Proposed Changes

Add the role to the enum variant


Zacholme7 (approved):
lgtm

PR #372, opened 2025-06-17
refactor: Revamp startup sequence and check sync status

dknopik:
## Issue Addressed

We want to be able to listen to the network without an Operator ID, for example to act as an exporter, or to already discover some peers during sync. 

Also, we want to make sure to not do anything punishable (invalid network messages, signing slashable objects) if we are out of sync.

## Proposed Changes

- Add `OwnValidatorId` to allow managers and network to start eagerly
- Provide watcher for synced status, unifying `historic_finished_notify` and `operational_status`
- Avoid sending messages and starting duties for slashable objects if not synced
- Add sync status logging to notifier

## Additional Info

Open Questions:

- Not sending messages anymore is a bit paranoid, but needed to avoid potentially sending invalid messages. But we could also argue that this does not matter as the "punishment" is ephemeral - we do not lose money as with slashing.
- To avoid issues when e.g. sync only briefly fails, we could instead set `synced` to false only if we are down for a certain time, instead of immediately.


anchor/eth/src/sync.rs
```diff
@@ -251,7 +242,7 @@ impl SsvEventSyncer {
                 Ok(_) => unreachable!("Sync should never finish successfully"),
                 Err(e) => {
                     error!(?e, "Sync failed, attempting recovery");
-                    self.operational_status.store(false, Ordering::Relaxed);
+                    self.synced.send_replace(false);
```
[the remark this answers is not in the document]
dknopik: ...`send_replace` does not return a result...

And if it did, we would get a "must use" clippy warning...

anchor/client/src/lib.rs
```diff
@@ -379,15 +376,14 @@ impl Client {
...
         )
         .await
         .map_err(|e| format!("Unable to create syncer: {e}"))?;
 
-        // Access to the operational status of the sync. This can be passed around to condition
-        // duties based on the current status of the sync
-        let _operational_status = syncer.operational_status();
+        // Access to the sync status. This can be passed around to condition duties based on whether
+        // we are synced.
+        let synced = syncer.synced();
```
diegomrsantos: nit: maybe is_synced?

anchor/eth/src/sync.rs
```diff
@@ -220,8 +211,8 @@ impl SsvEventSyncer {
     }
 
     // Get access to the current status of the sync
-    pub fn operational_status(&self) -> Arc<AtomicBool> {
-        self.operational_status.clone()
+    pub fn synced(&self) -> watch::Receiver<bool> {
```
diegomrsantos: Same here

anchor/eth/src/sync.rs
```diff
@@ -330,7 +319,7 @@ impl SsvEventSyncer {
         self.historical_sync(contract_address, deployment_block, SSV_EVENTS)
             .await?;
 
-        self.historic_finished_notify.take().map(|x| x.send(()));
+        self.synced.send_replace(true);
```
diegomrsantos: Before, it mentioned it was only the historical sync. Why has it changed?

anchor/client/src/lib.rs
dknopik: done

anchor/eth/src/sync.rs
dknopik: Previously, we had a one-shot message for the finished historic sync, and the "operational status", indicating whether we have issues with the connection to the EL. While working on the PR, I realized that this is redundant: 

a) Basically, if we have connection issues, we are not in sync. 
b) Before we finish historical sync, we are not in sync.
c) The code locations using the `historic_finished_notify` oneshot can also use a longer lived channel and then drop it.

Therefore, we can combine it into a single channel without losing functionality.

diegomrsantos: but is this only for the historical sync?

diegomrsantos: if so, maybe it'd be more descriptive to use something similar to the previous name, like `is_historical_synced" or "historical_sync_has_finished".

dknopik: I would say that there is just a single property: either we are synced, or not. There are no separate states of being "historical synced" or "live synced". Historical sync and live sync are just different methods of achieving sync: one is optimized for syncing multiple blocks at once (after downtime or when syncing from scratched), and the other one is optimized for syncing blocks as they come in.

diegomrsantos: You're right, thanks

anchor/database/src/lib.rs
```diff
@@ -228,3 +229,78 @@ impl NetworkDatabase {
...
+            Self::FromState { receiver, id } => {
+                if let Some(cached_id) = id.get() {
+                    return Some(*cached_id);
+                }
+                let operator_id = receiver.borrow().get_own_id();
+                if let Some(operator_id) = operator_id {
+                    // We ignore the error because we do not care that another thread was faster.
+                    let _ = id.set(operator_id);
+                }
+                operator_id
```
diegomrsantos: We can simplify this a bit if we use `OnceCell`  and `get_or_init`, wdyt?

anchor/database/src/lib.rs
```diff
@@ -228,3 +229,78 @@ impl NetworkDatabase {
...
+    /// The operator ID was known when the `OwnOperatorId` was created.
+    Known(OperatorId),
+    /// The operator ID was not known when the `OwnOperatorId` was created. It will be retrieved
+    /// from the `receiver` and cached in the `id` on first success.
+    FromState {
+        receiver: Receiver<NetworkState>,
+        /// We use a `OnceLock` so that `get` can be called without a mutable reference.
+        id: OnceLock<OperatorId>,
+    },
+}
```
diegomrsantos: Do we need an enum with two variants or could we set the id when the `OperatorId` is available?

anchor/database/src/lib.rs
diegomrsantos: And I think we wouldn't need to implement `Clone`?

dknopik: `OnceCell` does not implement `Sync` as it can not be shared across threads. We need this to be able to use it in `async` contexts.

anchor/database/src/lib.rs
dknopik: Strictly speaking, we can also have just the second variant. The first variant is for simplifying the case where we already know the ID.

anchor/database/src/lib.rs
diegomrsantos: I'm talking about this one https://docs.rs/once_cell/latest/once_cell/

dknopik: This was integrated into Rust's std a while ago: https://doc.rust-lang.org/std/cell/struct.OnceCell.html

dknopik: Feel free to push an improved version into this PR, maybe I just do not understand your approach :)

diegomrsantos: `get_or_try_init` is experimental in std. Why do we need to implement clone manually?

dknopik: We do not strictly need it, we can also use `#[derive(Clone)]` instead. I implemented it to be able to construct a `Known` in case we know the `OperatorId` when cloning, as an easy optimization. 

How do you want to use `get_or_try_init`?

diegomrsantos: something like

```rust
use once_cell::sync::OnceCell;
use tokio::sync::watch::Receiver;

#[derive(Clone)]
pub struct OwnOperatorId {
    receiver: Receiver<NetworkState>,
    id: OnceCell<OperatorId>,
}

impl OwnOperatorId {
    pub fn get(&self) -> Option<OperatorId> {
        self.id
            .get_or_try_init(|| self.receiver.borrow().get_own_id().ok_or(()))
            .ok()
            .copied()
    }
}
```

dknopik: Looks nice :)

`once_cell` is already a transitive dependency, so I guess this is fine by me, with a note to switch to `std` once `get_or_try_init` is stable.

However, I still like the two variant approach. In many cases, we have a operator ID already available on startup, so it will be used and avoids the synchronization needed for `OnceCell` a.k.a. `OnceLock`.

diegomrsantos: I believe it's better to favour the version that’s easiest to read and maintain, unless a profiler shows we have a real performance bottleneck. What do you think?

anchor/client/src/notifier.rs
```diff
@@ -60,14 +62,25 @@ async fn notify<E: EthSpec, T: SlotClock + 'static>(
...
+        (Some(operator_id), false) => {
+            info!(%operator_id, "Operator present on chain, waiting for sync");
+        }
+        (Some(operator_id), true) => {
+            if duties_service.total_validator_count() > 0 {
+                info!(%operator_id, cluster_count, "Operator active");
+                validator_services::notifier_service::notify(duties_service).await;
+            } else {
+                info!(%operator_id, "Operator ready, no validators assigned");
+            }
```
diegomrsantos: This is hard to understand imo

anchor/client/src/lib.rs
```diff
@@ -551,6 +546,24 @@ impl Client {
...
+            is_synced.clone(),
+            executor.clone(),
+            &spec,
+        );
+
+        // Wait for sync to complete before starting services
+        info!("Waiting for sync to complete before starting services...");
+        is_synced
+            .clone()
+            .wait_for(|&is_synced| is_synced)
```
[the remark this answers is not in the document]
diegomrsantos: A timeout might be useful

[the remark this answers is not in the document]
dknopik: > periodic logging

the notifier is logging periodically 

> timeout

what would happen on timeout? there is no way to handle it

anchor/eth/src/sync.rs
```diff
@@ -254,8 +245,6 @@ impl SsvEventSyncer {
                         }
                         _ => {} // these are logged where they occur
```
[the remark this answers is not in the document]
dknopik: A successful historic sync sets `is_synced` to true.

anchor/database/src/lib.rs
```diff
@@ -228,3 +229,54 @@ impl NetworkDatabase {
...
+
+    /// Get the operator ID if it is available. Caches the ID internally after the first successful
+    /// call to avoid locking the state in the future. This is possible because the own Operator ID
+    /// never changes.
+    pub fn get(&self) -> Option<OperatorId> {
+        match self {
+            Self::Known(id) => Some(*id),
+            Self::FromState { receiver, id } => {
+                // Switch to `std`'s OnceLock as soon as `get_or_try_init` is stable
+                id.get_or_try_init(|| receiver.borrow().get_own_id().ok_or(()))
```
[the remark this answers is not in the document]
dknopik: We get rid of the error via `.ok()` right after. `get_or_try_init` unfortunately does not support options.

anchor/client/src/lib.rs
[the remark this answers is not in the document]
diegomrsantos: How about we use a sufficiently large timeout so that if syncing hasn't finished after this, there's probably an issue, and then we panic?

anchor/client/src/notifier.rs
dknopik: pushed a suggestion

anchor/client/src/lib.rs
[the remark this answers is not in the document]
dknopik: I think in general we have the approach of trying our best to get to the target state without crashing. The approach of go-ssv to crash on recoverable problems has proven to cause complications in certain scenarios, such as kurtosis devnets, so I do not really want to adopt it.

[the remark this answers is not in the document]
diegomrsantos: > crash on recoverable problems

You said earlier that there's no way to handle it. The point is that for some reason, syncing gets stuck, the system will be unresponsive, and will require a forced kill. Actually, in this case, we should log that we couldn't sync, shutdown gracefully, and ask the user to report the problem.

anchor/client/src/notifier.rs
```diff
@@ -60,14 +62,22 @@ async fn notify<E: EthSpec, T: SlotClock + 'static>(
...
         } else {
-            info!(%operator_id, "Operator ready, no validators assigned");
+            info!(%operator_id, "Operator present on chain, waiting for sync");
         }
+    } else if is_synced {
+        info!("Synced, waiting for operator key to appear on chain");
     } else {
-        error!("No operator ID - the operator might have been removed via the contract");
+        info!("Syncing");
```
diegomrsantos: I think the match was better. It seems this is a hard one. Maybe?

```rust
    match (operator_id, is_synced) {
        (None, false) => info!("Syncing"),
        (None, true)  => info!("Synced, waiting for operator key to appear on chain"),
        (Some(op), false) => info!(%op, "Operator present on chain, waiting for sync"),
        (Some(op), true) if duties_service.total_validator_count() > 0 => {
            info!(%op, cluster_count, "Operator active");
            validator_services::notifier_service::notify(duties_service).await;
        }
        (Some(op), true) => info!(%op, "Operator ready, no validators assigned"),
    }
```

dknopik: I think `op` is not as easy to understand in the logs as `operator_id`

diegomrsantos: True, but some can be one line

dknopik: like this?

diegomrsantos: yes, thanks!

anchor/client/src/lib.rs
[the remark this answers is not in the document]
diegomrsantos: @dknopik, do you prefer to merge without this?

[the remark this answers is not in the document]
dknopik: The thing is that it is hard to differentiate "stuck" from "slow". We could pick some number now, but as history gets larger, we will need to adjust it. It is also hard to pick an appropriate number. For example, Anchor works fine on a Raspberry Pi, which suffers under low disk speed if the SD card is used for storage. However, except for reading/writing state data, the disk is not really needed. So sync may take a while on that platform, but that is fine, Anchor will work otherwise. So we could pick some high number to ensure all platforms are supported even if history continues growing (like several hours), but then it is questionable if this feature is really useful with timeouts that long.

In the end, I think the benefit is not high enough compared to the added complexity here under these circumstances. If you want, we can get some more opinions :)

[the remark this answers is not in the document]
diegomrsantos: Yeah we can discuss more about it

diegomrsantos (approved):
LGTM

PR #374, opened 2025-06-18
Initial work towards spec tests

AgeManning:
This PR needs thorough review (but only where it touches core code (there's not a lot ;) )

Key things that have changed to the code:
- Stricter Typing on our core types. Specifically to adhere to the fixed-size lengths of SSZ objects. This is done by replacing Vec's with VariableLists (with the correct lengths matching the spec)
- Once we had stricter typing, we could then derive TreeHash, so where applicable added this. 
- Some hacks to match specs - needs manual review. I will highlight these

Auxiliary Things:
- Initial framework for spec tests and where applicable passing all `create` kind of spec tests. I've extensively tested the tree hash and ssz encoding of all our types, so that is not going to be an issue going forward. We match the spec in that regard. 

I'll point out some things that we need to manually review that came up when matching the spec.


anchor/common/qbft/src/lib.rs
```diff
@@ -905,22 +920,52 @@ where
...
     ) -> UnsignedWrappedQbftMessage {
         let data = self.get_message_data(&msg_type, data_hash);
 
+        let round = if let Some(round) = round {
+            round
+        } else {
+            data.round.into()
+        };
+
+        // Clear full_data from justifications as these do not store full data.
```
AgeManning, line 933: I added this. 

Justifications are SignedMessages, but they should not store FullData. I wasn't super familiar with the code base. It kinda feels like we should handle this at the type-level and have separate types to avoid any possible chance of mixing this up. 

Anyway, for the spec tests, we have to strip the full data bytes off them when creating an unsigned message.

anchor/common/qbft/src/lib.rs
```diff
@@ -905,22 +920,52 @@ where
...
+
+        let round_change_justification = VariableList::from(round_change_justification_vec);
+        let prepare_justification = VariableList::from(prepare_justification_vec);
+
+        // HACK FOR TESTS
+        let root = if matches!(msg_type, QbftMessageType::Proposal) {
+            Hash256::from_slice(sha2::Sha256::digest(data.root.as_slice()).as_slice())
+        } else {
+            data.root
+        };
```
AgeManning, line 954: This is a hack that needs to be corrected. Looking for advice here. 

The spec tests have a `Value` field. If its a proposal, the root is actually the Sha256 hash of the value, whereas for other fields its just the value. 

The machinery in creating and signing messages in the tests didn't give me a lot of room to implement this outside of this function. But it doesn't belong here and needs to be moved

diegomrsantos:
Thanks a lot for handling this! Could we move the changes to the core code to a new PR?

anchor/common/qbft/src/lib.rs
```diff
@@ -905,22 +920,52 @@ where
...
+
         // Create the QBFT message
         let qbft_message = QbftMessage {
             qbft_message_type: msg_type,
             height: *self.instance_height as u64,
-            round: data.round,
+            round: round.into(),
             identifier: (&self.identifier).into(),
-            root: data.root,
+            root, // HACK
```
AgeManning, line 962: As above

anchor/common/ssv_types/src/message.rs
```diff
@@ -455,44 +676,14 @@ impl SignedSSVMessage {
...
-        self.operator_ids = sorted_operator_ids;
+        let (sorted_signatures, sorted_operator_ids) = sig_pairs.iter().cloned().unzip();
+        self.signatures = vec_to_variable_list!(sorted_signatures, TooManySignatures)?;
+        self.operator_ids = vec_to_variable_list!(sorted_operator_ids, TooManyOperatorIDs)?;
+        Ok(())
     }
 
     // Validate the signed message to ensure that it is well formed for qbft processing
     pub fn validate(&self) -> Result<(), SignedSSVMessageError> {
-        if self.signatures.len() > MAX_SIGNATURES {
```
AgeManning, line 465: I guess with correct typing we no longer need this. I personally didn't remove this, assume daniel did

anchor/message_sender/src/network.rs
```diff
@@ -141,10 +141,15 @@ impl<S: SlotClock, D: DutiesProvider> NetworkMessageSender<S, D> {
...
-    fn sign(&self, message: &UnsignedSSVMessage) -> Result<Vec<u8>, ErrorStack> {
+    fn sign(&self, message: &UnsignedSSVMessage) -> Result<[u8; 256], ErrorStack> {
         let serialized = message.ssv_message.as_ssz_bytes();
         let mut signer = Signer::new(MessageDigest::sha256(), &self.private_key)?;
         signer.update(&serialized)?;
-        signer.sign_to_vec()
+        let mut signature = [0u8; 256];
+        let len = signer.sign(&mut signature)?;
+        if len != 256 {
+            todo!("daniel")
```
AgeManning, line 151: Prob need to handle this

anchor/message_sender/src/network.rs
```diff
@@ -141,10 +141,15 @@ impl<S: SlotClock, D: DutiesProvider> NetworkMessageSender<S, D> {
         }
     }
 
-    fn sign(&self, message: &UnsignedSSVMessage) -> Result<Vec<u8>, ErrorStack> {
+    fn sign(&self, message: &UnsignedSSVMessage) -> Result<[u8; 256], ErrorStack> {
```
AgeManning, line 144: A signature type has to now be VariableList<u8, U256>. The treehash is different between the two types

anchor/spec_tests/src/qbft/mod.rs
```diff
@@ -0,0 +1,227 @@
...
+            round,
+        );
+
+        // If its a proposal, we don't store the SSZ version, but just the raw value as the full
+        // data.
+        if matches!(message_type, QbftMessageType::Proposal) {
+            // For proposal messages, the full_data should contain the actual proposal data
+            // The data_hash parameter for proposals actually contains the raw data bytes (32 bytes)
+            // not a hash, despite the name
+            message.unsigned_message.full_data = data_hash.as_slice().to_vec();
```
AgeManning, line 91: This is important. This is a hack, it implies that when we are creating SSVMessages we are using the wrong encoding. This puts the value straight in there to pass the tests. If you look at our implementation we are putting the ssz_bytes() in there. I think these are different. 

This is something we should look at and ideally remove this hack

AgeManning:
> Thanks a lot for handling this! Could we move the changes to the core code to a new PR?

Yep, i'll do this tomorrow

Zacholme7:
Closing as overlap with #411 and #377 and going to break everything up into new PRs now Thanks for the work here!

PR #379, opened 2025-06-20
Reduce memory usage in validator

dknopik:
- Clean up old unused duties from the duty map
- Avoid excessively large integers
- Store hash instead of proposal data

### DIFF
--- a/anchor/message_validator/src/duty_state.rs
+++ b/anchor/message_validator/src/duty_state.rs
@@ -109,7 +109,7 @@ impl DutyState {
     /// stored_slot_count`. This indicates that there has been no relevant activity for this duty
     /// recently and no relevant information is lost if this is dropped.
     pub(crate) fn outdated(&self, now: Slot) -> bool {
-        let earliest_relevant_slot = now - Slot::from(self.stored_slot_count);
+        let earliest_relevant_slot = now.saturating_sub(Slot::from(self.stored_slot_count));
         self.operators
             .values()
             .all(|operator_state| operator_state.max_slot < earliest_relevant_slot)
--- a/anchor/message_validator/src/duty_state.rs
+++ b/anchor/message_validator/src/duty_state.rs
@@ -336,12 +336,13 @@ mod tests {
     fn test_duty_state_update() {
         let mut duty_state = DutyState::new(10);
 
-        let qbft_message =
+        let mut qbft_message =
             QbftMessageBuilder::new(Role::Committee, QbftMessageType::Proposal).build();
 
         let operator_id = OperatorId(1);
 
         let full_data = vec![1, 2, 3];
+        *qbft_message.root = hash_data(&full_data);
         let signed_ssv_message = create_signed_consensus_message(
             qbft_message.clone(),
             vec![operator_id],
@@ -362,7 +363,7 @@ mod tests {
             assert_eq!(
                 signer_state.proposal_hash,
                 Some(hash_data(&full_data)),
-                "Proposal data should match the signed message data"
+                "Proposal data should match the hashed full data"
             );
 
             // Verify message counts were updated
--- a/anchor/message_validator/src/duty_state.rs
+++ b/anchor/message_validator/src/duty_state.rs
@@ -108,8 +108,9 @@ impl DutyState {
     /// Returns true if all operators within the map have a `max_slot` lower than `now -
     /// stored_slot_count`. This indicates that there has been no relevant activity for this duty
     /// recently and no relevant information is lost if this is dropped.
-    pub(crate) fn outdated(&self, now: Slot) -> bool {
-        let earliest_relevant_slot = now.saturating_sub(Slot::from(self.stored_slot_count));
+    pub(crate) fn outdated(&self, current_slot: Slot) -> bool {
+        let earliest_relevant_slot =
+            current_slot.saturating_sub(Slot::from(self.stored_slot_count));
         self.operators
             .values()
             .all(|operator_state| operator_state.max_slot < earliest_relevant_slot)
--- a/anchor/message_validator/src/duty_state.rs
+++ b/anchor/message_validator/src/duty_state.rs
@@ -104,6 +104,16 @@ impl DutyState {
 
         Ok(())
     }
+
+    /// Returns true if all operators within the map have a `max_slot` lower than `now -
+    /// stored_slot_count`. This indicates that there has been no relevant activity for this duty
+    /// recently and no relevant information is lost if this is dropped.
+    pub(crate) fn outdated(&self, now: Slot) -> bool {
+        let earliest_relevant_slot = now - Slot::from(self.stored_slot_count);
+        self.operators
+            .values()
+            .all(|operator_state| operator_state.max_slot < earliest_relevant_slot)
+    }
 }
 
 /// Tracks the state for a specific operator across multiple slots.
@@ -262,8 +272,8 @@ pub(crate) struct SignerState {
     pub(crate) round: u64,
     /// Records the count of each type of consensus message encountered.
     pub(crate) message_counts: MessageCounts,
-    /// Optionally holds proposal-related data if a proposal message was received.
-    pub(crate) proposal_data: Option<Vec<u8>>,
+    /// Holds the hash of the proposal data, if a proposal was received.
+    pub(crate) proposal_hash: Option<[u8; 32]>,
     /// A set of CommitteeIds indicating which committees have already been seen.
     seen_signers: HashSet<CommitteeId>,
 }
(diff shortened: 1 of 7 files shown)


anchor/message_validator/src/lib.rs
```diff
@@ -381,6 +389,38 @@ impl<S: SlotClock, D: DutiesProvider> Validator<S, D> {
...
             })
     }
+
+    async fn cleaner(self: Arc<Self>) {
+        let slot_clock = self.slot_clock.clone();
+        let slots_per_epoch = self.slots_per_epoch;
+
+        // Use a weak reference to exit when the other `Arc` are dropped.
+        let weak_self = Arc::downgrade(&self);
+        loop {
```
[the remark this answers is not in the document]
dknopik: It should be fine, we are not doing anything computationally heavy and `retain` locks shard by shard instead of locking them all.

anchor/message_validator/src/lib.rs
```diff
@@ -271,24 +274,29 @@ pub struct Validator<S: SlotClock, D: DutiesProvider> {
...
     pub fn new(
         network_state_rx: Receiver<NetworkState>,
         slots_per_epoch: u64,
         epochs_per_sync_committee_period: u64,
         sync_committee_size: usize,
         duties_provider: Arc<D>,
         slot_clock: S,
-    ) -> Self {
-        Self {
+        task_executor: &TaskExecutor,
```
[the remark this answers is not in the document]
dknopik: Not needed

anchor/message_validator/src/duty_state.rs
```diff
@@ -104,6 +104,16 @@ impl DutyState {
 
         Ok(())
     }
+
+    /// Returns true if all operators within the map have a `max_slot` lower than `now -
+    /// stored_slot_count`. This indicates that there has been no relevant activity for this duty
+    /// recently and no relevant information is lost if this is dropped.
+    pub(crate) fn outdated(&self, now: Slot) -> bool {
+        let earliest_relevant_slot = now - Slot::from(self.stored_slot_count);
```
diegomrsantos: What's the reasoning behind creating a slot from `stored_slot_count`?

dknopik: `Slot` implements `Sub<Slot>`, but not `Sub<usize>`. So we can either create a slot from `stored_slot_count` and use that to subtract from now, or create a `usize` from `now` and create a `Slot` again after subtracting.

I don't care at all which one we use, if you prefer the latter I can change it

diegomrsantos: I think the confusion was caused by the param's name `now` instead of `current_slot`

anchor/message_validator/src/lib.rs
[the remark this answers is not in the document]
diegomrsantos: > Passing a bare &TaskExecutor (or any non-owning handle to an async runtime) into a long-lived component is usually a foot-gun.
While it compiles, it lets the owner of the executor decide when it is dropped; if that happens while your component (or the detached tasks it spawns) is still alive, every future that tries to use the executor will panic or silently mis-behave. Tokio’s own docs make the problem explicit: an Arc<Runtime> keeps the runtime alive, whereas a lightweight Handle does not prevent shutdown because the runtime stops when its Drop runs.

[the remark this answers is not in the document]
dknopik: Where is that from :sweat_smile: 

>if that happens while your component (or the detached tasks it spawns) is still alive, every future that tries to use the executor will panic or silently mis-behave

We have no futures that "use " the executor in the cleaner. We simply do not need the executor anymore as we only need to spawn that one task. Once the task is spawned and running in the Tokio runtime, we no longer need the `TaskExecutor`. We truly do not care if it is dropped.



Besides, even *if* we were to use `Arc<TaskExecutor>` here, we would drop it again after the scope of this function, and then we have the same "problem" as before (which is not actually a problem).

anchor/message_validator/src/duty_state.rs
dknopik: updated

anchor/message_validator/src/lib.rs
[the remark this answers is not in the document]
diegomrsantos: > Once the task is spawned and running in the Tokio runtime, we no longer need the TaskExecutor. We truly do not care if it is dropped.

I think what the text is trying to say is that there's no clear guarantee the Runtime will be alive.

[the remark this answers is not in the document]
dknopik: Alive when?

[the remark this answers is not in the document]
diegomrsantos: https://docs.rs/tokio/latest/tokio/runtime/struct.Runtime.html#sharing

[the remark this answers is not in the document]
dknopik: But we only have a `Handle` or a `Weak<Runtime>` in the `TaskExecutor` anyway, regardless of whether we have a reference or an Arc, so I still don't get what the problem is. 

Please explain (yourself) in exactly what scenario `task_executor: &TaskExecutor,` is a problem, and `task_executor: Arc<TaskExecutor>,` isn't.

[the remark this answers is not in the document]
diegomrsantos: Yeah, in this case it doesn't matter.

anchor/message_validator/src/message_counts.rs
```diff
@@ -6,17 +6,17 @@ use ssv_types::{
...
 /// MessageCounts tracks different types of message counts per slot
 #[derive(Debug, Clone, Copy, Default)]
 pub(crate) struct MessageCounts {
-    pub(crate) pre_consensus: u64,
-    pub(crate) proposal: u64,
-    pub(crate) prepare: u64,
-    pub(crate) commit: u64,
-    pub(crate) round_change: u64,
-    pub(crate) post_consensus: u64,
+    pub(crate) pre_consensus: u8,
```
Zacholme7, line 14: Is this an issue? Can someone spam us, we wrap back to 0 and then end up accepting something we shouldnt?

anchor/message_validator/src/lib.rs
```diff
@@ -381,6 +389,38 @@ impl<S: SlotClock, D: DutiesProvider> Validator<S, D> {
...
+            let Some(until_next_epoch) = slot_clock.duration_to_next_epoch(slots_per_epoch) else {
+                sleep(slot_clock.slot_duration()).await;
+                continue;
+            };
+
+            // Wait until 5/6ths into the slot (to avoid busy phases within a slot).
+            let sleep_for = until_next_epoch + slot_clock.slot_duration() * 5 / 6;
+            sleep(sleep_for).await;
+
+            let Some(validator) = weak_self.upgrade() else {
```
Zacholme7, line 412: Didnt know about this pattern, super neat.

anchor/message_validator/src/message_counts.rs
dknopik: We only update the validator state if a message has successfully passed validation. Therefore these integers can only ever be 0 or 1. We could actually make these be `bool`s, or even a single bitfield containing flags for all message types. SSV did this here: https://github.com/ssvlabs/ssv/pull/1874

I decided to keep these `u8` as it is easier to understand IMO.

anchor/message_validator/src/lib.rs
```diff
@@ -381,6 +389,38 @@ impl<S: SlotClock, D: DutiesProvider> Validator<S, D> {
...
+        // Use a weak reference to exit when the other `Arc` are dropped.
+        let weak_self = Arc::downgrade(&self);
+        loop {
+            // Try to get the time to the next slot.
+            let Some(until_next_epoch) = slot_clock.duration_to_next_epoch(slots_per_epoch) else {
+                sleep(slot_clock.slot_duration()).await;
+                continue;
+            };
+
+            // Wait until 5/6ths into the slot (to avoid busy phases within a slot).
```
diegomrsantos: Why 5/6? What's a busy phase?

dknopik: With busy phases I mean times where there is much to do.

E.g. at 0 seconds we decide blocks and compute selection proofs, and at 4 seconds we decide attestations. We want to avoid blocking the validator during those times to process messages as fast as possible. These messages are most relevant to the validator rewards.

At 8 seconds, we decide aggregated attestations/sync contributions. This is not as important, as other aggregators also exist. As we are not always aggregators, there are also fewer messages in general compared to the selection proof or attestation phase earlier in the slot. 

Therefore, we do this maintenance work at 10 seconds, as there are likely almost no messages exchanged then.

Zacholme7:
im good with this pr, @diegomrsantos anything on your end or we good to approve this?

anchor/message_validator/src/lib.rs
diegomrsantos: Would it be good to add a comment explaining this?

dknopik: expanded the comment with a "in a nutshell" explanation

diegomrsantos (approved):

PR #387, opened 2025-06-23
feat(database): Add `metadata` table for upgradability and avoiding database switchups

### LINKED ISSUE #357: Prevent database mixup
Currently, nothing is preventing the user from mixing up the paths for data directories, causing sync to continue on a database from a different chain. 

Add some kind of metadata to the database containing the genesis fork version and contract address to uniquely identify an eth2 chain/ssv network and error on start up if the provided data dir mismatches

dknopik:
## Issue Addressed

- closes https://github.com/sigp/anchor/issues/357

## Proposed Changes

Introduce the `metadata` table where we track the currently used database schema. This will later allow us to implement "migrations", i.e. updating the database schema without requiring the user to resync.

Furthermore, when we create the database, we store the domain type used in order to avoid working with a database that exists, but was previously used on another network.

The existing `block` table is removed and it's column moved to `metadata` in order to have only one table with data that exists only once.

## Additional Info

Theoretically, we can already implement a migration from the schema as used in `v0.1.0` to the new schema `0`, which will be the schema of `v0.2.0`. But as Hoodi syncs pretty fast, I don't think it is worth the effort, so this PR causes the old database to be deleted for resync.


diegomrsantos:
Could you please split this PR into smaller ones?

dknopik:
> Could you please split this PR into smaller ones?

That is kind of hard as this all depends on each other. But I can split this up into multiple commits for easy commit-by-commit reviewing :+1:

anchor/database/src/sql_operations.rs
```diff
@@ -1,3 +1,8 @@
+// Metadata
+pub const INSERT_METADATA: &str = r#"INSERT INTO metadata (domain_type) VALUES (?1)"#;
```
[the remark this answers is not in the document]
dknopik: I like relying on the default as this directly couples the schema source code with the version: the default *is* the version and we do not have to pass it in (and forget to adjust it later).

anchor/database/src/table_schema.sql
```diff
@@ -1,7 +1,11 @@
-CREATE TABLE block (
+-- SCHEMA VERSION 0
+
+-- we should avoid removing columns from this to keep compatibility between anchor Versions
+CREATE TABLE metadata (
```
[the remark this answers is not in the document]
dknopik: done via trigger, the suggested approach is prone to error if `id` is modified

anchor/database/src/sql_operations.rs
```diff
@@ -1,3 +1,8 @@
...
+pub const GET_METADATA: &str = r#"SELECT * FROM metadata"#;
```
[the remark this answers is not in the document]
dknopik: done

anchor/database/src/schema.rs
```diff
@@ -0,0 +1,156 @@
...
+
+/// Ensure that there is an up-to-date database available at `db_path`. Also check or set the
+/// domain type to ensure the database is for the correct network.
+pub fn ensure_up_to_date(
+    db_path: impl AsRef<Path>,
+    domain: DomainType,
+) -> Result<(), DatabaseError> {
+    let db_path = db_path.as_ref();
+    let is_new_file = !db_path.exists();
+    let conn = rusqlite::Connection::open(db_path)?;
```
Zacholme7: nit: import at top

anchor/database/src/schema.rs
```diff
@@ -0,0 +1,156 @@
...
+    db_path: impl AsRef<Path>,
+    domain: DomainType,
+) -> Result<(), DatabaseError> {
+    // Explicitly drop the connection before deleting the file to ensure it is no longer in use.
+    drop(conn);
+    remove_file(db_path.as_ref())?;
+    ensure_up_to_date(db_path, domain)
+}
+
+fn determine_database_type(conn: &rusqlite::Connection, domain: DomainType) -> DatabaseType {
```
Zacholme7: nit: import at top

anchor/database/src/table_schema.sql
```diff
@@ -1,7 +1,17 @@
...
+    schema_version INTEGER NOT NULL DEFAULT 0, 
+    domain_type INTEGER NOT NULL,
     block_number INTEGER NOT NULL DEFAULT 0 CHECK (block_number >= 0)
```
Zacholme7, line 7: Think it makes sense to keep block number as part of metadata? I think of the metadata more of as describing the database and the block number doesnt really influence that, but not sure

anchor/database/src/schema.rs
```diff
@@ -0,0 +1,156 @@
...
+            let legacy = conn
+                .query_row(sql_operations::GET_LEGACY_BLOCK, [], |row| {
+                    // Check if there is the expected column and no further columns.
+                    Ok(
+                        row.get::<_, u64>("block_number").is_ok()
+                            && row.get::<_, Value>(1).is_err(),
+                    )
+                })
+                .unwrap_or(false);
+
```
Zacholme7, line 118: Im not sure I really understand the purpose of this legacy block, do we even want this? We are still at the stage where we can make breaking changes to the DB and tell people to just resync.

Zacholme7 (commented):
Nice looks good. few little things

anchor/database/src/schema.rs
```diff
@@ -0,0 +1,156 @@
+use std::{fs::remove_file, path::Path};
+
+use rusqlite::types::Value;
+use ssv_types::domain_type::DomainType;
+
+use crate::{DatabaseError, sql_operations};
+
+type SchemaVersion = u32;
+
+struct Metadata {
```
diegomrsantos, line 10: Do you know if there's a library we can use instead of writing this code?

diegomrsantos: How about https://github.com/rust-db/refinery

anchor/database/src/table_schema.sql
dknopik: Yeah, that is also reasonable. 

I decided to group it to keep all the data that is not really relational data, but just unique key/value data with only one row ever, in one place. However I am also open to keep it separated.

anchor/database/src/schema.rs
dknopik: Yeah, this code causes an automatic resync for old Anchor databases but keeps other databases intact. 

The reason why this is here is to avoid deleting some other SQLite db that has found its way here. But this is unlikely, so we could simplify this by (for now) always deleting, and later (mainnet release) changing to keeping it intact if no metadata table is found.

anchor/database/src/schema.rs
dknopik: Yeeeeah, we could. 

We have to add a metadata table anyway because of the `domain_type` storage, so I opportunistically implemented the upgrade mechanism as well as it did not add much complexity. 

From a glance, our custom implementation allows us more control, such as removing migrations at some point and updating the initial creation script instead of having to execute all migrations. While I also understand the appeal of having all migrations forever and having database initialisations just be applying all migrations, I also like to remove old code once it is reasonable to assume that older databases no longer exist (such as after the next hardfork).

diegomrsantos: > From a glance, our custom implementation allows us more control, such as removing migrations at some point and updating the initial creation script instead of having to execute all migrations

I believe it's also possible to do it using such libs

diegomrsantos: > We have to add a metadata table anyway because of the domain_type storage, so I opportunistically implemented the upgrade mechanism as well as it did not add much complexity.

I believe we could just add the additional column in the table created by the lib

dknopik: @Zacholme7 wdyt?

Zacholme7: IMO, I think we are fine to stick with what is going on here. Switching to the lib does not really give us much of a benefit, does it?

diegomrsantos: The benefit is not writing custom code that needs to be maintained and tested when a lib already provides it.

diegomrsantos: Keep in mind that's not only about the domain metadata, but about migrations. I recommend creating a PR to introduce the migration feature using a lib that does that. Then adding the domain and the logic around it in a new one.

Zacholme7: But the code itself here is not super complex. I guess im pretty neutral here. If the lib can simplify this a lot and maintain the functionality we need then I see no reason to go with that too

dknopik: >I recommend creating a PR to introduce the migration feature using a lib that does that. Then adding the domain and the logic around it in a new one.

Will try it out, then we can compare

dknopik:
I propose we unscope this from `v0.2.0` as there is still some need for discussion.

That means that we should avoid schema changes for `v0.2.0`, to avoid manual user intervention (and thus required support from our side). Regarding database switchup, as we do not support mainnet yet, the risk associated with that is rather low, so that can wait as well.

anchor/common/ssv_types/src/domain_type.rs
```diff
@@ -28,3 +33,19 @@ impl From<[u8; 4]> for DomainType {
         Self(bytes)
     }
 }
+
+impl FromSql for DomainType {
+    fn column_result(value: ValueRef<'_>) -> FromSqlResult<Self> {
+        let value = value.as_i64()?;
+        let value = u32::try_from(value).map_err(|_| FromSqlError::InvalidType)?;
```
diegomrsantos, line 40: Should we add more context to the error?

anchor/common/ssv_types/src/domain_type.rs
```diff
@@ -28,3 +33,19 @@ impl From<[u8; 4]> for DomainType {
         Self(bytes)
     }
 }
+
+impl FromSql for DomainType {
```
diegomrsantos, line 37: I'm not sure this crate is the right place to add database code.

anchor/database/src/schema.rs
```diff
@@ -0,0 +1,156 @@
...
+    } else {
+        match determine_database_type(&conn, domain) {
+            DatabaseType::Anchor(schema_version) => schema_version,
+            DatabaseType::Unknown => {
+                // We do not know what this is. Let's be safe and error out.
+                return Err(DatabaseError::AlreadyPresent(
+                    "Unknown database schema".to_string(),
+                ));
+            }
+            DatabaseType::IncorrectDomain(domain) => {
```
diegomrsantos, line 57: Could we create a new database for the domain?

anchor/database/src/schema.rs
```diff
@@ -0,0 +1,156 @@
...
+        match get_upgrade_action(schema_version) {
+            UpgradeAction::UpToDate => return Ok(()),
+            UpgradeAction::DoUpdate {
+                script,
+                new_version,
+            } => {
+                conn.execute_batch(script)?;
+                schema_version = Some(new_version);
+            }
+            UpgradeAction::Recreate => {
```
diegomrsantos: Could you elaborate more on when we want to do this?

anchor/common/ssv_types/src/domain_type.rs
dknopik: We must add the implementation of `FromSql` in the same crate where we define `DomainType`: https://doc.rust-lang.org/reference/items/implementations.html#orphan-rules

anchor/database/src/schema.rs
dknopik: IMO that is undesired. 

This occurrence indicates that a user mixed up data directories. Which could mean that the key in this directory is not appropriate for the network.

anchor/database/src/schema.rs
dknopik: Currently, we do this if we encounter a `0.1.0`/`0.2.0` Anchor database. In future, we might want to do this if we want to get rid of old migrations. This is reasonable e.g. after a hard-fork, when we can be sure that there are no more users on older version.

anchor/common/ssv_types/src/domain_type.rs
dknopik: How? We do not control `FromSqlError`

anchor/database/src/schema.rs
diegomrsantos: Hmm, so different networks should use different data dirs?

anchor/database/src/schema.rs
diegomrsantos: I'm not sure we should delete data without user consent. How about we error and tell the user the data should be fixed or deleted before continuing?

anchor/database/src/schema.rs
dknopik: Correct. The default datadir is `~/.anchor/<network>`

anchor/database/src/schema.rs
dknopik: Yeah, that seems reasonable, will change

dknopik: done

jking-aus (dismissed):
lot of discussion on this one previously, but I generally agree with the approach in here. It's early days so we could just nuke on release but forward planning for main net is a good idea. 

lgtm

jking-aus (approved):

PR #390, opened 2025-06-25
Add `operator_key` crate to bundle key serialization functionality

dknopik:
## Issue Addressed

We have decided to switch to the same key format as used by SSV. The previous key format used by us (herein referred to as `legacy`) is to be supported in the next release version of Anchor, where we automatically convert to the new format on startup.

## Proposed Changes

Add a new crate called `operator_key`. There, we support the following:

| Module Name| Format | Reading | Writing | Notes |
|-|-|:-:|:-:|-|
| `encrypted` | Encrypted private key | :white_check_mark: | :white_check_mark: | This is basically the `Crypto` object of an Ethereum keystore with a `pubKey` tacked on to it optionally | 
| `legacy` | Legacy Anchor key | :white_check_mark: | :x: | Writing not supported as Anchor `v0.2.0` will not write this kind of key anymore, just read it for conversion. | 
| `public` | Base64 public key | :white_check_mark: | :white_check_mark: | A base64 encoded PKCS8 PEM key with PKCS1 header :upside_down_face:. This is the format found onchain. | 
| `unencrypted` | Unencrypted private key | :white_check_mark: | :white_check_mark: | As can be specified in go-ssv's config. PKCS1 private key, base64 encoded. Correct header. | 

Existing functionality is adapted to use the new functions.

## Additional Info

Not in the scope of this PR is actually using the new functionality to use new keys. A second PR will adjust the key split tool and client start up logic to convert to and use the new key format.


dknopik:
will add some tests

anchor/keygen/src/lib.rs
```diff
@@ -76,20 +80,8 @@ pub fn run_keygen(keygen: Keygen) -> Result<Rsa<Private>, KeygenError> {
...
-        .replace(
-            "-----BEGIN PUBLIC KEY-----",
-            "-----BEGIN RSA PUBLIC KEY-----",
-        )
-        .replace("-----END PUBLIC KEY-----", "-----END RSA PUBLIC KEY-----");
-
-    // Encode them to onchain format
     let private_pem_encoded = Zeroizing::new(BASE64_STANDARD.encode(&private_pem));
-    let public_pem_encoded = BASE64_STANDARD.encode(&public_pem);
+    let public_pem_encoded = operator_key::public::to_base64(&private_key)?;
```
Zacholme7, line 84: is there a mixup here? Setting `public_pem_encoded` to the encoding of `private_key`

anchor/common/operator_key/src/lib.rs
```diff
@@ -0,0 +1,18 @@
+use std::str::Utf8Error;
```
Zacholme7, line 5: Can you add a small README or just document it in the lib.rs file giving a quick snippet of the formats and the purpose/reason for each. Even after working with this all for a while I was still getting confused working through this due to all of the encodings, decodings, formats, replacements, etc.

anchor/keygen/src/lib.rs
dknopik: `operator_key::public::to_base64` takes a public or private key, as both have the needed information to encode the public key. This is why I wrote the full path here to make clear what kind of key we are encoding here. 

I'll add some doc comments to the new functions to explain the exact behaviour, have been meaning to do that anyway but forgot.

anchor/common/operator_key/src/lib.rs
dknopik: Done, lmk if something is unclear

anchor/common/operator_key/src/encrypted.rs
```diff
@@ -0,0 +1,185 @@
...
+            }
+        }
+        Ok(key)
+    }
+
+    /// Encrypt a private key into a keystore.
+    ///
+    /// [`Cipher::Aes128Ctr`] is used as cipher, and `scrypt` as constructed by [`default_kdf`] is
+    /// used as key derivation function.
+    pub fn encrypt(key: &Rsa<Private>, password: &str) -> Result<EncryptedKey, EncryptionError> {
```
Zacholme7, line 114: This isnt hooked up yet, right? Hooking this up was the follow up you mentioned?

dknopik: Correct, I split it for an easier review.

Zacholme7 (approved):
Nice im good with this

anchor/keygen/src/lib.rs
```diff
@@ -33,10 +34,13 @@ pub enum KeygenError {
...
     #[error("{0}")]
     Custom(String),
+
+    #[error("Conversion Error: {0}")]
+    Conversion(#[from] ConversionError),
 }
 
 #[derive(Zeroize, ZeroizeOnDrop, PartialEq, Debug)]
-pub struct SecurePassword(String);
+pub struct SecurePassword(pub String);
```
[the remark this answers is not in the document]
dknopik: Reasonable point, but I'll revamp the password mechanism anyway to incorporate PW files with the next pr

PR #391, opened 2025-06-25
Combine owners and nonce tables

### LINKED ISSUE #385: Combine `owners` and `nonce` tables
First, we had the nonce table, and later added the owners table. Both have the owner address as primary key and no foreign key. We can combine them into a single table, removing "nonce" and adding a "nonce" column to "owners".

petarjuki7:
## Issue Addressed

Addresses Issue #385 

## Proposed Changes

There is now one table `owners` which has both the `fee_recipient` and `nonce` fields.

## Additional Info

Removed `NOT NULL` on `fee_recipient`, had to do it because on the `BUMP_NONCE` command we are only inserting on `owner` and `nonce`. 
Tested with `cargo test` in the `database` crate and all the tests pass.

Side comment:
As far as I can tell `cargo test` doesn't work on the `stable` branch, because in the top level `Cargo.toml` there is a `sync` feature missing in the `tokio` dependency import.

### DIFF
--- a/anchor/database/src/cluster_operations.rs
+++ b/anchor/database/src/cluster_operations.rs
@@ -166,16 +166,17 @@ impl NetworkDatabase {
             // bump the nonce in memory
             if !state.single_state.nonces.contains_key(owner) {
                 // if it does not yet exist in memory, then create an entry and set it to zero
-                state.single_state.nonces.insert(*owner, 0);
+                state.single_state.nonces.insert(*owner, Some(0));
             } else {
                 // otherwise, just increment the entry
-                let entry = state
+                let entry: &mut Option<u16> = state
                     .single_state
                     .nonces
                     .get_mut(owner)
                     .expect("This must exist");
-                *entry += 1;
-                nonce = *entry;
+                let new_value = entry.unwrap_or(0) + 1;
+                *entry = Some(new_value);
+                nonce = new_value;
             }
         });
         Ok(nonce)
--- a/anchor/database/src/cluster_operations.rs
+++ b/anchor/database/src/cluster_operations.rs
@@ -166,17 +166,16 @@ impl NetworkDatabase {
             // bump the nonce in memory
             if !state.single_state.nonces.contains_key(owner) {
                 // if it does not yet exist in memory, then create an entry and set it to zero
-                state.single_state.nonces.insert(*owner, Some(0));
+                state.single_state.nonces.insert(*owner, 0);
             } else {
                 // otherwise, just increment the entry
-                let entry: &mut Option<u16> = state
+                let entry = state
                     .single_state
                     .nonces
                     .get_mut(owner)
                     .expect("This must exist");
-                let new_value = entry.unwrap_or(0) + 1;
-                *entry = Some(new_value);
-                nonce = new_value;
+                *entry += 1;
+                nonce = *entry;
             }
         });
         Ok(nonce)
--- a/anchor/database/src/lib.rs
+++ b/anchor/database/src/lib.rs
@@ -109,7 +109,7 @@ struct SingleState {
     /// All of the Clusters that we are a memeber of
     clusters: HashSet<ClusterId>,
     /// Nonce of the owner account
-    nonces: HashMap<Address, u16>,
+    nonces: HashMap<Address, Option<u16>>,
 }
 
 #[derive(Debug)]
--- a/anchor/database/src/lib.rs
+++ b/anchor/database/src/lib.rs
@@ -109,7 +109,7 @@ struct SingleState {
     /// All of the Clusters that we are a memeber of
     clusters: HashSet<ClusterId>,
     /// Nonce of the owner account
-    nonces: HashMap<Address, Option<u16>>,
+    nonces: HashMap<Address, u16>,
 }
 
 #[derive(Debug)]
--- a/anchor/database/src/sql_operations.rs
+++ b/anchor/database/src/sql_operations.rs
@@ -85,9 +85,9 @@ pub const UPDATE_BLOCK_NUMBER: &str = r#"UPDATE block SET block_number = ?1"#;
 pub const GET_BLOCK_NUMBER: &str = r#"SELECT block_number FROM block"#;
 
 // Nonce
-pub const GET_ALL_NONCES: &str = r#"SELECT * FROM nonce"#;
-pub const GET_NONCE: &str = r#"SELECT nonce FROM nonce WHERE owner = ?1"#;
+pub const GET_ALL_NONCES: &str = r#"SELECT owner, nonce FROM owners"#;
+pub const GET_NONCE: &str = r#"SELECT nonce FROM owners WHERE owner = ?1"#;
 pub const BUMP_NONCE: &str = r#"
-    INSERT INTO nonce (owner, nonce) VALUES (?1, 0)
+    INSERT INTO owners (owner, nonce) VALUES (?1, 0)
     ON CONFLICT (owner) DO UPDATE SET nonce = nonce + 1
 "#;
--- a/anchor/database/src/sql_operations.rs
+++ b/anchor/database/src/sql_operations.rs
@@ -89,5 +89,5 @@ pub const GET_ALL_NONCES: &str = r#"SELECT owner, nonce FROM owners"#;
 pub const GET_NONCE: &str = r#"SELECT nonce FROM owners WHERE owner = ?1"#;
 pub const BUMP_NONCE: &str = r#"
     INSERT INTO owners (owner, nonce) VALUES (?1, 0)
-    ON CONFLICT (owner) DO UPDATE SET nonce = COALESCE(nonce, -1) + 1
+    ON CONFLICT (owner) DO UPDATE SET nonce = COALESCE(nonce + 1, 0)
 "#;
--- a/anchor/database/src/sql_operations.rs
+++ b/anchor/database/src/sql_operations.rs
@@ -89,5 +89,5 @@ pub const GET_ALL_NONCES: &str = r#"SELECT owner, nonce FROM owners"#;
 pub const GET_NONCE: &str = r#"SELECT nonce FROM owners WHERE owner = ?1"#;
 pub const BUMP_NONCE: &str = r#"
     INSERT INTO owners (owner, nonce) VALUES (?1, 0)
-    ON CONFLICT (owner) DO UPDATE SET nonce = nonce + 1
+    ON CONFLICT (owner) DO UPDATE SET nonce = COALESCE(nonce, -1) + 1
 "#;
--- a/anchor/database/src/table_schema.sql
+++ b/anchor/database/src/table_schema.sql
@@ -4,7 +4,7 @@ CREATE TABLE block (
 INSERT INTO block (block_number) VALUES (0);
 
 CREATE TABLE owners (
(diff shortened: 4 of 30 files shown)


dknopik:
Thanks for the PR! 

>Side comment:
As far as I can tell cargo test doesn't work on the stable branch, because in the top level Cargo.toml there is a sync feature missing in the tokio dependency import.

Yeah, I also noticed that the tests for the database can not be run on their own due to this. There is some other PR where I add the "sync" feature.

>Removed NOT NULL on fee_recipient, had to do it because on the BUMP_NONCE command we are only inserting on owner and nonce.

Makes sense. But I think we might need to adjust this: https://github.com/sigp/anchor/blob/e671f385dbe8f6f0fa3c4f1eec5cfba048d654d9/anchor/database/src/validator_operations.rs#L46 to also accept NULL values by reading an `Option<String>`.

I think defaulting the nonce to `0` might cause issues, as setting the fee recipient will set the nonce to 0 if there was no row before. This is different from an unset nonce, as a set nonce of 0 implies that nonce 0 was used already. I suspect this might cause issues with syncing for owners that set the recipient before registering any validators.

I will add two test cases to illustrate these issues :)

petarjuki7:
I changed the in-memory storage a bit.
With previous in-memory storage we would assume the nonce is always there:
`nonces: HashMap<Address, u16>`

I changed nonce storage to `HashMap<Address, Option<u16>>` to properly represent:
  - `None` = owner exists but has never used a nonce
  - `Some(0)` = owner has used nonce 0
  - `Some(n)` = owner has used nonce n

Also modified all the related nonce code to use `Option<u16>` instead of `u16`
I also updated `BUMP_NONCE` SQL to use `COALESCE(nonce, -1) + 1` to handle `NULL` nonces

I'm happy changing it either way you see fit.

A simpler solution is to leave it like it was, and handle situations with NULL nonces like this:
`let nonce: u64 = row.get::<_, Option<u64>>(0)?.unwrap_or(0);`

anchor/database/src/cluster_operations.rs
```diff
@@ -166,16 +166,17 @@ impl NetworkDatabase {
...
                 // otherwise, just increment the entry
-                let entry = state
+                let entry: &mut Option<u16> = state
                     .single_state
                     .nonces
                     .get_mut(owner)
                     .expect("This must exist");
-                *entry += 1;
-                nonce = *entry;
+                let new_value = entry.unwrap_or(0) + 1;
```
petarjuki7: Here we are sure this value exists so `unwrap()`ing may be Ok. If needed I can change to a `match`.

anchor/database/src/sql_operations.rs
```diff
@@ -85,9 +85,9 @@ pub const UPDATE_BLOCK_NUMBER: &str = r#"UPDATE block SET block_number = ?1"#;
...
 // Nonce
-pub const GET_ALL_NONCES: &str = r#"SELECT * FROM nonce"#;
-pub const GET_NONCE: &str = r#"SELECT nonce FROM nonce WHERE owner = ?1"#;
+pub const GET_ALL_NONCES: &str = r#"SELECT owner, nonce FROM owners"#;
+pub const GET_NONCE: &str = r#"SELECT nonce FROM owners WHERE owner = ?1"#;
 pub const BUMP_NONCE: &str = r#"
-    INSERT INTO nonce (owner, nonce) VALUES (?1, 0)
-    ON CONFLICT (owner) DO UPDATE SET nonce = nonce + 1
+    INSERT INTO owners (owner, nonce) VALUES (?1, 0)
+    ON CONFLICT (owner) DO UPDATE SET nonce = COALESCE(nonce, -1) + 1
```
dknopik: Oooh, good idea, I like it :)

I think `COALESCE(nonce + 1, 0)` might be easier to understand and should also work because `NULL + 1` computes to `NULL`.

anchor/database/src/lib.rs
```diff
@@ -109,7 +109,7 @@ struct SingleState {
     /// All of the Clusters that we are a memeber of
     clusters: HashSet<ClusterId>,
     /// Nonce of the owner account
-    nonces: HashMap<Address, u16>,
+    nonces: HashMap<Address, Option<u16>>,
```
dknopik: Hm, I think it is preferable to keep this as before. My reasoning is that at this point there should be a single representation for a certain state regardless of it's representation in the SQL database - when changed this way, an owner that has not registered a validator yet can be either represented as not being in the map or as mapping to `None`.

I feel like it may be preferable to `filter_map` in `fetch_nonces` to not insert `None`s and keep this type as before.

anchor/database/src/table_schema.sql
```diff
@@ -5,12 +5,8 @@ INSERT INTO block (block_number) VALUES (0);
 
 CREATE TABLE owners (
     owner TEXT PRIMARY KEY,
```
dknopik: Btw, not directly related to your changes, but I think we need to add `NOT NULL` here as SQLite [allows `NULL` in primary keys](https://sqlite.org/lang_createtable.html#the_primary_key).

anchor/database/src/lib.rs
petarjuki7: Noted, fixed.

diegomrsantos:
> I changed the in-memory storage a bit. With previous in-memory storage we would assume the nonce is always there: `nonces: HashMap<Address, u16>`
> 
> I changed nonce storage to `HashMap<Address, Option<u16>>` to properly represent:
> 
> * `None` = owner exists but has never used a nonce
> * `Some(0)` = owner has used nonce 0
> * `Some(n)` = owner has used nonce n
> 
> Also modified all the related nonce code to use `Option<u16>` instead of `u16` I also updated `BUMP_NONCE` SQL to use `COALESCE(nonce, -1) + 1` to handle `NULL` nonces
> 
> I'm happy changing it either way you see fit.
> 
> A simpler solution is to leave it like it was, and handle situations with NULL nonces like this: `let nonce: u64 = row.get::<_, Option<u64>>(0)?.unwrap_or(0);`

Can we say an unused nonce is 0?

petarjuki7:
> Can we say an unused nonce is 0?

Since nonces in Ethereum start with 0 (the first transactions nonce is 0), then maybe putting it at 0 could read as "already had a first transaction as nonce 0" and not that it hasn't been set yet

anchor/database/src/sql_operations.rs
```diff
@@ -85,9 +85,9 @@ pub const UPDATE_BLOCK_NUMBER: &str = r#"UPDATE block SET block_number = ?1"#;
...
 // Nonce
-pub const GET_ALL_NONCES: &str = r#"SELECT * FROM nonce"#;
-pub const GET_NONCE: &str = r#"SELECT nonce FROM nonce WHERE owner = ?1"#;
+pub const GET_ALL_NONCES: &str = r#"SELECT owner, nonce FROM owners"#;
+pub const GET_NONCE: &str = r#"SELECT nonce FROM owners WHERE owner = ?1"#;
 pub const BUMP_NONCE: &str = r#"
-    INSERT INTO nonce (owner, nonce) VALUES (?1, 0)
-    ON CONFLICT (owner) DO UPDATE SET nonce = nonce + 1
+    INSERT INTO owners (owner, nonce) VALUES (?1, 0)
+    ON CONFLICT (owner) DO UPDATE SET nonce = COALESCE(nonce + 1, 0)
```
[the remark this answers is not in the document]
dknopik: These are different.

anchor/database/src/table_schema.sql
```diff
@@ -4,13 +4,9 @@ CREATE TABLE block (
...
-    owner TEXT PRIMARY KEY,
-    fee_recipient TEXT NOT NULL
-);
-
-CREATE TABLE nonce (
-    owner TEXT NOT NULL PRIMARY KEY,
-    nonce INTEGER DEFAULT 0
+    owner TEXT PRIMARY KEY NOT NULL,
+    fee_recipient TEXT,
+    nonce INTEGER
```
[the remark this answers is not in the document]
dknopik: No, we need to have the NULL (not present) value

diegomrsantos:
> > Can we say an unused nonce is 0?
> 
> Since nonces in Ethereum start with 0 (the first transactions nonce is 0), then maybe putting it at 0 could read as "already had a first transaction as nonce 0" and not that it hasn't been set yet

Are you sure? The yellow paper defines nonce as the number of transactions sent from the address.

diegomrsantos:
> > > Can we say an unused nonce is 0?
> > 
> > 
> > Since nonces in Ethereum start with 0 (the first transactions nonce is 0), then maybe putting it at 0 could read as "already had a first transaction as nonce 0" and not that it hasn't been set yet
> 
> Are you sure? The yellow paper defines nonce as the number of transactions sent from the address.

It also says "An account is empty when it has no code, zero nonce and zero balance"

dknopik:
We are not talking about Ethereum nonces here, but nonces as used in the SSV network. 

In the database, we store the previous, not the next nonce. Therefore there is a difference between 0 (which means the last `ValidatorAdded` processed was supposed to have a nonce of 0) and NULL (which means that there was no `ValidatorAdded` message yet).

dknopik:
To clear up the confusion why I put " :+1: " to 
>Since nonces in Ethereum start with 0 (the first transactions nonce is 0), then maybe putting it at 0 could read as "already had a first transaction as nonce 0" and not that it hasn't been set yet

While we do not work with Ethereum nonces, but with nonces as used in SSV's `ValidatorAdded` events, I liked that comment as it highlighted that we store the nonce of the already seen event, not of the expected nonce of the next event (which is also the current event count).

I will test this change on some nodes to ensure it's working fine, then we can merge this :) Thanks @petarjuki7!

dknopik:
I found another small issue: we needed to account for the null value when `fetch_clusters` is run during startup. I fixed it myself, hope you don't mind :)

We will merge this soon after deciding what will go into the next version :+1:

petarjuki7:
> We will merge this soon after deciding what will go into the next version 👍

Great, thanks! :)

Thought of something else that might be worth changing:
currently in the `bump_and_get_nonce()` function, we update the nonce manually in memory:
https://github.com/sigp/anchor/blob/94d9d4a678f17f1b9341c7965a1aad3891f4302a/anchor/database/src/cluster_operations.rs#L164-L180

It could be changed to:
```rust
let mut stmt = tx.prepare_cached(sql_operations::GET_NONCE)?;
let nonce: u16 = stmt.query_row(params![owner.to_string()], |row| row.get(0))?;

// Update in memory with the truth from the database
self.modify_state(|state| {
    state.single_state.nonces.insert(*owner, nonce);
});
Ok(nonce)
```
So we are sure it is the same in the database as in memory. This removes any risk of the in-memory logic diverging from the SQL logic.

This passes all the tests also, let me know what do you think, I can include it in a commit.

dknopik (approved):
LGTM, thanks @petarjuki7!

dknopik:
@petarjuki7 

Thanks for your suggestion!

Our caching layer operates under the assumption that the cache update is implemented consistently with the actual database upgrade. Also, no other processes modify the database - which will be enforced by #358. So I think to keep the approach to DB updates consistent, we should keep the current state.

@ThreeHrSleep maybe this is a good target for fuzz testing to ensure the update functions are correct in all cases: 
- create some input events and apply them
- capture the state
- reopen database
- check if the reopened state matches

PR #392, opened 2025-06-26
test out refinery

dknopik:
Testing how migrations might work with refinery.

Missing is functionality to deprecate old migrations. As migrations are checksummed, I do not see any way to achieve that currently.


diegomrsantos:
Maybe this could be used https://docs.rs/refinery/latest/refinery/struct.Runner.html#method.set_abort_missing. We should read more about it and understand if there's a good reason for their default behavior.

PR #395, opened 2025-06-27
Use the correct role and encoding for sync committee contributions

### LINKED ISSUE #263: Sync committee aggregations
Sync committee aggregations seem to not really work for unknown reasons. Investigate.

dknopik:
## Issue Addressed

- closes #263 

## Proposed Changes

Use the correct Role for sync committee aggregations. I think I got confused here during implementation.

Additionally, I realized Go-SSV does not encode the data correctly, so I introduced a wrapper that treats the contribution as variable length to fix the encoding

### DIFF
--- a/anchor/common/ssv_types/src/consensus.rs
+++ b/anchor/common/ssv_types/src/consensus.rs
@@ -297,6 +297,44 @@ pub struct Contribution<E: EthSpec> {
     pub contribution: SyncCommitteeContribution<E>,
 }
 
+/// This type is a workaround for the fact that Go-SSV encodes lists of `Contribution` incorrectly:
+/// it treats `Contribution` as if it had a variable length, but it does not. This wrapper
+/// implements `Encode` and `Decode` to set `is_ssz_fixed_len` to `false` and delegates to the
+/// macro impls of `Encode and `Decode` on `Contribution` for the actual serialization and
+/// deserialization.
+#[derive(Clone, Debug, Into, From)]
+pub struct ContributionWrapper<E: EthSpec> {
+    pub contribution: Contribution<E>,
+}
+
+impl<E: EthSpec> Encode for ContributionWrapper<E> {
+    fn is_ssz_fixed_len() -> bool {
+        false
+    }
+
+    fn ssz_append(&self, buf: &mut Vec<u8>) {
+        self.contribution.ssz_append(buf)
+    }
+
+    fn ssz_bytes_len(&self) -> usize {
+        self.contribution.ssz_bytes_len()
+    }
+}
+
+impl<E: EthSpec> Decode for ContributionWrapper<E> {
+    fn is_ssz_fixed_len() -> bool {
+        false
+    }
+
+    fn from_ssz_bytes(bytes: &[u8]) -> Result<Self, DecodeError> {
+        Ok(Self {
+            contribution: Contribution::from_ssz_bytes(bytes)?,
+        })
+    }
+}
+
+pub type Contributions<E> = VariableList<ContributionWrapper<E>, U13>;
+
 #[derive(Clone, Debug, TreeHash, PartialEq, Eq, Encode, Decode)]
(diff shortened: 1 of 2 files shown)


anchor/validator_store/src/lib.rs
```diff
@@ -1406,7 +1400,7 @@ impl<T: SlotClock, E: EthSpec> ValidatorStore for AnchorValidatorStore<T, E> {
                             validator_committee_index: aggregator_index,
                             validator_sync_committee_indices: Default::default(),
                         },
-                        version: ForkName::Base.into(),
+                        version: ForkName::Altair.into(),
```
diegomrsantos, line 1403: Could you explain why this changed?

dknopik: Because the previous impl was incorrect. Go-SSV always uses Altair: https://github.com/ssvlabs/ssv/blob/285cd605466539e7cf7de3bea62de4276b9c467d/beacon/goclient/sync_committee_contribution.go#L123

anchor/common/ssv_types/src/consensus.rs
```diff
@@ -297,6 +297,44 @@ pub struct Contribution<E: EthSpec> {
     pub contribution: SyncCommitteeContribution<E>,
 }
 
+/// This type is a workaround for the fact that Go-SSV encodes lists of `Contribution` incorrectly:
+/// it treats `Contribution` as if it had a variable length, but it does not. This wrapper
+/// implements `Encode` and `Decode` to set `is_ssz_fixed_len` to `false` and delegates to the
+/// macro impls of `Encode and `Decode` on `Contribution` for the actual serialization and
+/// deserialization.
+#[derive(Clone, Debug, Into, From)]
+pub struct ContributionWrapper<E: EthSpec> {
```
diegomrsantos, line 306: Could you also explain why, from a technical perspective, this is necessary? Wasn't the previous code working correctly?

dknopik: @diegomrsantos The encoding mismatched. First, let me explain some background info.

Basically, the way SSZ encoding works, is by having a "fixed" part of an encoded container, and a "variable" part. In the fixed part, all values (for example integers, or nested containers) with a fixed size are contained, as well as offsets pointing to the starts of all variable length values (such as a byte string of variable length).

So, if we have a list of variable length items, we have a bunch of offsets, followed by the actual items. Basically, for a list with two variable length items, it looks like this:
```
offset1 |offset2 |data1   |data2
08000000|0b000000|abcdef00|12323434
```
Here, `offset1` is the little endian encoding of `8`, so it is pointing to the start of `data1`, and offset2 is `12`, pointing to the start of `data2`.

For a list of fixed sized values, no offsets are necessary: as the values are of fixed size, we can just divide the length of the message by the length of a single value to determine the number of items. 

So much for the background. This is hard to explain well in a nutshell here, for more info ping me or take a look at [the spec](https://github.com/ethereum/consensus-specs/blob/dev/ssz/simple-serialize.md). Now to our issue here:

`Contribution<E>` has a fixed lengh. So a `VariableList<Contribution<E>, U13>` needs no offsets.

I noticed that Go-SSV incorrectly sends lists with offsets - basically encoding the list as if `Contribution<E>` had a variable length. The wrapper with the custom impls simply sets the length of a `Contribution<E>` to be variable, in order to have `VariableList` add the offsets. Note that the encoding of a single  `Contribution<E>` is not affected by this.

Unfortunately this is needed, as we can't simply fix this on SSV side, as such a fix would be a breaking protocol change.

jking-aus (approved):

anchor/common/ssv_types/src/consensus.rs
diegomrsantos: Thanks for the explanation. But how was the code behaving before this PR?

dknopik: It could not decode the message, causing sync committee aggregations to fail

PR #397, opened 2025-06-27
Merge immediately if branch is up to date and passes tests

dknopik:
## Issue Addressed

Mergify, in it's current config, adds PRs to the merge queue even if it is tested and up to date with the current branch. That is unfortunate because it adds additional waiting time for us.

## Proposed Changes

Only add a PR to the merge queue if it is behind the target branch. 

Otherwise, wait until tests have passed, and then merge directly.

## Additional Info

CC @diegomrsantos

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -82,14 +82,35 @@ pull_request_rules:
       review:
         type: APPROVE
 
-  - name: Add ready-to-merge labeled PRs to merge queue
+  - name: Add ready-to-merge labeled PRs that are not up to date with target to merge queue
     conditions:
       # All branch protection rules are implicit: https://docs.mergify.com/conditions/#about-branch-protection
       - base!=stable
       - label=ready-for-merge
       - label!=do-not-merge
+      - "#commits-behind > 0"
     actions:
       queue:
+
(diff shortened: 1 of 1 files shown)


.github/mergify.yml
```diff
@@ -82,15 +82,36 @@ pull_request_rules:
       review:
         type: APPROVE
 
-  - name: Add ready-to-merge labeled PRs to merge queue
+  - name: Add ready-to-merge labeled PRs that are not up to date with target to merge queue
```
diegomrsantos: Following up on the copilot comment, does this name appear somewhere? If so, maybe we can use only "Add to merge queue" and add the rest as a comment. Otherwise, looks good to me. There's some duplication, but not a big deal imo.

dknopik: > Following up on the copilot comment, does this name appear somewhere? 

I think it appears in small letters in the Github actions window below the PR.

I made the names shorter :+1:

diegomrsantos (approved):
Thanks for addressing this

dknopik:
Need to point this to `stable` as mergify wants the config to be in the default branch :melting_face: 

>Thanks for addressing this

Np, hopefully it makes it a bit more convenient to use :)

PR #399, opened 2025-06-28
Refactor `MultiIndexMap`

### LINKED ISSUE #253: Refactor `MultiIndexMap`
The multi index map has become unwieldy. 

Disadvantages that could be removed by refactoring it:

- It has 8 generic parameters, this could be improved by combining the key and uniqueness tags into one each.
- It always clones values on retrieval, which is a performance pitfall.

PoulavBhowmick03:
## Issue Addressed

Fixes #253

### DIFF
--- a/anchor/database/src/cluster_operations.rs
+++ b/anchor/database/src/cluster_operations.rs
@@ -2,7 +2,9 @@ use rusqlite::{Transaction, params};
 use ssv_types::{Cluster, ClusterId, OperatorId, Share, ValidatorMetadata};
 use types::{Address, PublicKeyBytes};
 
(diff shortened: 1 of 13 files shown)


PoulavBhowmick03:
@diegomrsantos i have fixed the tests, can you check again

dknopik:
Hi @PoulavBhowmick03! Thank you for your PR, this is a good start.

Please take a look at the Copilot suggestions above. Also, while I like the reduction of generic parameters, it would be also great if we find a way to avoid the redundant `impl (Non)UniqueIndexAccess for ...`.

dknopik:
With #463, we decided to go another way, but thank you for the PR anyway!

Issue #371, opened 2025-06-12
Add topic score parameters


diegomrsantos:
Add topic score params defined in the Go implementation

PR #402, opened 2025-07-01
Add message rate

diegomrsantos:
## Issue Addressed

https://github.com/sigp/anchor/issues/371

## Proposed Changes

This PR integrates a new message rate calculation into topic scoring by introducing a `message_rate` module and updating topic scoring configuration to use it with `ChainSpec` and an `EthSpec` generic.

- Add `message_rate.rs` to compute expected message rates for gossipsub topics based on committee configurations.
- Update `TopicScoringOptions::new` and `topic_score_params_for_subnet` to accept a `ChainSpec` and use `calculate_message_rate_for_topic`.
- Expose the new `message_rate` module in `scoring/mod.rs`.

## Additional Info

Please provide any additional information. For example, future considerations
or information useful for reviewers.


anchor/network/src/scoring/message_rate.rs
```diff
@@ -0,0 +1,722 @@
...
+// Ethereum network parameters (these could be made configurable in the future)
+const ETHEREUM_VALIDATORS: f64 = 1_000_000.0;
+
+// Derived probabilities
+const ESTIMATED_ATTESTATION_COMMITTEE_SIZE: f64 = ETHEREUM_VALIDATORS / 2048.0;
+const AGGREGATOR_PROBABILITY: f64 = 16.0 / ESTIMATED_ATTESTATION_COMMITTEE_SIZE;
+const PROPOSAL_PROBABILITY: f64 = 1.0 / ETHEREUM_VALIDATORS;
+
+// Committee size limits
+const MAX_VALIDATORS_PER_COMMITTEE: usize = 560;
```
diegomrsantos: Should we parameterize this?

anchor/network/src/scoring/message_rate.rs
```diff
@@ -0,0 +1,722 @@
...
+            * duties_with_pre_consensus;
+
+        debug!(
+            committee_size = committee_size,
+            num_validators = num_validators,
+            attestation_duties = attestation_duties,
+            sync_committee_duties = sync_committee_duties,
+            aggregator_duties = aggregator_duties,
+            proposal_duties = proposal_duties,
+            sync_agg_duties = sync_agg_duties,
```
dknopik: If key and value is the same, you can just specify it once, 
```
committee_size,
num_validators,
...
```

anchor/network/src/scoring/message_rate.rs
```diff
@@ -0,0 +1,722 @@
...
+const ETHEREUM_VALIDATORS: f64 = 1_000_000.0;
+
+// Derived probabilities
+const ESTIMATED_ATTESTATION_COMMITTEE_SIZE: f64 = ETHEREUM_VALIDATORS / 2048.0;
+const AGGREGATOR_PROBABILITY: f64 = 16.0 / ESTIMATED_ATTESTATION_COMMITTEE_SIZE;
+const PROPOSAL_PROBABILITY: f64 = 1.0 / ETHEREUM_VALIDATORS;
+
+// Committee size limits
+const MAX_VALIDATORS_PER_COMMITTEE: usize = 560;
+const SINGLE_SC_DUTIES_LIMIT: f64 = 0.0;
```
dknopik, line 40: Is `SINGLE_SC_DUTIES_LIMIT` supposed to be `0.0`?

EDIT: Ah yes, this is because then all the messages for the committee duty will already be covered by `expected_committee_duties_per_epoch_due_to_attestation`. Maybe add a comment explaining this.

anchor/network/src/scoring/message_rate.rs
```diff
@@ -0,0 +1,722 @@
...
+    let chance_of_at_least_one_validator_being_in_sync_committee =
+        1.0 - chance_that_all_validators_are_not_in_sync_committee;
+
+    // Expected number of slots with no attestation duty
+    let expected_slots_with_no_duty = slots_per_epoch as f64
+        - expected_committee_duties_per_epoch_due_to_attestation(num_validators, slots_per_epoch);
+
+    // Expected number of committee duties per epoch created due to only sync committee duties
+    chance_of_at_least_one_validator_being_in_sync_committee * expected_slots_with_no_duty
+}
```
dknopik: This is comment is not a critique of your implementation, but of the approach used by Go-SSV.

This does not make sense to me, especially for small committees. We calculate a miniscule chance here and multiply it by slots. The result might be accurate if we are considering the message rate over YEARS, but as soon as a validator in a small committee hits a sync committee duty, we easily exceed the calculated rate for several hours. So, if we get unlucky, the calculated rate is far off the actual rate.

anchor/network/src/scoring/message_rate.rs
```diff
@@ -0,0 +1,722 @@
...
+        let sync_committee_duties = expected_single_sc_committee_duties_per_epoch(
+            num_validators,
+            slots_per_epoch,
+            sync_committee_size,
+        ) * duties_without_pre_consensus;
+
+        // Calculate sync committee probabilities dynamically
+        let sync_committee_probability = sync_committee_size / ETHEREUM_VALIDATORS;
+        let sync_committee_agg_prob =
+            sync_committee_probability * 16.0 / (sync_committee_size / 4.0);
```
dknopik: The `4.0` is `types::consts::altair::SYNC_COMMITTEE_SUBNET_COUNT`,  the `16.0` is`types::consts::altair::TARGET_AGGREGATORS_PER_SYNC_SUBCOMMITTEE`. We should use those for readability.

anchor/network/src/scoring/message_rate.rs
```diff
@@ -0,0 +1,722 @@
...
+
+use ssv_types::CommitteeInfo;
+use tracing::debug;
+
+// Ethereum network parameters (these could be made configurable in the future)
+const ETHEREUM_VALIDATORS: f64 = 1_000_000.0;
+
+// Derived probabilities
+const ESTIMATED_ATTESTATION_COMMITTEE_SIZE: f64 = ETHEREUM_VALIDATORS / 2048.0;
+const AGGREGATOR_PROBABILITY: f64 = 16.0 / ESTIMATED_ATTESTATION_COMMITTEE_SIZE;
```
dknopik: `2048.0` is hardcoding `E::MaxValidatorsPerCommittee`, `16.0` is hardcoding `ChainSpec::target_aggregators_per_committee`.

anchor/network/src/scoring/message_rate.rs
dknopik: I do not think we need to parametrize this, they probably picked `560` because by then the likelihood to have a slot without an attestation duty in an epoch is extremely close to `0`.

anchor/network/src/scoring/topic_score_config.rs
```diff
@@ -308,18 +318,22 @@ impl TopicScoringOptions {
...
 pub fn topic_score_params_for_subnet(
-    one_epoch_duration: Duration,
     subnet: SubnetId,
     validator_count: u64,
     subnet_count: u64,
     committees: &[CommitteeInfo],
+    slots_per_epoch: u32,
+    slot_duration: Duration,
+    sync_committee_size: f64,
 ) -> TopicScoreParams {
```
dknopik: Still unused, right? How often will this be recomputed?

anchor/network/src/scoring/message_rate.rs
```diff
@@ -0,0 +1,722 @@
...
+/// * `sync_committee_size` - Size of the sync committee
+///
+/// # Returns
+/// Expected message rate in messages per second  
+pub fn calculate_message_rate_for_topic(
+    committees: &[CommitteeInfo],
+    slots_per_epoch: u32,
+    slot_duration: Duration,
+    sync_committee_size: f64,
+) -> f64 {
```
dknopik: Especially with more parameters being needed (as there are more non-constant parameters needed as noted above), I recommend passing a `&ChainSpec` and an `E: EthSpec` into this instead of passing needed parameters one-by-one.

anchor/network/src/scoring/message_rate.rs
```diff
@@ -0,0 +1,722 @@
...
+            * PROPOSAL_PROBABILITY
+            * duties_with_pre_consensus;
+
+        // Sync committee aggregation duties (with pre-consensus)
+        let sync_agg_duties = num_validators as f64
+            * slots_per_epoch_f64
+            * sync_committee_agg_prob
+            * duties_with_pre_consensus;
+
+        debug!(
```
dknopik: `trace` maybe?

anchor/network/src/scoring/message_rate.rs
diegomrsantos: If that's the reason, we can improve the name of the constant and add a comment. @nkryuchkov could you please clarify?

anchor/network/src/scoring/message_rate.rs
diegomrsantos: > We calculate a miniscule chance here

Could you please elaborate more?

diegomrsantos: expected_single_committee_duties

You might be right. That's a graph generated with AI, and based on some test runs, it seems correct. I'm not sure if I understand your point, but the calculated value is always very small.

cc @MatheusFranco99

anchor/network/src/scoring/message_rate.rs
diegomrsantos: This is gonna run every time we join a topic, so it shouldn't be that often?

anchor/network/src/scoring/message_rate.rs
diegomrsantos: Ok, I get it now 

> The helper that returns an epoch-level mean is the wrong tool for sizing a live gossipsub topic.  Sync-committee traffic is a bursty, low-frequency, high-magnitude phenomenon: for most epochs it is zero, but the instant one of your validators hits the 27-hour sync-committee “lottery,” you get 32 extra messages every slot (8 192 messages in the period) — several orders of magnitude above the average.

diegomrsantos: cc @AgeManning

anchor/network/src/scoring/message_rate.rs
dknopik: And every time the committees change: We need to recalculate for a topic every time a validator is added to/removed from a committee.

I am just proposing to maybe change this one in the loop, we can keep the outer one debug IMO.

anchor/network/src/scoring/message_rate.rs
nkryuchkov: We recently added a comment for this value, it's still only on stage branch though: https://github.com/ssvlabs/ssv/blob/2aa303402f725849b7aaab9e6e3d1d489e5aaa09/network/topics/params/message_rate.go#L19

The value was determined as optimal for 32 slots per epoch, according to a Monte Carlo simulation done by @MatheusFranco99

anchor/network/src/scoring/message_rate.rs
diegomrsantos: Makes sense, I'll change it

anchor/network/src/scoring/message_rate.rs
diegomrsantos: Thanks!

anchor/network/src/scoring/message_rate.rs
diegomrsantos: @nkryuchkov these values are still hardcoded in `stage`, is it an issue?

anchor/network/src/scoring/message_rate.rs
diegomrsantos: Is this always the same for every environment?

anchor/network/src/scoring/topic_score_config.rs
diegomrsantos: I'll use it in another PR. I believe it'll be called every time we join the topic, and potentially every time any `CommitteeInfo` info in the topic changes?

anchor/network/src/scoring/message_rate.rs
nkryuchkov: @diegomrsantos thanks for noticing this. It's not critical and we will unlikely ever change it in test networks, but probably we'll need to consider moving this to network config

nkryuchkov: https://github.com/ssvlabs/ssv/issues/2371

anchor/network/src/scoring/message_rate.rs
MatheusFranco99: I see your point, but this discrepancy is somewhat irrelevant. Also, the operator sending more messages than expected presents no harm. Let me walk through why I think that:

So, the total GossipSub score for a peer is

$$min(TopicCap, \sum_{t_i\in Topics} TopicWeight_{t_i} \cdot \Big( w_1(t_i) \cdot P_1(t_i) + w_2(t_i) \cdot P_2(t_i) + w_3(t_i) \cdot P_3(t_i) + w_4(t_i) \cdot P_4(t_i) \Big)\Big) + w_5P_5 + w_6P_6 + w_7P_7$$

Above, I abstracted how each counter is used, but more precisely:
- $P_1 = min(P_{1c}, Cap)$
- $P_2 = min(P_{2c}, Cap)$
- $P_3 = (Threshold - P_{3c})^2$
- $P_4 = P_{4c}^2$

Anyway, the only two positive subscores are P1 and P2, which may contribute to a maximum positive score of ~32.
Any negative score, when contributing, will decrease the final score dramatically. E.g., in case of a malicious message, $w_4 = -1280$ entirely cuts any positive contribution.

The P2 subscore increases the score of the peer based on "first seen deliveries", and that's where the function `expected_single_sc_committee_duties_per_epoch` will be used to help compute the topic's expected msg rate. 
More precisely, letting:
- $d$ be the decay
- $S$ the target mesh size
- $r$ the expected message rate

We compute the P2 cap by

$$Cap = \sum_{i = 0}^{\infty} d^i \cdot \frac{2r}{S} = \frac{2r}{S}\cdot \frac{1}{1-d}$$

Then, the weight is computed as

$$w_2 = \frac{80}{Cap}$$

where 80 is the maximum contribution of P2.

Alright, going back to `expected_single_sc_committee_duties_per_epoch`, the influence of it on $r$ is minimal compared to other contributions (e.g. from committee duties with attestation or aggregator duties). In an epoch with no sync committee duties, it will send a little less than expected. In an epoch with SC duties, it will send more. But sending more doesn't harm anything as the score will reach the Cap. And sending a bit less doesn't harm anything because the cap is computed taking into account an infinite contribution of epochs.

Sorry for the long text 😅 
I tried to explain as clearly as possible why, from my perspective, this discrepancy is a bit irrelevant.

> [!NOTE]
> Expected message rate may also be used for P3, which penalises the score, but it's currently unused.

diegomrsantos: Thanks a lot for the clear explanation! I don't mind the long text at all ;)
If `expected_single_sc_committee_duties_per_epoch` is irrelevant in practical terms, do you think it'd be better to remove it and have less code to understand and maintain?

anchor/network/src/scoring/message_rate.rs
diegomrsantos: done

anchor/network/src/scoring/message_rate.rs
```diff
@@ -0,0 +1,661 @@
+//! Message rate calculations for SSV topics based on committee configurations.
+//!
+//! This module calculates expected message rates for gossipsub topics based on
+//! the number of validators and operators in committees, following the SSV
+//! reference implementation from Go.
+
+use ssv_types::CommitteeInfo;
+use tracing::{debug, trace};
+use types::{
+    ChainSpec, EthSpec, Unsigned,
```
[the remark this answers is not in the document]
diegomrsantos: Necessary for `E::MaxValidatorsPerCommittee::to_u64()`

anchor/network/src/scoring/message_rate.rs
dknopik: Yes

anchor/network/src/scoring/message_rate.rs
```diff
@@ -4,21 +4,39 @@
...
+
+/// Calculate aggregator probability based on EthSpec
+fn aggregator_probability<E: EthSpec>() -> f64 {
+    16.0 / estimated_attestation_committee_size::<E>()
+}
+
+// For values that exceed this, the expected number of committee duties approaches zero
+// TODO: It depends on duties per epoch, 32 duties per epoch maps to
+// 560. If the value of duties per epoch changes, this value needs
+// to be adjusted (need to run Monte Carlo simulation for that number).
```
dknopik, line 33: Good catch! Need to keep this is mind especially if the number of slots increases (though this is unlikely IMO).

dknopik (approved):
LGTM, thanks a lot for this! Feel free to merge if there is no outstanding TODO from your point of view.

Not tested on a node yet as we'll wire it up with the next PR.

PR #404, opened 2025-07-01
Use Go-SSV key formats, convert legacy keys

### LINKED ISSUE #323: Migration Tooling
We should add tooling to allow easy migration from go-ssv to Anchor - this basically only consists of converting their `keys.json` format to our `key.pem` format. We can consider doing this implicitly on startup (if the former is in the data dir, but the latter is not) or add an explicit cli tool to do it

dknopik:
## Issue Addressed

- closes #323 

## Proposed Changes

- Create new key format in keygen, removing support for writing legacy keys.
- Add support for password file for both keygen and node.
- If only a legacy key is found, convert it to the new format
- Add support for explicitly specifying a key file location.

## Additional Info

I got rid of the password read retry loop, as it made the logic too awkward for too little benefit.


anchor/client/src/key.rs
```diff
@@ -0,0 +1,181 @@
...
+use tracing::{debug, info, warn};
+use zeroize::Zeroizing;
+
+pub(crate) fn read_or_generate_private_key(
+    data_dir: &Path,
+    key_file: Option<&Path>,
+    password_file: Option<&Path>,
+) -> Result<Rsa<Private>, String> {
+    // First, we have to read a file and decide what to do.
+    // TODO: do not hardcode paths here: https://github.com/sigp/anchor/issues/403
```
jking-aus, line 14: yeah agree - let's get this in 0.3

anchor/client/src/key.rs
```diff
@@ -0,0 +1,181 @@
...
+    if let Some(password) = password {
+        let file = data_dir.join("encrypted_private_key.json");
+        info!(file = %file.display(), "Saving encrypted private key");
+        let encrypted_key =
+            EncryptedKey::encrypt(key, password.as_str()).map_err(|e| e.to_string())?;
+        let serialized_key = String::try_from(encrypted_key)
+            .map_err(|e| format!("Unable to serialize encrypted key: {e}"))?;
+        File::create_new(file)
+            .and_then(|mut file| file.write_all(serialized_key.as_ref()))
```
jking-aus: consider .sync_all() to prevent key being lost if something goes wrong here since we remove_file immediately on return.

anchor/keygen/src/lib.rs
```diff
@@ -91,53 +78,61 @@ pub fn run_keygen(keygen: Keygen) -> Result<Rsa<Private>, KeygenError> {
...
 }
 
-pub fn read_password_from_user(confirm: bool) -> Result<SecurePassword, KeygenError> {
+pub fn read_password_from_user(confirm: bool) -> Result<Zeroizing<String>, KeygenError> {
     loop {
         // Prompt for password
-        let password = SecurePassword(
-            rpassword::prompt_password("Enter password for RSA keyfile: ")
+        let password = Zeroizing::new(
+            rpassword::prompt_password("Enter password for keyfile: ")
```
jking-aus, line 135: does rpassword trim whitespace? we're doing that explicitly in the file read passwords. Just want to prevent different behaviour if accidentally entering trailing space or newline on key input.

anchor/client/src/key.rs
```diff
@@ -0,0 +1,181 @@
...
+    password_file: Option<&Path>,
+) -> Result<Rsa<Private>, String> {
+    // Try to read as an encrypted key
+    let key = convert(key, EncryptedKey::try_from)?;
+    let password = if let Some(password_file) = password_file {
+        read_password_from_file(password_file)
+    } else {
+        read_password_from_user()
+    }?;
+    key.decrypt(password.as_str()).map_err(|e| e.to_string())
```
jking-aus: careful not to leak key fragments in error message -- consider generic error string. same at 139 and 176

anchor/client/src/key.rs
```diff
@@ -0,0 +1,181 @@
...
+        return None;
+    }
+
+    debug!(file = %key_file.display(), "Reading private key");
+    let file_contents = match fs::read(key_file) {
+        Ok(contents) => Zeroizing::new(contents),
+        Err(e) => return Some(Err(format!("Unable to read {}: {e}", key_file.display()))),
+    };
+
+    Some(match key_file.extension().and_then(OsStr::to_str) {
```
jking-aus: add Some(ext) => match ext.to_ascii_lowercase().as_str() to prevent capitalisation failures of supported extensions

jking-aus (dismissed):
nice, couple small nits to consider but I'm ok either way since a lot of this is transitory

anchor/keygen/src/lib.rs
dknopik: I introduced the trim especially because of newlines, which can not occur in `prompt_password`, as the return key submits the password.

I instead changed the trim to trim only `\r` and `\n`. What do you think?

jking-aus: yep that's a good point and solution

jking-aus (approved):
lgtm -- nice

PR #406, opened 2025-07-03
Set topic parameters

diegomrsantos:
## Issue Addressed

https://github.com/sigp/anchor/issues/371

## Proposed Changes

This PR replaces the old `subnet_tracker` module with a new `subnet_service` that tracks subnets and calculates dynamic message rates on epoch boundaries. It updates downstream consumers (network, client, scoring) to use the new API and pre-calculated rates.

- Introduces `subnet_service` crate with dynamic message-rate computation and epoch scheduling
- Updates topic-scoring API to accept pre-calculated rates (`new_with_rate`, `topic_score_params_for_subnet_with_rate`)
- Refactors all crates (network, client, message_sender, fuzz, peer_manager, handshake, discovery) to depend on `subnet_service` instead of `subnet_tracker`
- creates `disable_gossipsub_topic_scoring` and sets it to `false`by default

## Additional Info

Please provide any additional information. For example, future considerations
or information useful for reviewers.


dknopik:
When adding the subnet logic, we decided to separate it into the `subnet_service` in order to avoid using the network state in the network code. This has had the advantage of avoiding blocking on the lock in the network loop and separating the application logic from the network layer.

What do you think of moving some of the logic here into the `subnet_service`? Either the full calculation, introducing a `UpdateMessageRate(Topic, Rate)` event, or just `TopicUpdated(Topic, Vec<CommitteeInfo>)` which would trigger the actual calculation in the network crate.

diegomrsantos:
> When adding the subnet logic, we decided to separate it into the `subnet_service` in order to avoid using the network state in the network code. This has had the advantage of avoiding blocking on the lock in the network loop and separating the application logic from the network layer.
> 
> What do you think of moving some of the logic here into the `subnet_service`? Either the full calculation, introducing a `UpdateMessageRate(Topic, Rate)` event, or just `TopicUpdated(Topic, Vec<CommitteeInfo>)` which would trigger the actual calculation in the network crate.

True, the current implementation needs to be changed. How about? 

```rust
pub enum SubnetEvent {
    Join(SubnetId, Vec<CommitteeInfo>),
    Leave(SubnetId),
}
```

dknopik:
@diegomrsantos

This would require us to send `Join` again to pass the committee info if it changes. While this may simplify some handling, an explicit `Update(SubnetId, Vec<CommitteeInfo>)` might be a bit clearer. I am on the fence about which is better, so feel free to try out what fits best.

diegomrsantos:
> @diegomrsantos
> 
> This would require us to send `Join` again to pass the committee info if it changes. While this may simplify some handling, an explicit `Update(SubnetId, Vec<CommitteeInfo>)` might be a bit clearer. I am on the fence about which is better, so feel free to try out what fits best.

How about both events with the vec? But the second one maybe `UpdateScore`

dknopik:
I am also unsure about where we should place the actual calculation of the rate - it kind of fits both crates IMO, depending how you look at it. It feels like an application specific calculation (just like the decision what subnets to subscribe to), and maybe should therefore also live in the `subnet_service`, with just the message rate being passed to `network`. But I am kind of torn on that as well, so I leave it up to you

diegomrsantos:
> I am also unsure about where we should place the actual calculation of the rate - it kind of fits both crates IMO, depending how you look at it. It feels like an application specific calculation (just like the decision what subnets to subscribe to), and maybe should therefore also live in the `subnet_service`, with just the message rate being passed to `network`. But I am kind of torn on that as well, so I leave it up to you

Good observation. I think I'll move `message_rate` to the subnet_service. Thanks!

anchor/network/src/network.rs
```diff
@@ -272,10 +278,71 @@ impl<R: MessageReceiver> Network<R> {
...
+            }
+        }
+    }
+
     fn on_subnet_tracker_event(&mut self, event: SubnetEvent) {
         let (subnet, subscribed) = match event {
             SubnetEvent::Join(subnet) => {
-                if let Err(err) = self.gossipsub().subscribe(&subnet_to_topic(subnet)) {
+                let topic = subnet_to_topic(subnet);
+                if let Err(err) = self.gossipsub().subscribe(&topic) {
```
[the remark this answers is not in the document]
diegomrsantos: done

dknopik:
The current state does not update the params on updated committee data, right? Also, based on the the discussion above, we have to adjust the events passed to `network`

anchor/subnet_service/src/lib.rs
```diff
@@ -0,0 +1,275 @@
...
+        executor.spawn(subnet_tracker(tx, db, subnet_count), "subnet_service");
+        rx
+    } else {
+        let (tx, rx) = mpsc::channel(subnet_count);
+        for subnet in (0..(subnet_count as u64)).map(SubnetId) {
+            // For the "all subnets" case, we don't have specific committee info, so pass an empty
+            // vec
+            if let Err(err) = tx.try_send(SubnetEvent::Join(subnet, Vec::new())) {
+                error!(?err, "Impossible error while subscribing to all subnets");
+            }
```
dknopik: We can still provide information on the committee in this case.

anchor/subnet_service/src/lib.rs
```diff
@@ -0,0 +1,275 @@
...
+/// - Emits `CommitteeUpdate` events when committee information changes for existing subnets.
+async fn subnet_tracker(
+    tx: mpsc::Sender<SubnetEvent>,
+    mut db: watch::Receiver<NetworkState>,
+    subnet_count: usize,
+) {
+    // `previous_subnets` tracks which subnets were joined in the last iteration.
+    let mut previous_subnets = HashSet::new();
+    // Track committee info hash for each subnet to detect changes efficiently
+    let mut previous_committee_hashes: HashMap<SubnetId, u64> = HashMap::new();
```
dknopik: I think these can be unified by hashing subnets, not committees.

But regardless, can we not just calculate the score and compare if the old score is different from the new score?

anchor/subnet_service/src/lib.rs
```diff
@@ -0,0 +1,275 @@
...
+        let mut members: Vec<_> = committee.committee_members.iter().collect();
+        members.sort_unstable();
+        members.hash(&mut hasher);
+
+        // Hash validator indices
+        committee.validator_indices.hash(&mut hasher);
+    }
+
+    hasher.finish()
+}
```
dknopik: It might be interesting to check if this is considerably faster than just calculating the score.

dknopik (commented):
Nice!

anchor/subnet_service/src/lib.rs
jking-aus: you guys probably know better than me, but I think this approach maintains cleaner boundaries: subnet service handles membership, network handles scoring. Wouldn't hashing the subnet mean we cant distinguish joining and committee changes? I dont mind either way just my 0.02.

jking-aus (dismissed):
lgtm - just address daniel's comments

anchor/subnet_service/src/lib.rs
dknopik: Sorry, my comment above is incorrect, we **are** hashing subnets here. Still, the set and the map can be unified: After an iteration, a subnet is always either in both `previous_subnets` and 
`previous_committee_hashes`, or neither. So `previous_subnets` does not provide value anymore, unless I am missing something.

> I think this approach maintains cleaner boundaries

IMO it does not. If we have the `network` crate calculate the message rate, it needs access to the network state as provided by the database, which we avoided until now. This means one more thing passed into it, one more place where we may wait on a lock, and more domain specific logic (the message rate calculation algorithm) inside it. Note that by `inside it` I mean the code flow, not the location of the file where the message rate is being calculated.

Basically, the way I think of it is: the network crate handles scoring and subnet membership based on the parameters it is passed. It does not care about how we decide about which subnets to describe on, or how we calculate the subnet message rate, as it does not have any domain logic. We do the same approach for messages: it immediately passes them to other components to avoid any domain logic computations in the code flow of the `network` loop.

dknopik:
There seems to be an issue, I tried this on `interop1` and all peers get banned quite soon.

anchor/subnet_service/src/lib.rs
jking-aus: if you unified the hashmap and set could you distinguish for committeeupdate events for committee information changes not just when we join/leave subnet? If we can all good. Thanks for the clarification and dont disagree - the separation as it is here is what I was referring to as a preference and dont want to add state dependencies. Apologies if I oversimplified.

diegomrsantos:
> There seems to be an issue, I tried this on `interop1` and all peers get banned quite soon.

Any idea what the issue might be?

dknopik:
> Any idea what the issue might be?

Maybe an incorrect calculation by us, or some issue of the values not being set incorrectly

anchor/subnet_service/src/lib.rs
diegomrsantos: Recalculating is faster than computing the hashes, so I'll recalculate everything at every epoch

anchor/subnet_service/src/lib.rs
diegomrsantos: done

anchor/subnet_service/src/lib.rs
```diff
@@ -0,0 +1,326 @@
...
+                );
+
+                if let Err(err) = tx.try_send(SubnetEvent::Join(subnet, message_rate)) {
+                    error!(
+                        ?err,
+                        subnet = *subnet,
+                        "Failed to send subnet join event during initialization"
+                    );
+                }
+            }
```
dknopik: Currently, this only runs on startup - so we never update the rate when in `--subscribe-all-subnets` mode

anchor/subnet_service/src/lib.rs
```diff
@@ -0,0 +1,326 @@
...
+        let remaining_slots_in_epoch = if current_slot_in_epoch == 0 {
+            // We're at epoch boundary, next epoch is one full epoch away
+            slots_per_epoch
+        } else {
+            // Calculate slots remaining in current epoch
+            slots_per_epoch - current_slot_in_epoch
+        };
+
+        // Calculate time to next epoch boundary
+        slot_duration * remaining_slots_in_epoch as u32
```
dknopik: `slot_clock.duration_to_next_epoch` should work, no?

anchor/subnet_service/src/lib.rs
```diff
@@ -0,0 +1,326 @@
...
+}
+
+/// Calculate message rate for a specific subnet from the current network state
+pub fn calculate_message_rate_for_subnet<E: EthSpec>(
+    subnet: &SubnetId,
+    network_state: impl Deref<Target = NetworkState>,
+    chain_spec: &ChainSpec,
+) -> f64 {
+    let committees_info = get_committee_info_for_subnet(subnet, network_state);
+    message_rate::calculate_message_rate_for_topic::<E>(&committees_info, chain_spec)
```
dknopik, line 287: drop the `network_state` between these lines to free up the lock earlier

dknopik (commented):
Very nice! Only a few nitpicks left

anchor/subnet_service/src/lib.rs
```diff
@@ -0,0 +1,326 @@
...
+
+    loop {
+        tokio::select! {
+            // Handle database changes for subnet join/leave
+            _ = db.changed() => {
+                handle_subnet_changes::<E>(&tx, &mut db, &mut previous_subnets, subnet_count, &chain_spec).await;
+            }
+
+            // Handle scheduled epoch boundaries
+            _ = sleep(next_epoch_delay) => {
```
dknopik: you need to construct the sleep outside of the loop, as currently you restart the sleep from the beginning everytime the database changes

anchor/subnet_service/src/lib.rs
diegomrsantos: done

anchor/subnet_service/src/lib.rs
diegomrsantos: Hasn't it been moved?

anchor/subnet_service/src/lib.rs
diegomrsantos: goosd catch, fixed

anchor/subnet_service/src/lib.rs
dknopik: ah. yeah you're right, sorry

dknopik:
Neat! Thanks!

One more round of testing over night, fingers crossed!

diegomrsantos:
> Neat! Thanks!
> 
> One more round of testing over night, fingers crossed!

The topic params are disabled in the current code

anchor/subnet_service/src/lib.rs
diegomrsantos: Fixed

anchor/subnet_service/src/lib.rs
```diff
@@ -0,0 +1,337 @@
...
+            }
+
+            // Handle scheduled epoch boundaries (for both modes, but only if scoring is enabled)
+            _ = sleep(next_epoch_delay), if !disable_gossipsub_topic_scoring => {
+                handle_epoch_committee_update::<E>(&tx, &mut db, &previous_subnets, &chain_spec).await;
+                // Recalculate the next epoch delay only after we've processed the epoch boundary
+                next_epoch_delay = calculate_duration_to_next_epoch::<E>(&slot_clock);
+            }
+        }
+    }
```
dknopik, line 174: select panics if all branches are disabled via `if`, unless an `else` branch is provided

I suggest adding an else branch with `return`, as the network crate should handle this gracefully.

anchor/network/src/config.rs
```diff
@@ -54,8 +54,11 @@ pub struct Config {
     /// List of nodes to initially connect to, on Multiaddr format.
     pub boot_nodes_multiaddr: Vec<Multiaddr>,
 
-    /// Disables peer scoring altogether.
-    pub disable_peer_scoring: bool,
+    /// Disables gossipsub peer scoring altogether.
+    pub disable_gossipsub_peer_scoring: bool,
+
+    /// Disables gossipsub topic scoring and message rate calculations.
+    pub disable_gossipsub_topic_scoring: bool,
```
dknopik, line 61: Do we want to add a CLI flag to re-enable this, maybe with a note that it is experimental?

anchor/subnet_service/src/lib.rs
diegomrsantos: fixed

dknopik (approved):
LGTM, thank you for tackling this!

diegomrsantos:
> LGTM, thank you for tackling this!

Thank you for the great review

PR #377, opened 2025-06-19
Add strict lengths to internal data types

AgeManning:
## Issue Addressed

The specification indicates specific strict lengths on some of our types. 

This PR adds these lengths at a type level enforcing spec compliance and compatibility at the type level. It improves validation because the type system prevents invalid states. 

This is a cut down version of https://github.com/sigp/anchor/pull/374

Extra changes I added into this PR (sorry):
- We have a hash tree root implemented on some core types
- I split SignedSSVMessage and SSVMessage into their own files (a lot of changes are due to that split). I thought it would be more organized.

PR #411, opened 2025-07-07
Type spec tests & supporting infrastructure

Zacholme7:
Implementation for the types spec tests. Gave myself a week for these and think it may be time to just clean this all up and call them good.  This PR contains the work in #374 and #377 as everything got a bit our of line so I decided to just merge everything together. This is a giga PR so I can also try to break it up. 

I have also deemed a few to be skippable
- The [committee member](https://github.com/ssvlabs/ssv-spec/tree/main/types/spectest/tests/committeemember) tests as our types differ substantially here and it would be a couple hundred lines of just parsing for basic quorum validation. I can just write a unit tests if this is really needed
- The [beacon deposit data test](https://github.com/ssvlabs/ssv-spec/tree/main/types/spectest/tests/beacon). This generates ETH deposit data, but this is never actually even used in the client.
- The [duty tests](https://github.com/ssvlabs/ssv-spec/tree/main/types/spectest/tests/duty). These are for their internal runner mapping which is an implementation detail, not spec level. 
- The [share encoding test](https://github.com/ssvlabs/ssv-spec/tree/main/types/spectest/tests/share). We use a different structure for share data so it does not make sense to test this encoding.

There are also some tests with data issues that I have raised
- The [ssv message](https://github.com/ssvlabs/ssv-spec/tree/main/types/spectest/tests/ssvmsg) tests use message ids that should not be considered valid for the testing scenario. I have addresses this [here](https://github.com/ssvlabs/ssv-spec/pull/560)
- the [ssz test](https://github.com/ssvlabs/ssv-spec/tree/main/types/spectest/tests/ssz) uses data from a beacon block with mocked signatures which when we try to parse into `BeaconBlock`, fails with `BLST_BAD_ENCODING`. Raised the issue [here](https://github.com/ssvlabs/ssv-spec/issues/563)


Zacholme7:
Broke this up into smaller PRs

PR #412, opened 2025-07-08
Also allow comma separated CLI values

dknopik:
## Issue Addressed

Our docs and `--help` claim that e.g. `--beacon-nodes` works with comma separated values. It does not.

## Proposed Changes

Allow specification of comma separated values for all `Vec` CLI options.

## Additional Info

The old way of having multiple instances of `--beacon-nodes` etc. continues to work.


Zacholme7 (approved):
lgtm

PR #413, opened 2025-07-08
Avoid penalizing decided messages

dknopik:
## Issue Addressed

This fixes the issues arising with #406. The root cause is as follows:

We send messages anonymously, setting a random author ID through `MessageAuthenticity::RandomAuthor`. Libp2p by default penalizes getting sent messages that originate from ourselves. However, as we do not track under which author we published each message, libp2p simply punishes messages that we have published, identified by message ID. The message ID depends only on the message contents. 

The famed "decided message" is generated after a QBFT instance completes - and as it simply is an aggregation of existing signatures and messages, it can be generated by all participants and actually ends up having identical content in many cases. Nodes generate this message at the same time and send it almost simultaneously.

This causes nodes to be penalized for "self" messages, which are not actually self messages.

## Proposed Changes

Disable penalizing getting sent messages sent by "self" - which may actually be generated by someone else but happens to be identical.

## Additional Info

This is a bit unfortunate, as we no longer punish getting sent messages that are actually self messages (not creatable by other parties).

See https://github.com/libp2p/rust-libp2p/blob/1aa43375c3b05b84de985b84a2ae0caa9efa3940/protocols/gossipsub/src/behaviour.rs#L1718-L1724


jking-aus:
It's worth noting that decided messages may be removed at the next fork. Or at least that's the advice we were given. If that happens we can revert this.

diegomrsantos:
I second adding the comment, otherwise LGTM.

dknopik:
Thanks, added comment but used other wording.

diegomrsantos (approved):
Thanks for investigating this issue

PR #414, opened 2025-07-08
Switch to Go-SSV default ports

dknopik:
- 12001 UDP for discv5
- 13001 TCP for libp2p
- 13002 UDP for libp2p via QUIC
- 15000 TCP for metrics


anchor/client/src/cli.rs
```diff
@@ -258,7 +259,7 @@ pub struct Node {
         long,
         value_name = "PORT",
         help = "The UDP port that discovery will listen on over IPv6 if listening over \
-                      both IPv4 and IPv6. Defaults to `port6`",
+                      both IPv4 and IPv6. Defaults to `discovery_port`",
```
Zacholme7, line 262: Should probably explicitly say 12001 here too, otherwise people have to look at the help for `discovery_port`

Zacholme7 (approved):
lgtm

PR #415, opened 2025-07-08
recursive clone for submodules

Zacholme7:
## Issue Addressed

We need the json files from the spec tests to actually run the tests, so we have to adjust CI so that it also clones all of the needed files


Zacholme7:
Cannot seem to fix the space added, is there a way to format this?

dknopik:
>Cannot seem to fix the space added, is there a way to format this? 

eh, the removed space is fine by me :)

dknopik (approved):
LGTM

PR #423, opened 2025-07-09
Fix port selection

dknopik:
## Issue Addressed

My previous PR introduced a regression, as `--port` was only respected for the TCP port, but not the discovery port.

Also I noticed that `--port6` was not correctly set at all.

## Proposed Changes

Ensure that:

- ports are set as in Go-SSV when `--port` is not specified.
- if only `--port` is specified, that is used for both TCP and discovery UDP.
- if `--discovery-port` is specified, that value is used for discovery UDP regardless of whether `--port` is specified.
- `--port6` is respected

### DIFF
--- a/anchor/client/src/cli.rs
+++ b/anchor/client/src/cli.rs
@@ -259,11 +259,11 @@ pub struct Node {
         help = "The TCP/UDP ports to listen on. There are two UDP ports. \
                       The discovery UDP port will be set to this value and the Quic UDP port will be set to this value + 1. The discovery port can be modified by the \
                       --discovery-port flag and the quic port can be modified by the --quic-port flag. If listening over both IPv4 and IPv6 the --port flag \
-                      will apply to the IPv4 address and --port6 to the IPv6 address.",
-        default_value = "13001",
+                      will apply to the IPv4 address and --port6 to the IPv6 address. If this flag is not set, the default values will be 12001 for discovery and 13001 for TCP, \
+                      matching go-ssv's default values.",
         action = ArgAction::Set,
     )]
-    pub port: u16,
+    pub port: Option<u16>,
 
     #[clap(
         long,
@@ -277,8 +277,7 @@ pub struct Node {
     #[clap(
         long,
         value_name = "PORT",
-        help = "The UDP port that discovery will listen on. Defaults to `12001`",
-        default_value = "12001",
+        help = "The UDP port that discovery will listen on. Defaults to --port if explicitly specified, and `12001` otherwise.",
         action = ArgAction::Set,
     )]
     pub discovery_port: Option<u16>,
--- a/anchor/client/src/cli.rs
+++ b/anchor/client/src/cli.rs
@@ -259,8 +259,7 @@ pub struct Node {
         help = "The TCP/UDP ports to listen on. There are two UDP ports. \
                       The discovery UDP port will be set to this value and the Quic UDP port will be set to this value + 1. The discovery port can be modified by the \
                       --discovery-port flag and the quic port can be modified by the --quic-port flag. If listening over both IPv4 and IPv6 the --port flag \
-                      will apply to the IPv4 address and --port6 to the IPv6 address. If this flag is not set, the default values will be 12001 for discovery and 13001 for TCP, \
-                      matching go-ssv's default values.",
+                      will apply to the IPv4 address and --port6 to the IPv6 address. If this flag is not set, the default values will be 12001 for discovery and 13001 for TCP.",
         action = ArgAction::Set,
     )]
     pub port: Option<u16>,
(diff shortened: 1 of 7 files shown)


dknopik:
Added some docs to highlight the default ports and sneaked in some more details :eyes:

anchor/client/src/cli.rs
```diff
@@ -259,11 +259,11 @@ pub struct Node {
         help = "The TCP/UDP ports to listen on. There are two UDP ports. \
                       The discovery UDP port will be set to this value and the Quic UDP port will be set to this value + 1. The discovery port can be modified by the \
                       --discovery-port flag and the quic port can be modified by the --quic-port flag. If listening over both IPv4 and IPv6 the --port flag \
-                      will apply to the IPv4 address and --port6 to the IPv6 address.",
-        default_value = "13001",
+                      will apply to the IPv4 address and --port6 to the IPv6 address. If this flag is not set, the default values will be 12001 for discovery and 13001 for TCP, \
+                      matching go-ssv's default values.",
```
Zacholme7: IMO, I wouldnt mention this is matching go-ssv.

anchor/client/src/cli.rs
```diff
@@ -277,8 +277,7 @@ pub struct Node {
     #[clap(
         long,
         value_name = "PORT",
-        help = "The UDP port that discovery will listen on. Defaults to `12001`",
-        default_value = "12001",
+        help = "The UDP port that discovery will listen on. Defaults to --port if explicitly specified, and `12001` otherwise.",
```
Zacholme7: I think this description is a little confusing. I would just add in "--port if --port is explicitly specified". Otherwise it makes it sound like this will default to --port if --discovery-port is specified.

Zacholme7 (dismissed):
tiny wording. lgtm

book/src/running_node.md
```diff
@@ -30,23 +32,27 @@ To register an operator, follow the instructions for the official
 
 **Step 3: Configure and run your Anchor node**
 
-Create a directory for Anchor-related data and move the generated private key into the directory.
+Create a directory for Anchor-related data and move the generated private key into the directory. Per default, Anchor
```
diegomrsantos:
```suggestion
Create a directory for Anchor-related data and move the generated private key into the directory. By default, Anchor
```

anchor/client/src/cli.rs
diegomrsantos: or maybe "It has the value passed in --port, if --port is explicitly specified"

book/src/cli.md
```diff
@@ -73,7 +73,7 @@ anchor node [OPTIONS]
 | `--listen-address <ADDRESS>` | Network address to listen for UDP & TCP connections | `0.0.0.0` |
 | `--port <PORT>` | Base port for all network connections | `13001` |
 | `--port6 <PORT>` | Base port for IPv6 network connections | Same as `--port` |
-| `--discovery-port <PORT>` | UDP port for discovery | `12001` |
+| `--discovery-port <PORT>` | UDP port for discovery | Same as ``--port` if specified, otherwise `12001` |
```
diegomrsantos:
```suggestion
| `--discovery-port <PORT>` | UDP port for discovery | Same as `--port` if specified, otherwise `12001` |
```

anchor/client/src/cli.rs
```diff
@@ -259,11 +259,10 @@ pub struct Node {
         help = "The TCP/UDP ports to listen on. There are two UDP ports. \
                       The discovery UDP port will be set to this value and the Quic UDP port will be set to this value + 1. The discovery port can be modified by the \
                       --discovery-port flag and the quic port can be modified by the --quic-port flag. If listening over both IPv4 and IPv6 the --port flag \
-                      will apply to the IPv4 address and --port6 to the IPv6 address.",
-        default_value = "13001",
+                      will apply to the IPv4 address and --port6 to the IPv6 address. If this flag is not set, the default values will be 12001 for discovery and 13001 for TCP.",
```
diegomrsantos: It's a bit confusing We mention TCP two times, but the text is mostly about discovery UDP and Quic UDP.

anchor/client/src/config.rs
```diff
@@ -364,16 +364,18 @@ pub fn parse_listening_addresses(cli_args: &Node) -> Result<ListenAddress, Strin
                 .use_zero_ports
                 .then(unused_port::unused_tcp6_port)
                 .transpose()?
-                .unwrap_or(cli_args.port);
+                .or(cli_args.port)
+                .unwrap_or(DEFAULT_TCP_PORT);
 
             // use zero ports if required. If not, use the specific udp port. If none given, use
-            // the tcp port.
+            // the specified tcp port. If not given, use the default disc port.
```
diegomrsantos: // Select the discovery port in the following order of precedence:
// 1. If `use_zero_ports` is set, use an unused UDP6 port.
// 2. Else, if `discovery_port` is specified in CLI args, use it.
// 3. Else, if `port` is specified, use it as the fallback.
// 4. If none of the above are set, use the default discovery port (`DEFAULT_DISC_PORT`).

anchor/client/src/cli.rs
dknopik: changed it a bit, wdyt?

diegomrsantos:
Small nits, otherwise looks great.

diegomrsantos (approved):
Thank you

PR #427, opened 2025-07-14
Block-list for peers based on gossipsub score

diegomrsantos:
## Proposed Changes

Introduces a peer block-list based on libp2p gossipsub scores with automatic unblocking after a configurable epoch and exposes manual block/unblock APIs.

- Integrates `allow_block_list` in `PeerManager` to deny connections and track block timestamps.
- Adds periodic score checks in `Network` to auto-block low-scoring peers and disconnect them.
- Updates `PeerManager::new` signature to accept epoch duration, adjusts behaviour initialization, and adds tests.

### DIFF
--- a/anchor/network/src/network.rs
+++ b/anchor/network/src/network.rs
@@ -400,20 +400,6 @@ impl<R: MessageReceiver> Network<R> {
         &mut self.swarm.behaviour_mut().discovery
     }
 
-    /// Block a peer from connecting to the node.
-    /// All existing connections to this peer will be immediately closed.
-    pub fn block_peer(&mut self, peer_id: PeerId) -> bool {
-        info!(%peer_id, "Blocking peer");
-        let was_inserted = self.swarm.behaviour_mut().peer_manager.block_peer(peer_id);
-
-        // Close any existing connections to this peer
-        if self.swarm.is_connected(&peer_id) {
-            let _ = self.swarm.disconnect_peer_id(peer_id);
-        }
-
-        was_inserted
-    }
-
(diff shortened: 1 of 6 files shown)


anchor/network/src/peer_manager.rs
```diff
@@ -405,6 +499,9 @@ impl NetworkBehaviour for PeerManager {
     }
 
     fn on_swarm_event(&mut self, event: FromSwarm) {
+        // Handle block list events first
+        self.block_list.on_swarm_event(event);
```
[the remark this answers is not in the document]
diegomrsantos: `FromSwarm` implements `Copy`

anchor/network/src/network.rs
```diff
@@ -258,6 +267,10 @@ impl<R: MessageReceiver> Network<R> {
                         }
                     }
                 }
+                _ = self.peer_score_check_timer.tick() => {
+                    // Periodic peer score checks
+                    self.check_and_block_peers_by_score();
+                }
```
dknopik: It would be nice to have only one heartbeat. I understand that we need access to the swarm for `check_and_block_peers_by_score`. Wdyt about moving the peer store heartbeat here also?

dknopik (commented):
Neat, I like it!

anchor/network/src/network.rs
diegomrsantos: Great idea. I changed it, let me know what you think.

diegomrsantos:
Closing in favor of https://github.com/sigp/anchor/pull/432

Issue #349, opened 2025-05-29
Introduce convetional commits


diegomrsantos:
More info on https://www.conventionalcommits.org/en/v1.0.0/. This will automate changelog creation for releases

PR #430, opened 2025-07-16
ci: Add CI checks for conventional commits

dknopik:
## Issue Addressed

-  #349 

## Proposed Changes

Add CI for enforcing conventional commit PR titles

## Additional Info

This intentionally targets stable - we need to use `on: pull_request_target`  (which reads the CI definition from the default branch) so that the CI can post a comment if the title is incorrect. `on: pull_request` can't.

I also moved the merge target check to that new workflow to bundle all workflows that inspect the PR state instead of the code.

### DIFF
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -1,7 +1,7 @@
 name: Pull Request
 
 on:
-  pull_request:
+  pull_request_target:
     types:
       - opened
       - reopened
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -1,0 +1,99 @@
+name: Pull Request
+
+on:
+  pull_request:
+    types:
+      - opened
+      - reopened
+      - edited
+      - synchronize
+
+permissions:
+  pull-requests: read
+  contents: read
+
+concurrency:
+  # Have a lock per ref to avoid multiple runs on the same branch.
+  group: ${{ github.workflow }}-${{ github.ref }}
+  # Only the latest workflow run will be kept in progress.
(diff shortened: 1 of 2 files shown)


diegomrsantos:
Maybe we could use this? https://github.com/wagoid/commitlint-github-action

dknopik:
>Maybe we could use this? https://github.com/wagoid/commitlint-github-action

That seems more like something to validate the messages of the commits within the PR. But we do not need validation there as they will be squashed away, right? But maybe I am missing something.

diegomrsantos:
> > Maybe we could use this? https://github.com/wagoid/commitlint-github-action
> 
> That seems more like something to validate the messages of the commits within the PR. But we do not need validation there as they will be squashed away, right? But maybe I am missing something.

We can do both https://commitlint.js.org/guides/ci-setup.html#github-actions. Maybe it's better to follow what's recommended on the official website instead

diegomrsantos:
> > > Maybe we could use this? https://github.com/wagoid/commitlint-github-action
> > 
> > 
> > That seems more like something to validate the messages of the commits within the PR. But we do not need validation there as they will be squashed away, right? But maybe I am missing something.
> 
> We can do both https://commitlint.js.org/guides/ci-setup.html#github-actions. Maybe it's better to follow what's recommended on the official website instead

My motivation was to have fewer things to maintain, but this isn't adding too much. If it already works, it's fine to use it.

diegomrsantos (approved):

diegomrsantos:
@Zacholme7 could you please give a final quick review?

Zacholme7 (approved):

PR #434, opened 2025-07-17
fix: Avoid duplicate entries in `MultiIndexMap`

dknopik:
## Issue Addressed

When using `insert` on a map with a non-unique key to update a value, the primary key is reinserted into the corresponding `_multi` map even if it already exists. This causes values to be returned multiple times when `get_all_by` is used.

## Proposed Changes

Use a HashSet to enforce uniqueness.

## Additional Info

I found this randomly, interesting that it did not cause any larger problems :thinking:


Zacholme7 (approved):
Ah, nice catch

PR #435, opened 2025-07-17
refactor: Switch to let-chains where possible

dknopik:
Let-if chains are now available in stable Rust :tada: 

This saves a level of indentation in many cases.

For this, we need to bump MSRV to 1.88.


Zacholme7 (approved):
Niceeee, thx rust compiler devs

PR #437, opened 2025-07-17
refactor(signature_collector): Remove `base_hash`

dknopik:
Consider the comment:
```rust
        /// A hash that identifies what we are signing. Note that the actual signing root might be
        /// different - for example, because we are in different beacon chain attestation
        /// committees, and the attestation data differs therefore.
        base_hash: Hash256,
```

This is example incorrect. Only the actual attestation data is signed - which is the same for all validators. Therefore there is no reason for this `base_hash` field to exist - we can simply use the signing root for the same purpose.

## Additional Info

Also renamed `ValidatorSigningData` to `SigningData` now that we also use the data within for committee signature tracking purposes.

### DIFF
--- a/anchor/validator_store/src/lib.rs
+++ b/anchor/validator_store/src/lib.rs
@@ -24,8 +24,7 @@ use qbft_manager::{
 };
 use safe_arith::{ArithError, SafeArith};
 use signature_collector::{
-    CollectionError, SignatureCollectorManager, SignatureMetadata, SignatureRequester,
-    ValidatorSigningData,
+    CollectionError, SignatureCollectorManager, SignatureMetadata, SignatureRequester, SigningData,
 };
 use slashing_protection::{NotSafe, Safe, SlashingDatabase};
 use slot_clock::SlotClock;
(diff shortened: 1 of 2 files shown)


anchor/validator_store/src/lib.rs
```diff
@@ -355,7 +353,7 @@ impl<T: SlotClock, E: EthSpec> AnchorValidatorStore<T, E> {
             committee_id,
         };
 
-        let requester = if let Some(base_hash) = base_hash {
+        let requester = if role == Role::Committee {
```
diegomrsantos, line 356: Could you please explain this?

dknopik: If we are signing for the `Committee` role, we have to use `SignatureRequester::Committee` to opt into the signature collection logic.

Previous, we were using the `base_hash` as indicator for that - but we were setting it to `Some` if and only if we were passing `Role::Committee`. So the condition is effectively equivalent. Let me know if I should go into more details.

diegomrsantos (approved):

PR #347, opened 2025-05-27
refactor(cli): Unify CLI parameters and refactor CLI definition

dknopik:
## Issue Addressed

- #266

## Proposed Changes

- Refactor logging crate
- Introduce global config crate that contains shared options
- Pass global config to subcommands to avoid duplicated logic about datadir and ssv network
- Set default network to Hoodi
- Adjust docs

## Additional Info

This is not as scary to review as the lines changed seem to indicate, a lot comes from moving the logging crate around.

PR #442, opened 2025-07-18
refactor(logging): Cleaner logging setup 

dknopik:
## Proposed Changes

- Move more logging related logic to the `logging` crate
- Logging crate had a logging module which was completely reexported. Define items in crate root instead.
- Keep guards unified and separate from logging layers.
- Create discv5 and libp2p layers only once.
- Implement compression of discv5 and libp2p layers

## Additional Info

This was originally part of #347


anchor/src/main.rs
```diff
@@ -65,7 +59,7 @@ fn main() {
         None
     };
 
-    let _guards = match enable_logging(file_logging_flags, &global_config) {
+    let _guards = match logging::enable_logging(file_logging_flags, &global_config) {
```
diegomrsantos, line 54: Why is it not used?

dknopik:
rude

PR #443, opened 2025-07-18
fix(validator_store): Use database directly in validator store

### LINKED ISSUE #114: Align `database` and `validator_store`
Right now, the `validator_store` copies validator data from the database to put it in its expected format. We could populate that directly in the database to avoid this unnecessary duplication.

https://github.com/sigp/anchor/pull/109#discussion_r1923728726

dknopik:
## Issue Addressed

I noticed a fun bug: When in a single batch (for historic sync) or block (for live sync) a validator is removed and readded again, the validator store does not notice the change. This is bad, as the private key share likely has changed. So we continue signing stuff with the old partial share, while correct nodes sign with the new partial share. This causes signatures to be invalid when recombined. 

## Proposed Changes

First, I was tempted to fix this while maintaining the current structure, but I realized that iterating over the whole database every time and comparing it to everything in the validator store is quite nonsensical if we can use the database directly instead.

So this PR gets rid of the `DashMap`s in the validator store and reads needed info directly from the database state.

## Additional Info

This closes #114

Currently, I decrypt shares on demand and cache them. What do you think? We could also decrypt them as they are loaded from the DB during startup and as they come in during sync.


anchor/validator_store/src/lib.rs
```diff
@@ -78,6 +79,9 @@ use validator_store::{
 /// This acts as a maximum safe-guard against clock drift.
 const SLASHING_PROTECTION_HISTORY_EPOCHS: u64 = 512;
 
+const MAX_VALIDATORS_PER_OPERATOR: NonZeroUsize =
```
diegomrsantos, line 82: Why can't we use an integer?

anchor/validator_store/src/lib.rs
```diff
@@ -156,171 +153,53 @@ impl<T: SlotClock, E: EthSpec> AnchorValidatorStore<T, E> {
...
+            let validators: Vec<_> = watch_state
+                .borrow()
+                .shares()
+                .values()
+                .map(|share| share.validator_pubkey)
+                .collect::<Vec<_>>();
+
+            let count = validators.len() as i64;
+            validator_metrics::set_gauge(&validator_metrics::ENABLED_VALIDATORS_COUNT, count);
+            validator_metrics::set_gauge(&validator_metrics::TOTAL_VALIDATORS_COUNT, count);
```
diegomrsantos: Do we need both with the same value?

anchor/validator_store/src/lib.rs
dknopik: Because `lru_cache` wants a `NonZeroUsize`

anchor/validator_store/src/lib.rs
```diff
@@ -355,43 +236,70 @@ impl<T: SlotClock, E: EthSpec> AnchorValidatorStore<T, E> {
...
+                let decrypted_key_share =
+                    decrypt_key_share(operator_key, encrypted_private_key, validator.public_key)
+                        .map_err(|_| SpecificError::KeyShareDecryptionFailed)?;
+                self.decrypted_keys
+                    .lock()
+                    .push(encrypted_private_key, decrypted_key_share.clone());
+                Some(decrypted_key_share)
+            } else {
+                cached_share
+            }
```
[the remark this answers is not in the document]
dknopik: Hmm. Not quite correct. We get it only once - but then lock it again to set the value after computing it.

We could use `get_or_insert` instead - but that keeps the map locked during key decryption. This is unfortunate, because this prevents us from accessing the map while decrypting, so we do it separately to release the lock while decrypting.

anchor/validator_store/src/lib.rs
```diff
@@ -798,19 +748,23 @@ impl<T: SlotClock, E: EthSpec> ValidatorStore for AnchorValidatorStore<T, E> {
...
     }
 
     fn get_fee_recipient(&self, validator_pubkey: &PublicKeyBytes) -> Option<Address> {
-        self.validator(*validator_pubkey)
-            .ok()
-            .map(|v| v.cluster.fee_recipient)
+        self.database
+            .state()
+            .clusters()
+            .get_by(validator_pubkey)
```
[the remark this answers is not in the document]
dknopik: Nope, being able to lookup directly here is the reason why we have the multi index map

anchor/validator_store/src/lib.rs
```diff
@@ -831,49 +785,57 @@ impl<T: SlotClock, E: EthSpec> ValidatorStore for AnchorValidatorStore<T, E> {
...
                     if old_idx != index {
                         error!(
                             ?validator_pubkey,
                             db=?old_idx,
                             got=?index,
                             "Inconsistent validator index - database corrupt?"
                         );
-                        index_set.remove(&old_idx);
+                    }
+                } else if v.index.is_none() {
```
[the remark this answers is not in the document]
dknopik: oof, right

anchor/validator_store/src/lib.rs
dknopik: Hmm, if we want to support both of these metrics originating from LH, then yes. This has the advantage of supporting existing Lighthouse dashboards.

diegomrsantos: What is the value provided?

anchor/validator_store/src/lib.rs
```diff
@@ -88,16 +92,9 @@ const SYNC_SELECTION_PROOF_LOG_NAME: &str = "sync selection proof";
...
-    cluster: Cluster,
-    metadata: ValidatorMetadata,
-    decrypted_key_share: Option<SecretKey>,
-}
-
 pub struct AnchorValidatorStore<T: SlotClock + 'static, E: EthSpec> {
-    validators: DashMap<PublicKeyBytes, InitializedValidator>,
-    validators_per_committee: DashMap<CommitteeId, HashSet<ValidatorIndex>>,
+    database: Arc<NetworkDatabase>,
+    decrypted_keys: Mutex<LruCache<[u8; ENCRYPTED_KEY_LENGTH], SecretKey>>,
```
Zacholme7, line 97: any reason for `Mutex` over `parking_lot::RwLock`?

diegomrsantos:
I created this as an alternative. Let me know what you guys think https://github.com/sigp/anchor/pull/445

anchor/validator_store/src/lib.rs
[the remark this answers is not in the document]
diegomrsantos: > This is unfortunate, because this prevents us from accessing the map while decrypting, so we do it separately to release the lock while decrypting.

I believe it might be premature optimization, and it'd be better to focus on codebase maintainability if there's no evidence of a performance issue.

anchor/validator_store/src/lib.rs
dknopik: `RwLock` is nice for situations where we may only need a shared reference. Both times where we use `decrypted_keys`, we need a `&mut`, so I chose `Mutex`.

anchor/validator_store/src/lib.rs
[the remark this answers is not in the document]
dknopik: We are talking about like 5 more lines. I think premature optimization is a good argument against larger architectural restructuring before measuring where the issue, but this only affects this single code location.

I don't find above code to be unmaintainable.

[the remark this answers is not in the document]
diegomrsantos: I didn't say the code is unmaintainable.

IMO, premature optimization is about prioritizing code that is harder to understand, trying to make optimizations that aren't necessary.

anchor/validator_store/src/lib.rs
```diff
@@ -1535,9 +1511,13 @@ impl<T: SlotClock, E: EthSpec> ValidatorStore for AnchorValidatorStore<T, E> {
...
-        self.validator(*pubkey).ok().map(|v| ProposalData {
-            validator_index: v.metadata.index.map(|idx| *idx as u64),
-            fee_recipient: Some(v.cluster.fee_recipient),
+        let state = self.database.state();
+        state.metadata().get_by(pubkey).map(|v| ProposalData {
+            validator_index: v.index.map(|idx| *idx as u64),
+            fee_recipient: state
+                .clusters()
+                .get_by(&v.cluster_id)
+                .map(|v| v.fee_recipient),
```
diegomrsantos:
```rust
let state = self.database.state();
let metadata = state.metadata();
let maybe_validator = metadata.get_by(pubkey);

maybe_validator.map(|validator| {
    let validator_index = validator.index.map(|idx| idx as u64);
    let cluster = state.clusters().get_by(&validator.cluster_id);
    let fee_recipient = cluster.map(|c| c.fee_recipient);

    ProposalData {
        validator_index,
        fee_recipient,
        gas_limit: self.gas_limit,
        builder_proposals: self.builder_proposals,
    }
})
```

anchor/validator_store/src/lib.rs
```diff
@@ -156,171 +153,61 @@ impl<T: SlotClock, E: EthSpec> AnchorValidatorStore<T, E> {
...
+        let state = self.database.state();
+        let validator = state
+            .metadata()
+            .get_by(&validator_pubkey)
+            .ok_or(Error::UnknownPubkey(validator_pubkey))?
+            .clone();
+        let cluster = state
+            .clusters()
+            .get_by(&validator.cluster_id)
+            .ok_or(SpecificError::InconsistentDatabase)?
```
diegomrsantos: Not sure about this error and how useful it'd be without more info. Should we panic?

anchor/validator_store/src/lib.rs
```diff
@@ -332,21 +219,23 @@ impl<T: SlotClock, E: EthSpec> AnchorValidatorStore<T, E> {
...
         slot: Slot,
     ) -> Result<Signature, Error> {
-        let committee_id = validator.cluster.committee_id();
+        let committee_id = cluster.committee_id();
         let metadata = SignatureMetadata {
             kind: signature_kind,
             role,
-            threshold: validator
-                .cluster
+            threshold: cluster
```
diegomrsantos, line 213: Do we have a function somewhere to calculate this?

dknopik: I don't think so

anchor/validator_store/src/lib.rs
dknopik: The database might still be fine for other validators, so I don't think we should panic here.

We can add context for the user here, but I am not sure how helpful it would be.

anchor/validator_store/src/lib.rs
dknopik: Good idea! I pushed another idea which is even flatter.

anchor/validator_store/src/lib.rs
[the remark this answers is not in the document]
dknopik: Done

anchor/validator_store/src/lib.rs
diegomrsantos: But would the user want to continue running the system in this state?

anchor/validator_store/src/lib.rs
diegomrsantos: Much better, thanks!

anchor/validator_store/src/lib.rs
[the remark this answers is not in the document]
dknopik: That would be a premature optimization.

anchor/validator_store/src/lib.rs
dknopik: I'd argue yes. The user wants maximum uptime, and slashing protection is a separate mechanism, so continuing (in the hope of the situation fixing itself or being able to continue with other validators) seem reasonable to me.

diegomrsantos: Is this related to slashing protection?

anchor/validator_store/src/lib.rs
```diff
@@ -355,43 +259,63 @@ impl<T: SlotClock, E: EthSpec> AnchorValidatorStore<T, E> {
...
+            (num_signatures_to_collect, encrypted_private_key)
+        };
+
+        let decrypted_key_share = if let Some(operator_key) = &self.private_key {
+            let key = self
+                .decrypted_keys
+                .lock()
+                .try_get_or_insert(encrypted_private_key, || {
+                    decrypt_key_share(operator_key, encrypted_private_key, validator.public_key)
+                        .map_err(|_| SpecificError::KeyShareDecryptionFailed)
```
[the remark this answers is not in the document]
dknopik: The error type is `()`, it makes no sense to capture it

anchor/validator_store/src/lib.rs
```diff
@@ -678,6 +602,42 @@ async fn run_and_update_metrics<T>(
...
+    .map_err(|err| {
+        error!(
+            ?err,
+            validator = %pubkey_bytes,
+            "Decrypted share is not a hex string of size 64"
+        )
+    })?;
+
+    SecretKey::deserialize(&secret_key)
+        .map_err(|err| error!(?err, validator = %pubkey_bytes, "Invalid secret key decrypted"))
```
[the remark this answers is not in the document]
dknopik: We already logged the error, we do not necessarily need to propagate it

anchor/validator_store/src/lib.rs
```diff
@@ -798,19 +769,23 @@ impl<T: SlotClock, E: EthSpec> ValidatorStore for AnchorValidatorStore<T, E> {
...
 
     fn get_fee_recipient(&self, validator_pubkey: &PublicKeyBytes) -> Option<Address> {
-        self.validator(*validator_pubkey)
-            .ok()
-            .map(|v| v.cluster.fee_recipient)
+        self.database
+            .state()
+            .clusters()
+            .get_by(validator_pubkey)
+            .map(|cluster| cluster.fee_recipient)
```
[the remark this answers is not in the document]
dknopik: Again, that is the purpose of the multi index map

diegomrsantos (dismissed):
Thanks a lot!

anchor/validator_store/src/lib.rs
diegomrsantos: How would this situation fix itself?

dknopik: Through e.g. the validator getting removed

diegomrsantos: Should we have a `FOREIGN KEY` for `cluster_id` in the `validators` table?

diegomrsantos: I see we already have it, so this situation can't happen, right?

dknopik:
Pending testing

dknopik:
Something with this is messing up sync committees and attestation aggregation... will need a bit more in the oven

diegomrsantos:
> Something with this is messing up sync committees and attestation aggregation... will need a bit more in the oven

The title might be `fix` rather than `refactor`

dknopik:
> The title might be `fix` rather than `refactor`

Hmm, while the idea for this PR originated in order to fix something, I feel like the changes are complex enough to warrant `refactor`. But I don't feel strongly.

diegomrsantos:
> > The title might be `fix` rather than `refactor`
> 
> Hmm, while the idea for this PR originated in order to fix something, I feel like the changes are complex enough to warrant `refactor`. But I don't feel strongly.

I believe the main argument is that in a refactor, the behavior shouldn't change, but this PR does change the behavior.

anchor/validator_store/src/lib.rs
```diff
@@ -156,171 +153,76 @@ impl<T: SlotClock, E: EthSpec> AnchorValidatorStore<T, E> {
...
+            validator_metrics::set_gauge(&validator_metrics::ENABLED_VALIDATORS_COUNT, count);
+            validator_metrics::set_gauge(&validator_metrics::TOTAL_VALIDATORS_COUNT, count);
+
+            let unregistered_validators = validators
+                .difference(&registered_validators)
+                .collect::<Vec<_>>();
+            if !unregistered_validators.is_empty() {
+                if let Err(err) = self
+                    .slashing_protection
+                    .register_validators(unregistered_validators.into_iter())
```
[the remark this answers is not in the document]
dknopik: this code is gone by now, resolving

anchor/validator_store/src/lib.rs
dknopik: It can not happen on the Sqlite level (assuming sqlite is correct). However we need to consider the case here, in case e.g. we store something incorrectly on startup.

I am still on the fence on panic vs error in this case. This boils down to the good old liveness vs safety. As it might be possible to continue for other validators, and are guarded by the fact that we are operating in a committee and that the slashing protection is a separate data structure, I lean towards leaving it as is, but I am not feeling strongly

dknopik:
>Something with this is messing up sync committees and attestation aggregation... will need a bit more in the oven

I believe this was an issue with another PR, so this is good to go from my POV now.

anchor/validator_store/src/lib.rs
```diff
@@ -794,28 +794,25 @@ impl<T: SlotClock, E: EthSpec> ValidatorStore for AnchorValidatorStore<T, E> {
...
 
     fn set_validator_index(&self, validator_pubkey: &PublicKeyBytes, index: u64) {
-        match self
-            .database
-            .state()
-            .metadata()
-            .get_by_validator_pubkey(validator_pubkey)
-        {
+        let state = self.database.state();
+        match state.metadata().get_by_validator_pubkey(validator_pubkey) {
```
diegomrsantos: Could we use a let for the metadata here and drop the state?

anchor/validator_store/src/lib.rs
```diff
@@ -794,28 +794,25 @@ impl<T: SlotClock, E: EthSpec> ValidatorStore for AnchorValidatorStore<T, E> {
...
+                            db=?old_idx,
+                            got=?index,
+                            "Inconsistent validator index - database corrupt?"
+                        );
+                    }
                 } else {
+                    drop(state);
                     let result = self
                         .database
                         .set_validator_indices(HashMap::from([(*validator_pubkey, index)]));
```
diegomrsantos: What is the "unnecessary setting of index" mentioned in the commit?

anchor/validator_store/src/lib.rs
dknopik: The return value  of `get_by_validator_pubkey` contains a reference, so the state must live longer than the metadata is used.

That being said, we can map the metadata to an owned copy of the index before the match. I'll push a suggestion

anchor/validator_store/src/lib.rs
dknopik: Before the commit, we called `set_validator_indices` when `v.metadata.index` was `Some` and equaled the index passed into this function. This was useless as the index is already set.

This resulted from a non-conflicting merge.

anchor/validator_store/src/lib.rs
diegomrsantos: How does it look like after the revert?

dknopik: Not that much different. `get_by_validator_pubkey` becomes `get_by`, and the returned type is `Option<&Metadata>` instead of `Option<&MetadataIndexed>`

diegomrsantos (dismissed):

diegomrsantos (dismissed):

diegomrsantos (approved):

PR #448, opened 2025-07-24
refactor: Implement data dir module for central management of paths

### LINKED ISSUE #403: Create common accessor for various data dir paths
Instead of hardcoding filenames all over the repo, create a common module that provides access to the various files in the data dir.

This could e.g. be a `common` crate with a `DataDir` struct that has several methods to get `PathBuf`s to the corresponding files.

dknopik:
## Issue Addressed

We have many hardcoded paths all over the codebase.

## Proposed Changes

Move all of them and directory creation to a `data_dir` module in global config.

There we can also soon add a lockfile mechanism!


anchor/client/src/config.rs
```diff
@@ -86,6 +86,7 @@ impl Config {
         ];
         let execution_nodes_websocket = SensitiveUrl::parse(DEFAULT_EXECUTION_NODE_WS)
             .expect("execution_nodes_websocket must always be a valid url.");
+        let network = network::Config::new(global_config.data_dir.network_dir());
```
Zacholme7: General nit, think we should prefix all of the configs with their context instead of relying on the import path? We have like 7 different Config structs all named `Config`

dknopik: Hmm, that is a matter of taste/philosophy. 

There is a (`allow` by default) clippy lint about this: https://rust-lang.github.io/rust-clippy/master/index.html#module_name_repetitions

I guess the argument for `network::Config` is that within the network, we can use `Config` (and it is apparent from context that it is the network config), and outside, we can use `network::Config`, which is not very different from writing `NetworkConfig`.

That being said, it is a matter of taste, and I don't feel strongly. Curious about @diegomrsantos's opinion.

diegomrsantos: It took me some time to get used to it, too, but I believe using `network::Config` is idiomatic Rust. Now I kinda like it.

diegomrsantos: `let network_config`?

anchor/common/global_config/src/data_dir.rs
```diff
@@ -0,0 +1,92 @@
...
+        self.path.join("unencrypted_private_key.txt")
+    }
+
+    pub fn encrypted_private_key_file(&self) -> PathBuf {
+        self.path.join("encrypted_private_key.json")
+    }
+
+    pub fn database_file(&self) -> PathBuf {
+        self.path.join("anchor_db.sqlite")
+    }
```
diegomrsantos, line 62: Should this be in the database module?

anchor/common/global_config/src/data_dir.rs
```diff
@@ -0,0 +1,92 @@
...
+    #[error("Failed to create data directory")]
+    Create(#[from] std::io::Error),
+}
+
+impl DataDir {
+    pub fn new(path: PathBuf) -> Result<Self, DataDirError> {
+        let ret = DataDir { path };
+
+        create_dir_all(&ret.path)?;
+        create_dir_all(&ret.network_dir().path)?;
```
diegomrsantos, line 25: I don't believe side effects in a constructor are a good practice.

anchor/common/global_config/src/data_dir.rs
diegomrsantos: Ok, we could argue that this module owns every filesystem concern; the database layer only needs a string/PathBuf to open a connection.

diegomrsantos (commented):
Great improvement, thanks for addressing it. Just a small comment.

anchor/common/global_config/src/data_dir.rs
dknopik: right, that was my thought behind it

anchor/common/global_config/src/data_dir.rs
dknopik: The idea is that holding `DataDir` is holding the resource ownership to a valid data dir.

This is basically using [RAII](https://en.wikipedia.org/wiki/Resource_acquisition_is_initialization): When we create a `DataDir`, we assure that we have a valid data dir, including existing underlying directories and (later) locking to assure that we are the only one holding a `DataDir`. When we drop the `DataDir`, we release the file lock. This is comparable to e.g. acquiring a mutex lock which is kept as long as we are holding the guard, and similar to [`tempfile`](https://crates.io/crates/tempfile), where [`TempDir::new`](https://docs.rs/tempfile/latest/tempfile/struct.TempDir.html#method.new) creates a new `TempDir`, which is cleaned up as soon as the `TempDir` value is dropped.

diegomrsantos: Makes sense, thanks for elaborating. I'm not familiar with the pattern, so I'll read more about it, as there might be tradeoffs, especially when testing.

diegomrsantos (approved):

PR #449, opened 2025-07-24
refactor(keysplit): Use `eth2_keystore` from LH instead of reimplementing it

dknopik:
Lighthouse has code for reading and decrypting eth2 keystores, so let's use it instead of our own


Zacholme7 (dismissed):
Not sure how I missed that, would have saved a lot of pain. Lgtm

dknopik:
Yeah, also did not think of it back then. Oh well :man_shrugging:

dknopik:
@mergify requeue

dknopik:
@Zacholme7 I needed to fix CI, can you please reapprove?

Zacholme7 (approved):

PR #451, opened 2025-07-24
chore: add claude md

diegomrsantos:
## Proposed Changes

This adds a CLAUDE.md file that provides guidance to Claude Code when working with this repository. The file includes:
- Common commands for building, testing, and linting
- Architecture overview
- Code organization details
- Development tips

🤖 Generated with [Claude Code](https://claude.ai/code)


Zacholme7:
Im not sure on best practices for LLM .md files, but I have had really good success using the `init-ai-docs` script from the ethpandaops team https://github.com/ethpandaops/ai-cookbook. Maybe we should run this instead

diegomrsantos:
> Im not sure on best practices for LLM .md files, but I have had really good success using the `init-ai-docs` script from the ethpandaops team https://github.com/ethpandaops/ai-cookbook. Maybe we should run this instead

This file was created by Claude itself :D

Zacholme7:
I know, but the ethpandaops scripts are much more guided and in depth which makes it work much better. For example their `create-implementation-plan` is miles better than the default claude planning mode

diegomrsantos:
> I know, but the ethpandaops scripts are much more guided and in depth which makes it work much better. For example their `create-implementation-plan` is miles better than the default claude planning mode

Could you run it here, and then we can compare?

Zacholme7:
ran it here #455, might be way overkill for what we want

dknopik (approved):

PR #452, opened 2025-07-25
chore: Update dependencies

dknopik:
Update all dependencies.

Notably,

- The lighthouse dependencies to the current state of the `anchor` branch. This is includes an update to recent unstable as well as
  - https://github.com/sigp/lighthouse/pull/7783 to enable QBFT value checks for attestations
  - https://github.com/sigp/lighthouse/pull/7761 to decouple us from `lighthouse_network`
  - updating https://github.com/sigp/lighthouse/pull/7016, which was already in `anchor` previously
- Alloy. This caused some changes around the creation of `Provider`s, renaming `on_*` to `connect_*`.
- Libp2p, gossipsub and the peer store. There were some API changes with the latter.

I tried to split this up, but all above are linked: LH and alloy need to be updated together as the native library `c-kzg` is used by both. LH and libp2p unfortunately need to be updated together due to our use of dependency version patching for the unreleased peer store.


diegomrsantos:
Are the code changes necessary after the update?

anchor/network/src/discovery.rs
```diff
@@ -585,31 +582,31 @@ pub fn build_enr(
...
+    match File::create(dir.join(Path::new("enr.dat")))
         .and_then(|mut f| f.write_all(enr.to_base64().as_bytes()))
     {
         Ok(_) => {
             debug!("ENR written to disk");
         }
         Err(e) => {
             warn!(
-                file = format!("{:?}{:?}",dir, ENR_FILENAME),
+                file = format!("{:?}{:?}",dir, "enr.dat"),
```
dknopik: Hardcoded temporarily until the `datadir` is merged

jking-aus (dismissed):
lgtm

dknopik:
Sorry, realized last second that this introduced a bug due to slightly changed semantics of the sync selection proof config. Fixed in https://github.com/sigp/anchor/pull/452/commits/fe8f956da8adf34e20276de02966b948b1f3e93e

jking-aus (approved):

PR #455, opened 2025-07-25
chore: ai cookbook init-ai-docs

Zacholme7:
## Issue Addressed
N/A

## Proposed Changes

This is an alternative to #451 that uses the [init ai docs](https://github.com/ethpandaops/ai-cookbook/tree/master/scripts) script from the ethpandaops team. With this, you start a claude session with [prime-context](https://github.com/ethpandaops/ai-cookbook/tree/master?tab=readme-ov-file#prime-context)

This is extremely in depth, not sure if we want something this large and comprehensive or keep it simple


diegomrsantos:
19k lines?

Zacholme7:
@diegomrsantos yeah I didnt know it was so in depth, thats why I said im not sure we want something this in depth

diegomrsantos:
> @diegomrsantos yeah I didnt know it was so in depth, thats why I said im not sure we want something this in depth

IMO it's way too much. What's the issue with the other approach? If it's too simple, we can ask Claude to add more content.

Zacholme7:
No issue, I just didnt know if we wanted something more in depth and I figured we could give this a try. Obviously an LLM will be way more effective priming its context with this approach and hence do better work on the codebase, but ya 20k lines is a lot

Issue #255, opened 2025-04-14
Update ports in ENR


dknopik:
We want to keep our ENR up to date, so we want to listen to discv5's SocketUpdated and libp2p's NewListenAddr and update our ENRs accordingly, unless the user has requested values explicitly. This can happen e.g. for solo stakers that get a fresh IP from their ISP.

Check Lighthouse for inspiration. This could also be done as part of an effort to generalise the discovery behaviour for inclusion in libp2p.

PR #457, opened 2025-07-28
feat(networking): update ports in ENR

petarjuki7:
## Issue Addressed

Addresses issue #255 

## Proposed Changes

Much of the logic insipred by Lighthouse.
Listening on discv5's `SocketUpdated` event and updating the local ENR socket.
Listening on libp2p's `NewListenAddr` and updating the ENR accordingly.

## Additional Info

Needed to finish #444


anchor/network/src/discovery.rs
```diff
@@ -257,14 +260,14 @@ impl Discovery {
             }
         }
 
-        // TODO: update local ports from libp2p events
+        // Update local ports from libp2p events
         // https://github.com/sigp/anchor/issues/255
```
dknopik: The link to the issue can also be removed now that it is addressed.

anchor/network/src/discovery.rs
```diff
@@ -337,6 +340,84 @@ impl Discovery {
...
+                // The field is already set to the same value, nothing to do
+                return Ok(false);
+            }
+            "quic"
+        };
+        let current_field = self.discv5.external_enr().read().quic4();
+        if current_field == Some(port) {
+            // The current field is already set, no need to update.
+            return Ok(false);
+        }
```
dknopik: I think this is accidentally duplicated with above?

anchor/network/src/discovery.rs
```diff
@@ -337,6 +340,84 @@ impl Discovery {
...
+    /// This is with caution. Discovery should automatically maintain this. This should only be
+    /// used when automatic discovery is disabled.
+    pub fn update_enr_udp_socket(&mut self, socket_addr: SocketAddr) -> Result<(), String> {
+        const IS_TCP: bool = false;
+        if self.discv5.update_local_enr_socket(socket_addr, IS_TCP) {
+            // persist modified enr to disk
+            save_enr_to_disk(Path::new(&self.enr_dir), &self.discv5.local_enr());
+        }
+        Ok(())
+    }
```
dknopik: This seems to be unused? Interestingly it is also unused in Lighthouse... I think we can get rid of it.

anchor/network/src/network.rs
```diff
@@ -215,15 +215,100 @@ impl<R: MessageReceiver> Network<R> {
...
                                 peer_id
                             );
-                        }
+                        },
+                        SwarmEvent::NewListenAddr { listener_id, address } => {
+                            trace!(
+                                ?listener_id,
+                                ?address,
+                                "Received NewListenAddr event from swarm"
+                            );
```
dknopik: Let's move everything in this branch to a new function.

anchor/network/src/discovery.rs
```diff
@@ -497,6 +582,30 @@ impl NetworkBehaviour for Discovery {
             // return the result to the peer manager
             return Poll::Ready(ToSwarm::GenerateEvent(DiscoveredPeers { peers }));
         }
+
+        match self.event_stream {
+            EventStream::Present(ref mut receiver) => {
+                while let Poll::Ready(Some(event)) = receiver.poll_recv(cx) {
+                    match event {
+                        discv5::Event::Discovered(_enr) => {}
```
dknopik: Remove this empty branch, will be caught by the fallback `_ => {}` below.

anchor/network/src/discovery.rs
```diff
@@ -337,6 +340,84 @@ impl Discovery {
...
+        // persist modified enr to disk
+        save_enr_to_disk(Path::new(&self.enr_dir), &self.discv5.local_enr());
+        Ok(true)
+    }
+
+    // TODO: Group these functions here once the ENR is shared across discv5 and lighthouse and
+    // Lighthouse can modify the ENR directly.
+    // This currently doesn't support ipv6. All of these functions should be removed and
+    // addressed properly in the following issue.
+    // https://github.com/sigp/lighthouse/issues/4706
```
dknopik: I think this LH specific comment can be removed. It references the `NetworkGlobals`, which we do not have.

dknopik (changes requested):
Thanks for the PR!

anchor/network/src/discovery.rs
petarjuki7: Yeah true, noticed that as well, just left it for completness. Will remove

anchor/network/src/discovery.rs
```diff
@@ -484,6 +545,7 @@ impl NetworkBehaviour for Discovery {
     ) {
     }
 
+    #[allow(clippy::single_match)]
```
petarjuki7, line 550: I'd rather keep it a single match, so it shows there are more variants we are ignoring for now.
I can change it to a `if let Some(event)...` if needed, no problem.

anchor/network/src/discovery.rs
```diff
@@ -337,6 +339,61 @@ impl Discovery {
         }
     }
 
+    /// Updates the local ENR TCP port.
```
diegomrsantos:
```rust
    /// Internal helper that updates a single ENR port field (IPv4 or IPv6) if it differs.
    ///
    /// - `desired_port` is the port we want to have in the ENR.
    /// - `is_ipv6` selects the v6 accessor/key when `true`, otherwise v4.
    /// - `key_v4` / `key_v6` are the ENR key names to write (e.g. `"tcp"` / `"tcp6"`).
    /// - `current_v4` / `current_v6` are accessors that read the current port
    ///   from the external ENR (`tcp4()`, `tcp6()`, `quic4()`, `quic6()`, etc.).
    ///
    /// Returns:
    /// - `Ok(true)`  — field updated and ENR persisted to disk.
    /// - `Ok(false)` — no change required (already set to `desired_port`).
    /// - `Err(_)`    — failed to write into the ENR.
    fn update_enr_port<Fv4, Fv6>(
        &mut self,
        desired_port: u16,
        is_ipv6: bool,
        key_v4: &'static str,
        key_v6: &'static str,
        current_v4: Fv4,
        current_v6: Fv6,
    ) -> Result<bool, String>
    where
        Fv4: Fn(&Enr<CombinedKey>) -> Option<u16>,
        Fv6: Fn(&Enr<CombinedKey>) -> Option<u16>,
    {
        // Check if the value is already set.
        {
            let external_enr = self.discv5.external_enr().read();
            let already_set = if is_ipv6 {
                current_v6(&external_enr) == Some(desired_port)
            } else {
                current_v4(&external_enr) == Some(desired_port)
            };
            if already_set {
                return Ok(false);
            }
        }

        // Update the appropriate ENR key.
        let enr_key = if is_ipv6 { key_v6 } else { key_v4 };
        self.discv5
            .enr_insert(enr_key, &desired_port)
            .map_err(|e| format!("{e:?}"))?;

        // Persist modified ENR to disk.
        save_enr_to_disk(Path::new(&self.enr_dir), &self.discv5.local_enr());
        Ok(true)
    }

    /// Update the ENR **TCP** port (IPv4 or IPv6).
    ///
    /// This only updates the port field in the ENR and **does not** modify the address.
    /// Discovery is expected to update the external address automatically.
    /// If you need to change the external address, use `update_enr_udp_socket`.
    ///
    /// Returns `Ok(true)` if the ENR was changed and persisted, `Ok(false)` if the
    /// existing value already matches `port`.
    pub fn update_enr_tcp_port(&mut self, port: u16, is_ipv6: bool) -> Result<bool, String> {
        self.update_enr_port(port, is_ipv6, "tcp", "tcp6", |e| e.tcp4(), |e| e.tcp6())
    }

    /// Update the ENR **QUIC** port (IPv4 or IPv6).
    ///
    /// This only updates the port field in the ENR and **does not** modify the address.
    /// Discovery is expected to update the external address automatically.
    /// If you need to change the external address, use `update_enr_udp_socket`.
    ///
    /// Returns `Ok(true)` if the ENR was changed and persisted, `Ok(false)` if the
    /// existing value already matches `port`.
    pub fn update_enr_quic_port(&mut self, port: u16, is_ipv6: bool) -> Result<bool, String> {
        self.update_enr_port(port, is_ipv6, "quic", "quic6", |e| e.quic4(), |e| e.quic6())
    }
}
```

dknopik: wdyt about keeping the `if is_ipv6` in `update_enr_tcp_port` and `update_enr_quic_port`? Then we only have the parameters `desired_port`, `key` and `current` to the main function, which seems easier to understand

anchor/network/src/discovery.rs
```diff
@@ -497,6 +559,29 @@ impl NetworkBehaviour for Discovery {
...
+            EventStream::Present(ref mut receiver) => {
+                while let Poll::Ready(Some(event)) = receiver.poll_recv(cx) {
+                    match event {
+                        discv5::Event::SocketUpdated(socket_addr) => {
+                            info!(ip = %socket_addr.ip(), udp_port = %socket_addr.port(),"Address updated");
+
+                            if (self.update_ports.tcp4 && socket_addr.is_ipv4())
+                                || (self.update_ports.tcp6 && socket_addr.is_ipv6())
+                            {
+                                self.discv5.update_local_enr_socket(socket_addr, true);
```
diegomrsantos: Why do we have to call `update_local_enr_socket` if `SocketUpdated` is described as  `/// Our local ENR IP address has been updated.`

anchor/network/src/discovery.rs
```diff
@@ -337,6 +339,61 @@ impl Discovery {
         }
     }
 
+    /// Updates the local ENR TCP port.
+    /// There currently isn't a case to update the address here. We opt for discovery to
+    /// automatically update the external address.
+    ///
+    /// If the external address needs to be modified, use `update_enr_udp_socket.
```
diegomrsantos: Why does Discovery automatically update only the IP, and do we need to update the port?

anchor/network/src/discovery.rs
diegomrsantos: Then we would keep some duplication. IMO, the current code isn't that hard to understand, but I don't have a strong opinion.

anchor/network/src/discovery.rs
petarjuki7: If I'm reading the `discv5` code correctly, the `SocketUpdated` event is emitted when only the UDP port is updated, with this we update our TCP port.

https://github.com/sigp/discv5/blob/ac91ad48e398665ef251746adcc68c5ed490854c/src/service.rs#L917-L933

anchor/network/src/discovery.rs
petarjuki7: If I understand correctly, discv5 automatically updates the IP and the UDP port in the ENR, but we have to manually update the TCP port because discv5 only manages UDP sockets.

anchor/network/src/discovery.rs
petarjuki7: I combined the two aproaches, the closures were elegant but maybe a bit hard to reason about at first glance IMO, can change it either way

anchor/network/src/discovery.rs
diegomrsantos: I see, thanks. The event and comment in `discv5` should be changed to make it clear.

petarjuki7: True, I agree

diegomrsantos: I believe that after this event, the IP address and the UDP port have been updated. `self.discv5.update_local_enr_socket(socket_addr, true)` updates the IP address and the TCP port, but it doesn't update the QUIC one. Would it be sufficient to update only the TCP and QUIC ports instead?

petarjuki7: Afaik the `discv5` doesn't "speak" QUIC, that's why we dont update it. We update it when we get the `NewListenAddr` event from libp2p because we can directly check if the QUIC port changed. We can maybe assume and update it here also, but this `SocketUpdated` event doesn't tell us about QUIC

diegomrsantos: discv5 doesn't know anything about the TCP port either

petarjuki7: True... This is the way it was done in Lighthouse so I took it as inspiration, should I also do `update_enr_quic_port()` here?

diegomrsantos: Thinking more about it, we can't set the discv5 UDP and QUIC port to the same value. I believe LH sets the TCP port here as a way to improve connectivity for nodes behind a NAT. But does nothing for QUIC. @AgeManning could you please clarify?

diegomrsantos:
@petarjuki7 

What do you think about using only one function to update the port?

```rust
    /// Try to update an ENR port based on port type and configuration.
    ///
    /// This method centralizes all port update logic in one place:
    /// 1. Checks if updates are allowed for this port type
    /// 2. Gets current port value from ENR
    /// 3. Updates the port if needed
    /// 4. Persists changes to disk
    ///
    /// Parameters:
    /// - `is_tcp`: Whether this is a TCP port (true) or QUIC port (false)
    /// - `is_ipv6`: Whether this is an IPv6 port (true) or IPv4 port (false)
    /// - `port`: The new port value to set
    ///
    /// Returns:
    /// - `Ok(true)`: Port was updated and persisted to disk
    /// - `Ok(false)`: No update was needed (config disallows it or port already matches)
    /// - `Err(String)`: Update failed with the given error message
    pub fn try_update_port(&mut self, is_tcp: bool, is_ipv6: bool, new_port: u16) -> Result<bool, String> {
        
        let (read_fn, key): (fn(&_) -> Option<u16>, &str) = match (is_tcp, is_ipv6) {
            (true, false) if self.update_ports.tcp4 => (Enr::tcp4, "tcp"),
            (true, true) if self.update_ports.tcp6 => (Enr::tcp6, "tcp6"),
            (false, false) if self.update_ports.quic4 => (Enr::quic4, "quic4"),
            (false, true) if self.update_ports.quic6 => (Enr::quic6, "quic6"),
            _ => return Ok(false)
        };
        let port_opt = read_fn(&self.discv5.external_enr().read());

        if port_opt == Some(new_port) {
            return Ok(false);
        }
        
        self.discv5
            .enr_insert(key, &new_port)
            .map_err(|e| format!("{e:?}"))?;

        save_enr_to_disk(Path::new(&self.enr_dir), &self.discv5.local_enr());
        
        Ok(true)
    }
  ```

**Comment by diegomrsantos** [comment]
Hi @petarjuki7 , are you still interested in completing this, or can I go ahead and continue?

**Comment by petarjuki7** [comment]
> Hi @petarjuki7 , are you still interested in completing this, or can I go ahead and continue?

Hi, if it's fine you can continue. Thanks @diegomrsantos

File: anchor/network/src/discovery.rs [hunk 11]
```diff
@@ -497,6 +545,42 @@ impl NetworkBehaviour for Discovery {
...
+            EventStream::Present(receiver) => {
+                while let Poll::Ready(Some(event)) = receiver.poll_recv(cx) {
+                    match event {
+                        discv5::Event::SocketUpdated(socket_addr) => {
+                            info!(ip = %socket_addr.ip(), udp_port = %socket_addr.port(),"Address updated");
+
+                            let was_updated = if socket_addr.is_ipv4() {
+                                self.try_update_port(true, false, socket_addr.port())
+                            } else {
+                                self.try_update_port(true, true, socket_addr.port())
```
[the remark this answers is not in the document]
diegomrsantos: The UDP has been updated already. We are assuming the external port is the same for TCP.

anchor/network/src/discovery.rs
```diff
@@ -489,6 +561,29 @@ impl NetworkBehaviour for Discovery {
             // return the result to the peer manager
             return Poll::Ready(ToSwarm::GenerateEvent(DiscoveredPeers { peers }));
         }
+
+        while let Some(event) = self.event_stream.recv(cx) {
```
dknopik, line 565: The syntax is now nicer here by pulling the event stream handling into its own function

anchor/network/src/discovery.rs
```diff
@@ -105,6 +107,24 @@ enum EventStream {
     InActive,
 }
 
+impl EventStream {
+    fn recv(&mut self, cx: &mut Context) -> Option<discv5::Event> {
+        if let EventStream::Awaiting(future) = self
+            && let Poll::Ready(Ok(receiver)) = future.as_mut().poll(cx)
+        {
+            *self = EventStream::Present(receiver);
+        }
```
dknopik, line 116: I added this as we else never actually have a present `EventStream`

dknopik (commented):
I found a bug while improving the code and decide to just push it instead of describing it:

anchor/network/src/discovery.rs
diegomrsantos: Could you elaborate more?

diegomrsantos (approved):
Thanks @petarjuki7 for starting this.

anchor/network/src/discovery.rs
dknopik: When we construct `Discovery`, we put an `EventStream::Awaiting` into `event_stream`. We need to convert it into an `EventStream::Present` somewhere.

dknopik (approved):

PR #458, opened 2025-07-28
refactor: Remove pre-genesis polling

dknopik:
## Proposed Changes

Replace `poll_whilst_waiting_for_genesis` with a simple sleep to genesis. This function will never be useful in production, as SSV requires an already existing network to bootstrap the SSV contract (and then supply the address of that contract to Anchor). The only reason against removing `wait_for_genesis` altogether is that someone might pre-deploy the contracts in a local devnet and spin up the node before genesis has occurred. However, this scenario does not require the extensive logging offered by `poll_whilst_waiting_for_genesis`.

## Additional Info

This is useful for updating dependencies in a later step as `get_lighthouse_staking` was removed upstream. (see #452)

### DIFF
--- a/anchor/client/src/lib.rs
+++ b/anchor/client/src/lib.rs
@@ -42,8 +42,9 @@ use subnet_service::{SUBNET_COUNT, SubnetId, start_subnet_service};
 use task_executor::TaskExecutor;
 use tokio::{
     net::TcpListener,
+    select,
     sync::{mpsc, mpsc::unbounded_channel},
-    time::sleep,
+    time::{interval, sleep},
 };
 use tracing::{debug, error, info, warn};
 use types::{EthSpec, Hash256};
@@ -764,7 +765,19 @@ async fn wait_for_genesis(genesis_time: u64) -> Result<(), String> {
             "Starting node prior to genesis",
         );
 
-        sleep(genesis_time - now).await;
+        let genesis_sleep = sleep(genesis_time - now);
+        tokio::pin!(genesis_sleep);
+        let mut log_interval = interval(Duration::from_secs(30));
+
+        loop {
+            select! {
+                _ = &mut genesis_sleep => break,
+                _ = log_interval.tick() => info!(
+                    seconds_to_wait = (genesis_time - now).as_secs(),
+                    "Waiting for genesis",
+                ),
+            }
+        }
 
         info!("Genesis has occurred");
     } else {
--- a/anchor/client/src/lib.rs
+++ b/anchor/client/src/lib.rs
@@ -761,14 +761,13 @@ async fn wait_for_genesis(genesis_time: u64) -> Result<(), String> {
     //
     // If the validator client starts before genesis, it will get errors from
     // the slot clock.
-    let now = get_now()?;
-    if now < genesis_time {
+    if get_now()? < genesis_time {
         info!(
-            seconds_to_wait = (genesis_time - now).as_secs(),
+            seconds_to_wait = (genesis_time - get_now()?).as_secs(),
             "Starting node prior to genesis",
         );
 
-        let genesis_sleep = sleep(genesis_time - now);
+        let genesis_sleep = sleep(genesis_time - get_now()?);
         tokio::pin!(genesis_sleep);
         let mut log_interval = interval(Duration::from_secs(30));
 
--- a/anchor/client/src/lib.rs
+++ b/anchor/client/src/lib.rs
@@ -749,9 +749,11 @@ async fn init_from_beacon_node<E: EthSpec>(
 }
 
 async fn wait_for_genesis(genesis_time: u64) -> Result<(), String> {
-    let now = SystemTime::now()
-        .duration_since(UNIX_EPOCH)
-        .map_err(|e| format!("Unable to read system time: {e:?}"))?;
+    let get_now = || {
+        SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .map_err(|e| format!("Unable to read system time: {e:?}"))
+    };
     let genesis_time = Duration::from_secs(genesis_time);
 
     // If the time now is less than (prior to) genesis, then delay until the
@@ -759,6 +761,7 @@ async fn wait_for_genesis(genesis_time: u64) -> Result<(), String> {
     //
     // If the validator client starts before genesis, it will get errors from
     // the slot clock.
+    let now = get_now()?;
     if now < genesis_time {
         info!(
             seconds_to_wait = (genesis_time - now).as_secs(),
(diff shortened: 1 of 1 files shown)


dknopik:
Meta: I was unsure what conventional commit type to pick for this one. `feat` feels wrong because we are removing a feature.

diegomrsantos:
> The only reason against removing wait_for_genesis altogether is that someone might pre-deploy the contracts in a local devnet and spin up the node before genesis has occurred

IMO, we should have code only for the current needs. If something is necessary in the future, we add it. Isn't the network already running anyway?

dknopik:
>IMO, we should have code only for the current needs. If something is necessary in the future, we add it. Isn't the network already running anyway?

The approach I described is valid already for construction of devnets - in particular, it can be used for `ssv-mini` as soon as there is development capacity to implement that approach there. Artificially introducing a blocker for that now seems counter-productive.

> Isn't the network already running anyway?

In the scenario I described, genesis might lie in the future.

anchor/client/src/lib.rs
```diff
@@ -750,10 +747,7 @@ async fn init_from_beacon_node<E: EthSpec>(
     Ok((genesis.genesis_time, genesis.genesis_validators_root))
 }
 
-async fn wait_for_genesis(
-    beacon_nodes: &BeaconNodeFallback<SystemTimeSlotClock>,
-    genesis_time: u64,
-) -> Result<(), String> {
+async fn wait_for_genesis(genesis_time: u64) -> Result<(), String> {
```
diegomrsantos, line 751: how about something like

```rust
use chrono::{DateTime, Utc};
use tokio::time::{interval, Duration};
use tracing::{info, debug};

/// Wait until `genesis_at`, emitting a progress line every 30 s.
/// This future is **cancellable** simply by dropping it (handled by TaskExecutor).
pub async fn wait_for_genesis(genesis_at: DateTime<Utc>) {
    // Restarted after genesis?
    if Utc::now() >= genesis_at {
        info!(?genesis_at, "Genesis already passed – skipping wait");
        return;
    }

    let mut ticker = interval(Duration::from_secs(30));

    loop {
        ticker.tick().await;                       // <-- only awaited future
        let now = Utc::now();
        if now >= genesis_at {
            info!(?genesis_at, "Genesis reached – continuing startup");
            break;
        }
        debug!(
            remaining = (genesis_at - now).num_seconds(),
            "Waiting for genesis…"
        );
    }
}
```

diegomrsantos: `Utc::now` comes from https://docs.rs/chrono/latest/chrono/

diegomrsantos: `Utc::now` isn't necessary, although it could be a better option. But logging from time to time while waiting seems a good idea.

dknopik: Pushed a suggestion that avoids chono, wdyt?

diegomrsantos: Looks good. I suggested chrono because it seems odd to have to handle an error when getting the current time

anchor/client/src/lib.rs
```diff
@@ -770,17 +765,21 @@ async fn wait_for_genesis(
...
+        tokio::pin!(genesis_sleep);
+        let mut log_interval = interval(Duration::from_secs(30));
+
+        loop {
+            select! {
+                _ = &mut genesis_sleep => break,
+                _ = log_interval.tick() => info!(
+                    seconds_to_wait = (genesis_time - now).as_secs(),
+                    "Waiting for genesis",
+                ),
```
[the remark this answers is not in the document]
dknopik: oh thanks mate

anchor/client/src/lib.rs
```diff
@@ -749,16 +749,19 @@ async fn init_from_beacon_node<E: EthSpec>(
...
+            .map_err(|e| format!("Unable to read system time: {e:?}"))
+    };
     let genesis_time = Duration::from_secs(genesis_time);
 
     // If the time now is less than (prior to) genesis, then delay until the
     // genesis instant.
     //
     // If the validator client starts before genesis, it will get errors from
     // the slot clock.
+    let now = get_now()?;
```
diegomrsantos: Do we need to create it, or can we use the fn directly?

dknopik: done

diegomrsantos (dismissed):

anchor/client/src/lib.rs
```diff
@@ -750,93 +748,50 @@ async fn init_from_beacon_node<E: EthSpec>(
...
     // If the time now is less than (prior to) genesis, then delay until the
     // genesis instant.
     //
     // If the validator client starts before genesis, it will get errors from
     // the slot clock.
-    if now < genesis_time {
+    if get_now()? < genesis_time {
         info!(
-            seconds_to_wait = (genesis_time - now).as_secs(),
+            seconds_to_wait = (genesis_time - get_now()?).as_secs(),
```
diegomrsantos, line 766: Can we return this result or zero if it's negative? Then we don't need the if/else, and we can have a log "Genesis occurred" inside the loop. Wdyt?

anchor/client/src/lib.rs
```diff
@@ -750,93 +748,50 @@ async fn init_from_beacon_node<E: EthSpec>(
...
             "Starting node prior to genesis",
         );
 
-        // Start polling the node for pre-genesis information, cancelling the polling as soon as the
-        // timer runs out.
-        tokio::select! {
-            result = poll_whilst_waiting_for_genesis(beacon_nodes, genesis_time) => result?,
-            () = sleep(genesis_time - now) => ()
+        let genesis_sleep = sleep(genesis_time - get_now()?);
```
diegomrsantos: Can we return this result or zero if it's negative? Then we don't need the if/else, and we can have a log "Genesis occurred" inside the loop. Wdyt?

dknopik: nice idea. wdyt?

anchor/client/src/lib.rs
```diff
@@ -750,93 +748,32 @@ async fn init_from_beacon_node<E: EthSpec>(
...
-        // timer runs out.
-        tokio::select! {
-            result = poll_whilst_waiting_for_genesis(beacon_nodes, genesis_time) => result?,
-            () = sleep(genesis_time - now) => ()
+    loop {
+        select! {
+            biased;
+            _ = &mut genesis_sleep => break,
+            _ = log_interval.tick() => {
+                let seconds_to_wait = genesis_sleep.deadline().duration_since(Instant::now()).as_secs();
```
diegomrsantos, line 767: Sorry for another one, could we use `Instant::now()` for the initial `now` too? Also, do you think we need a log for when genesis happens?

dknopik: I do not think so, as we need some association to the system time. `Instant` in opaque, we can't really derive the current time of the system through it. It works further below because we piggyback off the sleep, which has an instant as deadline.

I do not care strongly about the log.

diegomrsantos (approved):
Thanks a lot!

PR #459, opened 2025-07-28
refactor(network): Avoid specialized Lighthouse types

dknopik:
In #452, we will update dependencies, including Lighthouse. Following this update, we will not use `lighthouse_network` anymore (to avoid redundant dependencies, including libp2p). Certain types that are useful for both projects have been moved to `network_utils` - others are too specialized to warrant being shared. In preparation for that, this PR discontinues use of certain things `lighthouse_network` exports and defines its own variants.

This PR is split into individual changes, so reviewing commit by commit is likely easier.


anchor/network/src/discovery.rs
```diff
@@ -417,8 +417,7 @@ impl Discovery {
                             "Grouped subnet discovery query yielded no results.",
                         );
                     }
-                    Ok(r) => {
-                        let results = r.into_iter().map(|enr| (enr, None)).collect();
+                    Ok(results) => {
```
diegomrsantos, line 413: The Lighthouse code returns an Option<Instant> here. When I first coded it, I ignored it because its purpose was not clear. I'll take a look at it.

diegomrsantos: @AgeManning does the following make sense?

In Lighthouse, min_ttl manages ephemeral subnet connections based on specific validator duties and timing requirements. It's crucial because:

  1. Specific implementation: Tracks when peers can be disconnected after their subnet duties end
  2. Code detail: The logic in process_completed_queries (lines 926-953) maps ENRs to their furthest-future expiration time when a peer serves multiple subnets

  For SSV after Alan fork:

  1. Fixed subnet assignment: Unlike Lighthouse's dynamic duty-based subnet connections, SSV operators are permanently assigned to specific subnets based on committee
   ID hash (out of 128 total subnets)
  2. Committee-based: Operators form committees with fixed assignments; stakers register to these committees
  3. Deterministic placement: The hash of operator IDs determines subnet placement, with potential collisions

  Given these differences, min_ttl would be less relevant for Anchor because:
  - SSV operators maintain persistent subnet assignments rather than temporary duty-based ones
  - Subnet membership is deterministic and long-term, not requiring expiration tracking
  - Network connectivity needs are more static and predictable

diegomrsantos: If this makes sense, we also want to remove `min_ttl` from 
```rust
#[derive(Debug, Clone, PartialEq)]
struct SubnetQuery {
    subnet: SubnetId,
    min_ttl: Option<Instant>,
    retries: usize,
}
```

dknopik: The general notion is correct. I will remove the `min_ttl` there in another small PR
.

diegomrsantos: I think it's fine to remove it here

dknopik: Oh yeah, that was a smaller change than I thought. done :)

diegomrsantos (approved):

PR #462, opened 2025-07-31
Set bls_lagrange rand dependency to workspace.

### LINKED ISSUE #461: Recommend using workspace rand version in bls_lagrange
The bls_lagrange crate uses rand 0.8.5 when the workspace specifies 0.9.
A bit more fiddly when trying to pass in an RNG from another crate that also uses other workspace crates.

https://github.com/sigp/anchor/blob/dac9efc4bf1505770025af10317a9e654bd8c6b3/anchor/common/bls_lagrange/Cargo.toml#L17

gnattishness:
Update to rand@0.9 and replace deprecated names.

## Issue Addressed

Fixes #461 

## Proposed Changes

Use rand 0.9 and set to workspace version.

## Additional Info

Its a very small change, understandable if you don't want to accept it :)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1559,7 +1559,7 @@ dependencies = [
  "bls",
  "blst",
  "blstrs_plus",
- "rand 0.8.5",
+ "rand 0.9.1",
  "vsss-rs",
  "zeroize",
 ]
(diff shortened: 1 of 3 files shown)


dknopik:
This breaks when building with `feature = blsful`, as we pass the RNG to a dependency which still uses `rand` 0.8.

However, we have been using the custom `blst` variant for a while now and I am open to just removing the `blsful` variant. Then we could merge this.

gnattishness:
Ah thanks! I'd not checked with the blsful feature.
I'll have a bit of a play around and get back to you. If feasible to maintain, having the variant (which is supposed to be equivalent) seems helpful to test correctness

dknopik:
The old impl was removed and the dependency changed in #725

PR #463, opened 2025-07-31
refactor: replace custom MultiIndexMap with multi_index_map Crate

diegomrsantos:
## Issue Addressed

This PR replaces our custom MultiIndexMap implementation with the third-party [multi_index_map](https://crates.io/crates/multi_index_map) crate (v0.15.0). This change improves code quality and maintainability through:

### Benefits

  - Reduced Code Complexity: Eliminates our custom implementation in favor of a well-maintained library
  - Improved Type Safety: Uses derive macros to generate strongly typed indices
  - Enhanced Performance: Leverages optimized data structures for multi-indexed access patterns
  - Better Maintainability: Reduces the need to maintain custom data structures

## Proposed Changes

  - Added annotation-based index definitions with #[multi_index] attributes
  - Created indexed types (ShareIndexed, MetadataIndexed, ClusterIndexed) with appropriate index configurations
  - Updated API calls throughout the codebase to use the new indexing patterns
  - Adapted access patterns to use type-specific index functions (e.g., get_by_validator_pubkey() instead of generic get_by())
  - Modified field access to account for the new indexed wrapper types

### Technical Approach

  The implementation uses the macro-generated map types directly rather than creating additional abstraction layers. Index fields are marked with appropriate annotations (hashed_unique for unique indices
   and hashed_non_unique for non-unique indices) to maintain the same access patterns as the original implementation while taking advantage of the crate's optimized indexing structures

## Additional Info

closes https://github.com/sigp/anchor/issues/253
closes https://github.com/sigp/anchor/pull/399


anchor/database/src/lib.rs
```diff
@@ -16,16 +17,64 @@ use tokio::sync::{
...
+
+    #[multi_index(hashed_non_unique)]
+    pub committee_id: CommitteeId,
+
+    pub metadata: ValidatorMetadata,
+}
+
+/// All the clusters in the network
+#[derive(Debug, Clone, MultiIndexMap)]
+pub struct ClusterIndexed {
```
diegomrsantos, line 62: @Zacholme7 I think `validator_pubkey` wasn't necessary here

anchor/eth/src/index_sync.rs
```diff
@@ -137,12 +138,15 @@ async fn validator_index_syncer(
...
-            .get_by(&metadata.cluster_id)
-            .is_some_and(|c| !c.liquidated))
-    .then_some(metadata.public_key)
+        && clusters.get_by_cluster_id(&metadata.cluster_id).is_some()
+        && !clusters
+            .get_by_cluster_id(&metadata.cluster_id)
+            .unwrap()
+            .cluster
+            .liquidated)
+        .then_some(metadata.public_key)
```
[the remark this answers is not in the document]
diegomrsantos: fixed

anchor/database/src/lib.rs
```diff
@@ -16,16 +17,64 @@ use tokio::sync::{
...
-    state::NetworkState,
-};
+pub use crate::{error::DatabaseError, state::NetworkState};
+
+/// All the shares that belong to the current operator.
+/// IMPORTANT: There are parts of the code that assume this only contains shares that belong to the
+/// current operator. If this ever changes, make sure to update the code accordingly.
+#[derive(Debug, Clone, MultiIndexMap)]
+pub struct ShareIndexed {
+    #[multi_index(hashed_unique)]
```
Zacholme7, line 27: Id keep our comments giving some quick info on the purpose of each index and specific its uniquness

anchor/database/src/cluster_operations.rs
```diff
@@ -58,33 +60,43 @@ impl NetworkDatabase {
...
-            // Save the metadata for the validators
-            state.multi_state.validator_metadata.insert(
-                &validator.public_key,   // The public key of the validator
-                &cluster.cluster_id,     // The id of the cluster the validator belongs to
-                &cluster.owner,          // The owner of the cluster
-                &cluster.committee_id(), // The committee id of the cluster
-                validator.to_owned(),    // The metadata of the validator
-            );
+            // Check if we already have this cluster
+            let existing = state
```
Zacholme7, line 74: Can we define this as a helper on the struct?

Zacholme7 (commented):
Niceee, really great find on the this crate. This should probably be taken for a spin live before merging

anchor/database/src/lib.rs
diegomrsantos: Isn't it auto-explanatory in the current code?

Zacholme7: I guess because we are familiar with it right now, but If I look at this 6 months later im probably going to forget the relationship between all of these.

diegomrsantos: Isn't the relationship described by the uniqueness of the index?

anchor/database/src/cluster_operations.rs
```diff
@@ -58,33 +60,43 @@ impl NetworkDatabase {
...
+
+            // Only insert if it doesn't exist yet
+            if !existing {
+                state.multi_state.clusters.insert(ClusterIndexed {
+                    cluster_id: cluster.cluster_id,
+                    owner: cluster.owner,
+                    committee_id: cluster.committee_id(),
+                    cluster: cluster.to_owned(),
+                });
+            }
```
dknopik, line 88: Is checking first whether it already exists necessary? If it is uniquely indexed by the cluster ID, I'd assume it'd simply update the value if it already existed.

anchor/database/src/lib.rs
```diff
@@ -16,16 +17,64 @@ use tokio::sync::{
...
+    #[multi_index(hashed_non_unique)]
+    pub cluster_id: ClusterId,
+
+    #[multi_index(hashed_non_unique)]
+    pub owner: Address,
+
+    #[multi_index(hashed_non_unique)]
+    pub committee_id: CommitteeId,
+
+    pub share: Share,
```
dknopik, line 39: Wdyt about flattening this? 

Basically adding the attributes and missing fields directly to `Share` and removing `ShareIndexed`. 

Advantages:
- No redundant information, which might (due to some bug) deviate, or confuse the developer due to the redundancy.
- Smaller data size

Disadvantage:
- Introduces more fields to `Share`, which might not be necessary everywhere.
- The `ssv_types` crate has now a bunch of `multi_index` annotations, which can be argued to actually belong to the database.

Same of course for `ValidatorMetadata` and `Cluster` below.

Just an idea, I am not 100% sure on this.

anchor/eth/src/event_processor.rs
```diff
@@ -691,11 +694,26 @@ impl EventProcessor {
...
+                    "Failed to fetch validator metadata from database"
+                );
+                return Err(ExecutionError::Database(
+                    "Failed to fetch validator metadata from database".to_string(),
+                ));
+            }
+        };
+
+        // Now get the cluster using the cluster ID from metadata
+        let cluster_idx = match state.clusters().get_by_cluster_id(&metadata_idx.cluster_id) {
```
dknopik, line 712: This is now necessary due to removing the `validator_pubkey` index (see your comment alleging that it was unnecessary).

anchor/message_receiver/src/manager.rs
```diff
@@ -131,9 +131,9 @@ impl<S: SlotClock + 'static, D: DutiesProvider> MessageReceiver
                         // of operators.
                         let is_member = state
                             .clusters()
-                            .get_all_by(&committee_id)
-                            .next()
-                            .map(|c| c.cluster_members.contains(&own_id))
+                            .get_by_committee_id(&committee_id)
```
dknopik, line 134: can we please use `iter_by_` and `next` here to avoid allocating? This function is executed on every message, so it is a very hot path.

anchor/validator_store/src/lib.rs
```diff
@@ -175,37 +175,40 @@ impl<T: SlotClock, E: EthSpec> AnchorValidatorStore<T, E> {
...
-        let db_clusters = state.get_own_clusters().iter().collect::<Vec<_>>();
+        let db_clusters_ids = state.get_own_clusters().iter().collect::<Vec<_>>();
 
-        for (cluster, validator) in db_clusters
+        for (cluster_idx, validator) in db_clusters_ids
             .into_iter()
-            .filter_map(|id| state.clusters().get_by(id))
-            .filter(|cluster| !cluster.liquidated)
-            .flat_map(|cluster| {
+            .flat_map(|id| state.clusters().get_by_cluster_id(id))
```
dknopik, line 182: `iter_by_` as you turn it into an iterator anyway. (implicitly through `flat_mat`)

anchor/validator_store/src/lib.rs
```diff
@@ -175,37 +175,40 @@ impl<T: SlotClock, E: EthSpec> AnchorValidatorStore<T, E> {
...
+            .filter(|cluster_idx| !cluster_idx.cluster.liquidated)
+            .flat_map(|cluster_idx| {
                 state
                     .metadata()
-                    .get_all_by(&cluster.cluster_id)
-                    .map(move |metadata| (cluster, metadata))
+                    .get_by_cluster_id(&cluster_idx.cluster_id)
+                    .into_iter()
```
dknopik, line 188: `iter_by_` as you turn it into an iterator anyway.

anchor/subnet_service/src/lib.rs
```diff
@@ -297,21 +298,24 @@ pub fn get_committee_info_for_subnet(
...
         })
-        .map(|cluster| {
+        .map(|cluster_idx| {
             // Convert cluster to CommitteeInfo by getting validator indices
             let validator_indices = network_state
                 .metadata()
-                .get_all_by(&cluster.cluster_id)
-                .flat_map(|metadata| metadata.index)
+                .get_by_cluster_id(&cluster_idx.cluster_id)
+                .iter()
```
dknopik, line 313: use `iter_by_` as you turn it into an iterator anyway.

dknopik (commented):
Great find, I love it! 

I've put it on interop, looking good so far.

anchor/database/src/cluster_operations.rs
diegomrsantos: `cluster_id` is unique, and it means we can't insert the same ID twice. Should we update the other values in this case?

anchor/database/src/lib.rs
diegomrsantos: This index isn't used much, but when it is, it's for getting the share of a validator given a pub key. How would it be done instead?

anchor/eth/src/event_processor.rs
diegomrsantos: Yes. We can get the validator's cluster using the metadata index. This should be fast, and this isn't a critical path anyway. Wdyt?

anchor/database/src/cluster_operations.rs
dknopik: I think in practice it makes no difference as the data will be the same. Personally, I find the previous version simpler, but that is matter of taste. I don't feel strongly, so I'll mark this resolved

anchor/eth/src/event_processor.rs
dknopik: >We can't get the validator's cluster using metadata

Well, previously we could using the `validator_pubkey` within the metadata.

But as this really seems to be the only location where we used that index, I guess the manual lookup is fine.

anchor/database/src/cluster_operations.rs
diegomrsantos: Just clarify, it will be an error if we try to insert the same ID twice

anchor/eth/src/event_processor.rs
diegomrsantos: Sorry, I meant "we can", fixed it in my previous comment. I think you meant that previously we could get the `cluster` directly from the cluster index using the `validator_pubkey`.

anchor/database/src/lib.rs
dknopik: Sorry, I think my comment was unclear, let me give an example:

Instead of defining `ShareIndexed`, change `Share` from 

```rust
#[derive(Debug, Clone, Eq, PartialEq, Hash)]
pub struct Share {
    /// Public Key of the validator
    pub validator_pubkey: PublicKeyBytes,
    /// Operator this share belongs to
    pub operator_id: OperatorId,
    /// Cluster the operator who owns this share belongs to
    pub cluster_id: ClusterId,
    /// The public key of this Share
    pub share_pubkey: PublicKeyBytes,
    /// The encrypted private key of the share
    pub encrypted_private_key: [u8; ENCRYPTED_KEY_LENGTH],
}
```

to

```rust
#[derive(Debug, Clone, Eq, PartialEq, Hash, MultiIndexMap)]
pub struct Share {
    /// Public Key of the validator
    #[multi_index(hashed_unique)]
    pub validator_pubkey: PublicKeyBytes,
    /// Operator this share belongs to
    pub operator_id: OperatorId,
    /// Cluster the operator who owns this share belongs to
    #[multi_index(hashed_non_unique)]
    pub cluster_id: ClusterId,
    /// The public key of this Share
    pub share_pubkey: PublicKeyBytes,
    /// The encrypted private key of the share
    pub encrypted_private_key: [u8; ENCRYPTED_KEY_LENGTH],
}
```

The indices on `owner` and `committee_id` actually seem to be not used.

For the other two structs, we might move some fields into the underlying structs, but I think that might be fine.

diegomrsantos: We could also remove all the other fields apart from `pub share_pubkey: PublicKeyBytes` and ` pub encrypted_private_key: [u8; ENCRYPTED_KEY_LENGTH]` and leave them only in the index. Wdyt?

diegomrsantos: But not sure tbh. We can discuss it here, but I believe it can and should be done in a new PR anyway. I removed the `pub validator_pubkey: PublicKeyBytes` from the cluster index cause one test was failing as the semantics of uniqueness in our custom multi index were different.

anchor/message_receiver/src/manager.rs
diegomrsantos: Great catch! Done

anchor/subnet_service/src/lib.rs
diegomrsantos: Done

anchor/validator_store/src/lib.rs
diegomrsantos: Done

anchor/validator_store/src/lib.rs
diegomrsantos: Done. Let me know about the changes

anchor/database/src/cluster_operations.rs
dknopik: Ah! Sorry, was not aware. Then of course this is 100% correct.

anchor/validator_store/src/lib.rs
diegomrsantos: reverted https://github.com/lun3x/multi_index_map/issues/64

anchor/message_receiver/src/manager.rs
diegomrsantos: reverted https://github.com/lun3x/multi_index_map/issues/64

anchor/database/src/lib.rs
dknopik: Yeah, let's leave that for another PR. I'll put up an issue so that we do not forget it

dknopik (approved):
lgtm, thanks!

PR #477, opened 2025-08-11
fix: revert removal of `base_hash`

dknopik:
## Issue Addressed

#437 introduced a bug, where anchor fails to collect signatures for committees with a sync duty.

## Proposed Changes

This reverts #437.

Also change the incorrect comment, in order to prevent this from being removed again. Sorry!


diegomrsantos:
Imo it's better to stick only with the revert commit in this PR

diegomrsantos:
> Imo it's better to stick only with the revert commit in this PR

Actually, for this one it's not necessary, but imo it's a good practice. If, for some reason, we want to revert this PR, it'd also revert the other changes.

diegomrsantos (approved):

PR #478, opened 2025-08-11
feat: verify gossipsub subnets via observed subscriptions

diegomrsantos:
## Issue Addressed

This PR introduces subnet-based peer filtering by tracking observed gossipsub subscriptions to verify which subnets peers are actually subscribed to, rather than relying solely on ENR advertisements. The implementation prevents connections to peers that don't offer any needed subnets.

## Proposed Changes

- Tracks peer gossipsub subscription states through observed subscription/unsubscription events
- Adds connection filtering to reject peers not subscribed to needed subnets
- Implements periodic disconnection of peers that no longer offer needed subnets


anchor/network/src/peer_manager/connection.rs
```diff
@@ -68,6 +70,24 @@ impl ConnectionManager {
...
             max_with_priority_peers: max_priority_peers,
+            observed_peer_subnets: HashMap::new(),
+        }
+    }
+
+    /// External update from gossipsub events about peer subscription state
+    pub fn set_peer_subscribed(&mut self, peer: PeerId, subnet: SubnetId, subscribed: bool) {
+        let entry = self.observed_peer_subnets.entry(peer).or_default();
+
+        let idx = *std::ops::Deref::deref(&subnet) as usize;
```
dknopik: Why not the usual postfix `.deref()` notation?

anchor/network/src/peer_manager/connection.rs
```diff
@@ -197,6 +246,12 @@ impl ConnectionManager {
             .handle_established_inbound_connection(connection_id, peer, local_addr, remote_addr);
 
         let Err(denied) = limit_result else {
+            // Before accepting, ensure the peer offers at least one needed subnet
+            if !self.offers_any_needed(&peer, peer_store, needed_subnets) {
+                return Err(ConnectionDenied::new(std::io::Error::other(
```
dknopik: Why `std::io::Error::other`?

anchor/network/src/peer_manager/connection.rs
```diff
@@ -130,20 +150,47 @@ impl ConnectionManager {
...
         &self,
         peer: &PeerId,
         peer_store: &MemoryStore<Enr>,
     ) -> Option<Bitfield<Fixed<U128>>> {
+        if let Some(observed) = self.observed_peer_subnets.get(peer) {
+            return Some(observed.clone());
+        }
+        // Fallback to ENR-advertised subnets
         let enr = peer_store.get_custom_data(peer)?;
         discovery::committee_bitfield(enr).ok()
```
dknopik: As `observed_peer_subnets` is created on the first subscription, peers that do not subscribe at all always fall back to ENR advertised subnets.

anchor/network/src/peer_manager/connection.rs
diegomrsantos: Done, thanks. Why do we need to use it?

anchor/network/src/peer_manager/connection.rs
diegomrsantos: We need to use `std::io::Error`. What do you recommend?

anchor/network/src/peer_manager/connection.rs
diegomrsantos: But what if they want to, but haven't subscribed yet?

anchor/network/src/peer_manager/connection.rs
dknopik: `ConnectionDenied:new` accepts anything implementing `Into<Box<dyn error::Error + Send + Sync + 'static>>`.

This might e.g. also be `Box<NoOverlappingSubnets>`, where `NoOverlappingSubnets` is a custom type that impls `Error`.

anchor/network/src/peer_manager/connection.rs
dknopik: In order to get from `SubnetId` to the underlying integer type stored within

anchor/network/src/peer_manager/connection.rs
dknopik: Hmm. Maybe we need some kind of grace period? A couple of seconds?

anchor/network/src/peer_manager/connection.rs

anchor/network/src/peer_manager/connection.rs
diegomrsantos: Shouldn't a simple get be enough?

anchor/network/src/peer_manager/connection.rs
```diff
@@ -130,20 +161,47 @@ impl ConnectionManager {
                 continue;
             };
             for (&subnet_id, count) in subnet_ids.iter().zip(&mut peer_subnet_counts) {
-                if subnets.get(*subnet_id as usize).unwrap_or(false) {
+                let idx = *std::ops::Deref::deref(&subnet_id) as usize;
```
dknopik: Same deref thing here

anchor/network/src/peer_manager/connection.rs
dknopik: We need to convert `SubnetId` to `usize`. We do this here by calling our `Deref` implementation and then converting the `u64` to `usize`. What do you mean by "simple get"?

anchor/network/src/peer_manager/connection.rs
diegomrsantos: I had changed it, forgot to push

anchor/network/src/peer_manager/connection.rs
diegomrsantos: We can perform the fallback only when we connect and always use the observed peer subnets for the heartbeat.

diegomrsantos: let me know what you think

dknopik: neat! will run it on interop

dknopik (approved):
LGTM, thank you very much for tackling this!

Issue #436, opened 2025-07-17
Missing attestation validation


Zacholme7:
There are [spec tests](https://github.com/ssvlabs/ssv-spec/tree/main/ssv/spectest/tests/valcheck/valcheckattestations) for validating attestations which we do not seem to have

Issue #418, opened 2025-07-08
Consensus data validation


Zacholme7:
Have to explore where this is actually used, but the spec tests calls [validate](https://github.com/ssvlabs/ssv-spec/blob/dd0581c5d2a57063f04ea641e6b4f0ae2b2d4d4e/types/spectest/tests/validatorconsensusdata/test.go#L25) on the consensus data which is implemented [here](https://github.com/ssvlabs/ssv-spec/blob/dd0581c5d2a57063f04ea641e6b4f0ae2b2d4d4e/types/consensus_data.go#L164).

PR #480, opened 2025-08-11
refactor: Create `QbftDataValidator` for flexible QBFT data validation

dknopik:
## Issue Addressed

This is prerequisite for the proper implementation of:

- #418 
- #436

## Proposed Changes

Instead of having a `validate` function on `QbftData`, create a new trait for validation. This allows using contextual data for validation

### DIFF
--- a/anchor/common/qbft/src/lib.rs
+++ b/anchor/common/qbft/src/lib.rs
@@ -9,7 +9,7 @@ pub use qbft_types::{
 };
 use ssv_types::{
     OperatorId, Round,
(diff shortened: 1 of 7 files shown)


diegomrsantos:
Do both issues need to be addressed in the same PR, or can we create one for each?

dknopik:
>Do both issues need to be addressed in the same PR, or can we create one for each?

Sure. There are some common infrastructure changes for this, so I'll use this for those changes and then two on top for the actual issues.

dknopik:
Made smaller, other two PRs will be based on the changes within this one.

Zacholme7 (approved):
Nice, I like this

Issue #269, opened 2025-04-15
Make signature collector more robust


dknopik:
Currently, the signature collector happily combines the received signatures as soon as the threshold is reached. The resulting signature is assumed to be valid. However, if one of the partial signature is not correct, we recover an invalid signature - which is not detected. So, a malicious/faulty operator can easily cause us to publish signed objects with incorrect signatures.

We could adjust the signature collector to 
- verify the combined signature against the root to sign
- as soon as we have more partial signatures than the threshold, try out varying combinations of available partial signatures to find a combination that works

Alternatively, track the partial public keys, and verify the partial signatures against them.

dknopik:
See https://github.com/ssvlabs/ssv-spec/blob/5d6c4a70f831304e623f04f9976c2dae14c7973a/ssv/runner_validations.go#L34-L45

PR #486, opened 2025-08-12
feat: Make signature collector retry reconstruction with all possible combinations

dknopik:
## Issue Addressed

- #269

## Proposed Changes

- Verify the reconstructed signature
- On reconstruction failure, retry with different combinations of available partial signatures


dknopik:
Closing this - the risk of this behaviour being abused for denial of service is far too high, at worst we are trying $C(9, 13)$, which is too much, and can be easily triggered for public operators

PR #488, opened 2025-08-12
feat: QBFT value checks for `ValidatorConsensusData`

dknopik:
## Issue Addressed

- #418


anchor/common/ssv_types/src/consensus.rs
```diff
@@ -194,6 +201,172 @@ impl QbftData for ValidatorConsensusData {
...
+
+        // Always do this check, even if we're not validating slashing. This is to ensure that we
+        // have a decodable value.
+        let header = BlindedBeaconBlock::<E>::from_ssz_bytes_for_fork(&value.data_ssz, fork)
+            .map(|block| block.block_header())
+            .or_else(|_| {
+                FullBlockContents::<E>::from_ssz_bytes_for_fork(&value.data_ssz, fork)
+                    .map(|block| block.block().block_header())
+            })
+            .map_err(DataValidationError::DecodeError)?;
```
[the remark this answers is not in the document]
dknopik: Unfortunately, we can not determine this from the context. :/

anchor/common/ssv_types/src/consensus.rs
```diff
@@ -194,6 +201,172 @@ impl QbftData for ValidatorConsensusData {
...
+        if value.duty.validator_index != our_value.duty.validator_index {
+            return Err(DataValidationError::IndexMismatch {
+                expected: our_value.duty.validator_index,
+                got: value.duty.validator_index,
+            });
+        }
+
+        match value.duty.r#type {
+            BEACON_ROLE_AGGREGATOR => {
+                if value.version < DataVersion(ForkName::Electra) {
```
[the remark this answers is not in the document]
dknopik: IMO, the current variant is reasonably readable.

anchor/common/ssv_types/src/consensus.rs
```diff
@@ -194,6 +201,172 @@ impl QbftData for ValidatorConsensusData {
...
+            _phantom: PhantomData,
+        }
+    }
+
+    pub fn do_validation(
+        &self,
+        value: &ValidatorConsensusData,
+        our_value: &ValidatorConsensusData,
+    ) -> Result<(), DataValidationError> {
+        // Check whether the slot matches
```
Zacholme7, line 249: just looks like we are missing an epoch check here.
https://github.com/ssvlabs/ssv-spec/blob/d4c734909677b5fd3304e1d9220963cc9d794f09/ssv/value_check.go#L21

dknopik: Our check (exactly matching the slot) is tighter than their check of "somewhere in the current or next epoch".

Not sure why they chose that approach - the duty slot is fixed, and can be easily determined. Implementing their check would mean we have to add the slot clock here, for a worse check. @jking-aus what do you think?

Zacholme7: Ah yes, I think it is best to avoid the slot clock here and go with the tighter check

Zacholme7 (approved):

PR #493, opened 2025-08-13
feat: type spec test & spec test infra

Zacholme7:
## Issue Addressed

#39 

## Proposed Changes

This introduces the ssz infrastructure on our core types, the framework for setting up and running spec tests, and an implementation of the relevant type spec tests

1) We have migrated over to `VariableList` for key fields so that we are able to derive `TreeHash` and hence get the root for the struct. This is needed as this is the primary method of comparison for the tests.

2) This sets up a register/loader based approach for the spec tests where each test implements the `SpecTest` trait so that we able to properly load in the correct input json files and have a simple way to setup and run each test file

3) This nearly completes the type spec tests that are relevant to our implementation.  These tests are all pretty straightforwards and I dont anticipate any errors with the remaining ones so we may want to just remove them or finish them up if time permitting. 

Copied from closed issue regarding type tests

I have also deemed a few to be skippable
- The [committee member](https://github.com/ssvlabs/ssv-spec/tree/main/types/spectest/tests/committeemember) tests as our types differ substantially here and it would be a couple hundred lines of just parsing for basic quorum validation. I can just write a unit tests if this is really needed
- The [beacon deposit data test](https://github.com/ssvlabs/ssv-spec/tree/main/types/spectest/tests/beacon). This generates ETH deposit data, but this is never actually even used in the client.
- The [duty tests](https://github.com/ssvlabs/ssv-spec/tree/main/types/spectest/tests/duty). These are for their internal runner mapping which is an implementation detail, not spec level. 
- The [share encoding test](https://github.com/ssvlabs/ssv-spec/tree/main/types/spectest/tests/share). We use a different structure for share data so it does not make sense to test this encoding.

There are also some tests with data issues that I have raised
- the [ssz test](https://github.com/ssvlabs/ssv-spec/tree/main/types/spectest/tests/ssz) uses data from a beacon block with mocked signatures which when we try to parse into `BeaconBlock`, fails with `BLST_BAD_ENCODING`. Raised the issue [here](https://github.com/ssvlabs/ssv-spec/issues/563)


dknopik:
this seems to behave well on `interop1` btw :)

PR #500, opened 2025-08-14
fix: Pass in slot metadata to avoid errors on late collection start

dknopik:
## Issue Addressed

Fetching the `slot_metadata` in `collect_signature` is problematic: As the preceding QBFT instance may terminate only after the slot, the slot metadata may be expired by then. This causes "`ERROR anchor_validator_store: Got newer metadata - performance issues?`"

## Proposed Changes

Instead, have the caller fetch the slot metadata early and pass it in. Only require this when collecting in a `Committee` context - for single validator collection it is unneeded.

## Additional Info

This PR has the added benefit of only fetching the metadata when it is needed - this e.g. shortens the critical path for block proposals!


jking-aus (approved):
lgtm

PR #502, opened 2025-08-14
fix: Always save ENR on address update, do not set TCP port

dknopik:
## Issue Addressed

When `discv5` emits `SocketUpdated`, it has already updated the IP and UDP port in the ENR.

We need to persist this. Also, we should not set the TCP port, as the TCP port is not necessarily the same as the discovery port.

## Proposed Changes

- Always save the ENR to disk when receiving `SocketUpdated`
- Never store the TCP port when receiving `SocketUpdated`


jking-aus (approved):
yep, socket_addr.port() is the udp port and try_update_port(1,0) is for tcp. good correction.

PR #503, opened 2025-08-14
fix: Specify an empty list on index fetch for Teku compat

dknopik:
## Issue Addressed

Teku does not conform to the Beacon API spec: https://github.com/Consensys/teku/issues/9687

Violated line: https://github.com/ethereum/beacon-APIs/blob/69d2feb12a6047c4f21e628dea8b0135b7e9013a/apis/beacon/states/validators.yaml#L107

## Proposed Changes

Instead rely on the behaviour described here: https://github.com/ethereum/beacon-APIs/blob/69d2feb12a6047c4f21e628dea8b0135b7e9013a/apis/beacon/states/validators.yaml#L130

That is, supply an empty list instead of null/None.

## Further information

This is not a fix for Anchor, but rather a workaround.

### DIFF
--- a/anchor/eth/src/index_sync.rs
+++ b/anchor/eth/src/index_sync.rs
@@ -111,7 +111,7 @@ async fn validator_index_syncer(
                         .collect::<Vec<_>>();
                     async move {
                         client
-                            .post_beacon_states_validators(StateId::Head, Some(batch), None)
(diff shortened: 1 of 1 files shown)


jking-aus (approved):
yeah correct work around -- assume change does not impact other clients as per spec should just return all validators

Issue #446, opened 2025-07-24
Improve eth_getLogs contract event processing


yorickdowne:
Anchor 0.2 on an unsynced EL (Nethermind) will log `Processed contract events from block 573560 log_count=0`

Recommend a change:
- Start `eth_getLogs` only when `eth_syncing` returns `false`

Outlook to the future: With history expiry, `eth_getLogs` cannot be relied upon. Either add another parameter for a separate logs/receipts RPC, or work with SSV to get these data without relying on receipts.

dknopik:
Thanks for raising!

PR #504, opened 2025-08-14
fix: Only historic sync using a synced endpoint

dknopik:
## Issue Addressed

- #446

## Proposed Changes

Before starting historic sync, check for sync status of the connected EL client, and stall if not yet synced.


jking-aus (approved):
lgtm

PR #513, opened 2025-08-18
fix: Do not punish peers during sync

dknopik:
## Issue Addressed

During sync, we are happily processing messages. This is fine - it is worthwhile to keep up with the network and start peering ASAP. We will never sign anything during this period, but forward messages.

However, as our state might not be up-to-date yet, validation of valid messages might fail. We do not forward these messages and punish those peers.

## Proposed Changes

Do not forward invalid messages but do not punish peers as the messages might actually be valid.

## Additional Info

This contributes to the issues in `v0.3.0`.


jking-aus (approved):
lgtm

PR #514, opened 2025-08-18
fix: Calculate the sleep duration in every iteration

dknopik:
## Issue Addressed

Calling `sleep` in a loop means starting the sleep from the beginning on every iteration. When targeting a specific point in time, we must therefore recalculate the time until then every iteration, or start the sleep before the loop.

## Proposed Changes

Calculate the sleep duration in every iteration.


jking-aus (approved):
yep lgtm

PR #522, opened 2025-08-22
fix: Do not log exit event as unknown during historic sync

dknopik:
## Issue Addressed

During historic sync, we get a lot of "Unknown event signature, skipping", which are actually `ValidatorExited` events.

## Proposed Changes

Pull the `if live` into the variant to avoid triggering the fallback variant below.

### DIFF
--- a/anchor/eth/src/event_processor.rs
+++ b/anchor/eth/src/event_processor.rs
@@ -116,8 +116,12 @@ impl EventProcessor {
                     self.process_fee_recipient_updated(log, &tx)
                 }
 
-                SSVContract::ValidatorExited::SIGNATURE_HASH if live => {
-                    self.process_validator_exited(log)
+                SSVContract::ValidatorExited::SIGNATURE_HASH => {
+                    if live {
+                        self.process_validator_exited(log)
+                    } else {
+                        Ok(())
+                    }
                 }
                 _ => {
                     debug!(?topic0, "Unknown event signature, skipping");
--- a/anchor/eth/src/event_processor.rs
+++ b/anchor/eth/src/event_processor.rs
@@ -602,12 +602,12 @@ impl EventProcessor {
 
         if !live {
             if is_our_validator {
-                warn!(
+                debug!(
                     %validator_index,
                     "Ignoring historic validator exit for validator assigned to us"
                 );
             } else {
-                debug!(%validator_index, "Ignoring historic validator exit");
+                trace!(%validator_index, "Ignoring historic validator exit");
             }
             return Ok(());
         }
--- a/anchor/eth/src/event_processor.rs
+++ b/anchor/eth/src/event_processor.rs
@@ -117,11 +117,7 @@ impl EventProcessor {
                 }
 
                 SSVContract::ValidatorExited::SIGNATURE_HASH => {
-                    if live {
-                        self.process_validator_exited(log)
-                    } else {
-                        Ok(())
-                    }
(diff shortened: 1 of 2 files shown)


anchor/eth/src/event_processor.rs
```diff
@@ -116,8 +116,12 @@ impl EventProcessor {
                     self.process_fee_recipient_updated(log, &tx)
                 }
 
-                SSVContract::ValidatorExited::SIGNATURE_HASH if live => {
-                    self.process_validator_exited(log)
+                SSVContract::ValidatorExited::SIGNATURE_HASH => {
+                    if live {
+                        self.process_validator_exited(log)
+                    } else {
+                        Ok(())
```
diegomrsantos: Should we completely remove the log or add a more specific one about the exit?

dknopik: Matter of taste :man_shrugging: 

personally I think as we want to reduce logging for the happy path anyway, if we do add another log, it should be `trace`.

diegomrsantos: What would happen if this validator exit isn't processed because there aren't enough nodes live when it happens?

dknopik: the user could re-request it

diegomrsantos: However, we don't have any logs at all to inform us that those events are being ignored.

dknopik: yeah, we could add something that logs if and only if we skip an event for a validator that is relevant for us. otherwise its too noisy

diegomrsantos: If it's easy to do this check, we could log it as debug, otherwise as trace. Wdyt?

dknopik: done, wdyt? :)

anchor/eth/src/event_processor.rs
```diff
@@ -604,6 +600,18 @@ impl EventProcessor {
...
         let is_our_validator = self.is_our_validator(&validator_pubkey);
 
+        if !live {
+            if is_our_validator {
+                warn!(
+                    %validator_index,
+                    "Ignoring historic validator exit for validator assigned to us"
+                );
+            } else {
+                debug!(%validator_index, "Ignoring historic validator exit");
```
diegomrsantos: This should be normal most of the time, shouldn't it? I'd use debug for our validator and trace otherwise.

dknopik: done :)

anchor/eth/src/event_processor.rs
```diff
@@ -616,6 +616,18 @@ impl EventProcessor {
...
+            if is_our_validator {
+                debug!(
+                    %validator_index,
+                    "Ignoring historic validator exit for validator assigned to us"
+                );
+            } else {
+                trace!(%validator_index, "Ignoring historic validator exit");
+            }
+            return Ok(());
+        }
```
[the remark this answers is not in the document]
dknopik: yeah nah, because it is `is_our_validator` is also used further below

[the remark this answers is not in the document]
diegomrsantos: But when it's not live, we always return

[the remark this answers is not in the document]
diegomrsantos: Yeah I think you're right actually

diegomrsantos (approved):
Thanks!

PR #523, opened 2025-08-22
fix: Remove error log for missing operators

dknopik:
## Issue Addressed

In this function we log an error regarding missing operators even though we also return one.

As there is an event triggering this on mainnet, we have the unfortunate sideeffect to always log an ERROR during historic sync on mainnet. This is to be avoided.

## Proposed Changes

Remove the log.

### DIFF
--- a/anchor/eth/src/util.rs
+++ b/anchor/eth/src/util.rs
@@ -11,7 +11,7 @@ use reqwest::Client;
 use sensitive_url::SensitiveUrl;
 use ssv_types::{ClusterId, ENCRYPTED_KEY_LENGTH, OperatorId, Share, ValidatorMetadata};
 use tower::ServiceBuilder;
-use tracing::{debug, error, trace};
+use tracing::{debug, trace};
 use types::{Graffiti, PublicKeyBytes, Signature};
 
 use crate::{error::ExecutionError, sync::MAX_OPERATORS};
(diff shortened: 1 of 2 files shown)


anchor/eth/src/util.rs
```diff
@@ -168,7 +168,6 @@ pub fn validate_operators(
         .iter()
         .any(|id| !network_state.operator_exists(id))
     {
-        error!(cluster_id = ?cluster_id, "One or more operators do not exist");
```
diegomrsantos, line 171: Could you provide more details on this? We are still returning an error.

dknopik: Yes, the event is in fact still misformed, so it is appropriate to return an error. However, if we log an error, users will assume that their node encountered an error. But we are in fact **correctly** processing the message by discarding it. So it does not make sense to log an ERROR (alerting the user).

diegomrsantos: Should we log it with a different log level?

dknopik: I don't think so, the error returned is bubbled up and logged here appropriately: https://github.com/sigp/anchor/blob/7e1da12ecd124a184ff10eff614267e8ab578844/anchor/eth/src/event_processor.rs#L134-L138

diegomrsantos: Do you think we should add more info about the event itself for debugging purposes?

dknopik: Added the transaction hash to those logs, so that the log can be looked up later. WDYT?

diegomrsantos (approved):

Issue #524, opened 2025-08-23
docs: ` --output-path` is not a subcommand in keygen


shubhexists:
Hi, the `keygen` command of the `stable` branch doens't have a `--output-path` command.
The [docs](https://anchor.sigmaprime.io/running_an_operator#what-is-an-ssv-operator) still mention that.
```
# Generate encrypted keys (recommended for production)
anchor keygen --encrypt --output-path /path/to/keys/directory
```
Not sure is the name is changed or something?

<img width="947" height="604" alt="Image" src="https://github.com/user-attachments/assets/553857f5-5c83-414c-ac79-7328af86a4ed" />

dknopik:
Hi, thanks for raising this issue!

Apologies, it seems the docs were not correctly adjusted here. The `keygen` subcommand now generates the key into the `--data-dir` (or the default data dir if not explicitly specified). 

We will adjust the docs. We are actually considering adjusting the docs to be generated from the CLI definition to avoid such mistakes in future.

PS: nice shell prompt!

shubhexists:
Cool ! If it's not strictly for the internal team, I would love to help on this : D
Should I continue on with a sample implementation over this?

dknopik:
CC @chong-he: you mentioned you were going to look into the CLI docs update script, have you started on it yet?

chong-he:
> CC [@chong-he](https://github.com/chong-he): you mentioned you were going to look into the CLI docs update script, have you started on it yet?

It was near completion with the old `.md` file that I was working on previously: https://github.com/chong-he/anchor/commit/97c237005c3cdb6aeae80199782e9bcf00ac98ec, but then the book shifted the format, so I haven't got chance to look into it again. I will look into this again some time soon

PR #526, opened 2025-08-26
feat: added docgen for doc autogeneration

shubhexists:
## Issue Addressed
#524 

Which issue # does this PR address?
Auto generation of docs from clap cli args

## Additional Info
I thought implementing it directly via clap's `CommandFactory` is a much cleaner way to get all the arguments and generate docs out of it. This is my implementation of how I think it could be done.

Tried to match the exact structure of the docs which were currently there. Current issues/points that could be discussed could be - 
1) The description of some clap commands are pretty long which are eventually reflected in the `md` docs. 
2) Default values of some commands are not provided via `clap` which is making it hard to get the default values.....
3) Not sure if `examples` in md could be autogenerated. Haven't thought much about that yet.

PS - The current docs in this branch are generated via the `docgen`. 

cc for reviews: @chong-he @dknopik


anchor/src/docs.rs
```diff
@@ -0,0 +1,270 @@
...
+        doc.push_str("  --port 9100 \\\n");
+        doc.push_str("  --http \\\n");
+        doc.push_str("  --http-address 127.0.0.1 \\\n");
+        doc.push_str("  --http-port 9200 \\\n");
+        doc.push_str("  --unencrypted-http-transport \\\n");
+        doc.push_str("  --metrics \\\n");
+        doc.push_str("  --metrics-address 127.0.0.1 \\\n");
+        doc.push_str("  --metrics-port 9300 \\\n");
+        doc.push_str("  --password-file /path/to/your/password\n");
+        doc.push_str("```\n");
```
chong-he: This feels a bit of a clutter. Can we do this in one `doc.push_str()`, or maybe something like:
```
let example = ""
doc.push_str(example)
```

I think this looks a bit cleaner and easier to maintain. This applies to other examples that involves repeated `doc.push_str()` command

anchor/src/docs.rs
```diff
@@ -0,0 +1,270 @@
...
+        doc.push_str("## Onchain Keysplit Subcommand\n\n");
+        doc.push_str("```bash\n");
+        doc.push_str("anchor keysplit onchain [OPTIONS]\n");
+        doc.push_str("```\n\n");
+        doc.push_str(&Self::build_options_table(
+            Onchain::command(),
+            "Onchain-specific Options",
+        ));
+
+        doc.push_str("## Examples\n\n");
```
chong-he: Can we move this manual key split example to just after the `Manual Keysplit Subcommand`. I feel that the flow is better this way.
(I understand that you are putting it this way because the current doc does so)

anchor/src/docs.rs
```diff
@@ -0,0 +1,270 @@
+use clap::{Command, CommandFactory};
```
chong-he: I was thinking maybe we can put this file under `/anchor/client/src`? Mainly because this folder also  also sits `cli.rs` which is related to `doc.rs`, it can also make the import command in `/anchor/src/main.rs` a bit cleaner, since `use client::{}` is already in `main.rs`.

anchor/src/docs.rs
```diff
@@ -0,0 +1,270 @@
...
+        doc.push_str("### Encrypted key generation\n\n");
+        doc.push_str("This will create a `encrypted_private_key.json` file encrypted with the provided password ");
+        doc.push_str("and a `public_key.txt` file with the BASE64 encoded public key used for registering the ");
+        doc.push_str("operator. The password must be provided via `--password-file` or interactively when running ");
+        doc.push_str("Anchor.\n\n");
+        doc.push_str("```bash\n");
+        doc.push_str("anchor keygen --encrypt --output-path /path/to/keys\n");
+        doc.push_str("```\n\n");
+
+        doc.push_str("## Key Storage\n\n");
```
chong-he: This is not really necessary? (referring to line 126)

anchor/src/docs.rs
```diff
@@ -0,0 +1,270 @@
...
+        doc.push_str(
+            "Anchor will look for the key file in the default directory `~/.anchor/{network}`, ",
+        );
+        doc.push_str("or the directory specified by `--datadir`.\n");
```
chong-he: Similar to the example comment, for this type of docs, we can do 1 push_str per paragraph. That will look a bit cleaner

docs/docs/pages/cli-node.mdx
```diff
@@ -94,7 +97,7 @@ anchor node \
   --beacon-nodes https://beacon1.example.com,https://beacon2.example.com \
   --execution-rpc https://execution1.example.com,https://execution2.example.com \
   --execution-ws wss://execution1.example.com \
-  --listen-addresses 10.0.0.10 \
+  --listen-address 10.0.0.10 \
```
chong-he: In the `cli.rs`, it is `listen-addresses`. The same applies to the examples in the help text

anchor/src/docs.rs
```diff
@@ -0,0 +1,270 @@
...
+            let default = if arg.get_default_values().is_empty() {
+                "None".to_string()
+            } else {
+                let values: Vec<String> = arg
+                    .get_default_values()
+                    .iter()
+                    .map(|v| v.to_string_lossy().to_string())
+                    .collect();
+                format!("`{}`", values.join(", "))
+            };
```
chong-he: This is about #2 in your original post on default values, I can see that the `docgen` can capture the default values if there is `default_value` in the clap, example: https://github.com/sigp/anchor/blob/a79943b1ecd43c3453b75f9540327d0ea454e956/anchor/client/src/cli.rs#L217

There are some flags with `default_value_if` that are not captured. Not sure if we can have another `else if` arm to check for these flags? The `get_default_values` is returning this field: https://docs.rs/clap_builder/4.5.47/src/clap_builder/builder/arg.rs.html#4341

Although I don't see the method to "get default values if" but I wonder if we can do something similar like `arg.default_vals_ifs`?

chong-he (commented):
Thank you for the PR. This is nice! I like the use of `clap` built-in functions to extract the CLI help text. Some comments:
1. `cargo fmt` is failing. You can run `cargo +nightly fmt --all` to fix this

2. The goal of the docgen is to have it integrated in the CI, so that whenever a doc is modified, it must be updated or fails the CI. 

Therefore, could you also modify [this script](https://github.com/sigp/anchor/blob/stable/scripts/cli.sh) and [integrate in the CI](https://github.com/sigp/anchor/blob/ff84d7b1559631e1c261eceaecf569bfdae6e1ee/.github/workflows/test-suite.yml#L188-L201)? The script is copied from Lighthouse where we generate the CLI help text using `--help`, but in this case, we can do so with `anchor docgen`, and then compare with the existing docs. In the future, when devs are updating the doc, they just have to run the script and the doc can be updated. 

Thanks!

anchor/client/Cargo.toml
```diff
@@ -24,6 +24,7 @@ http_api = { workspace = true }
 http_metrics = { workspace = true }
 hyper = { workspace = true }
 keygen = { workspace = true }
+keysplit = { workspace = true }
```
diegomrsantos, line 27: Do we need to add keysplit?

dknopik: Agree. It would be nicer to have the docgen as a separate binary target of the top level crate IMO.

shubhexists: Hi what would be better, a separate binary target for docgen or have a top level library crate that it then called into the main anchor binary as a CLI command?

shubhexists:
Hi guys, any suggestions on the 2 unsolved comments?

anchor/client/Cargo.toml
diegomrsantos: My recommendation:  make docgen a separate workspace binary (e.g. tools/docgen) rather than an anchor client subcommand. This keeps the shipping CLI lean, avoids linking extra deps (e.g., keysplit) into the client crate.

Sketch:
	•	tools/docgen/Cargo.toml (bin = docgen)
	•	Depends on the crates that define the clap trees (e.g. anchor-client, keygen, keysplit) via path = "../anchor/client" etc.
	•	Each of those crates exposes a small pub fn cli() -> clap::Command (or impl CommandFactory) so docgen can introspect without booting the whole app.

anchor/src/docs.rs
diegomrsantos: Can we put the condition in `long_help`?

diegomrsantos:
Would using `clap_markdown` and a slim post-processor for the MDX layout be a better solution than creating a custom renderer? The less code unrelated to our project goals that we have to maintain, the better.

chong-he:
From my understanding, with this PR, the CLI flags itself, together with the description texts, are already auto-generated from `CommandFactory`, using the command `anchor docgen`. The examples are not auto-generated (I think the examples can't be auto-generated using either `CommandFactory` or `clap_markdown` as they are manually written). The current CLI reference contains table format, which the author has adjusted the doc to cater to the table format. 

The questions are:
- the examples require manual update if they change
- `doc.rs` may require manual update in the future if the table/section changes

One way we can do this is to just generate the help text using `anchor --help` (like [Lighthouse](https://lighthouse-book.sigmaprime.io/help_general.html) does and move the examples into a separate page. This also means the current table format of the help text will be gone - not sure if the current table format is necessary?

What does the Anchor team think?

diegomrsantos:
Hi @shubhexists, thank you for the work on this — the idea of auto-generating CLI docs is valuable and something we want to pursue.

Since this PR has been open for a while and the team discussion raised some design questions that remain unresolved, we're going to close this and open a focused issue capturing the agreed-upon approach. The key takeaways from the discussion:

1. **Use `clap_markdown` (or similar) + a slim post-processor** rather than a custom renderer — less code to maintain and fewer edge cases to handle.
2. **CI integration is essential** — the main value is a CI check that fails when CLI options change but docs aren't regenerated (as @chong-he outlined).
3. **The generator shouldn't live in the `client` crate** — it adds unnecessary deps to the runtime. A standalone tool or placing it alongside `main.rs` is cleaner.
4. **Hardcoded examples should stay hand-written** in the mdx files; the generator should only produce the options tables.

We'll reference this PR in the new issue. Thanks again for the contribution and sorry for the long wait on a clear direction.

PR #528, opened 2025-08-26
fix: Validation for prepared round changes before insertion

dknopik:
## Issue Addressed

If we have prepared a value, we include it in a round change messages and include the quorum of prepare justifications proving that value was prepared in the round_change_justifications field. These round change messages have to have the round_change_justifications prepare messages validated to ensure that these messages actually prove that the value was prepared. Otherwise, you could claim to prepare some value but not properly be able to support/prove it.

## Proposed Changes

Implement the missing validation and change existing code a bit to be able to reuse `is_valid_prepare_justification_for_round_and_root`.

### DIFF
--- a/anchor/common/qbft/src/lib.rs
+++ b/anchor/common/qbft/src/lib.rs
@@ -811,7 +811,6 @@ where
 
         let qbft_msg = &wrapped_msg.qbft_message;
         // If this is a "prepared" round change, we have to check the justifications.
-        // see https://github.com/ssvlabs/ssv/blob/7999b26308990b3f364d60e1618a8b6cf5b2d3e6/protocol/v2/qbft/instance/round_change.go#L304
         if qbft_msg.data_round > 0 {
             if !self.check_quorum(&qbft_msg.round_change_justification) {
                 debug!(
--- a/anchor/common/qbft/src/lib.rs
+++ b/anchor/common/qbft/src/lib.rs
@@ -834,11 +834,17 @@ where
             }
 
             for justification in qbft_msg.round_change_justification.iter() {
-                self.is_valid_prepare_justification_for_round_and_root(
+                if !self.is_valid_prepare_justification_for_round_and_root(
                     justification,
                     round,
                     &qbft_msg.root,
-                );
+                ) {
+                    debug!(
+                        from = *operator_id,
+                        "ROUNDCHANGE has invalid prepare justification"
+                    );
+                    return;
+                }
             }
         }
 
--- a/anchor/common/qbft/src/lib.rs
+++ b/anchor/common/qbft/src/lib.rs
@@ -1,4 +1,7 @@
-use std::{collections::HashMap, sync::Arc};
+use std::{
+    collections::{HashMap, HashSet},
+    sync::Arc,
+};
 
 // Re-Exports for Manager
 pub use config::{Config, ConfigBuilder};
@@ -234,6 +237,16 @@ where
         self.config.committee_members().contains(operator_id)
     }
 
+    /// Checks if we have a quorum of unique committee operators from these messages.
+    fn check_quorum<'a>(&self, msgs: impl IntoIterator<Item = &'a SignedSSVMessage>) -> bool {
+        let unique_operators = msgs
+            .into_iter()
+            .flat_map(|justification| justification.operator_ids())
+            .filter(|operator_id| self.check_committee(operator_id))
+            .collect::<HashSet<_>>();
+        unique_operators.len() >= self.config.quorum_size()
+    }
+
     // Perform base QBFT relevant message verification. This verfiication is applicable to all QBFT
     // message types
     // Return type expresses that we either have
@@ -434,7 +447,7 @@ where
 
         // If we are passed the first round, make sure that the justifications actually justify the
         // received proposal
-        if round > Round::default() && !self.validate_justifications(&wrapped_msg) {
+        if round > Round::default() && !self.validate_proposal_justifications(&wrapped_msg) {
             warn!(from = ?operator_id, "Justification verifiction failed");
             return;
         }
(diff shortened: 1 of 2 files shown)


anchor/common/qbft/src/lib.rs
```diff
@@ -788,6 +809,45 @@ where
             return;
         }
 
+        let qbft_msg = &wrapped_msg.qbft_message;
+        // If this is a "prepared" round change, we have to check the justifications.
+        // see https://github.com/ssvlabs/ssv/blob/7999b26308990b3f364d60e1618a8b6cf5b2d3e6/protocol/v2/qbft/instance/round_change.go#L304
```
diegomrsantos: Is this the right version to be used? In stage, it's in a different line https://github.com/ssvlabs/ssv/blob/stage/protocol/v2/qbft/instance/round_change.go#L272

dknopik: tbh i picked whatever was spit out by my ide's copy link feature :D

I'll just remove the link to avoid misleading

diegomrsantos: It's the same one used in the description. It's not only about using the proper link. As I raised some time ago, we need to agree on which SSV branch is the source of truth for our implementation, as there isn't a spec, and we consider the Go code as the one.

anchor/common/qbft/src/lib.rs
```diff
@@ -542,8 +554,6 @@ where
 
             // If the data_round > 1, that means we have prepared a value in previous rounds
             if round_change.data_round > 1 {
-                previously_prepared = true;
-
                 // also track the max prepared value and round
```
Zacholme7, line 602: I think we also need to validate all of the prepare justifications in the round change message. They have spec test that are testing this. I know there was talk about the nested justifications, so maybe this test is not even valid? 

[here](https://github.com/Zacholme7/anchor/blob/d2d0fb532df28d8c15ef7f283faf3bbbbd4a2abf/anchor/common/qbft/src/justification.rs#L175) is what I have for this. Still need to go through it closer

https://github.com/ssvlabs/ssv-spec/blob/d4c734909677b5fd3304e1d9220963cc9d794f09/qbft/spectest/tests/proposal/invalid_prepare_justification_round.go#L23

anchor/common/qbft/src/lib.rs
```diff
@@ -560,23 +579,60 @@ where
...
+            if round_change.height != *self.instance_height as u64 {
                 return false;
             }
 
-            // If the data_round > 1, that means we have prepared a value in previous rounds
-            if round_change.data_round > 1 {
+            // If the data_round > 0, that means we have prepared a value in previous rounds
+            // We also have to go through all of the prepare justifications in the round change to
+            // ensure that they are well formed and properly justify the prepared value
+            if round_change.data_round > 0 {
```
dknopik, line 601: Does the change from `> 1` to `> 0` mean that there is in fact a bug in `go-ssv`? Or did I incorrectly port behaviour/behavior?

Zacholme7: I took a look and I dont think go-ssv doing `> 1` anywhere or am I missing it? If they do that then it is a bug. If it was `> 1` then we would miss validation for values prepared in round 1.

Example:
Round one starts and we are not the leader but we reach a quorum of prepares and then time out so we send a round change with `data_round = 1` since it is the first round & we prepared a value and we include our prepare messages for the justifications.

Round two then starts and we are the leader. We will send a proposal with our round change messages that allowed us to move into round 2 and these will have `data_round = 1` since a value was prepared last round and we included all of the prepare messages with it. 

If we checked `> 1` then the validation of prepares from this first round would be ignored.

Does this make sense or do I have some flawed logic here?

dknopik: I mean this code location: https://github.com/ssvlabs/ssv/blob/7999b26308990b3f364d60e1618a8b6cf5b2d3e6/protocol/v2/qbft/instance/proposal.go#L158-L160

If the round is `== FirstRound` (which is 1), we immediately return `nil` (no error).

Zacholme7: that is `round` not `data_round`. The earliest we can have justification is round 2 but the earliest we can prepare a value is round 1 (`data_round = 1`)

dknopik: :facepalm: of course. Sorry!

diegomrsantos:
> ## Issue Addressed
> We're missing the validation at https://github.com/ssvlabs/ssv/blob/7999b26308990b3f364d60e1618a8b6cf5b2d3e6/protocol/v2/qbft/instance/round_change.go#L304
> 
> ## Proposed Changes
> Implement the missing validation and change existing code a bit to be able to reuse `is_valid_prepare_justification_for_round_and_root`.

Could you please describe what issue this PR is addressing and what changes are introduced? By this, I mean giving more context in English for people who aren't deeply familiar with the code and QBFT.

anchor/common/qbft/src/lib.rs
diegomrsantos: > that is `round` not `data_round`. The earliest we can have justification is round 2 but the earliest we can prepare a value is round 1 (`data_round = 1`)

@Zacholme7, could you elaborate on the reason for that and/or point to the part of the spec that discusses it?

Zacholme7: For which part? My example gives a walk through demonstrating it

The earliest we can prepare a value is round 1 because that is the first round so `data_round == 1` is the lowest it can be. The first time we can include justifications is when we send a round change in round 1 that is wanting to move to round 2, so the earliest we can have justifications is round 2

diegomrsantos: Thank you. It's clearer now for me.

Zacholme7:
> Could you please describe what issue this PR is addressing and what changes are introduced? By this, I mean giving more context in English for people who aren't deeply familiar with the code and QBFT.

If we have prepared a value, we include it in a round change messages and include the quorum of prepare justifications proving that value was prepared in the `round_change_justifications` field. These round change messages have to have the `round_change_justifications` prepare messages validated to ensure that these messages actually prove that the value was prepared. Otherwise, you could claim to prepare some value but not properly be able to support/prove it.

dknopik:
Thanks @Zacholme7 - I stole that for the description (and by extension for he commit message)

dknopik:
This one is good to go from my POV, but I can not approve ;( @Zacholme7 anything missing from your side?

diegomrsantos:
> This one is good to go from my POV, but I can not approve ;( @Zacholme7 anything missing from your side?

should we cherry-pick the test from https://github.com/sigp/anchor/pull/549?

dknopik:
Merged #549 into this PR.

Zacholme7:
Im going to run the qbft-agent at this to see if that picks anything up. Lots of little checks going on here

Zacholme7 (approved):
Nice

Zacholme7:
Only missing #559

PR #531, opened 2025-08-26
fix: disconnect blocked peers

diegomrsantos:
## Issue Addressed

This PR fixes an issue where blocked peers were not being disconnected from the network. The change ensures that peers with low scores are both blocked and immediately disconnected, rather than just being blocked.

## Proposed Changes

- Renamed variable for clarity and added immediate disconnection of blocked peers
- Extracted peer disconnection logic into a reusable helper method
- Ensured blocked peers are properly disconnected from the swarm


dknopik (approved):
bellissimo, thx

PR #537, opened 2025-08-27
fix: commit validation

Zacholme7:
## Issue Addressed

1) if we have accepted a proposal, we also have to check that the root of the commit message matches the accepted proposal root
2) If we have a quorum of commit messages
- if we are in commit state just make sure it matches the proposal root
- if we are in prepare state, make sure it matches proposal root and move into commit state
- if we are still awaiting a proposal, a quorum of commits moves us into the commit state

### DIFF
--- a/anchor/common/qbft/src/lib.rs
+++ b/anchor/common/qbft/src/lib.rs
@@ -730,6 +730,7 @@ where
                         warn!("COMMIT quorum root does not match accepted PROPOSAL root");
                         return;
                     }
+                    return;
                 }
                 InstanceState::Prepare { proposal_root } => {
                     // Transition to Commit state first
--- a/anchor/common/qbft/src/lib.rs
+++ b/anchor/common/qbft/src/lib.rs
@@ -746,16 +746,6 @@ where
                     }
                     self.state = InstanceState::Commit { proposal_root };
                 }
-                InstanceState::AwaitingProposal => {
-                    // Catch-up scenario: we have commit quorum without seeing proposal
-                    // This is valid - we can decide based on commit quorum alone
-                    // Transition directly to Commit state
-                    debug!("Received commit quorum without proposal - catch-up scenario");
-                    self.state = InstanceState::Commit {
-                        proposal_root: hash,
-                    };
-                    self.proposal_root = Some(hash);
-                }
                 _ => return,
             }
 
--- a/anchor/common/qbft/src/lib.rs
+++ b/anchor/common/qbft/src/lib.rs
@@ -696,11 +696,18 @@ where
             return;
         }
 
-        // Make sure that we have accepted a proposal for this round
-        if !self.proposal_accepted_for_current_round {
-            warn!(from=?operator_id, ?self.state, "Have not accepted Proposal for current round yet");
+        // If we have accepted a proposal, check that the commit matches it
+        // But allow commits without proposal (for catch-up scenarios)
+        if self.proposal_accepted_for_current_round {
+            if let Some(accepted_root) = self.proposal_root {
+                if wrapped_msg.qbft_message.root != accepted_root {
+                    return;
+                }
+            }
+        } else {
+            debug!(from=?operator_id, ?self.state, "Have not accepted Proposal for current round yet");
             return;
-        }
+        };
 
         debug!(from = ?operator_id, state = ?self.state, "COMMIT received");
 
@@ -716,17 +723,33 @@ where
         if let Some(hash) = self.commit_container.has_quorum(round) {
             // Make sure that the root of the data that we have come to a commit consensus on
             // matches the root of the proposal that we have accepted
-            let proposal_root = match self.state {
-                InstanceState::Prepare { proposal_root } => proposal_root,
-                InstanceState::Commit { proposal_root } => proposal_root,
-                _ => {
-                    warn!(from=?operator_id, ?self.state, "Not in PREPARE or COMMIT state");
-                    return;
+            match self.state {
+                InstanceState::Commit { proposal_root } => {
+                    // We already accepted a proposal and are in commit state
+                    if hash != proposal_root {
+                        warn!("COMMIT quorum root does not match accepted PROPOSAL root");
+                        return;
+                    }
                 }
-            };
-            if hash != proposal_root {
-                warn!("COMMIT quorum root does not match accepted PROPOSAL root");
-                return;
+                InstanceState::Prepare { proposal_root } => {
+                    // Transition to Commit state first
+                    if hash != proposal_root {
+                        warn!("COMMIT quorum root does not match accepted PROPOSAL root");
+                        return;
+                    }
+                    self.state = InstanceState::Commit { proposal_root };
+                }
+                InstanceState::AwaitingProposal => {
+                    // Catch-up scenario: we have commit quorum without seeing proposal
+                    // This is valid - we can decide based on commit quorum alone
+                    // Transition directly to Commit state
+                    debug!("Received commit quorum without proposal - catch-up scenario");
+                    self.state = InstanceState::Commit {
+                        proposal_root: hash,
+                    };
+                    self.proposal_root = Some(hash);
+                }
+                _ => return,
             }
 
             // Aggregate all of the commit messages
--- a/anchor/common/qbft/src/lib.rs
+++ b/anchor/common/qbft/src/lib.rs
@@ -699,10 +699,10 @@ where
         // If we have accepted a proposal, check that the commit matches it
         // But allow commits without proposal (for catch-up scenarios)
         if self.proposal_accepted_for_current_round {
-            if let Some(accepted_root) = self.proposal_root {
-                if wrapped_msg.qbft_message.root != accepted_root {
-                    return;
-                }
+            if let Some(accepted_root) = self.proposal_root
+                && wrapped_msg.qbft_message.root != accepted_root
+            {
+                return;
             }
         } else {
             debug!(from=?operator_id, ?self.state, "Have not accepted Proposal for current round yet");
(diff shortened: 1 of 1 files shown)


anchor/common/qbft/src/lib.rs
```diff
@@ -696,11 +696,18 @@ where
...
+            if let Some(accepted_root) = self.proposal_root
+                && wrapped_msg.qbft_message.root != accepted_root
+            {
+                return;
+            }
+        } else {
+            debug!(from=?operator_id, ?self.state, "Have not accepted Proposal for current round yet");
             return;
-        }
+        };
```
diegomrsantos:
```rust
// Handle commit message, checking proposal acceptance and catch-up scenarios.

// If we have NOT accepted a proposal for this round, this is a catch-up scenario.
// We allow commits without having seen a proposal in this case.
if !self.proposal_accepted_for_current_round {
    debug!(from=?operator_id, ?self.state, "Have not accepted Proposal for current round yet (catch-up scenario)");
    return;
}

// Proposal accepted: ensure commit matches the accepted proposal root.
if let Some(accepted_root) = self.proposal_root {
    if wrapped_msg.qbft_message.root != accepted_root {
        warn!(from=?operator_id, ?self.state, "Commit root does not match accepted Proposal root");
        return;
    }
} else {
    warn!(from=?operator_id, ?self.state, "Proposal accepted, but no proposal root found");
    return;
}
```

anchor/common/qbft/src/lib.rs
```diff
@@ -696,11 +696,18 @@ where
             return;
         }
 
-        // Make sure that we have accepted a proposal for this round
-        if !self.proposal_accepted_for_current_round {
-            warn!(from=?operator_id, ?self.state, "Have not accepted Proposal for current round yet");
+        // If we have accepted a proposal, check that the commit matches it
+        // But allow commits without proposal (for catch-up scenarios)
+        if self.proposal_accepted_for_current_round {
+            if let Some(accepted_root) = self.proposal_root
```
diegomrsantos: Should we return if `self.proposal_root` is `None`?

anchor/common/qbft/src/lib.rs
```diff
@@ -696,11 +696,18 @@ where
             return;
         }
 
-        // Make sure that we have accepted a proposal for this round
-        if !self.proposal_accepted_for_current_round {
-            warn!(from=?operator_id, ?self.state, "Have not accepted Proposal for current round yet");
+        // If we have accepted a proposal, check that the commit matches it
+        // But allow commits without proposal (for catch-up scenarios)
+        if self.proposal_accepted_for_current_round {
```
diegomrsantos: Do we need `proposal_accepted_for_current_round` at all, or could we use `self.proposal_root`?

anchor/common/qbft/src/lib.rs
Zacholme7: This looks good, thx

anchor/common/qbft/src/lib.rs
```diff
@@ -716,17 +723,33 @@ where
...
+                }
+                InstanceState::AwaitingProposal => {
+                    // Catch-up scenario: we have commit quorum without seeing proposal
+                    // This is valid - we can decide based on commit quorum alone
+                    // Transition directly to Commit state
+                    debug!("Received commit quorum without proposal - catch-up scenario");
+                    self.state = InstanceState::Commit {
+                        proposal_root: hash,
+                    };
+                    self.proposal_root = Some(hash);
```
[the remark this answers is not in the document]
Zacholme7: it is impossible to reach AwaitingProposal state if we are already checked if a proposal has been accepted. removed

anchor/common/qbft/src/lib.rs
diegomrsantos: Why has this been resolved?

Zacholme7: Whoops sorry I just saw the outdated. this is needed right now because when we send a round change we set proposal accepted to false but dont set proposal root back to None. 

Im not sure if there is a reason behind this, but if we reset both together then we can get rid of it

diegomrsantos: > when we send a round change we set proposal accepted to false but dont set proposal root back to None.

Wondering if this could create an issue

anchor/common/qbft/src/lib.rs
```diff
@@ -696,9 +696,23 @@ where
...
 
-        // Make sure that we have accepted a proposal for this round
+        // Handle commit message, checking proposal acceptance and catch-up scenarios.
+
+        // If we have NOT accepted a proposal for this round, this is a catch-up scenario.
+        // We allow commits without having seen a proposal in this case.
         if !self.proposal_accepted_for_current_round {
-            warn!(from=?operator_id, ?self.state, "Have not accepted Proposal for current round yet");
+            debug!(from=?operator_id, ?self.state, "Have not accepted Proposal for current round yet (catch-up scenario)");
+            return;
```
diegomrsantos, line 705: Inconsistent states are treated all the same with just a return; is this good enough?

anchor/common/qbft/src/lib.rs
```diff
@@ -696,9 +696,23 @@ where
...
+        }
+
+        // Proposal accepted: ensure commit matches the accepted proposal root.
+        if let Some(accepted_root) = self.proposal_root {
+            if wrapped_msg.qbft_message.root != accepted_root {
+                warn!(from=?operator_id, ?self.state, "Commit root does not match accepted Proposal root");
+                return;
```
diegomrsantos, line 760: Inconsistent states are treated all the same with just a return; is this good enough?

anchor/common/qbft/src/lib.rs
```diff
@@ -716,17 +730,23 @@ where
...
-                return;
+                InstanceState::Prepare { proposal_root } => {
+                    // Transition to Commit state first
+                    if hash != proposal_root {
+                        warn!("COMMIT quorum root does not match accepted PROPOSAL root");
+                        return;
+                    }
+                    self.state = InstanceState::Commit { proposal_root };
+                }
+                _ => return,
```
diegomrsantos, line 790: Is it ok to ignore all other states? No log necessary?

anchor/common/qbft/src/lib.rs
Zacholme7: Right now no because we always check to see if a proposal is accepted first before doing anything with the proposal root

anchor/common/qbft/src/lib.rs
Zacholme7: Wdym? if we have accepted a proposal and the commit does not match our proposal we do not want to do anything with it

anchor/common/qbft/src/lib.rs
Zacholme7: in the spec test branch I return an InvalidState error

anchor/common/qbft/src/lib.rs
diegomrsantos: We can leave it like this for now, but consider if we can use only `proposal_root`. Wdyt @dknopik ?

anchor/common/qbft/src/lib.rs
diegomrsantos: Right above it, we also return, and it seems to be a valid case, while here we also return and this seems to be an error.

anchor/common/qbft/src/lib.rs
diegomrsantos: So should we do it here?

Zacholme7: Ill add the log when I move everything to errors

anchor/common/qbft/src/lib.rs
Zacholme7: So youre saying I should just move that log above it to a warn too?

anchor/common/qbft/src/lib.rs
dknopik: Yeah there are several opportunities for improvement in the QBFT state type. 

I still hold onto the hope that we can properly spin out the QBFT crate - that is, make it more abstract over the actual contents and encoding of messages. This would also be a great opportunity to improve the internal types.

anchor/common/qbft/src/lib.rs
diegomrsantos: I'm wondering if we should return specific errors when it's actually an error. Is just returning for all cases good enough?

Zacholme7: I do return specific errors in the spec test branch when its an actual error. Im just holding off on getting that pr in until we have the core infrastructure complete

anchor/common/qbft/src/lib.rs
```diff
@@ -764,17 +778,23 @@ where
...
-            if hash != proposal_root {
-                warn!("COMMIT quorum root does not match accepted PROPOSAL root");
-                return;
+                InstanceState::Prepare { proposal_root } => {
+                    // Transition to Commit state first
+                    if hash != proposal_root {
+                        warn!("COMMIT quorum root does not match accepted PROPOSAL root");
+                        return;
+                    }
+                    self.state = InstanceState::Commit { proposal_root };
```
dknopik: Is switching to commit state correct? This basically means that we won't send a commit message ourself if enough prepares arrive before reaching commit quorum.

Zacholme7: You are right. it does feel weird not moving to a commit state if we have already seen a commit quorum but that is just the way of the state transition. fixed.

dknopik: Ah! in fact I missed that this is only checked on `commit_container.has_quorum` - in that case we are going to switch to `InstanceState::Complete` anyway, right? So it does not really matter anyway. Still, thanks :)

dknopik (approved):
LGTM

PR #538, opened 2025-08-27
fix: store messages without full data

dknopik:
## Issue Addressed

We do not need to store full data in the message containers - we store it separately in the instance. Not storing the full data with the message avoids sending it the message with full data when sending message with justifications.

## Proposed Changes

Replace the fulldata with an empty vec on send,

### DIFF
--- a/anchor/common/qbft/src/msg_container.rs
+++ b/anchor/common/qbft/src/msg_container.rs
@@ -43,6 +43,10 @@ impl MessageContainer {
             return false; // Duplicate message
         }
 
+        let mut msg = msg.clone();
+        // We have no longer have need for full data in these messages
+        msg.signed_message.set_full_data(vec![]);
+
         // Add message and track its value
         self.messages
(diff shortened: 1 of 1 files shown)


Zacholme7 (dismissed):
yep nice https://github.com/ssvlabs/ssv-spec/blob/d4c734909677b5fd3304e1d9220963cc9d794f09/qbft/messages.go#L134

anchor/common/qbft/src/msg_container.rs
```diff
@@ -42,13 +42,18 @@ impl MessageContainer {
             return false;
         }
 
-        self.messages.entry(round).or_default().push(msg.clone());
+        let mut msg = msg.clone();
+        // We no longer have need for full data in these messages
```
diegomrsantos: Imo it's better to add a comment describing the why

Zacholme7 (approved):

PR #541, opened 2025-08-27
fix: partial roundchange quorum

dknopik:
Partial round change quorums work by getting the lowest round which achieves partial quorum using all rounds above.
Additionally, the `current_round` is already set.

### DIFF
--- a/anchor/common/qbft/src/msg_container.rs
+++ b/anchor/common/qbft/src/msg_container.rs
@@ -76,16 +76,28 @@ impl MessageContainer {
         round: Round,
         partial: usize,
     ) -> Option<Round> {
-        let mut operators_seen = HashSet::new();
-        for (&round, operators) in self.senders_by_round.range((round + 1)..).rev() {
-            for &operator in operators {
-                operators_seen.insert(operator);
+        // Collect all operators from rounds > round
+        let mut all_operators = HashSet::new();
+        let mut min_future_round = None;
+
+        for (&r, operators) in self.senders_by_round.range((round + 1)..) {
+            // Track minimum round
+            if min_future_round.is_none() {
+                min_future_round = Some(r);
             }
-            if operators_seen.len() >= partial {
-                return Some(round);
+
+            // Add all operators from this round
+            for &operator in operators {
+                all_operators.insert(operator);
             }
         }
-        None
+
+        // If we have partial quorum, return the minimum round
+        if all_operators.len() >= partial {
+            min_future_round
+        } else {
+            None
+        }
     }
 
     /// If we have a quorum for the round, get all of the messages that correspond to that quorum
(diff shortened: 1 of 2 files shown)


anchor/common/qbft/src/msg_container.rs
```diff
@@ -71,11 +71,21 @@ impl MessageContainer {
...
-        self.messages
-            .get(&round)
-            .map(|msgs| msgs.len())
-            .unwrap_or(0)
+    pub fn highest_partial_quorum_above_round(
+        &self,
+        round: Round,
+        partial: usize,
+    ) -> Option<Round> {
+        let mut operators_seen = HashSet::new();
```
Zacholme7: Okay so I am pretty sure this should be returning the minimum round instead here. 

[here](https://github.com/ssvlabs/ssv-spec/blob/d4c734909677b5fd3304e1d9220963cc9d794f09/qbft/round_change.go#L71-L73) is where they get the min round

something like
```rust
pub fn highest_partial_quorum_above_round(
    &self,
    round: Round,
    partial: usize,
) -> Option<Round> {
    // Collect all operators from rounds > round
    let mut all_operators = HashSet::new();
    let mut min_future_round = None;
    
    for (&r, operators) in self.senders_by_round.range((round + 1)..) {
        // Track minimum round
        if min_future_round.is_none() {
            min_future_round = Some(r);
        }
        
        // Add all operators from this round
        for &operator in operators {
            all_operators.insert(operator);
        }
    }
    
    // If we have partial quorum, return the minimum round
    if all_operators.len() >= partial {
        min_future_round
    } else {
        None
    }
}
```

Zacholme7: I took this change for a run on the spec test branch and it was a drop in replacement for the previous helper I had

dknopik: doh - of course. Thanks, good spot!

Zacholme7 (approved):
Nice

anchor/common/qbft/src/msg_container.rs
```diff
@@ -11,7 +11,7 @@ pub struct MessageContainer {
     /// Messages stored as a Vec per round to preserve insertion order
     messages: HashMap<Round, Vec<WrappedQbftMessage>>,
     /// Track which operators have sent messages for each round
-    senders_by_round: HashMap<Round, HashSet<OperatorId>>,
+    senders_by_round: BTreeMap<Round, HashSet<OperatorId>>,
```
diegomrsantos, line 14: why a BTreeMap?

dknopik: Because we need to access the rounds in order. `HashMap` does not offer that. A `BTree` automatically maintains order.

PR #548, opened 2025-08-28
fix(qbft): commit message handling in catch-up scenarios

diegomrsantos:
## Issue Addressed

This PR addresses a bug in QBFT commit message handling where commit messages are dropped when received before a proposal, preventing consensus achievement in out-of-order message scenarios.

## Proposed Changes

- Add comprehensive test `test_consensus_with_commits_before_proposal` that verifies QBFT can achieve consensus when commit messages arrive before proposals
- The test demonstrates the expected behavior: nodes should achieve consensus when they have both a valid proposal and commit quorum, regardless of message arrival order
- This ensures QBFT provides proper message ordering independence as required by distributed consensus protocols

## Additional Info

The current implementation drops commit messages when `proposal_accepted_for_current_round` is false. The fix involves implementing proper message buffering and re-evaluation when proposals arrive.

**Test Scenario:**
1. Node receives commit messages from other nodes (commit quorum available)  
2. Node later receives the corresponding proposal
3. Node achieves consensus with the committed value

This ensures network resilience by allowing nodes to properly participate in consensus even when experiencing temporary network issues or joining late.

### DIFF
--- a/anchor/common/qbft/src/tests.rs
+++ b/anchor/common/qbft/src/tests.rs
@@ -223,3 +223,128 @@ fn test_node_recovery() {
     let num_consensus = test_instance.wait_until_end();
     assert_eq!(num_consensus, 5); // Should reach full consensus after recovery
 }
+
+#[test]
+/// Test that demonstrates QBFT incorrectly drops commit messages when no proposal accepted
+///
+/// In a proper QBFT implementation, commit messages should be buffered when they arrive
+/// before a proposal, allowing catch-up scenarios where a node can achieve consensus
+/// based on a commit quorum even without seeing the original proposal.
+///
+/// Current bug: Individual commit messages are dropped when
+/// proposal_accepted_for_current_round is false, preventing nodes from ever
+/// reaching commit quorum in catch-up scenarios.
+fn test_commit_messages_dropped_without_proposal_acceptance() {
+    if ENABLE_TEST_LOGGING {
+        let env_filter = EnvFilter::new("debug");
+        let _ = tracing_subscriber::fmt()
+            .compact()
+            .with_env_filter(env_filter)
+            .try_init();
+    }
+
+    use ssv_types::{
+        consensus::QbftMessage,
+        message::{MsgType, RSA_SIGNATURE_SIZE, SSVMessage, SignedSSVMessage},
+    };
+
+    // Create QBFT instance with 3 nodes (f=0, quorum=3)
+    let config = ConfigBuilder::<DefaultLeaderFunction>::new(
+        1.into(),
+        InstanceHeight::default(),
+        (1..4).map(OperatorId::from).collect(), // 3 nodes, quorum = 3
+    )
+    .with_operator_id(OperatorId::from(1))
+    .build()
+    .expect("config should be valid");
+
+    let test_data = TestData(789);
+    let mut qbft_instance = Qbft::new(
+        config,
+        test_data.clone(),
+        Box::new(NoDataValidation),
+        MessageId::from([0; 56]),
+        |_| {},
+    );
+
+    // Verify initial state: no proposal accepted
+    assert!(!qbft_instance.proposal_accepted_for_current_round);
+    assert!(matches!(
+        qbft_instance.state,
+        InstanceState::AwaitingProposal
+    ));
+
+    // STEP 1: Send commit messages BEFORE accepting any proposal (catch-up scenario)
+    // This simulates a node that missed the proposal but receives commit messages from other nodes
+
+    let _commit_messages_before = qbft_instance
+        .commit_container
+        .get_messages_for_round(1.into())
+        .len();
+
+    // Create 3 valid commit messages for the same data
+    for operator_id in [1, 2, 3] {
+        let commit_msg = QbftMessage {
+            qbft_message_type: QbftMessageType::Commit,
+            height: 0,
+            round: 1,
+            identifier: [0; 56].to_vec().into(),
+            root: test_data.hash(),
+            data_round: 0,
+            round_change_justification: vec![],
+            prepare_justification: vec![],
+        };
+
+        let commit_ssv_message = SSVMessage::new(
+            MsgType::SSVConsensusMsgType,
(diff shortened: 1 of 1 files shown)


Zacholme7:
So we should just store the commit in the container before checking if the proposal was accepted?

diegomrsantos:
> So we should just store the commit in the container before checking if the proposal was accepted?

From what I understand, we can receive the messages out of order, so we need to buffer the commit messages in case we receive the proposal later. Do you think it makes sense?

diegomrsantos:
The QBFT Specification Requirement

  According to the QBFT specification, if:
  1. We have a valid proposal for a value
  2. We receive a quorum of commit messages for that same value

  Then we should accept the value as committed, regardless of the order these messages arrived.

  Why Buffering is Necessary

  In distributed systems, messages can arrive out of order:

  Scenario 1 (Normal Order):

  1. Receive proposal for value X
  2. Accept proposal
  3. Receive commit messages for value X
  4. Achieve commit quorum → Success ✅

  Scenario 2 (Out of Order - Catch-up):

  1. Receive commit messages for value X first (from other nodes)
  2. Receive proposal for value X later
  3. Accept proposal
  4. Re-evaluate buffered commits → Achieve commit quorum → Success ✅

  Current Anchor Bug

  The current code at lines 748-750:
  ```rust
  if !self.proposal_accepted_for_current_round {
      warn!(from=?operator_id, ?self.state, "Have not accepted Proposal for current round yet");
      return; // ❌ DROPS commit messages instead of buffering
  }
```

  This violates the QBFT specification because it prevents Scenario 2 from working correctly.

  Correct Implementation Should:

  1. Always buffer commit messages (regardless of proposal status)
  2. When a proposal is accepted, re-evaluate all buffered commit messages
  3. Check for commit quorum using both new and previously buffered commits

  This ensures that message ordering doesn't affect consensus correctness, which is a fundamental requirement for robust distributed consensus protocols like QBFT.

File: anchor/common/qbft/src/tests.rs [hunk 1]
```diff
@@ -223,3 +223,184 @@ fn test_node_recovery() {
...
+    use ssv_types::{
+        consensus::QbftMessage,
+        message::{MsgType, RSA_SIGNATURE_SIZE, SSVMessage, SignedSSVMessage},
+    };
+
+    // Create QBFT instance with 3 nodes (f=0, quorum=3)
+    let config = ConfigBuilder::<DefaultLeaderFunction>::new(
+        1.into(),
+        InstanceHeight::default(),
+        (1..4).map(OperatorId::from).collect(), // 3 nodes, quorum = 3
```
dknopik: this should be `1..=4`

anchor/common/qbft/src/lib.rs
```diff
@@ -560,7 +560,7 @@ where
             }
 
             // If the data_round > 1, that means we have prepared a value in previous rounds
-            if round_change.data_round > 1 {
+            if round_change.data_round > 0 {
```
dknopik: How is this related to the test added?

anchor/common/qbft/src/tests.rs
diegomrsantos: Why? Shouldn't a committee of 3 also work? It's just not fault-tolerant, which isn't relevant for this test.

anchor/common/qbft/src/lib.rs
diegomrsantos: Yeah sorry, it shouldn't be here

anchor/common/qbft/src/tests.rs
dknopik: because later you are doing "`for operator_id in [2, 3, 4]`"

diegomrsantos: Indeed, good catch. Thanks! Why was there no error related to that?

dknopik: from the CI run:     `2025-08-29T11:22:07.688989Z  WARN qbft: Signer is not part of committee`

diegomrsantos: > from the CI run: `2025-08-29T11:22:07.688989Z WARN qbft: Signer is not part of committee`

nice

Zacholme7:
Okay good job, I did some more looking into this and it does appear to be a valid issue. This is spec compliance, but anchor and go-ssv both do not implement this. I think our first step would be to bring it up to them in the discord and coordinate a change. I dont think this change would be huge, but it does require refactoring our some pieces of our logic and thinking through every case where this is needed. 

We are just doing the same thing as go-ssv atm

diegomrsantos:
> Okay good job, I did some more looking into this and it does appear to be a valid issue. This is spec compliance, but anchor and go-ssv both do not implement this. I think our first step would be to bring it up to them in the discord and coordinate a change. I dont think this change would be huge, but it does require refactoring our some pieces of our logic and thinking through every case where this is needed.
> 
> We are just doing the same thing as go-ssv atm

I agree, let's do it

Zacholme7:
Think we should close these for now and make an issue for each linking these prs?

diegomrsantos:
> Think we should close these for now and make an issue for each linking these prs?

But thinking more about it, what would be the issue if Anchor fixes it while Go doesn't?

Zacholme7:
I dont think there is an issue, but atm I think our focus should be making sure we are 100% confident in our implementations correctness with respect to go-ssv and then coordinate these extra changes.

diegomrsantos:
> I dont think there is an issue, but atm I think our focus should be making sure we are 100% confident in our implementations correctness with respect to go-ssv and then coordinate these extra changes.

Shouldn't we be correct wrt to the spec? If this is really an issue that happens in production, although probably not often, an Anchor only cluster could have worse performance. But I agree it might not be high in the priority. @jking-aus @AgeManning what are your thoughts?

Zacholme7:
Are we good to close these and just put them into an issue?

diegomrsantos:
> Are we good to close these and just put them into an issue?

Should we raise this on discord?

Zacholme7:
opened issue #614

PR #551, opened 2025-08-29
fix(qbft): future round PREPARE message buffering for catch-up scenarios

diegomrsantos:
## Issue Addressed

This PR addresses future round PREPARE message buffering in QBFT consensus to ensure proper catch-up behavior during network partitions.

## Proposed Changes

- Add comprehensive test `test_future_round_prepare_messages_rejected` that verifies QBFT properly buffers PREPARE messages from future rounds
- The test demonstrates expected behavior: nodes should buffer PREPARE messages from future rounds and make them available when advancing to those rounds
- This ensures QBFT maintains liveness during network partitions where nodes may receive messages from different rounds out of order

## Additional Info

The test scenario covers:
1. Node in round 1 receives PREPARE messages from round 2 (future round)
2. PREPARE messages are properly buffered for future processing
3. When node advances to round 2, buffered messages become available for consensus

This buffering mechanism is essential for maintaining liveness during catch-up scenarios where nodes rejoin after network partitions or temporary outages.

### DIFF
--- a/anchor/common/qbft/src/tests.rs
+++ b/anchor/common/qbft/src/tests.rs
@@ -246,11 +246,11 @@ fn test_future_round_prepare_messages_rejected() {
         message::{MsgType, RSA_SIGNATURE_SIZE, SSVMessage, SignedSSVMessage},
     };
 
-    // Create QBFT instance with 3 nodes (f=0, quorum=3)
+    // Create QBFT instance with 4 nodes (f=1, quorum=3)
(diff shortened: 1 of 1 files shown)


diegomrsantos:
@Zacholme7 @dknopik should we also buffer other messages?

Zacholme7:
seems like this also makes sense. We would need to refactor and implement a new mechanism to try to transition as many states as possible on each message receipt.

Zacholme7:
opened issue #615

PR #554, opened 2025-08-29
feat: reduce validator operation logging noise with aggregated summaries

diegomrsantos:
## Issue Addressed

This PR addresses excessive logging noise in anchor.log during validator synchronization operations. No specific GitHub issue was filed, but the problem was identified through log analysis showing thousands of repetitive DEBUG messages during bulk validator processing.

## Proposed Changes

- **anchor/eth/src/event_processor.rs**: 
  - Move individual "Successfully added validator" and "Successfully removed validator" logs from DEBUG to TRACE level
  - Add simple counters to track validator operations per batch
  - Add summary DEBUG logs showing total validators added/removed after processing each batch
- **CLAUDE.md**: 
  - Add "Simplicity First" principle to general development guidelines
  - Add "Check Requirements First" principle to prevent template/guideline oversights
  - Add guidance for updating documentation after successful sessions
  - Emphasize reading PR template before creating PRs
- **.claude/agents/logging-subagent.md**: 
  - Add logging-specific principles for noise reduction strategies
  - Document preference for simple aggregation over complex data structures

## Additional Info

This change dramatically reduces log noise during validator synchronization while preserving all detailed information at TRACE level. The implementation uses simple counters rather than complex collections, following the principle of choosing the simplest solution that elegantly solves the problem. Operators will see clean summaries like "Added validators count=150" instead of 150 individual DEBUG messages, while detailed logs remain available when RUST_LOG=trace is enabled.

### DIFF
--- a/anchor/eth/src/event_processor.rs
+++ b/anchor/eth/src/event_processor.rs
@@ -67,6 +67,10 @@ impl EventProcessor {
         debug!(logs_count = logs.len(), "Starting log processing");
         let timer = metrics::start_timer(&metrics::EXECUTION_LOG_PROCESSING_TIME);
 
+        // Counters for summary logging
+        let mut validators_added = 0;
+        let mut validators_removed = 0;
+
         // Open a transaction for the log batch.
         let mut conn = self
             .db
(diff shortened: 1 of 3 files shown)


anchor/eth/src/event_processor.rs
```diff
@@ -97,10 +101,12 @@ impl EventProcessor {
                 }
 
                 SSVContract::ValidatorAdded::SIGNATURE_HASH => {
+                    validators_added += 1;
                     self.process_validator_added(log, &tx)
                 }
 
                 SSVContract::ValidatorRemoved::SIGNATURE_HASH => {
+                    validators_removed += 1;
```
dknopik: These should only be incremented on success IMO

diegomrsantos: done

dknopik (approved):
LGTM

PR #555, opened 2025-08-29
fix: properly justify round change

Zacholme7:
## Issue Addressed

We were making sure that we had reach consensus on the value that the round changes were preparing, which defeats the purpose of the round change justifications. 

Now, we will find the highest prepared and use that data from the round change directly. If for some reason it is prepared and there is no full data, we fall back to our data, otherwise we just use the start value

### DIFF
--- a/anchor/common/qbft/src/lib.rs
+++ b/anchor/common/qbft/src/lib.rs
@@ -117,9 +117,6 @@ where
     last_prepared_round: Option<Round>,
     last_prepared_value: Option<D::Hash>,
 
-    /// Past prepare consensus that we have reached
-    past_consensus: HashMap<Round, D::Hash>,
-
     /// Aggregated commit message
     aggregated_commit: Option<SignedSSVMessage>,
 
@@ -179,8 +176,6 @@ where
             proposal_root: None,
             last_prepared_round: None,
             last_prepared_value: None,
-
-            past_consensus: HashMap::new(),
 
             aggregated_commit: None,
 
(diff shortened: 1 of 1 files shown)


anchor/common/qbft/src/lib.rs
```diff
@@ -337,46 +332,58 @@ where
...
+        // If no one prepared anything, return None (will use start data)
+        let highest_prepared = match highest_prepared {
+            Some(msg) => msg,
+            None => return None,
+        };
+
+        let claimed_hash = highest_prepared.qbft_message.root;
+
+        // First, try to get data from the round change message itself
+        if !highest_prepared.signed_message.full_data().is_empty() {
```
dknopik, line 359: This will always be empty since my change to clear full data on insertion into the message container.

Which maybe points towards to that change not being a good idea in retrospect. Maybe we should remove the full data from justifications on send instead?

Zacholme7: Hm yes I think you are right. Did some looking and we are actually correctly removing the fulldata in #493 [here](https://github.com/Zacholme7/anchor/blob/f8d66b31d11f02d79e2702131c0236afc0479584/anchor/common/qbft/src/lib.rs#L1001-L1005) which they are also doing [here](https://github.com/ssvlabs/ssv-spec/blob/d4c734909677b5fd3304e1d9220963cc9d794f09/qbft/messages.go#L134). Trying to think if this leads to the same issue?

dknopik: Your approach in #493 seems better - I'll open a PR to switch to that.

dknopik (approved):
LGTM

PR #556, opened 2025-09-01
fix: Remove full data fomr justifications on send

dknopik:
## Issue Addressed

We actually need to keep the full data with messages in the message containers.

## Proposed Changes

- Revert #538 
- In `new_unsigned_message`, call `prepare_justification.set_full_data(vec![]);` on all passed justifications.

## Additional Info
Idea stolen from @Zacholme7


Zacholme7 (approved):
lgtm

PR #565, opened 2025-09-02
feat: ssz infrastructure

Zacholme7:
## Proposed Changes

This pulls out the ssz infrastructure in #493 into its own PR. Functionally, this should change nothing but it enables hashing for the checks required in the spec tests

### DIFF
--- a/anchor/common/qbft/src/lib.rs
+++ b/anchor/common/qbft/src/lib.rs
@@ -8,7 +8,7 @@ pub use qbft_types::{
     UnsignedWrappedQbftMessage, WrappedQbftMessage,
 };
 use ssv_types::{
-    OperatorId, Round,
+    OperatorId, Round, VariableList,
     consensus::{QbftData, QbftDataValidator, QbftMessage, QbftMessageType, UnsignedSSVMessage},
     message::{MsgType, SSVMessage, SignedSSVMessage},
     msgid::MessageId,
@@ -534,13 +534,20 @@ where
         // There was a quorum of round change justifications. We need to go though and verify each
         // one. Each will be a SignedSSVMessage
         for signed_round_change in &msg.qbft_message.round_change_justification {
-            // The qbft message is represented as a Vec<u8> in the signed message, deserialize this
-            // into a proper QbftMessage
-            let round_change: QbftMessage =
-                match QbftMessage::from_ssz_bytes(signed_round_change.ssv_message().data()) {
+            // The justification message is represented as a VariableList<u8> in the signed message,
+            // deserialize this into a proper QbftMessage
+            let Ok(typed_signed_round_change) =
+                SignedSSVMessage::from_ssz_bytes(signed_round_change)
+            else {
+                warn!("Invalid Signed Round change encoded within a message");
+                return false;
+            };
+            let round_change: QbftMessage = {
+                match QbftMessage::from_ssz_bytes(typed_signed_round_change.ssv_message().data()) {
                     Ok(data) => data,
                     Err(_) => return false,
-                };
+                }
+            };
 
             // Make sure this is actually a round change message
             if !matches!(round_change.qbft_message_type, QbftMessageType::RoundChange) {
@@ -550,7 +557,7 @@ where
 
             // Convert to a wrapped message and perform verification
             let wrapped = WrappedQbftMessage {
-                signed_message: signed_round_change.clone(),
+                signed_message: typed_signed_round_change.clone(),
                 qbft_message: round_change.clone(),
             };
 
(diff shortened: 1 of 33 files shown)


dknopik:
Note that this seems to make extensive use of `VariableList::from(Vec)`, which is unfortunately truncating and [being removed](https://github.com/sigp/ssz_types/pull/52)

anchor/common/qbft/src/lib.rs
```diff
@@ -8,7 +8,7 @@ pub use qbft_types::{
     UnsignedWrappedQbftMessage, WrappedQbftMessage,
 };
 use ssv_types::{
-    OperatorId, Round,
+    OperatorId, Round, VariableList,
```
diegomrsantos, line 14: Should changes in this file also be extracted to an individual PR, as they seem to be QBFT-related?

dknopik:
What is the motivation of merging this into `release-v0.3.0`?

anchor/common/qbft/src/lib.rs
Zacholme7: Changes are minimal and make no logic changes apart some extra validation so I think they are fine here. Also, qbft will not compile without the changes here

anchor/common/ssv_types/src/partial_sig.rs
```diff
@@ -23,22 +34,26 @@ pub enum PartialSignatureKind {
...
-            4 => Ok(PartialSignatureKind::ValidatorRegistration),
-            5 => Ok(PartialSignatureKind::VoluntaryExit),
-            _ => Err(()),
+            0 => PartialSignatureKind::PostConsensus,
+            1 => PartialSignatureKind::RandaoPartialSig,
+            2 => PartialSignatureKind::SelectionProofPartialSig,
+            3 => PartialSignatureKind::ContributionProofs,
+            4 => PartialSignatureKind::ValidatorRegistration,
+            5 => PartialSignatureKind::VoluntaryExit,
+            _ => panic!("Invalid PartialSignatureKind value: {value}"),
```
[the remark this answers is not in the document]
Zacholme7: Serde does not work with `TryFrom` which is why I had to change this. This is used for pasing in the spec test input files. No really good way to handle this. Maybe I should know add in an `Unknown(u64)` variant? @diegomrsantos @dknopik

Zacholme7:
This adds bounds to the fields via the variable list and makes sure they use the sizes described in go-ssv so I thought it was important to get in. Changes here are also needed in spec tests.

Think this should be unstable instead?

anchor/common/ssv_types/src/partial_sig.rs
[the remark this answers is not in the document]
diegomrsantos: If this is used only in spec tests it'd be fine, but then this code should be there and not allowed to be called from production code. Otherwise, we should avoid unwrap/panic, unless that's what we really want to do. It doesn't seem to be the case here.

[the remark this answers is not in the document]
Zacholme7: Rust does not allow you to implement a trait on something outside of its crate. Maybe I could wrap it in a `SpecPartialSignatureKind(PartialSignatureKind)` and use that in the spec tests?

anchor/common/qbft/src/lib.rs
diegomrsantos: I believe it's not a good idea to make any change in validation - even if small - in a big PR, especially if there's no explicit description of what is being changed and why.

Zacholme7: I confused myself, There is no changes in validation against release. All this pr does it change types around and enforces sizes at the type level. There should be no logical changes.

anchor/common/ssv_types/src/partial_sig.rs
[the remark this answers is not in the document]
diegomrsantos: Unknown(u64) might be a good enough solution. @dknopik ?

diegomrsantos:
> This adds bounds to the fields via the variable list and makes sure they use the sizes described in go-ssv so I thought it was important to get in. Changes here are also needed in spec tests.
> 
> Think this should be unstable instead?

If it's possible, I'd break this PR further. It'd make reverting in case of issues easier to reason.

Zacholme7:
> Note that this seems to make extensive use of VariableList::from(Vec), which is unfortunately truncating and being removed

hmm interesting. This pr is from a while ago so it does predate that. I think its probably best to just merge the `validate` into the `new` function then?

Zacholme7:
Will break up

Zacholme7:
re-opening as a lot of this stuff depends on each other and its messy to break it up. can do if if we really need it but the stuff here is not super dense

anchor/common/ssv_types/src/partial_sig.rs
```diff
@@ -76,22 +91,106 @@ impl Decode for PartialSignatureKind {
             });
         }
         let value = u64::from_le_bytes(bytes.try_into().unwrap());
-        value.try_into().map_err(|_| DecodeError::NoMatchingVariant)
+        match value {
```
diegomrsantos, line 94: Not introduced here, but please let's remove the unwrap above

dknopik:
> hmm interesting. This pr is from a while ago so it does predate that. I think its probably best to just merge the validate into the new function then?

Well, that does not fix it, there are other instances as well, e.g. here: https://github.com/sigp/anchor/blob/30c436f777379217fe7f6accbfa2790a02e4fd09/anchor/validator_store/src/lib.rs#L335

PR #570, opened 2025-09-02
fix: fix name and docs of `lowest_partial_quorum_above_round`

dknopik:
## Issue Addressed
During a late change in a previous pr, the behaviour of `highest_partial_quorum_above_round` was adjusted, but the name was not. The copy-pasted docs were also inaccurate

## Proposed Changes

Rename to `lowest_partial_quorum_above_round` and change docs.


Zacholme7 (dismissed):

Zacholme7 (approved):
nice

PR #571, opened 2025-09-02
fix(qbft): Check round change quorum disregarding root

dknopik:
## Issue Addressed

Round change messages may have a non-zero root (if we have previously prepared a value) or a zero root.

This causes us to not change round sometimes (where go-ssv does), as we check for quorum per round and root.

## Proposed Changes

Introduce a function to the message container just checking for a sufficient amount of messages in the container, without grouping by root.


Zacholme7 (approved):
yep nice catch

PR #574, opened 2025-09-03
fix(qbft): Validate the full data in round changes late

dknopik:
## Issue Addressed

The fuzzer found out that we actually are not supposed to validate the full data on every message: Full data attached to round change messages are only validated later when they are inspected while starting a new round.

## Proposed Changes

Check full data during validation only if the message is not a round change.

Restructure second half of `justify_round_change_quorum` for easier legibility. Then, remove fetching from data map, and finally add fulldata validation back in there.

## Additional Info

See discord for confirmation that this is actually intended behaviour.

### DIFF
--- a/anchor/common/qbft/src/lib.rs
+++ b/anchor/common/qbft/src/lib.rs
@@ -310,8 +310,11 @@ where
         // Message is not a decide message, we know there is only one signer
         let signer = wrapped_msg.signed_message.operator_ids().first()?;
 
-        // Fulldata may be empty. This is still considered valid though
-        if wrapped_msg.signed_message.full_data().is_empty() {
+        // Fulldata may be empty. This is still considered valid though. We also do not validate
+        // fulldata on round change messages.
+        if wrapped_msg.signed_message.full_data().is_empty()
+            || wrapped_msg.qbft_message.qbft_message_type == QbftMessageType::RoundChange
+        {
             let valid_data = Some(ValidData::new(None, wrapped_msg.qbft_message.root));
             return Some((valid_data, *signer));
         }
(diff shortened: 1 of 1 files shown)


diegomrsantos:
> ## Issue Addressed
> The fuzzer found out that we actually are not supposed to validate the full data on every message: Full data attached to round change messages are only validated later when they are inspected while starting a new round.
> 
> ## Proposed Changes
> Check full data during validation only if the message is not a round change.
> 
> Restructure second half of `justify_round_change_quorum` for easier legibility. Then, remove fetching from data map, and finally add fulldata validation back in there.
> 
> ## Additional Info
> See discord for confirmation that this is actually intended behaviour.

Can we paste here the conclusion of the Discord messages?

anchor/common/qbft/src/lib.rs
```diff
@@ -369,32 +372,28 @@ where
...
-            return Some(ValidData::new(Some(data.clone()), claimed_hash));
+        // The round change includes the full data - decode and use it
+        let Ok(data) = D::from_ssz_bytes(highest_prepared.signed_message.full_data()) else {
+            warn!("Failed to decode round change full data");
+            return None;
+        };
+
+        // Verify the data matches the claimed hash
+        if data.hash() != claimed_hash {
+            warn!("Round change full data doesn't match claimed hash");
```
diegomrsantos, line 387: I'm wondering what we consider a `warn` and an `error`

dknopik: I'd say: `error` means that either an **internal** invariant has been violated, or some other error that is likely only fixable by ourselves.

I'd use `warn` on any failure that may have occurred due to no fault by ourselves - e.g. a weird network message-

dknopik: (of course this is more like a rough framework instead of strict rules!)

anchor/common/qbft/src/lib.rs
```diff
@@ -369,32 +372,28 @@ where
...
-                    return Some(ValidData::new(Some(Arc::new(data)), claimed_hash));
-                } else {
-                    warn!("Round change full data doesn't match claimed hash");
-                }
-            } else {
-                warn!("Failed to decode round change full data");
-            }
+        if highest_prepared.signed_message.full_data().is_empty() {
+            return None;
         }
```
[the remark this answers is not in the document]
dknopik: well, the condition is flipped because the inner code was flipped: return early instead of performing the other checks

anchor/common/qbft/src/lib.rs
```diff
@@ -369,32 +372,28 @@ where
...
 
-        warn!(
-            "Missing data for highest prepared value with hash {:?}",
-            claimed_hash
-        );
+        if !self.data_validator.validate(&data, &self.start_data) {
+            warn!("Round change full data is invalid");
+            return None;
+        }
```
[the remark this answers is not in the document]
dknopik: Local storage lookup made no sense as the message must include the full data on prepare round changes

anchor/common/qbft/src/lib.rs
diegomrsantos: Makes sense, but on the other hand, what's the value of showing this warning to the user? Maybe it should be a debug?

dknopik: Well, when receiving such a log, something is wrong. Using the log you marked as an example, possible root causes:

- someone is trying to attack by sending invalid messages
- the hashing scheme has been changed by a network upgrade
- there is a bug in anchor that causes it

All of these are something that the user should be aware of. Logging it as debug hides it among many harmless messages. The user should see it and alert us if the issue persists.

diegomrsantos: There can be another buggy implementation as well. I agree, even though someone purposely sending invalid messages isn't necessarily actionable, it's something that requires attention if it happens too often.

diegomrsantos:
I believe it's better to let @Zacholme7 approve it

Zacholme7 (approved):
looks good on spec branch, nice

Issue #558, opened 2025-09-01
Do not perform duties for liquidated clusters


dknopik:
During #443, we introduced a regression causing Anchor to keep handling validators belonging to liquidated clusters. 

We should reintroduce filtering in our `ValidatorStore` implementation to avoid treating these validators as active.

dknopik:
@petarjuki7 tagging you as I can't assign you to issues yet. This might be a good one to get started :)

PR #576, opened 2025-09-03
fix(validator_store): filter inactive validators

petarjuki7:
## Issue Addressed

#558 
Makes sure if a cluster is liquidated that it doesn't get processed.


anchor/validator_store/src/lib.rs
```diff
@@ -166,6 +166,9 @@ impl<T: SlotClock, E: EthSpec> AnchorValidatorStore<T, E> {
 
         // First, attempt to get the cluster normally
         if let Some(cluster) = state.clusters().get_by(&validator.cluster_id) {
+            if cluster.liquidated {
+                return Err(Error::SpecificError(SpecificError::Unsupported));
```
petarjuki7: Not sure if this is the appropriate error to throw

Zacholme7: Should probably just add a new variant in, something like `ClusterLiquidated`

anchor/validator_store/src/lib.rs
```diff
@@ -166,6 +166,9 @@ impl<T: SlotClock, E: EthSpec> AnchorValidatorStore<T, E> {
...
+            }
```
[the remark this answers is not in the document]
petarjuki7: Fixed

diegomrsantos:
Can we add a test to avoid this regression from being added?

dknopik:
Nice! We also should filter in `voting_pubkeys`, so that the validator services stop attempting duties for those validators.

petarjuki7:
> Can we add a test to avoid this regression from being added?

Not sure if I understand, where to add a test?

dknopik:
>Can we add a test to avoid this regression from being added?

While I agree, right now it is really hard to test the validator store due to it's tightly coupled nature. I think it might be better to keep it in mind for when we add more tests.


Alternatively, our integration test approach currently is based on local Testnets. We could add a check there where we assert that a liquidated validator does not attest

anchor/validator_store/src/lib.rs
```diff
@@ -798,6 +802,7 @@ impl<T: SlotClock, E: EthSpec> ValidatorStore for AnchorValidatorStore<T, E> {
             .shares()
             .values()
             .filter_map(|v| filter_func(DoppelgangerStatus::SigningEnabled(v.validator_pubkey)))
+            .filter(|public_key| self.get_validator_and_cluster(*public_key).is_ok())
```
dknopik: This approach has a problem:

We try to lock (within `get_validator_and_cluster`) while we already hold a lock (here by calling `state`)

This can cause a deadlock, as the second lock might wait for a pending writer, which waits for the lock we're holding here first.

Instead, get the cluster here and check for liquidation.

diegomrsantos:
> > Can we add a test to avoid this regression from being added?
> 
> While I agree, right now it is really hard to test the validator store due to it's tightly coupled nature. I think it might be better to keep it in mind for when we add more tests.
> 
> Alternatively, our integration test approach currently is based on local Testnets. We could add a check there where we assert that a liquidated validator does not attest

Could you explain in more detail how our integration test approach works? Where are the tests located?

dknopik:
@diegomrsantos I am talking about #562. It runs a Testnet and then checks if the expected duties are performed. It seems feasible to add a liquidated cluster and check that it does not perform duties.

diegomrsantos:
> @diegomrsantos I am talking about #562. It runs a Testnet and then checks if the expected duties are performed. It seems feasible to add a liquidated cluster and check that it does not perform duties.

But where are the tests exactly? I see only a yaml file there

anchor/validator_store/src/lib.rs
petarjuki7: Ah okay, I understand, I meant to save from some duplication of code.

anchor/validator_store/src/lib.rs
```diff
@@ -586,6 +589,19 @@ impl<T: SlotClock, E: EthSpec> AnchorValidatorStore<T, E> {
...
+
+        if let Some(validator) = state.metadata().get_by(validator_pubkey)
+            && let Some(cluster) = state.clusters().get_by(&validator.cluster_id)
+        {
+            return !cluster.liquidated;
+        }
+
+        // We did not manage to fetch the cluster
+        false
+    }
```
dknopik: Unfortunately, same issue as before: 

By trying to borrow here (in line 594) while we already holding the lock through the `.state()` call in `voting_pubkeys`, we are at risk of a deadlock if a writer tries to acquire the lock between the two calls.

Furthermore, you can call `get_by` directly with the validator pubkey - no need to go via the `metadata`, as the multi index map allows access via the `validator_pubkey`.

anchor/validator_store/src/lib.rs
```diff
@@ -809,13 +796,20 @@ impl<T: SlotClock, E: EthSpec> ValidatorStore for AnchorValidatorStore<T, E> {
...
+        state
             .shares()
             .values()
             .filter_map(|v| filter_func(DoppelgangerStatus::SigningEnabled(v.validator_pubkey)))
-            .filter(|public_key| self.is_cluster_active(public_key))
+            .filter(|public_key| {
+                if let Some(clusters) = clusters.get_by(public_key) {
+                    return !clusters.liquidated;
+                }
+                false
```
dknopik: nice, exactly :)

still got one nitpick in me :P

I really like [`Option::is_some_and`](https://doc.rust-lang.org/std/option/enum.Option.html#method.is_some_and), but that is a matter of taste

dknopik (dismissed):
LGTM, thanks! I've got one nitpick, but that one is a matter of taste. Feel free to address or merge (by adding the `ready-to-merge` label)

dknopik:
Sorry, I just noticed that this was based on `unstable`. Most bugfixes should be based on the `releases-v0.3.0` branch, in order to separate experimental work (on `unstable`) from our fixes ahead of the big release. Can you please rebase your commits on top of `release-v0.3.0`?

petarjuki7:
Closing this in favour of #589 to keep a cleaner commit history

PR #577, opened 2025-09-03
feat: specific qbft errors

Zacholme7:
## Issue Addressed

This PR adds in specific error variants for qbft. This is important for the spec tests as we have to map from qbft error -> error string for success validation.

PR #580, opened 2025-09-04
feat: ssz infrastructure

Zacholme7:
## Issue Addressed

This pulls out the ssz infrastructure in #493 into its own PR. Functionally, this should change nothing but it enables hashing for the checks required in the spec tests

This uses new functions `to_variable_list` and `to_variable_list_with_error` which both take the length as a generic to check the upper bound of the data being validated. Allows us to use `From` while also ensuring that we reject invalid length data.

This is the base PR for all of the spec test as we must hash the messages for output comparison against go-ssv


anchor/common/ssv_types/src/message.rs
```diff
@@ -121,17 +136,7 @@ impl Decode for MsgType {
     }
 
     fn from_ssz_bytes(bytes: &[u8]) -> Result<Self, DecodeError> {
```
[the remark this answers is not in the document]
Zacholme7: from_ssz_bytes will handle the error

anchor/common/qbft/src/lib.rs
```diff
@@ -549,15 +553,23 @@ where
...
+            .qbft_message
+            .round_change_justification
+            .iter()
+            .filter_map(|bytes| SignedSSVMessage::from_ssz_bytes(bytes).ok())
+            .collect();
+
         // Make sure we have a quorum of round change messages
-        if !self.check_quorum(&msg.qbft_message.round_change_justification) {
+        if !self.check_quorum(&signed_rc_justifications) {
             warn!("Did not receive a quorum of round change messages");
```
diegomrsantos, line 645: I'm wondering about those warnings. Would this case be either a malicious message or a bug in the code?

Zacholme7: I would say just an incorrect message. You need a quorum to do anything malicious and I dont see how this would be a bug

diegomrsantos: So I'm not sure it adds value to show a warning in this case. Would a debug be better?

Zacholme7: There is something wacky going on if we hit that log so I think a warn is correct here

diegomrsantos: Like what, for example? Is this an actionable log?

Zacholme7: Someone is constructing an invalid proposal. This corresponds to an error in the spec tests so it could even be bumped to an error, but it is not actionable just worth taking note of.

diegomrsantos: By an invalid proposal, you mean there's a bug in the code that spec tests aren't catching? If so, it should be an error. If we aren't sure it's an error, but it might be, and we need to investigate if it happens often, it should be a warning. Otherwise, it should be debug or trace.

Zacholme7: I don't understand. The spec test for this is specifically testing the case where we have a proposal with justifications but do not have a quorum of rc justifications to support it and it should be rejected as it is an invalid message. So as this is an invalid message I can get behind this being an error as an error needs to be returned in my qbft error branch. I dont understand where the bug comes in, this is just validation.

diegomrsantos: What's the relationship of this with the spec tests? This isn't spec test code. 

In a P2P network we can get any message. An invalid message doesn't necessarily mean a bug in our code. We could consider the execution of this log an error only if we expect this kind of message being filtered by a higher layer.

diegomrsantos: Another important point is that production code shouldn't be modeled after tests. Ideally they have their own requirements according to the needs of the system and the tests adapt to it.

Zacholme7: [this](https://github.com/ssvlabs/ssv-spec/blob/main/qbft/spectest/generate/tests/tests.MsgProcessingSpecTest_qbft_message_processing_no_rc_quorum_(prepared).json) is the qbft test testing this validation and they consider this an [error](https://github.com/ssvlabs/ssv-spec/blob/d4c734909677b5fd3304e1d9220963cc9d794f09/qbft/spectest/generate/tests/tests.MsgProcessingSpecTest_qbft_message_processing_no_rc_quorum_(prepared).json#L73)

diegomrsantos: In a controlled environment, yes. But once the system is running in the wild we can get any kind of message. Maybe there's another buggy implementation or someone is trying to spam/crash the system. Not all messages will be created by our code.

diegomrsantos: Unless, as I said before, that at this point we expect all invalid messages to be filtered. So yeah, this would be a bug, but a validation one.

Zacholme7: So what are you suggesting I do then?

diegomrsantos: I think it's ok to leave as a warn. @dknopik wdyt?

dknopik: Sooo, I think this might be the first place where we check this, so this is not an internal sanity check failing or an internal invariant being broken, but just an invalid message. So IMO `warn` is fine :)

diegomrsantos: It was a good discussion tho ;)

anchor/common/qbft/src/lib.rs
```diff
@@ -549,15 +553,23 @@ where
         let mut max_prepared_round = 0;
         let mut max_prepared_msg = None;
 
+        // Deserialize round change justifications for validation
+        let signed_rc_justifications: Vec<SignedSSVMessage> = msg
+            .qbft_message
+            .round_change_justification
+            .iter()
+            .filter_map(|bytes| SignedSSVMessage::from_ssz_bytes(bytes).ok())
+            .collect();
```
dknopik: If deserialization fails, it will be silently ignored. In conjunction with #577, do we maybe catch this with an `UndecodableNestedMessage` error or something?

(also to be found in other places)

anchor/common/qbft/src/lib.rs
```diff
@@ -621,7 +633,14 @@ where
                     return false;
                 }
 
-                if !self.check_quorum(&round_change.round_change_justification) {
+                // Deserialize prepare justifications for validation
+                let signed_inner_rc_justifications: Vec<SignedSSVMessage> = round_change
+                    .round_change_justification
+                    .iter()
+                    .filter_map(|bytes| SignedSSVMessage::from_ssz_bytes(bytes).ok())
```
diegomrsantos: Are errors during deserialization being silently ignored?

anchor/common/qbft/src/lib.rs
```diff
@@ -904,11 +931,13 @@ where
...
                 .map(|msg| msg.signed_message.clone());
-            aggregated_commit.aggregate(signed_commits);
+            aggregated_commit.aggregate(signed_commits).ok()?;
 
             // Set full data
             let hash = first_commit.qbft_message.root;
-            aggregated_commit.set_full_data(self.data.get(&hash)?.as_ssz_bytes());
+            aggregated_commit
+                .set_full_data(self.data.get(&hash)?.as_ssz_bytes())
+                .ok()?;
```
diegomrsantos: I believe the error is also being silently ignored here. Worth checking other places too

anchor/common/ssv_types/src/message.rs
```diff
@@ -75,6 +70,26 @@ pub enum MsgType {
     SSVPartialSignatureMsgType = 1,
 }
 
+impl TreeHash for MsgType {
+    fn tree_hash_type() -> TreeHashType {
+        TreeHashType::Basic
+    }
+
+    fn tree_hash_packed_encoding(&self) -> PackedEncoding {
+        let value = self.clone() as u64;
```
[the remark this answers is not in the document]
dknopik: at copilot: MsgType does not impl Copy

(it probably should though)

anchor/common/ssv_types/src/message.rs
```diff
@@ -430,19 +496,49 @@ impl SignedSSVMessage {
...
+        Ok(())
+    }
+
+    /// Returns a clone of this SignedSSVMessage with empty full_data.
+    /// This matches the Go implementation's WithoutFullData() method used for justifications.
+    pub fn without_full_data(&self) -> Self {
+        let mut cloned = self.clone();
+        cloned.full_data = VariableList::empty();
+        cloned
     }
```
dknopik: This clones the whole message (including full data, which is potentially large), just to remove full data again.

Maybe we can either
1. manually clone everything except the full data, or
2. use `set_full_data(vec![])`

dknopik (commented):
Some minor notes

anchor/common/ssv_types/src/message.rs
```diff
@@ -36,35 +35,31 @@ const VALIDATOR_INDEX_SIZE: usize = 8;
...
-    + (MAX_PARTIAL_SIGNATURE_MSGS_SIZE / ENCODING_OVERHEAD_DIVISOR)
-    + 4;
+const MAX_FULL_DATA_SIZE: usize = SSVMessageFullDataLen::USIZE;
+
+/// SSVMessage.Data max size: 722412 (from Go spec)
+/// 722412 = 722 * 1000 + 412 = 722000 + 412
+pub type SSVMessageDataLen = Sum<Prod<U722, U1000>, U412>;
 
 /// Defines the types of messages with explicit discriminant values.
 #[derive(Debug, Clone, PartialEq, Eq)]
```
diegomrsantos: If the comment below is true, we can remove the `Clone`

anchor/common/ssv_types/src/message.rs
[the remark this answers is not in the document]
dknopik: no, as noted previously this is handled by `u64::from_ssz_bytes`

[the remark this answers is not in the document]
diegomrsantos: Could you please share the rationale?

anchor/common/ssv_types/src/message.rs
```diff
@@ -152,26 +157,26 @@ pub enum SSVMessageError {
...
 /// Represents a bare SSVMessage with a type, ID, and data.
-#[derive(Encode, Decode, Clone, PartialEq, Eq)]
+#[derive(Encode, Decode, Clone, PartialEq, Eq, TreeHash)]
 #[cfg_attr(feature = "arbitrary-fuzz", derive(arbitrary::Arbitrary))]
 pub struct SSVMessage {
     msg_type: MsgType,
-    msg_id: MessageId, // Fixed-size [u8; 56]
-    data: Vec<u8>,     // Variable-length byte array
+    msg_id: MessageId,
+    data: VariableList<u8, SSVMessageDataLen>,
```
diegomrsantos, line 165: Can this have a variable size?

anchor/common/ssv_types/src/message.rs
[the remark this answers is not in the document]
diegomrsantos: It says:

- Your MsgType enum is defined as #[repr(u64)] and has no fields, just variants.
- It derives Clone, but also (since all its fields/variants are Copy types and it is repr(u64)), the enum itself is Copy by default in Rust.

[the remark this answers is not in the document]
dknopik: > the enum itself is Copy by default in Rust.

No.

https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=36d7d7cd34a6e0e96f3f57aab4195a25

This fails to compile and is fixed when `Copy` is added to the derive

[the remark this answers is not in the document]
dknopik: Also, another educational note: using `Copy` "to avoid the clone overhead" is nonsense, after optimization the resulting binary would look the same in this case. The only reason to do it is code quality.

anchor/common/ssv_types/src/message.rs
Zacholme7: Why variable size?

anchor/common/qbft/src/lib.rs
Zacholme7: this branch doesnt have qbft errors so will adjust in that branch

anchor/common/ssv_types/src/message.rs
Zacholme7: need to impl clone if we impl copy

anchor/common/ssv_types/src/message.rs
Zacholme7: `set_full_data' needs `mut self`, so just cloned everything

anchor/common/ssv_types/src/message.rs
diegomrsantos: > need to impl clone if we impl copy

I don't get it

anchor/common/ssv_types/src/message.rs
diegomrsantos: I mean, why has this been changed to `VariableList`?

anchor/common/ssv_types/src/message.rs
Zacholme7: to be able to derive Copy, you must also derive Clone because `Copy: Clone` https://doc.rust-lang.org/std/marker/trait.Copy.html

anchor/common/ssv_types/src/message.rs
Zacholme7: In comparison to the `Vec`? We can derive `TreeHash` when its `VariableList` not with the `Vec`

anchor/common/ssv_types/src/message.rs
diegomrsantos: I didn't know, it seems weird tbh

anchor/common/ssv_types/src/message.rs
diegomrsantos: And is `data` allowed to have a variable size?

anchor/common/ssv_types/src/message.rs
Zacholme7: yea weird rust quirk

anchor/common/ssv_types/src/message.rs
Zacholme7: Yes its size is allowed to be up to `SSVMessageDataLen`

anchor/common/ssv_types/src/lib.rs
```diff
@@ -13,8 +13,45 @@ pub mod partial_sig;
...
+) -> Option<ssz_types::VariableList<T, N>> {
+    if vec.len() <= N::to_usize() {
+        Some(ssz_types::VariableList::from(vec))
+    } else {
+        None
+    }
+}
+
+/// Converts a Vec to VariableList, returning a custom error on failure.
+pub fn to_variable_list_with_error<T, N, E, F>(
```
diegomrsantos: How about `try_to_variable_list`?

anchor/validator_store/src/lib.rs
```diff
@@ -1466,7 +1477,11 @@ impl<T: SlotClock, E: EthSpec> ValidatorStore for AnchorValidatorStore<T, E> {
                             validator_sync_committee_indices: Default::default(),
                         },
                         version: ForkName::Altair.into(),
-                        data_ssz: data.as_ssz_bytes(),
+                        data_ssz: to_variable_list(data.as_ssz_bytes()).ok_or_else(|| {
```
diegomrsantos: Why don't we use the version with an error here?

Zacholme7: fixed

anchor/validator_store/src/lib.rs
```diff
@@ -1165,7 +1173,12 @@ impl<T: SlotClock, E: EthSpec> ValidatorStore for AnchorValidatorStore<T, E> {
                             validator_sync_committee_indices: Default::default(),
                         },
                         version,
-                        data_ssz: message.as_ssz_bytes(),
+                        data_ssz: try_to_variable_list(message.as_ssz_bytes(), |provided, max| {
+                            Error::SpecificError(SpecificError::DataTooLarge(format!(
+                                "Attestation data too large for consensus: {} > {}",
+                                provided, max
+                            )))
```
diegomrsantos, line 1189: Can this be moved inside the function, and we pass only the error message?

Zacholme7: this is used with multiple different error variants, not just the `DataTooLarge`

diegomrsantos: Like what? Are other errors necessary?

Zacholme7: Here are some 
https://github.com/Zacholme7/anchor/blob/21616f72ae1dda7bf54e1252f1e84c45f557257d/anchor/common/ssv_types/src/message.rs#L453-L471

anchor/message_validator/src/partial_signature.rs
```diff
@@ -343,9 +343,15 @@ mod tests {
             let p_key = PKey::from_rsa(pk.clone()).unwrap();
             let mut signer = Signer::new(MessageDigest::sha256(), &p_key).unwrap();
             signer.update(&ssv_msg.as_ssz_bytes()).unwrap();
-            vec![signer.sign_to_vec().expect("Failed to sign message")]
+            vec![
+                signer
+                    .sign_to_vec()
+                    .expect("Failed to sign message")
+                    .try_into()
+                    .expect("Signature should be 256 bytes"),
```
diegomrsantos, line 353: Do we want to use `expect` here?

Zacholme7: its just a test method to construct data so I figured it was fine

diegomrsantos: Ah, sorry, I see it in normal files and forgot we have tests in those

anchor/common/qbft/src/lib.rs
diegomrsantos: What's the resolution for this?

diegomrsantos: Now it's using a map, I guess. It's weird, I can't see these conversations in the code.

anchor/common/qbft/src/lib.rs
```diff
@@ -1108,18 +1171,29 @@ where
...
-            round_change_justification.set_full_data(vec![]);
-        }
-        for prepare_justification in &mut prepare_justification {
-            prepare_justification.set_full_data(vec![]);
-        }
+        let round_change_justification_vec: Vec<VariableList<u8, RoundChangeLength>> =
+            round_change_justification
+                .into_iter()
+                .map(|msg| msg.without_full_data())
+                .filter_map(|msg| to_variable_list(msg.as_ssz_bytes()))
```
diegomrsantos: error being ignored here?

anchor/common/qbft/src/lib.rs
```diff
@@ -1262,11 +1336,21 @@ where
...
+            let prepare_msgs: Vec<SignedSSVMessage> = if let Ok(justifications) = highest_rc
+                .qbft_message
+                .round_change_justification
+                .iter()
+                .map(|bytes| SignedSSVMessage::from_ssz_bytes(bytes))
+                .collect::<Result<Vec<_>, _>>()
+            {
+                justifications
+            } else {
+                return (vec![], None);
```
diegomrsantos: Should an error be returned?

anchor/common/qbft/src/lib.rs
```diff
@@ -1108,18 +1171,29 @@ where
         &self,
         msg_type: QbftMessageType,
         data_hash: D::Hash,
-        mut round_change_justification: Vec<SignedSSVMessage>,
-        mut prepare_justification: Vec<SignedSSVMessage>,
-    ) -> UnsignedWrappedQbftMessage {
+        round_change_justification: Vec<SignedSSVMessage>,
+        prepare_justification: Vec<SignedSSVMessage>,
+    ) -> Option<UnsignedWrappedQbftMessage> {
```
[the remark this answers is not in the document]
Zacholme7: handled at callsite

anchor/common/qbft/src/lib.rs
Zacholme7: If there is an error we should just return empty justifications so imo this is fine

diegomrsantos: Why not an error?

anchor/common/qbft/src/lib.rs
Zacholme7: fixed

anchor/common/qbft/src/lib.rs
Zacholme7: fixed

anchor/common/ssv_types/src/partial_sig.rs
```diff
@@ -80,18 +89,83 @@ impl Decode for PartialSignatureKind {
...
+#[derive(Debug, PartialEq)]
+pub enum PartialSignatureError {
+    NoMessages,
+    InconsistentSigners,
+    ZeroSigner,
+}
+
+impl PartialSignatureMessages {
+    /// Validates the partial signature messages
+    pub fn validate(&self) -> Result<(), PartialSignatureError> {
```
diegomrsantos: Was it somewhere else, or we never had it? Same for the other one

Zacholme7: Ah this should be moved to the type spec tests. Ill double check if the msg validator covers this but go implements this function specifically for the PartialSignatureMessage

anchor/common/qbft/src/lib.rs
```diff
@@ -1309,19 +1414,27 @@ where
...
-        let unsigned_msg =
-            self.new_unsigned_message(QbftMessageType::Commit, data_hash, vec![], vec![]);
-
-        self.message_sender.send(unsigned_msg);
+        match self.new_unsigned_message(QbftMessageType::Commit, data_hash, vec![], vec![]) {
+            Ok(unsigned_msg) => {
+                self.message_sender.send(unsigned_msg);
+            }
+            Err(e) => {
+                warn!("Failed to construct commit message: {:?}", e);
```
diegomrsantos: Can we silently fail in this function? Doesn't the caller care?

anchor/common/qbft/src/lib.rs
```diff
@@ -1309,19 +1414,27 @@ where
...
-        let unsigned_msg =
-            self.new_unsigned_message(QbftMessageType::Prepare, data_hash, vec![], vec![]);
-
-        self.message_sender.send(unsigned_msg);
+        match self.new_unsigned_message(QbftMessageType::Prepare, data_hash, vec![], vec![]) {
+            Ok(unsigned_msg) => {
+                self.message_sender.send(unsigned_msg);
+            }
+            Err(e) => {
+                warn!("Failed to construct prepare message: {:?}", e);
```
diegomrsantos: Same as below

anchor/common/qbft/src/lib.rs
```diff
@@ -1332,17 +1445,23 @@ where
...
+        ) {
+            Ok(unsigned_msg) => {
+                // forget that we accepted a proposal
+                self.proposal_accepted_for_current_round = false;
+                self.message_sender.send(unsigned_msg);
+            }
+            Err(e) => {
+                warn!("Failed to construct round change message: {:?}", e);
+                // Still reset the proposal accepted flag even if message construction failed
+                self.proposal_accepted_for_current_round = false;
```
diegomrsantos: Seems a bit confusing to do it here in this function. Like, why is that so?

anchor/common/ssv_types/src/partial_sig.rs
```diff
@@ -80,15 +89,35 @@ impl Decode for PartialSignatureKind {
     }
 }
 
+impl TreeHash for PartialSignatureKind {
```
diegomrsantos, line 92: Can't it be derived?

anchor/common/ssv_types/src/lib.rs
```diff
@@ -13,8 +13,40 @@ pub mod partial_sig;
...
+use ssz_types::typenum::Unsigned;
 pub use types::{Epoch, Slot, VariableList};
+
+// Shared constants used across message types
+pub const RSA_SIGNATURE_SIZE: usize = 256;
+pub const MAX_SIGNATURES: usize = 13;
+
+/// Converts a Vec to VariableList if it fits within the type's bounds.
+/// Returns None if the vec length exceeds the maximum capacity.
+pub fn to_variable_list<T, N: Unsigned + Clone>(vec: Vec<T>) -> Option<VariableList<T, N>> {
```
diegomrsantos: When do we need to use this one?

anchor/common/ssv_types/src/partial_sig.rs
dknopik: No. I think I suggested adding support for cases like this a while ago, need to check

anchor/common/qbft/src/lib.rs
diegomrsantos: We could return a proper error and do it at the caller

anchor/common/qbft/src/lib.rs
```diff
@@ -555,15 +556,28 @@ where
...
+            .qbft_message
+            .round_change_justification
+            .iter()
+            .map(|bytes| SignedSSVMessage::from_ssz_bytes(bytes))
+            .collect::<Result<Vec<_>, _>>()
+        {
+            justifications
+        } else {
+            return Err(QbftError::RoundChangeJustificationDecodeFailed);
+        };
```
diegomrsantos: How about we extract it to a function like (not tested):

```rust
fn try_decode_signed_ssv_messages(
    justification: &Vec<Vec<u8>>,
    decode_err: QbftError,
) -> Result<Vec<SignedSSVMessage>, QbftError> {
    justification
        .iter()
        .map(|bytes| SignedSSVMessage::from_ssz_bytes(bytes))
        .collect::<Result<Vec<_>, _>>()
        .map_err(|_| decode_err)
}
```

anchor/common/qbft/src/lib.rs
```diff
@@ -876,16 +936,12 @@ where
...
-                debug!(state = ?self.state, "Reached a COMMIT consensus. Success!");
-                self.aggregated_commit = aggregated_commit;
-                self.state = InstanceState::Complete;
-                self.completed = Some(Completed::Success(hash));
-            } else {
-                error!("Failed to aggregate commit quorum");
-                return Err(QbftError::FailedToAggregate);
-            }
+
+            let aggregated_commit = self.aggregate_commit_messages(commit_quorum)?;
```
diegomrsantos: Why we removed the log `error!("Failed to aggregate commit quorum");`?

anchor/common/qbft/src/lib.rs
```diff
@@ -905,22 +961,34 @@ where
...
+            match self.data.get(&hash) {
+                Some(data) => {
+                    if let Err(e) = aggregated_commit.set_full_data(data.as_ssz_bytes()) {
+                        return Err(QbftError::SignedSSVMessageError(e));
+                    }
+                }
+                None => {
+                    return Err(QbftError::MissingData);
+                }
+            }
```
diegomrsantos: We don't need logs here?

diegomrsantos: We don't need logs here?

anchor/common/ssv_types/src/consensus.rs
```diff
@@ -56,6 +56,17 @@ impl<D: QbftData> QbftDataValidator<D> for NoDataValidation {
     }
 }
 
+/// ValidatorConsensusData.DataSSZ max size: 8388608 bytes (2^23)
+/// This is calculated as 2^23 = 8,388,608
+/// We can represent this as 8 * 1000000 + 388 * 1000 + 608
+pub type ValidatorConsensusDataLen = Sum<Prod<U8, U1000000>, Sum<Prod<U388, U1000>, U608>>;
```
diegomrsantos: If the code doesn't explain why the number is decomposed like 8 * 1000000 + 388 * 1000 + 608, and just uses the type-level arithmetic for a known constant, it adds complexity without providing clarity or benefit to the reader.

Zacholme7: So I should just add a comment saying "typenum doesnt define U8388608"? We can decompose it numerous ways that is just the most straightforward

anchor/common/ssv_types/src/message.rs
```diff
@@ -36,45 +35,61 @@ const VALIDATOR_INDEX_SIZE: usize = 8;
...
     + ROOT_SIZE
     + ROUND_SIZE
-    + MAX_SIGNATURES * (MAX_NO_JUSTIFICATION_SIZE + MAX1_JUSTIFICATION_SIZE);
-
-const MAX_ENCODED_CONSENSUS_MSG_SIZE: usize =
-    MAX_CONSENSUS_MSG_SIZE + (MAX_CONSENSUS_MSG_SIZE / ENCODING_OVERHEAD_DIVISOR) + 4;
+    + (MAX_SIGNATURES * (RoundChangeLength::USIZE + ssz::BYTES_PER_LENGTH_OFFSET)
+        + ssz::BYTES_PER_LENGTH_OFFSET)
+    + (MAX_SIGNATURES * (JustificationLength::USIZE + ssz::BYTES_PER_LENGTH_OFFSET)
+        + ssz::BYTES_PER_LENGTH_OFFSET);
```
diegomrsantos: Why has this changed?

anchor/common/ssv_types/src/consensus.rs
diegomrsantos: I thought there was a domain reason for that. So it's just cause we can't define `U8388608` and it must be a type instead of a const? 

Can't we also do `2^23`?

Zacholme7: yes it is not defined in typenum so we have to construct it.

`2^23` for what? we have to define this length using typenum types so we have access to the derived tree hash

diegomrsantos: `Pow<U2, U23>;` (if it's possible) is much simpler than Sum<Prod<U8, U1000000>, Sum<Prod<U388, U1000>, U608>>;

anchor/common/ssv_types/src/message.rs
Zacholme7: This was changed back in https://github.com/sigp/anchor/pull/377
Not sure the exact reason why but im sure there was a reason. Maybe @dknopik remembers?

anchor/common/ssv_types/src/consensus.rs
Zacholme7: You can define it as `pub type ValidatorConsensusDataLen = <U2 as Pow<U23>>::Output;`

IMO, that looks more confusing.

diegomrsantos: Why do you think it's more confusing?

Zacholme7: I mean I think it is simple and perfectly understandable either way, `<U2 as Pow<U23>>::Output` is just more complicated syntax than just some  `Sum` and `Prod` with the simple arithmetic formula right above it but this shouldn't make any difference so ill change it

diegomrsantos: Imo it's worth it to add a comment explaining the need for this. I thought this calculation was a representation of some domain behavior and each number was the value of some field

anchor/message_validator/src/consensus_message.rs
```diff
@@ -1134,12 +1134,14 @@ mod tests {
         let signature = signer.sign_to_vec().expect("Failed to create signature");
 
         // Pad signature to RSA_SIGNATURE_SIZE if needed
-        let padded_signature = if signature.len() < RSA_SIGNATURE_SIZE {
-            let mut padded = vec![0; RSA_SIGNATURE_SIZE];
+        let padded_signature: [u8; RSA_SIGNATURE_SIZE] = if signature.len() < RSA_SIGNATURE_SIZE {
```
diegomrsantos: how about `pub type RsaSignature = [u8; RSA_SIGNATURE_SIZE];`?

anchor/common/ssv_types/src/message.rs
diegomrsantos: @dknopik do you recall?

anchor/common/qbft/src/lib.rs
```diff
@@ -432,7 +468,11 @@ where
             debug!(hash = ?valid_data.hash, "Current leader proposing data");
 
             // Send the initial proposal and then the following prepare
-            self.send_proposal(valid_data.hash, valid_data.data.expect("Start data exists"));
+            if let Err(e) =
+                self.send_proposal(valid_data.hash, valid_data.data.expect("Start data exists"))
```
diegomrsantos: @dknopik @Zacholme7 should we get the release branch merged into unstable? The expect has been removed

Zacholme7: at this point it might just be best to wait until release in finalized and then merged into unstable before getting this in

anchor/message_validator/src/consensus_message.rs
diegomrsantos: This was for creating a type alias for `[u8; RSA_SIGNATURE_SIZE]`

Zacholme7: yes but this was the only place we explicitly defining the type and the compiler can infer it so its not needed here

diegomrsantos: We use it in other places too. But we also do `vec![[0u8; RSA_SIGNATURE_SIZE]]` which is actually defining elements instead of a type

anchor/common/ssv_types/src/message.rs
dknopik: I believe this was changed because the old calculation was incorrect.

Note that the new test function `ensure_message_sizes_correct` verifies that this new calculation is correct.

diegomrsantos: Thanks

diegomrsantos (commented):
From my side, I'm good with it. Sorry for the lengthy review. As I'm not that familiar with the code, it's better if @dknopik can give it a final review.

dknopik:
Will review once tomorrow's post-release backmerge has been merged here

dknopik (approved):
LGTM, thank you so much!

PR #581, opened 2025-09-04
feat: type spec tests 

Zacholme7:
## Issue Addressed

#39 

## Proposed Changes

1) This sets up a register/loader based approach for the spec tests where each test implements the `SpecTest` trait so that we able to properly load in the correct input json files and have a simple way to setup and run each test file

2) This nearly completes the type spec tests that are relevant to our implementation.  These tests are all pretty straightforwards and I dont anticipate any errors with the remaining ones so we may want to just remove them or finish them up if time permitting. 

## Change summary
1) All of the changes in `common/ssv_types/src` just deal with de-serializing the types for the spec tests
2) in `spec_tests/src/types` are all of the spec tests and they follow the same pattern. Have a struct for parsing in the data that implement the `SpecTest` trait for a clean interface to run the test and get its type
3) `spec_tests/src/utils` just contains some extra de-serializers that are needed
4) `spec_tests/src/lib.rs` is the core brain for loading and running the tests. A test gets registers in the `TEST_LOADERES` which makes it so that the `run_tests` function can properly extract the correct files to load in for each test variant

Note: We need `allow(dead_code)` at the top of `lib.rs` as this code is just used for testing so clippy marks it as un-used when it is actually used. Ive done a walkthrough and manually checked the usage of everything


shane-moore:
Closing since I'm revamping this in separate PR's, thanks @Zacholme7, has been super useful to reference

PR #583, opened 2025-09-04
fix: reduce external library logging noise

diegomrsantos:
## Issue Addressed

Excessive logging noise from external libraries libp2p_gossipsub and discv5 affecting log readability and storage.

## Proposed Changes

- Set external library log levels to WARN in libp2p_discv5_layer filter
- Changed from DEBUG to `EnvFilter::new("libp2p_gossipsub=warn,discv5=warn")`
- Added documentation explaining the noise reduction strategy

## Additional Info

Preserves Anchor's debug logging while filtering external library noise. Targeted change affecting only the specialized logging layer.

### DIFF
--- a/anchor/src/main.rs
+++ b/anchor/src/main.rs
@@ -207,11 +207,13 @@ pub fn enable_logging(
         if let Some(libp2p_discv5_layer) = libp2p_discv5_layer {
             // Create filter that reduces external library noise to WARN level while preserving
             // the configured file log level for Anchor crates
-            let external_libs_filter = "libp2p_gossipsub=warn,discv5=warn";
 
             logging_layers.push(
                 libp2p_discv5_layer
-                    .with_filter(EnvFilter::new(&external_libs_filter))
+                    .with_filter(
+                        EnvFilter::try_new("warn,libp2p_gossipsub::peer_score=debug,libp2p_gossipsub::gossip_promises=debug")
+                            .unwrap_or_else(|_| EnvFilter::new("debug")),
+                    )
                     .boxed(),
             );
         }
--- a/anchor/src/main.rs
+++ b/anchor/src/main.rs
@@ -205,13 +205,13 @@ pub fn enable_logging(
         let file_logging_layer = init_file_logging(&logs_dir, file_logging_flags.clone());
 
         if let Some(libp2p_discv5_layer) = libp2p_discv5_layer {
+            // Create filter that reduces external library noise to WARN level while preserving
+            // the configured file log level for Anchor crates
+            let external_libs_filter = format!("libp2p_gossipsub=warn,discv5=warn");
+            
             logging_layers.push(
                 libp2p_discv5_layer
-                    .with_filter(
-                        EnvFilter::builder()
-                            .with_default_directive(Level::DEBUG.into())
-                            .from_env_lossy(),
-                    )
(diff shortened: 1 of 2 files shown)


dknopik:
A lot of the "[Penalty]" logmessages in libp2p are on `debug` level. IMO these are quite useful for debugging in scenarios where Anchor unexpectedly blocks a lot of peers. wdyt?

diegomrsantos:
> A lot of the "[Penalty]" logmessages in libp2p are on `debug` level. IMO these are quite useful for debugging in scenarios where Anchor unexpectedly blocks a lot of peers. wdyt?

They can be quite noisy, can't they? I'm actually thinking those should be trace and we should make a PR. Wdyt?

dknopik:
Sure, they can be noisy, but also very useful for troubleshooting, especially as they represent an abnormal state. I think they are fine.

diegomrsantos:
> Sure, they can be noisy, but also very useful for troubleshooting, especially as they represent an abnormal state. I think they are fine.

I think you're right. So maybe the current content of libp2p should be actually `trace`:
  - "Message already received, ignoring" - routine deduplication
  - "Sending message to peer" - low-level message routing
  - "Completed forwarding message" - internal state transitions
  - "Handling message from peer" - message processing steps
  
Wdyt?

dknopik:
While I am inclined to agree, I am not sure what the `rust-libp2p` guidelines WRT logging are. So I guess João might help here.

diegomrsantos:
> While I am inclined to agree, I am not sure what the `rust-libp2p` guidelines WRT logging are. So I guess João might help here.

@jxs what are your thoughts?

jxs:
Hi folks, thanks for pinging and sorry for the delay.
I am not sure if I understand the question, but, there aren't repo wide logging guidelines in libp2p that I am aware of, each Behaviour has its own. And I think putting discovery and gossipsub to `warn` might suppress useful information

anchor/src/main.rs
```diff
@@ -205,12 +205,14 @@ pub fn enable_logging(
...
+            // Create filter that reduces external library noise to WARN level while preserving
+            // the configured file log level for Anchor crates
+
             logging_layers.push(
                 libp2p_discv5_layer
                     .with_filter(
-                        EnvFilter::builder()
-                            .with_default_directive(Level::DEBUG.into())
-                            .from_env_lossy(),
+                        EnvFilter::try_new("warn,libp2p_gossipsub::peer_score=debug,libp2p_gossipsub::gossip_promises=debug")
```
diegomrsantos, line 214: @dknopik let me know what you think

diegomrsantos: @dknopik I tried to test it with the impostor mode. Didn't see any logs on discv5 and libp2p as the node doesn't seem to find any peers.

diegomrsantos:
> Hi folks, thanks for pinging and sorry for the delay. I am not sure if I understand the question, but, there aren't repo wide logging guidelines in libp2p that I am aware of, each Behaviour has its own. And I think putting discovery and gossipsub to `warn` might suppress useful information

Hi @jxs, sorry for the lack of context. We were talking about the content of this file: [libp2p.txt](https://github.com/user-attachments/files/22238585/libp2p.txt) in this message https://github.com/sigp/anchor/pull/583#issuecomment-3258364766

dknopik (approved):
LGTM

PR #585, opened 2025-09-05
fix: check for multi signer on prepare justifications

Zacholme7:
## Issue Addressed

make sure the prepare justification only has one signer/signature

### DIFF
--- a/anchor/common/qbft/src/lib.rs
+++ b/anchor/common/qbft/src/lib.rs
@@ -679,6 +679,11 @@ where
         round: Round,
         root: &Hash256,
     ) -> bool {
+        // Make sure there is only one signer
+        if justification.operator_ids().len() > 1 {
+            return false;
+        }
+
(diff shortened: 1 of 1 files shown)


dknopik:
Nice. What do you think about changing each `>` to `!=`?

Zacholme7:
fixed

dknopik (approved):
LGTM tyvm

anchor/common/qbft/src/lib.rs
```diff
@@ -679,6 +681,11 @@ where
         round: Round,
         root: &Hash256,
     ) -> bool {
+        // Make sure there is only one signer
+        if justification.operator_ids().len() != 1 || justification.signatures().len() != 1 {
+            return false;
+        }
+
```
diegomrsantos, line 688: nit: we could extract it to a function and avoid the duplication

PR #586, opened 2025-09-05
fix: check for rc quorum before and after

Zacholme7:
## Proposed Changes

Follow the [go logic here](https://github.com/ssvlabs/ssv-spec/blob/d4c734909677b5fd3304e1d9220963cc9d794f09/qbft/round_change.go#L19-L31). Make sure to check if we have a quorum before adding a rc message to ensure we dont double trigger a proposal.


dknopik (approved):
LGTM, thanks!

PR #589, opened 2025-09-08
fix(validator_store): filter inactive validators

petarjuki7:
## Issue Addressed

#558 
Makes sure if a cluster is liquidated that it doesn't get processed.

Closes #576


dknopik (approved):
LGTM

PR #590, opened 2025-09-08
feat: Add a deterministic (BIP-39) keygen

antondlr:
## Proposed Changes

Adds a deterministic BIP-39 based option to `anchor keygen`. 
As a node operator, I highly prefer securing seed phrases over raw key material. 

## Additional Info

This is 100% Claude's work. I have no idea what I'm doing. Seems to work fine but may be ultra dangerous. etc.


anchor/keygen/src/lib.rs
```diff
@@ -62,12 +75,188 @@ pub struct Keygen {
...
+}
+
+// Generate a deterministic RSA key from a mnemonic seed
+fn generate_deterministic_rsa_key(mnemonic: &Mnemonic, index: u32) -> Result<Rsa<Private>, KeygenError> {
+    // Convert mnemonic to seed using BIP39 derivation
+    let seed = mnemonic.to_seed("");
+    let seed_bytes = &seed;
+
+    // Create info string for HKDF using the derivation index
+    let info = format!("anchor-rsa-key-{}", index);
```
dknopik: The index does not seem to be actually used to create a derivation path for the seed?

antondlr: hmm, I had generated an index 0 and an index 1 key and they were different. same index, same key luckily haha.

dknopik: Ah I guess the "info" does that

test_keygen_standalone/src/main.rs
```diff
@@ -0,0 +1,137 @@
...
+    generate_rsa_from_seed(&key_material)
+}
+
+fn generate_rsa_from_seed(seed: &[u8]) -> Result<Rsa<Private>, Box<dyn std::error::Error>> {
+    // Create a deterministic RNG from the seed
+    let mut rng = StdRng::from_seed({
+        let mut seed_array = [0u8; 32];
+        seed_array.copy_from_slice(&seed[..32]);
+        seed_array
+    });
```
dknopik, line 33: We're expanding the seed to 512 bytes just to cut it down to 32 again.

I feel like it makes more sense to use the expanded key material directly instead of passing it through a RNG again

diegomrsantos:
Should the tests be inside the keygen dir?

anchor/keygen/src/lib.rs
```diff
@@ -62,12 +75,225 @@ pub struct Keygen {
...
+        help = "Generate deterministic key from BIP39 mnemonic seed. Mnemonic will be generated if not provided."
+    )]
+    pub deterministic: bool,
+
+    #[clap(
+        long,
+        help = "BIP39 mnemonic phrase for deterministic key generation (12 or 24 words)",
+        requires = "deterministic"
+    )]
+    pub mnemonic: Option<String>,
```
AgeManning, line 90: I dont think we should allow mnemonics on the CLI. They will end up in the processes list and shell history.

AgeManning (commented):
I have had a small look through this. 

Although I think this has great utility, I dont think its worth the security risk and we shouldn't promote it to the public. Here are my reasons:
- It builds the RSA components itself, against the general rule of "Dont roll your own crypto"
- RSA relies fairly heavily on generating random primes. It looks like this uses the seed for a PRNG but then sequentially looks for primes from this starting point. It not obvious to me that this method is as secure and randomly searching for primes, I suspect not. 
- This PR would require significant review before we could think of adding this to production.


One way of doing this that I'd personally feel a bit better about, is just straight using a rust-based RSA library, removing all component generation and just creating the RNG trait which is seeded by the mnemonic seed. i,e we use this function: https://docs.rs/rsa/latest/src/rsa/key.rs.html#217-220 

The mnemonic seed bytes is only 64 bytes, although it uses pbkdf  and can be modified to generate larger key sizes, which may be necessary. 

EDIT: There exists this thing: https://docs.rs/deterministic-keygen/latest/deterministic_keygen/fn.derive_rsa_key_from_phrase.html

Which looks like it creates the PRNG from the seed.

dknopik:
Closing this for now. It is clear that (if we want to add such a feature), it would need an overhaul. See @AgeManning's comment above.

Issue #559, opened 2025-09-01
Verify signatures of justifications


dknopik:
QBFT messages that we receive from the network may have justifications attached. These justifications must have valid signatures. Adjust the `message_validator` crate to verify the signatures of all attached justifications, ideally in the existing `validate_justifications` function

dknopik:
Another one that might be a good one for @petarjuki7 to get started

diegomrsantos:
Should we label it with "good first issue"?

dknopik:
@diegomrsantos People outside of the org tend to look for the "good first issue" tag, I think this one would be good for Petar to handle, so I didn't tag it "good first issue"

PR #591, opened 2025-09-08
fix: verify signatures of justifications

petarjuki7:
## Issue Addressed

Addresses issue #559

## Proposed Changes

Go through all of the justifications chained together and verify their signatures.

## Additional Info

Added operators public keys to the `validate_justifications()` function because they are needed for justification verification.

### DIFF
--- a/anchor/message_validator/src/consensus_message.rs
+++ b/anchor/message_validator/src/consensus_message.rs
@@ -1,6 +1,7 @@
 use std::{convert::Into, sync::Arc, time::Duration};
 
 use duties_tracker::DutiesProvider;
+use openssl::{pkey::Public, rsa::Rsa};
 use slot_clock::SlotClock;
 use ssv_types::{
     CommitteeInfo, IndexSet, OperatorId, Round, Slot, VariableList,
@@ -34,6 +35,7 @@ pub(crate) fn validate_consensus_message(
         validation_context.signed_ssv_message,
         &consensus_message,
         validation_context.committee_info,
+        Some(validation_context.operators_pk),
     )?;
 
     validate_qbft_logic(&validation_context, &consensus_message, duty_state)?;
@@ -64,6 +66,7 @@ pub(crate) fn validate_consensus_message_semantics(
     signed_ssv_message: &SignedSSVMessage,
     consensus_message: &QbftMessage,
     committee_info: &CommitteeInfo,
+    operators_pks: Option<&[Rsa<Public>]>,
 ) -> Result<(), ValidationFailure> {
     let signers = signed_ssv_message.operator_ids().len();
 
@@ -139,13 +142,14 @@ pub(crate) fn validate_consensus_message_semantics(
         });
     }
 
-    validate_justifications(consensus_message)?;
+    validate_justifications(consensus_message, operators_pks)?;
 
     Ok(())
 }
 
 pub(crate) fn validate_justifications(
     consensus_message: &QbftMessage,
+    operators_pks: Option<&[Rsa<Public>]>,
 ) -> Result<(), ValidationFailure> {
     // Rule: Can only exist for Proposal messages
     let prepare_justifications = &consensus_message.prepare_justification;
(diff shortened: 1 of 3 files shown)


anchor/message_validator/src/consensus_message.rs
```diff
@@ -139,13 +142,14 @@ pub(crate) fn validate_consensus_message_semantics(
...
 
-    validate_justifications(consensus_message)?;
+    validate_justifications(consensus_message, operators_pks)?;
 
     Ok(())
 }
 
 pub(crate) fn validate_justifications(
     consensus_message: &QbftMessage,
+    operators_pks: Option<&[Rsa<Public>]>,
```
dknopik: I think it is better to have this be a non-option and adjust the tests instead to sign the justifications properly

anchor/message_validator/src/consensus_message.rs
```diff
@@ -164,6 +168,13 @@ pub(crate) fn validate_justifications(
         return Err(ValidationFailure::UnexpectedRoundChangeJustifications);
     }
 
+    if let Some(pks) = operators_pks {
+        prepare_justifications
+            .iter()
+            .chain(round_change_justifications.iter())
+            .try_for_each(|signed_message| verify_message_signatures(signed_message, pks))?;
+    }
```
dknopik: `verify_message_signatures` assumes that the provided public keys are for the correct operators - however, `validation_context.operators_pk` is currently filled with the pubkey(s) for the root message. I think it might make sense to change `validation_context.operators_pk` to a map from operator id to public key and fill it with the needed pubkeys (either every operator id found in the root message + justifications or just all pubkeys of the relevant committee).

anchor/message_validator/src/consensus_message.rs
```diff
@@ -139,13 +142,14 @@ pub(crate) fn validate_consensus_message_semantics(
         });
     }
 
-    validate_justifications(consensus_message)?;
+    validate_justifications(consensus_message, operators_pks)?;
```
diegomrsantos: I believe pk is more used to mean private key

anchor/message_validator/src/partial_signature.rs
```diff
@@ -41,8 +41,9 @@ pub(crate) fn validate_partial_signature_message(
     )?;
 
     let operator_pk = validation_context
-        .operators_pk
-        .first()
+        .operator_public_keys
+        .values()
+        .next()
```
dknopik: This does not work as the first key returned by `values` is not necessarily the one we need.

dknopik (commented):
LGTM, but would appreciate @diegomrsantos's opinion

anchor/message_validator/src/consensus_message.rs
```diff
@@ -47,7 +49,7 @@ pub(crate) fn validate_consensus_message(
 
     verify_message_signatures(
         validation_context.signed_ssv_message,
-        validation_context.operators_pk,
+        validation_context.operator_public_keys,
```
diegomrsantos: Thanks for the change!

anchor/message_validator/src/lib.rs
```diff
@@ -956,6 +968,14 @@ mod tests {
         MessageId::new(&domain, role, &duty_executor)
     }
 
+    // Helper to create a HashMap of CommitteeId -> PublicKey for tests
+    pub(crate) fn create_hashmap_for_test(
```
diegomrsantos: How about `create_operator_pub_keys`?

anchor/message_validator/src/lib.rs
```diff
@@ -755,19 +764,20 @@ pub(crate) fn compute_quorum_size(committee_size: usize) -> usize {
     f * 2 + 1
 }
 
-fn get_operator_pks(
+fn get_operator_public_keys_map(
```
diegomrsantos: How about `get_operator_pub_keys`? The returning type is already described in the signature

anchor/message_validator/src/consensus_message.rs
diegomrsantos: very nit: `operator_pub_keys`

diegomrsantos:
> LGTM, but would appreciate @diegomrsantos's opinion

Looks good, just small nits. Thanks!

diegomrsantos (approved):
Thanks!

PR #592, opened 2025-09-09
fix: qbft spec parity

diegomrsantos:
## Issue Addressed

The QBFT implementation diverged from the spec in several safety / liveness impacting areas:
- Leader could fall back to proposing the start data when the highest prepared RoundChange's full data was missing, risking proposing an unprepared (potentially conflicting) value.
- RoundChange messages with prepared_round equal to the target round were accepted (spec requires prepared_round < round), weakening a safety invariant.
- A RoundChange carrying full_data did not strictly enforce that the hash (root) matched the data bytes, permitting an inconsistent (root, data) pair.
- Proposal justification demanded all RoundChanges share the same root, which is stricter than the spec and could block progress when only the highest prepared RC has the data/root.

This PR aligns behavior with the spec: only the highest prepared RC can justify the value, prepared_round must be strictly less than the round, hash/data integrity is enforced, and root uniformity is relaxed except for the highest prepared RC.

## Proposed Changes

1. Introduce RcJustificationOutcome enum; leader proposes only from highest prepared RC and never falls back if that data is missing.
2. Enforce prepared_round < round (reject equality).
3. Require full_data hash matches provided root (single decode + hash verification path).
4. Relax root uniformity requirement: only highest prepared RC must match proposal root; others may lack data.
5. Refactor validate_message to remove redundant decoding/logging; minor test harness logging adjustment to avoid global subscriber panic.

## Additional Info

Added spec parity tests (inline in `qbft/src/lib.rs`) covering:
- Highest prepared RC selection logic.
- Rejection of prepared_round == round.
- Hash mismatch rejection for RoundChange with full_data.
- Progress when only highest prepared RC supplies data/root.

### DIFF
--- a/anchor/common/qbft/src/lib.rs
+++ b/anchor/common/qbft/src/lib.rs
@@ -383,39 +383,28 @@ where
 
         let claimed_hash = highest_prepared.qbft_message.root;
 
-        // First, try to get data from the round change message itself
-        if !highest_prepared.signed_message.full_data().is_empty() {
-            // The round change includes the full data - decode and use it
-            if let Ok(data) = D::from_ssz_bytes(highest_prepared.signed_message.full_data()) {
-                // Verify the data matches the claimed hash
-                if data.hash() == claimed_hash {
-                    // Validate against the provided validator
-                    if self.data_validator.validate(&data, &self.start_data) {
-                        return RcJustificationOutcome::HighestPrepared(ValidData::new(
-                            Some(Arc::new(data)),
-                            claimed_hash,
-                        ));
-                    } else {
-                        warn!("Round change full data is invalid");
-                    }
-                } else {
-                    warn!("Round change full data doesn't match claimed hash");
-                }
-            } else {
-                warn!("Failed to decode round change full data");
-            }
+        // We must have valid full_data on the highest prepared RC itself. If not, do not propose.
+        if highest_prepared.signed_message.full_data().is_empty() {
+            return RcJustificationOutcome::PreparedExistsButDataMissing(claimed_hash);
         }
 
-        // If we don't have the data in the round change, try our local storage
-        if let Some(data) = self.data.get(&claimed_hash) {
-            return RcJustificationOutcome::HighestPrepared(ValidData::new(
-                Some(data.clone()),
-                claimed_hash,
-            ));
+        // Decode and validate the data from the RC
+        let Ok(data) = D::from_ssz_bytes(highest_prepared.signed_message.full_data()) else {
+            warn!("Failed to decode round change full data");
+            return RcJustificationOutcome::PreparedExistsButDataMissing(claimed_hash);
+        };
+
+        if data.hash() != claimed_hash {
+            warn!("Round change full data doesn't match claimed hash");
+            return RcJustificationOutcome::PreparedExistsButDataMissing(claimed_hash);
         }
 
-        // Spec: highest prepared exists but we don't have the data yet. Do not propose start data.
-        RcJustificationOutcome::PreparedExistsButDataMissing(claimed_hash)
+        if !self.data_validator.validate(&data, &self.start_data) {
+            warn!("Round change full data is invalid");
+            return RcJustificationOutcome::PreparedExistsButDataMissing(claimed_hash);
+        }
+
+        RcJustificationOutcome::HighestPrepared(ValidData::new(Some(Arc::new(data)), claimed_hash))
     }
 
     // Handles the beginning of a round.
--- a/anchor/common/qbft/src/lib.rs
+++ b/anchor/common/qbft/src/lib.rs
@@ -634,7 +634,7 @@ where
                     max_prepared_msg = Some(round_change.clone());
                 }
 
-                // Check that prepared round is not greater than current round
+                // Check that prepared round is less than the round change round
                 if round_change.data_round >= round_change.round {
                     warn!(
                         "Round change has prepared round {} >= round {}",
--- a/anchor/common/qbft/src/lib.rs
+++ b/anchor/common/qbft/src/lib.rs
@@ -56,6 +56,13 @@ pub struct ValidData<D: QbftData<Hash = Hash256>> {
     data: Option<Arc<D>>,
 }
 
+// Outcome when justifying a proposal from a RoundChange quorum
+enum RcJustificationOutcome<D: QbftData<Hash = Hash256>> {
+    HighestPrepared(ValidData<D>),
+    PreparedExistsButDataMissing(Hash256),
+    NoPrepared,
+}
+
 impl<D: QbftData<Hash = Hash256>> ValidData<D> {
     fn new(data: Option<Arc<D>>, hash: Hash256) -> Self {
         Self { hash, data }
@@ -310,7 +317,7 @@ where
         // Message is not a decide message, we know there is only one signer
         let signer = wrapped_msg.signed_message.operator_ids().first()?;
 
-        // Fulldata may be empty. This is still considered valid though
+        // Fulldata may be empty. This is still considered valid.
         if wrapped_msg.signed_message.full_data().is_empty() {
             let valid_data = Some(ValidData::new(None, wrapped_msg.qbft_message.root));
             return Some((valid_data, *signer));
@@ -319,18 +326,21 @@ where
         // Try to decode the data. If we can decode the data, then also validate it
         let data = match D::from_ssz_bytes(wrapped_msg.signed_message.full_data()) {
             Ok(data) => data,
-            _ => {
-                error!(
-                    msg = %wrapped_msg,
-                    "Invalid full data received",
-                );
+            Err(_) => {
+                error!(msg = %wrapped_msg, "Invalid full data received");
                 debug!(
                     full_data = hex::encode(wrapped_msg.signed_message.full_data()),
-                    "Raw invalid full data",
+                    "Raw invalid full data"
                 );
                 return None;
             }
         };
+
+        // Ensure the decoded data's hash matches the advertised root
+        if data.hash() != wrapped_msg.qbft_message.root {
+            warn!("Full data hash mismatches message root");
+            return None;
+        }
 
         if !self.data_validator.validate(&data, &self.start_data) {
             return None;
@@ -347,14 +357,14 @@ where
     /// Justify the round change quorum
     /// Finds the highest prepared value from round change messages and returns it
     /// for the proposal.
-    fn justify_round_change_quorum(&self) -> Option<ValidData<D>> {
+    fn justify_round_change_quorum(&self) -> RcJustificationOutcome<D> {
         let round_change_messages = self
             .round_change_container
             .get_messages_for_round(self.current_round);
 
         // Need quorum to proceed
         if round_change_messages.len() < self.config.quorum_size() {
-            return None;
+            return RcJustificationOutcome::NoPrepared;
         }
 
         // Find the round change with the highest prepared round
@@ -364,7 +374,9 @@ where
             .max_by_key(|msg| msg.qbft_message.data_round);
 
         // If no one prepared anything, return None (will use start data)
-        let highest_prepared = highest_prepared?;
+        let Some(highest_prepared) = highest_prepared else {
+            return RcJustificationOutcome::NoPrepared;
+        };
 
         let claimed_hash = highest_prepared.qbft_message.root;
 
@@ -374,7 +386,10 @@ where
             if let Ok(data) = D::from_ssz_bytes(highest_prepared.signed_message.full_data()) {
                 // Verify the data matches the claimed hash
                 if data.hash() == claimed_hash {
-                    return Some(ValidData::new(Some(Arc::new(data)), claimed_hash));
+                    return RcJustificationOutcome::HighestPrepared(ValidData::new(
+                        Some(Arc::new(data)),
+                        claimed_hash,
+                    ));
                 } else {
                     warn!("Round change full data doesn't match claimed hash");
                 }
@@ -385,7 +400,10 @@ where
 
         // If we don't have the data in the round change, try our local storage
         if let Some(data) = self.data.get(&claimed_hash) {
-            return Some(ValidData::new(Some(data.clone()), claimed_hash));
+            return RcJustificationOutcome::HighestPrepared(ValidData::new(
+                Some(data.clone()),
+                claimed_hash,
+            ));
         }
 
         warn!(
@@ -393,8 +411,7 @@ where
             claimed_hash
         );
 
-        // Return None - will fall back to start data
-        None
+        RcJustificationOutcome::PreparedExistsButDataMissing(claimed_hash)
     }
 
     // Handles the beginning of a round.
@@ -415,9 +432,15 @@ where
 
             // Check justification of round change quorum. If there is a justification, we will use
             // that data. Otherwise, use the initial state data
-            let valid_data = self
-                .justify_round_change_quorum()
-                .unwrap_or_else(|| self.valid_start_data.clone());
+            let valid_data = match self.justify_round_change_quorum() {
+                RcJustificationOutcome::HighestPrepared(vd) => vd,
+                RcJustificationOutcome::NoPrepared => self.valid_start_data.clone(),
+                RcJustificationOutcome::PreparedExistsButDataMissing(h) => {
+                    // Spec: must propose highest prepared if exists; if missing bytes, wait.
+                    warn!(hash = ?h, "Highest prepared exists but data is missing; not proposing this round");
+                    return;
+                }
+            };
 
             debug!(hash = ?valid_data.hash, "Current leader proposing data");
 
@@ -608,17 +631,11 @@ where
                 }
 
                 // Check that prepared round is not greater than current round
-                if round_change.data_round > round_change.round {
+                if round_change.data_round >= round_change.round {
                     warn!(
-                        "Round change has prepared round {} > round {}",
+                        "Round change has prepared round {} >= round {}",
                         round_change.data_round, round_change.round
                     );
-                    return false;
-                }
-
-                // Verify that if round change has full data, it matches the root
-                if msg.qbft_message.root != round_change.root {
-                    warn!("Proposal root doesn't match round change prepared root");
                     return false;
                 }
 
@@ -948,7 +965,7 @@ where
                 return;
             }
 
-            if qbft_msg.data_round > qbft_msg.round {
+            if qbft_msg.data_round >= qbft_msg.round {
                 debug!(
                     from = *operator_id,
                     data_round = qbft_msg.data_round,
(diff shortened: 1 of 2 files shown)


anchor/common/qbft/src/lib.rs
```diff
@@ -394,7 +412,7 @@ where
         );
 
         // Return None - will fall back to start data
-        None
+        RcJustificationOutcome::PreparedExistsButDataMissing(claimed_hash)
```
diegomrsantos: When the highest prepared RC’s bytes weren’t locally available it returned None

Zacholme7:
havent looked into this yet, but on a first glance it clashes with #574 and I am going over that right now

Also, even thought his might be spec we need to be sure this does not deviate from go-ssv behavior

anchor/common/qbft/src/lib.rs
```diff
@@ -415,9 +432,15 @@ where
...
-                .unwrap_or_else(|| self.valid_start_data.clone());
+            let valid_data = match self.justify_round_change_quorum() {
+                RcJustificationOutcome::HighestPrepared(vd) => vd,
+                RcJustificationOutcome::NoPrepared => self.valid_start_data.clone(),
+                RcJustificationOutcome::PreparedExistsButDataMissing(h) => {
+                    // Spec: must propose highest prepared if exists; if missing bytes, wait.
+                    warn!(hash = ?h, "Highest prepared exists but data is missing; not proposing this round");
+                    return;
+                }
+            };
```
diegomrsantos, line 436: Before it unconditionally fell back to the instance’s start data when justify_round_change_quorum returned `None`

Why this is incorrect per spec

- In the spec’s UponRoundChange (1.1 node.dfy):
   - A leader may propose only if hasReceivedProposalJustificationForLeadingRound(current) holds, which requires isProposalJustificationForLeadingRound(...).
   - isProposalJustificationForLeadingRound delegates to isProposalJustification, which constrains the chosen block:
     - The proposed block must be in the set of available blocks carried by the collected ROUND-CHANGE messages (often described as block ∈ setOfAvailableBlocks where setOfAvailableBlocks = receivedBlocksInRoundChanges(roundChanges)).
     - If any prepared certificate exists among the RCs, the proposal must match the highest prepared certificate’s digest and be backed by a quorum of matching PREPAREs.
   - If the leader lacks a valid proposal justification (e.g., the highest-prepared block’s bytes aren’t present in any RC), the spec’s second branch applies: only f+1 ROUND-CHANGE are used to advance to a new round, and the node emits a ROUND-CHANGE (not a proposal).
 - Where to look:
   - 1.1 node.dfy → predicate UponRoundChange: first branch guarded by hasReceivedProposalJustificationForLeadingRound(current) produces a Proposal; the else-if branch (f+1 RCs) produces a ROUND-CHANGE.
   - The helper predicates named in that first branch: hasReceivedProposalJustificationForLeadingRound → isProposalJustificationForLeadingRound → isProposalJustification (see 1.1 node.dfy and 1.2 node_auxiliary_functions.dfy).
   - The “block-in-RC-set” condition and “highest prepared certificate” matching are part of isProposalJustification/isProposalJustificationForLeadingRound semantics (node_auxiliary_functions.dfy), including the receivedBlocksInRoundChanges construction.
   - Public landing page with inlined Dafny and anchors:
     - 1.1 node.dfy (UponRoundChange, hasReceivedProposalJustificationForLeadingRound): [https://entethalliance.github.io/client-spec/qbft_spec.html#nodedfysec](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html)
     - Linked sources for auxiliary functions: node_auxiliary_functions.dfy (receivedBlocksInRoundChanges, isProposalJustification)
   - Implication: Proposing “start data” when a prepared value exists but bytes are missing is not a valid proposal justification; honest receivers should reject it.

diegomrsantos:
> havent looked into this yet, but on a first glance it clashes with #574 and I am going over that right now

Could you elaborate?

> Also, even thought his might be spec we need to be sure this does not deviate from go-ssv behavior

Could you also elaborate on why the go-ssv behavior is the source of correctness for the algorithm?

Zacholme7:
> Could you elaborate?

#574 will just introduce some merge conflicts

> Could you also elaborate on why the go-ssv behavior is the source of correctness for the algorithm?

This is what we have been running into. The spec is the source, but if we introduce "spec-correct" changes that have incompatibilities with go-ssv then that is an issue. The current implementation was based off of go-ssv so we just need to be sure this pr does not introduce any differences.

anchor/common/qbft/src/lib.rs
diegomrsantos: cc @Zacholme7

anchor/common/qbft/src/lib.rs
```diff
@@ -607,20 +635,14 @@ where
                 }
 
                 // Check that prepared round is not greater than current round
-                if round_change.data_round > round_change.round {
+                if round_change.data_round >= round_change.round {
```
diegomrsantos, line 627: see `predicate validRoundChange` on https://entethalliance.org/specs/qbft/v1/#nodeauxdfysec

anchor/common/qbft/src/lib.rs
```diff
@@ -322,19 +329,22 @@ where
...
                 );
                 return None;
             }
         };
 
+        // Ensure the decoded data's hash matches the advertised root
+        if data.hash() != wrapped_msg.qbft_message.root {
+            warn!("Full data hash mismatches message root");
+            return None;
+        }
```
diegomrsantos, line 346: The spec reasons over digests and binds them consistently across PROPOSAL/PREPARE/PCs (e.g., digest(proposedBlock) and preparedValue). It doesn’t model raw bytes, so an implementation must ensure any carried bytes hash to the digest to preserve those semantics. See 1.1 node.dfy for digest consistency in UponProposal/UponPrepare and the proposal justification path in UponRoundChange: https://entethalliance.github.io/client-spec/qbft_spec.html#nodedfysec

anchor/common/qbft/src/lib.rs
```diff
@@ -606,21 +634,15 @@ where
...
                     );
                     return false;
                 }
 
-                // Verify that if round change has full data, it matches the root
-                if msg.qbft_message.root != round_change.root {
-                    warn!("Proposal root doesn't match round change prepared root");
-                    return false;
-                }
-
```
diegomrsantos, line 623: Before

- While iterating RCs with data_round > 0, it enforced:
  - if msg.qbft_message.root != round_change.root { return false; }
- This demanded every RC match the proposal’s root, which can block progress.

Now

- For RCs: validate each individually, track max_prepared_round and its message.
- After the loop: if a max prepared exists, require:
  - proposal.root == max_prepared_msg.root, and
  - proposal.prepare_justification forms a quorum for (max_prepared_round, max_prepared_root).
- Other RCs may be unprepared or reference different values; no uniformity requirement.

Spec reference:
  - UponRoundChange requires a valid proposal justification: when a prepared certificate exists in the RC set, the leader must propose that prepared value and include its PC. It doesn’t require all RCs to share the same value—only that the proposed block be one of the receivedBlocksInRoundChanges and, if prepared exists, that it match the highest prepared certificate and be backed by a quorum of PREPAREs. See the first branch of UponRoundChange and helper predicates (isProposalJustificationForLeadingRound, isProposalJustification), plus the “received blocks in RCs” construction described in the auxiliary functions: https://entethalliance.github.io/client-spec/qbft_spec.html#nodedfysec

Zacholme7:
To be honest I am extremely confused with everything that it going on here. Could you ELI5 these changes and why they are necessary? Is this all confirmed in the spec/go-ssv code or just stuff that the qbft-subagent is spitting out?

diegomrsantos:
> To be honest I am extremely confused with everything that it going on here. Could you ELI5 these changes and why they are necessary? Is this all confirmed in the spec/go-ssv code or just stuff that the qbft-subagent is spitting out?

What are you confused about? Didn't all the comments I added help at all?

anchor/common/qbft/src/lib.rs
diegomrsantos: Here’s the intuition and the formal-ish reasoning for why the spec does not require all RoundChange (RC) messages in the quorum to carry the same root/value, and instead only insists that:

1. there exists at least one RC that carries a prepared certificate with the highest round among the RCs collected,
2. the proposal’s value matches that highest-prepared value (with the proposal’s round number updated), and
3. the proposal includes a quorum of PREPAREs for that (preparedRound, preparedValue).

Why “existence of a highest-prepared RC” is enough for safety

- Prepared certificate = 2f+1 PREPAREs at some round r. Once such a certificate exists for value V at round r, no conflicting value can get a prepared certificate at the same round, because any two size-(2f+1) quorums intersect in ≥ f+1 honest validators, and honest validators won’t sign conflicting PREPAREs in the same round. This is the classic quorum-intersection safety invariant inherited from PBFT.

- Across view/round changes, the new leader is required to learn the highest prepared certificate (highest round among the RCs it gathers) and propose that value. This is exactly the rule in IBFT 2.0 (the closest, well-documented ancestor of QBFT): if any RCs include prepared certificates, the proposal must match the value attached to the highest prepared certificate among them. Uniformity of all RCs is not required; one RC carrying the highest prepared cert is sufficient, as long as the proposal also carries the PREPARE quorum for that pair. 

- This mirrors the “highest QC”/“highest lock**” handover rule used in modern BFT families (HotStuff/Tendermint lineage): leaders extend (or repropose) the highest certified value they learn during the view change. The safety argument is the same — highest certificate + quorum intersection prevents conflicting commits

Why requiring all RCs to match the same root hurts liveness (and isn’t needed)

- In real networks, honest validators may have seen different subsets of messages before timing out. Some honest RCs might be unprepared (no prepared cert), others might carry lower prepared rounds, and (crucially) at least one honest RC will carry the highest prepared certificate if such a certificate exists.

- If you demanded that every RC in the RC-quorum show the same prepared value/root, you could fail to assemble any RC-quorum even when safety would allow progress — because many honest nodes simply didn’t see those PREPAREs yet (they cannot “invent” them). The spec therefore only requires existence of a highest-prepared RC and a matching PREPARE quorum attached to the proposal. That’s sufficient for safety (see above) and improves liveness under partial synchrony. This exact rule (pick the highest-prepared among the RCs; don’t require uniformity) is explicitly stated for IBFT 2.0 proposals > round 0.

Where this appears in the spec lineage

- QBFT v1 (EEA) expresses proposal acceptance with predicates like isProposalJustification… and drives leader behavior on a RoundChange via UponRoundChange — i.e., propose only when the justification for the leading round holds (which includes the “prepared-value matches proposal” condition). The doc is organized with Dafny predicates and an existential style: there exists a highest-prepared RC that matches, and the proposal carries the necessary PREPAREs. Uniformity across all RCs is not required.  

- IBFT 2.0 paper (Section on Proposal acceptance for r>0): “otherwise, the hash of the (block, highest-prepared-round) in the proposal must match the hash in any of the prepared certificates with the highest round among those in the RC-certificate.” This is the precise statement that only the highest prepared needs to line up; other RCs may be empty or carry different (lower) prepared values. 

A tiny worked example (n=4, f=1)

- Suppose round 5 had a prepared certificate for value A (3 PREPAREs).

- At round change to 6, the leader collects 3 RCs:
  – RC#1: includes prepared (r=5, value A)
  – RC#2: unprepared (no prepared value)
  – RC#3: prepared at lower round 3 for value B
  
- Spec rule: choose the highest prepared (r=5, value A), and propose A with the PREPARE quorum for (5, A) attached.
- Safety holds (quorum intersection with the (5, A) PREPARE certificate), and progress is possible. If we required all RCs to mention A, we could be stuck even though the system is safe to proceed.

anchor/common/qbft/src/lib.rs
diegomrsantos: Why the strict inequality matters (intuitively + rigorously)

- Temporal sanity / no circular justifications
A RoundChange for round r is a request to enter r using information that happened before r. If you let prepared_round == r, the “evidence” (the prepare-certificate) would be from the very round you are trying to enter, creating a causality loop (“I can enter round r because I already prepared in round r”). The spec forbids this by requiring prepared_round < r.

- Monotone-lock invariant
The prepared pair (preparedRound, preparedValue) is a historical lock that leaders must carry forward when rounds advance. Making it strictly older than the target round ensures you only leverage past prepares to guide future proposals. Allowing equality weakens this invariant and mixes the “evidence” of the target round with the round’s own proposal/prepare flow, which the spec 

- Prevents self-referential proposals at the new round
Proposal justification rules for rounds r > 0 rely on taking the highest prepared from the RoundChange quorum and re-proposing that value (with the round updated). That mechanism assumes the certificate came from a strictly earlier round; otherwise the proposal for r could try to justify itself using prepares from r, which is circular. The strict < r side condition is what prevents that. (EEA spec’s RoundChange/Proposal justification functions encode this flow.) 

- Matches the IBFT 2.0 lineage
QBFT inherits the RC structure from IBFT 2.0: ROUND-CHANGE⟨λ, r, pr, pv⟩, where pr is the round in which a value was already prepared and r is the next round to start; conceptually, pr < r. Using evidence from the same round you are trying to enter contradicts the round-change protocol’s intention.  

- Bottom line: Accepting prepared_round == round contradicts the spec’s validRoundChange predicate and weakens the protocol’s causal and locking invariants. The correct behavior is to reject such RoundChange messages (i.e., enforce prepared_round < round), which is exactly what the spec requires.

anchor/common/qbft/src/lib.rs
```diff
@@ -955,7 +977,7 @@ where
                 return;
             }
 
-            if qbft_msg.data_round > qbft_msg.round {
+            if qbft_msg.data_round >= qbft_msg.round {
```
diegomrsantos, line 969: @Zacholme7 @dknopik why do we duplicate this check?

anchor/common/qbft/src/lib.rs
dknopik: prepared roundchanges should always carry the full data, shouldn't they?

anchor/common/qbft/src/lib.rs
Zacholme7: They are different contexts. This one is for just a round change message and the other is for proposal justifications

dknopik: where is the check duplicated?

anchor/common/qbft/src/lib.rs
diegomrsantos: so what the lines starting at 409 mean?

anchor/common/qbft/src/lib.rs
diegomrsantos: Here’s the (spec-driven) intuition for why a leader must not fall back to “start data” when there is a highest-prepared RoundChange (RC) but the block bytes aren’t available:

- What a valid proposal needs when any RC is prepared (round > 0).
The spec’s isProposalJustification requires: (i) a quorum of RCs; (ii) the proposed block must be one of the blocks actually carried in the RC set (block ∈ setOfAvailableBlocks, where setOfAvailableBlocks = receivedBlocksInRoundChanges(roundChanges)); (iii) existence of an RC with the highest prepared round whose preparedValue equals the digest of the proposed block with its round replaced by the prepared round; and (iv) a quorum of PREPAREs matching that (preparedRound, preparedValue). Uniformity across all RCs is not required—only the highest-prepared RC plus the PREPARE quorum must match. 

- Why “start data” is not allowed in that situation.
If a highest-prepared certificate exists but none of the RCs include the corresponding block bytes, then that block is not in setOfAvailableBlocks; the block ∈ setOfAvailableBlocks clause fails, so isProposalJustification is false and the leader has no license to propose anything (including “start data”). In this case the spec’s round-advance logic (the UponRoundChange path) drives nodes to accumulate RCs (f+1 → send RC; quorum → move rounds), not to emit an unjustified proposal. 

- Safety intuition (ties back to PBFT/IBFT2).
The “highest-prepared certificate locks the value” rule prevents a leader from proposing an unrelated value once any value is prepared; otherwise you risk conflicting prepares/commits across rounds. This is the same lock-in principle used by PBFT/IBFT2, where a new leader must propose the block proven by the highest prepared certificate. 

So, per the spec: when a highest-prepared RC exists but its block bytes aren’t present in any RC, the leader must not propose (certainly not “start data”); nodes should continue the round-change process until a valid proposal justification (including the actual block bytes from the RC set) is possible.

anchor/common/qbft/src/lib.rs
Zacholme7: `>` is also being used here https://github.com/ssvlabs/ssv-spec/blob/d4c734909677b5fd3304e1d9220963cc9d794f09/qbft/round_change.go#L304

diegomrsantos: > where is the check duplicated?

https://github.com/sigp/anchor/pull/592/files#diff-1480a846af2b359779733974f2f345dadfe4e3380369d7cc67218cf8494286d8R638

dknopik: that is a different function

anchor/common/qbft/src/lib.rs
dknopik: Too long; didn't read fully

But we are in a codepath where `round_change.data_round > 0` - so the sender is claiming that the value was previously prepared, so it seems reasonable that the messages used to justify that in fact do refer to the same data

for the case where the RC's do NOT prepare a value, we have the code path where data_round == 0

anchor/common/qbft/src/lib.rs
dknopik: of the old version or the new version?

anchor/common/qbft/src/lib.rs
diegomrsantos: I know :D, but I was wondering if this could be unified somehow. Very easy to forget to update both places

anchor/common/qbft/src/lib.rs
diegomrsantos: That's not what the text is about

anchor/common/qbft/src/lib.rs
diegomrsantos: I see my PR was based in an older version of the code that your PR modified. Maybe my merge is incorrect, as the code doesn't use `self.data` anymore. But your code still returns `None` if the highest prepared doesn't have `full_data` https://github.com/sigp/anchor/pull/592/files#diff-1480a846af2b359779733974f2f345dadfe4e3380369d7cc67218cf8494286d8L382

diegomrsantos: Apparently, in this case, we shouldn't propose `self.valid_start_data`, only if there's no prepared

diegomrsantos: simplified the code to match the latest code

anchor/common/qbft/src/lib.rs
```diff
@@ -1384,3 +1406,293 @@ where
...
+mod spec_parity_tests {
+    //! Spec parity tests
+    //!
+    //! These tests pin behaviour required by the QBFT spec:
+    //! - Only the highest prepared RoundChange must match the Proposal value.
+    //! - A RoundChange with a prepared value must satisfy `prepared_round < round`.
+    //! - Any message carrying `full_data` must have bytes whose hash equals the advertised root.
+    //!
+    //! Why in this file? They touch internal helpers like `validate_proposal_justifications`
+    //! and `validate_message`. Keeping them here avoids widening the public API only for tests.
```
dknopik, line 1409: it does not matter for access whether a submodule is in the same or a different file, so this reasoning is faulty

anchor/common/qbft/src/lib.rs
dknopik: Ok sorry, after a night of sleep and fresh coffee I think I get it now.

I have thrown it into the fuzzer, but I am optimistic that you are right. thanks!

anchor/common/qbft/src/lib.rs
dknopik: makes sense

anchor/common/qbft/src/lib.rs
Zacholme7: Looks like we are already checking this in the message validator [here](https://github.com/sigp/anchor/blob/a79943b1ecd43c3453b75f9540327d0ea454e956/anchor/message_validator/src/consensus_message.rs#L99-L103)

Zacholme7:
So as i understand it now the main crux of this pr is that if we have a prepared value but no accompanying data, dont propose at all when before I was just falling back to start data? I think that makes sense

diegomrsantos:
> So as i understand it now the main crux of this pr is that if we have a prepared value but no accompanying data, dont propose at all when before I was just falling back to start data? I think that makes sense

This is one of the issues. It might be better to break it down into smaller issues, as this has become too complex. I'll do it.

Zacholme7:
Sounds good. I think I am on board with everything here

anchor/common/qbft/src/lib.rs
diegomrsantos: You think we need to add it here too?

diegomrsantos:
This PR was split into 3 new ones

PR #596, opened 2025-09-10
fix(qbft): check whether the leader is correct wrt the round sent with the message

dknopik:
## Issue Addressed

On receiving a PROPOSE, we check whether the leader sent it, based on the current round.

This is not correct when receiving a PROPOSE that advances the round through the provided justification. 

## Proposed Changes

Use the `round` from the message. Later in the function we will update the `current_round` to that value (if it is not equal yet), so this seems reasonable.

## Additional Info
oof

### DIFF
--- a/anchor/common/qbft/src/lib.rs
+++ b/anchor/common/qbft/src/lib.rs
@@ -218,10 +218,10 @@ where
     }
 
     // Validation and check functions.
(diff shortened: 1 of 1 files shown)


Zacholme7 (approved):
Nice!!

PR #598, opened 2025-09-11
fix(qbft): prevent unsafe leader fallback when highest prepared data missing

diegomrsantos:
## Issue Addressed

Addresses one critical QBFT specification compliance issue from the broader work in PR #592.

Previously, when the leader had a highest prepared RoundChange but the full_data was missing, it would incorrectly fall back to proposing the start_data. This violates the QBFT specification which requires the leader to only propose from the highest prepared value when it exists.

## Proposed Changes

This change introduces `RcJustificationOutcome` enum to distinguish between three scenarios:
- `HighestPrepared(ValidData<D>)`: Valid prepared data available for proposal
- `PreparedExistsButDataMissing(Hash256)`: Prepared value exists but data is inaccessible  
- `NoPrepared`: No prepared values, safe to use start_data

When `PreparedExistsButDataMissing` occurs, the leader now correctly waits instead of unsafely proposing start_data, ensuring QBFT specification compliance.

### Key Changes:
1. **RcJustificationOutcome enum**: Replaces boolean return with explicit outcome states
2. **Leader logic update**: Handles missing prepared data by waiting instead of fallback
3. **Comprehensive test**: Verifies leader waits when highest prepared data is missing

## Additional Info

This is part of the broader QBFT specification compliance effort detailed in PR #592. For complete technical context and rationale, refer to that PR's description and comments.

This focused change addresses only the leader proposal fallback issue, with additional specification compliance fixes to follow in separate PRs.

### DIFF
--- a/anchor/common/qbft/src/lib.rs
+++ b/anchor/common/qbft/src/lib.rs
@@ -50,22 +50,41 @@ impl<D: QbftData<Hash = Hash256>> MessageData<D> {
 }
 
 // Store hash and deserialized data together to avoid redundant lookups
-#[derive(Debug, Default, Clone)]
+#[derive(Debug, Clone)]
 pub struct ValidData<D: QbftData<Hash = Hash256>> {
     hash: D::Hash,
-    data: Option<Arc<D>>,
+    data: Arc<D>,
+}
+
+// Store only hash when data is not available
+#[derive(Debug, Clone)]
+pub struct HashOnly {
+    hash: Hash256,
+}
+
+// Enum to represent message content - either complete data or just the hash
+#[derive(Debug, Clone)]
+pub enum MessageContent<D: QbftData<Hash = Hash256>> {
+    Complete(ValidData<D>),
+    HashOnly(HashOnly),
 }
 
 // Outcome when justifying a proposal from a RoundChange quorum
 enum RcJustificationOutcome<D: QbftData<Hash = Hash256>> {
     HighestPrepared(ValidData<D>),
-    PreparedExistsButDataMissing(Hash256),
+    PreparedExistsButDataMissing(HashOnly),
     NoPrepared,
 }
 
 impl<D: QbftData<Hash = Hash256>> ValidData<D> {
-    fn new(data: Option<Arc<D>>, hash: Hash256) -> Self {
+    fn new(hash: Hash256, data: Arc<D>) -> Self {
         Self { hash, data }
+    }
+}
+
+impl HashOnly {
+    fn new(hash: Hash256) -> Self {
+        Self { hash }
     }
 }
 
@@ -162,7 +181,7 @@ where
 
         let start_data = Arc::new(start_data);
         let start_data_hash = start_data.hash();
-        let valid_start_data = ValidData::new(Some(start_data.clone()), start_data_hash);
+        let valid_start_data = ValidData::new(start_data_hash, start_data.clone());
 
         let mut qbft = Qbft {
             config,
@@ -256,12 +275,12 @@ where
     // message types
     // Return type expresses that we either have
     // 1) An invalid message via None
-    // 2) A valid message with empty fulldata via Some(None, ID)
-    // 3) A valid message with fulldata via Some(data, ID)
+    // 2) A valid message with hash only via Some(MessageContent::HashOnly, ID)
+    // 3) A valid message with fulldata via Some(MessageContent::Complete, ID)
     fn validate_message(
         &self,
         wrapped_msg: &WrappedQbftMessage,
-    ) -> Option<(Option<ValidData<D>>, OperatorId)> {
+    ) -> Option<(MessageContent<D>, OperatorId)> {
         // Ensure that this message is for the correct round
         if wrapped_msg.qbft_message.round < self.current_round.into() {
             debug!(
@@ -313,8 +332,9 @@ where
             // The message validator already checked this is a decided message (a commit message
             // with > 1 signers). Do not care about data here, just that we had a
             // success
-            let valid_data = Some(ValidData::new(None, wrapped_msg.qbft_message.root));
-            return Some((valid_data, OperatorId::from(0)));
+            let validated_msg =
+                MessageContent::HashOnly(HashOnly::new(wrapped_msg.qbft_message.root));
+            return Some((validated_msg, OperatorId::from(0)));
         }
 
         // Message is not a decide message, we know there is only one signer
@@ -325,8 +345,9 @@ where
         if wrapped_msg.signed_message.full_data().is_empty()
             || wrapped_msg.qbft_message.qbft_message_type == QbftMessageType::RoundChange
         {
-            let valid_data = Some(ValidData::new(None, wrapped_msg.qbft_message.root));
-            return Some((valid_data, *signer));
+            let validated_msg =
+                MessageContent::HashOnly(HashOnly::new(wrapped_msg.qbft_message.root));
+            return Some((validated_msg, *signer));
         }
 
         // Try to decode the data. If we can decode the data, then also validate it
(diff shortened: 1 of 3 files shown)


anchor/common/qbft/src/tests.rs
```diff
@@ -356,3 +356,173 @@ fn test_round_change_validation_skips_round_one_prepared_values() {
...
+        let round_change = QbftMessage {
+            qbft_message_type: QbftMessageType::RoundChange,
+            height: 0,
+            round: 2, // Moving to round 2
+            identifier: [0; 56].to_vec().into(),
+            root: prepared_hash,                // Claims this hash was prepared
+            data_round: 1,                      // Claims preparation happened in round 1
+            round_change_justification: vec![], // No RC justifications needed for this test
+            prepare_justification: vec![],      /* Should have prepare messages but we'll skip
+                                                 * validation */
```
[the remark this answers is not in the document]
diegomrsantos: fmt does that

anchor/common/qbft/src/lib.rs
```diff
@@ -414,14 +423,29 @@ where
...
             // that data. Otherwise, use the initial state data
-            let valid_data = self
-                .justify_round_change_quorum()
-                .unwrap_or_else(|| self.valid_start_data.clone());
-
-            debug!(hash = ?valid_data.hash, "Current leader proposing data");
-
-            // Send the initial proposal and then the following prepare
-            self.send_proposal(valid_data.hash, valid_data.data.expect("Start data exists"));
+            match self.justify_round_change_quorum() {
```
Zacholme7: could do something like 
```rust
let (hash, data) = match self.justify_round_change_quorum() {
// matching logic in here, return corresponding data or from function
}
self.send_proposal(hash, data);
```

diegomrsantos: Done. Also removed the expects

anchor/common/qbft/src/lib.rs
```diff
@@ -50,18 +50,44 @@ impl<D: QbftData<Hash = Hash256>> MessageData<D> {
...
+pub struct HashOnly {
+    hash: Hash256,
+}
+
+// Enum to represent message content - either complete data or just the hash
+#[derive(Debug, Clone)]
+pub enum MessageContent<D: QbftData<Hash = Hash256>> {
+    Complete(ValidData<D>),
+    HashOnly(HashOnly),
+}
```
Zacholme7: IMO I feel like it was more clean and simple before this change, this feels a bit over-engineering the problem

diegomrsantos: We were forced to either call expect or handle the None case when there shouldn't be any None

Zacholme7: Got it

anchor/common/qbft/src/lib.rs
```diff
@@ -50,18 +50,44 @@ impl<D: QbftData<Hash = Hash256>> MessageData<D> {
...
+#[derive(Debug, Clone)]
+pub struct HashOnly {
+    hash: Hash256,
+}
+
+// Enum to represent message content - either complete data or just the hash
+#[derive(Debug, Clone)]
+pub enum MessageContent<D: QbftData<Hash = Hash256>> {
+    Complete(ValidData<D>),
+    HashOnly(HashOnly),
```
Zacholme7: can we get rid of `HashOnly` and just use `Hash256` here?

diegomrsantos: Done

anchor/common/qbft/src/lib.rs
```diff
@@ -50,14 +50,28 @@ impl<D: QbftData<Hash = Hash256>> MessageData<D> {
...
+// Outcome when justifying a proposal from a RoundChange quorum
+enum RcJustificationOutcome<D: QbftData<Hash = Hash256>> {
+    HighestPrepared(ValidData<D>),
+    PreparedExistsButDataMissing(D::Hash),
+    NoPrepared,
 }
 
 impl<D: QbftData<Hash = Hash256>> ValidData<D> {
-    fn new(data: Option<Arc<D>>, hash: Hash256) -> Self {
+    fn new(hash: Hash256, data: Arc<D>) -> Self {
```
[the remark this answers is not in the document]
diegomrsantos: order restored

anchor/common/qbft/src/lib.rs
```diff
@@ -50,14 +50,28 @@ impl<D: QbftData<Hash = Hash256>> MessageData<D> {
...
 
 // Store hash and deserialized data together to avoid redundant lookups
-#[derive(Debug, Default, Clone)]
+#[derive(Debug, Clone)]
 pub struct ValidData<D: QbftData<Hash = Hash256>> {
     hash: D::Hash,
-    data: Option<Arc<D>>,
+    data: Arc<D>,
+}
+
```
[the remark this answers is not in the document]
diegomrsantos: data isn't Option anymore

anchor/common/qbft/src/lib.rs
```diff
@@ -50,14 +50,28 @@ impl<D: QbftData<Hash = Hash256>> MessageData<D> {
...
+// Enum to represent message content - either complete data or just the hash
+#[derive(Debug, Clone)]
+pub enum MessageContent<D: QbftData<Hash = Hash256>> {
```
diegomrsantos, line 61: But I'm not happy with this name. @dknopik what does `Valid` mean in `ValidData`?

Zacholme7: Just that it passed `validate_message`

diegomrsantos: But we don't do much validation in the `data` itself there, do we?

dknopik: we do the value checks, i.e. whatever was passed here https://github.com/sigp/anchor/blob/65bce108d0abc881e7f9bfdc625a80aa40cb8e23/anchor/common/qbft/src/lib.rs#L143

Zacholme7: Not sure I am fully sold on the `MessageContent` name, but I unfortunately cannot think of something better

anchor/common/qbft/src/lib.rs
```diff
@@ -414,21 +430,32 @@ where
...
+            };
+            self.send_proposal(hash, data);
         }
     }
 
     /// Receive a new message from the network
     pub fn receive(&mut self, wrapped_msg: WrappedQbftMessage) {
         // Perform base qbft releveant verification on the message
-        let Some((Some(valid_data), signer)) = self.validate_message(&wrapped_msg) else {
+        let Some((validated_msg, signer)) = self.validate_message(&wrapped_msg) else {
```
Zacholme7, line 461: Maybe this? then we could remove the check below
```rust
let Some((MessageContent::Complete(valid_data), signer)) = self.validate_message(&wrapped_msg)
```

anchor/common/qbft/src/lib.rs
dknopik: wdyt about `DataOrHash`?

anchor/common/qbft/src/lib.rs
dknopik: We cant do that as some message types do not have complete data. 

The check below is only on the `Proposal` variant

anchor/common/qbft/src/lib.rs
Zacholme7: Im on board with that

dknopik (approved):
LGTM, but feel free to apply the enum name change suggested above

anchor/common/qbft/src/lib.rs
diegomrsantos: I'm not satisfied with the name either. But we can get this in as it is and create an issue to improve the naming in general, as imo it's a bit confusing:

```rust

pub trait QbftData: Debug + Clone + Encode + Decode {
    type Hash: Debug + Clone + Eq + Hash;

    fn hash(&self) -> Self::Hash;
}

pub trait QbftDataValidator<D: QbftData>: Send + Sync {
    fn validate(&self, value: &D, start_value: &D) -> bool;
}
```

wdyt?

diegomrsantos:
I've checked the Go code, and it does seem to behave like in this PR. Does anyone think it's worth it if someone else doubles check it?

dknopik:
done. It is kind of hard to assert, as they store things differently, but in the end, I believe the `data not found` case should not really happen anyway because of previous validation. Lets see. We'll stay vigilant and i'll try to catch any deviation in more fuzzing.

PR #599, opened 2025-09-11
fix(qbft): enforce strict prepared_round < round validation for RoundChange messages

diegomrsantos:
## Issue Addressed

The QBFT implementation accepted RoundChange messages where prepared_round was equal to the target round. The spec requires prepared_round < round to avoid self-referential justifications.

## Proposed Changes

- Enforce prepared_round < round by changing validation from `>` to `>=`
- Update validation in both proposal justification and message processing paths  
- Add tests verifying rejection of messages with prepared_round >= round

## Additional Info

Part of the broader QBFT spec compliance work detailed in PR #592.

PR #601, opened 2025-09-11
fix: relax QBFT proposal validation for mixed RoundChange values

diegomrsantos:
## Issue Addressed

This PR addresses part of the QBFT data integrity verification improvements described in the larger PR #592. Specifically, it fixes proposal validation that was incorrectly rejecting valid scenarios where RoundChange messages have different prepared values.

## Proposed Changes

**Problem:** Proposal validation required ALL RoundChange messages with prepared values to share the same root as the proposal. This violated the QBFT specification and could block legitimate consensus progress.

**Solution:** 
- Modified validation to only check that the proposal root matches the highest prepared RoundChange
- Moved validation logic to occur after processing all RoundChange messages
- Added comprehensive test covering mixed RoundChange scenarios with different prepared values

**QBFT Specification:** The spec allows RoundChange messages from different rounds to have different roots. Only the value from the highest prepared round should determine the proposal.

## Additional Info

This change ensures spec compliance while maintaining all existing security validations. See PR #592 for broader context on QBFT data integrity improvements.

### DIFF
--- a/anchor/common/qbft/src/lib.rs
+++ b/anchor/common/qbft/src/lib.rs
@@ -636,19 +636,6 @@ where
             }
         }
 
-        // After processing all round changes, verify the proposal root matches the highest prepared
-        // value According to QBFT spec, proposal should use value from highest prepared
-        // round
-        if let Some(max_prepared_msg) = &max_prepared_msg
-            && msg.qbft_message.root != max_prepared_msg.root
-        {
-            warn!(
-                "Proposal root doesn't match highest prepared round change root. Proposal root: {:?}, Highest prepared root: {:?}, prepared round: {}",
-                msg.qbft_message.root, max_prepared_msg.root, max_prepared_msg.data_round
-            );
-            return false;
-        }
-
         // If there was a value that was also previously prepared, we must also verify all of the
         // prepare justifications
         if let Some(max_prepared_msg) = max_prepared_msg {
@@ -663,7 +650,10 @@ where
 
             // Make sure that the roots match
             if msg.qbft_message.root != max_prepared_msg.root {
-                warn!("Highest prepared does not match proposed data");
+                warn!(
+                    "Proposal root doesn't match highest prepared round change root. Proposal root: {:?}, Highest prepared root: {:?}, prepared round: {}",
+                    msg.qbft_message.root, max_prepared_msg.root, max_prepared_msg.data_round
+                );
                 return false;
             }
 
--- a/anchor/common/qbft/src/lib.rs
+++ b/anchor/common/qbft/src/lib.rs
@@ -615,12 +615,6 @@ where
                     return false;
                 }
 
-                // Verify that if round change has full data, it matches the root
-                if msg.qbft_message.root != round_change.root {
-                    warn!("Proposal root doesn't match round change prepared root");
-                    return false;
-                }
-
                 if !self.check_quorum(&round_change.round_change_justification) {
                     warn!(
                         num_justifications = round_change.round_change_justification.len(),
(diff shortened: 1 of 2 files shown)


anchor/common/qbft/src/lib.rs
```diff
@@ -642,6 +636,19 @@ where
...
+        // value According to QBFT spec, proposal should use value from highest prepared
+        // round
+        if let Some(max_prepared_msg) = &max_prepared_msg
+            && msg.qbft_message.root != max_prepared_msg.root
+        {
+            warn!(
+                "Proposal root doesn't match highest prepared round change root. Proposal root: {:?}, Highest prepared root: {:?}, prepared round: {}",
+                msg.qbft_message.root, max_prepared_msg.root, max_prepared_msg.data_round
+            );
+            return false;
```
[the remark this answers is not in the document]
diegomrsantos: I don't agree

anchor/common/qbft/src/tests.rs
```diff
@@ -356,3 +356,373 @@ fn test_round_change_validation_skips_round_one_prepared_values() {
...
+///
+/// This complements the previous test by checking that data_round > round is also rejected,
+/// ensuring the validation covers the full >= condition.
+fn test_round_change_rejects_prepared_round_greater_than_current_round() {
+    if ENABLE_TEST_LOGGING {
+        let env_filter = EnvFilter::new("debug");
+        let _ = tracing_subscriber::fmt()
+            .compact()
+            .with_env_filter(env_filter)
+            .try_init();
```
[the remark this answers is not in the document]
diegomrsantos: In this specific case for test logging setup, ignoring the error is actually the correct approach because:

  1. try_init() fails if a global subscriber is already set (which happens when running multiple tests)
  2. We don't want tests to fail just because logging is already initialized
  3. The let _ pattern is intentionally discarding the error, which is appropriate here

So the current code is correct and the copilot suggestion to "properly handle the error" would actually make the code worse in this context

anchor/common/qbft/src/lib.rs
```diff
@@ -606,21 +606,15 @@ where
                     max_prepared_msg = Some(round_change.clone());
                 }
 
-                // Check that prepared round is not greater than current round
```
Zacholme7: This is the same PR as #599

diegomrsantos: Yeah claude made a mess here. Fixing it, sorry.

anchor/common/qbft/src/lib.rs
```diff
@@ -642,6 +636,19 @@ where
...
+        // round
+        if let Some(max_prepared_msg) = &max_prepared_msg
+            && msg.qbft_message.root != max_prepared_msg.root
+        {
+            warn!(
+                "Proposal root doesn't match highest prepared round change root. Proposal root: {:?}, Highest prepared root: {:?}, prepared round: {}",
+                msg.qbft_message.root, max_prepared_msg.root, max_prepared_msg.data_round
+            );
+            return false;
+        }
```
Zacholme7: This is checked right below this

diegomrsantos: True, removed

Zacholme7:
So is this basically saying we can have round change justifications that dont agree with the proposal root but all that matters is that the highest one agrees with the root?

diegomrsantos:
> So is this basically saying we can have round change justifications that dont agree with the proposal root but all that matters is that the highest one agrees with the root?

Yes. Look for `predicate isProposalJustification` on https://entethalliance.org/specs/qbft/v1/#nodeauxdfysec. In the second `else` we have 

```
&& exists rcm | rcm in roundChanges ::
  && isHighestPrepared(rcm,roundChanges)
  && var proposedBlockWithOldRound :=
    replaceRoundInBlock(
      block,
      optionGet(rcm.unsignedPayload.preparedRound)
    );
  && optionGet(rcm.unsignedPayload.preparedValue) == digest(proposedBlockWithOldRound)

Zacholme7 (approved):
Makes sense now, thanks!

dknopik (approved):
Great find, thanks Diego!

PR #604, opened 2025-09-12
fix(validation): Check the signatures of the justifications' justifications

dknopik:
## Issue Addressed

Messages used in round change justifications can have justifications themselves. Check those.

## Proposed Changes

Decode the justifications and call `validate_justifications` on them.


anchor/message_validator/src/consensus_message.rs
```diff
@@ -172,7 +172,13 @@ pub(crate) fn validate_justifications(
         .iter()
         .chain(round_change_justifications.iter())
         .try_for_each(|signed_message| {
-            verify_message_signatures(signed_message, operator_pub_keys)
+            verify_message_signatures(signed_message, operator_pub_keys)?;
+            // Also check the justifications' justifications
+            validate_justifications(
```
Zacholme7: Love recursion in the wild

Zacholme7 (dismissed):
nice!

diegomrsantos:
is this supported by the spec?

anchor/message_validator/src/consensus_message.rs
```diff
@@ -172,7 +172,13 @@ pub(crate) fn validate_justifications(
...
+                &QbftMessage::from_ssz_bytes(signed_message.ssv_message().data())
+                    .map_err(|_| ValidationFailure::MalformedJustifications)?,
+                operator_pub_keys,
+            )
```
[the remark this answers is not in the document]
Zacholme7: Im not sure this is possible. We cap the data lengths and I think circular references are hard to create in rust without some fancy `RefCell` stuff?

diegomrsantos (dismissed):
Let's confirm first this is supported by the spec? Is go doing it?

diegomrsantos:
how about https://leastauthority.com/wp-content/uploads/2024/01/Coin_Dash_Ltd._SSV_Specification_Final_Audit_Report_Updated_Least-Authority-.pdf

Zacholme7:
A message with an invalid signature is not something we want to accept

anchor/message_validator/src/consensus_message.rs
diegomrsantos: is `validate_justifications` validating only signatures?

diegomrsantos: Prepare messages never carry justifications. So recursing “inside a Prepare” is a dead end by spec design

Zacholme7: it will only recurse if it is non-empty, and since prepare messages never carry justifications it will not recurse

diegomrsantos: Imo this isn't a good idea

diegomrsantos: For example, if someone adds, for some reason, invalid justifications inside a prepare message, our code will reject it, but it's "valid". If we want to be very strict, imo it'd be better to assert that a prepare message doesn't have any justification instead.

diegomrsantos: How about something like

```rust
pub(crate) fn validate_justifications(
    consensus_message: &QbftMessage,
    operator_pub_keys: &HashMap<OperatorId, Rsa<Public>>,
) -> Result<(), ValidationFailure> {
    // ----- Presence rules (where each justification set is allowed) -----
    // Prepare justifications MUST appear only on Proposals.
    if !consensus_message.prepare_justification.is_empty()
        && consensus_message.qbft_message_type != QbftMessageType::Proposal
    {
        return Err(ValidationFailure::UnexpectedPrepareJustifications);
    }

    // "round_change_justification" is:
    // - on a Proposal: the RC set (each item must be a RoundChange)
    // - on a RoundChange: the prepares for the last prepared (each item must be a Prepare)
    // - otherwise: must be empty
    if !consensus_message.round_change_justification.is_empty()
        && consensus_message.qbft_message_type != QbftMessageType::Proposal
        && consensus_message.qbft_message_type != QbftMessageType::RoundChange
    {
        return Err(ValidationFailure::UnexpectedRoundChangeJustifications);
    }

    // ----- Type + signature checks for each set -----
    match consensus_message.qbft_message_type {
        QbftMessageType::Proposal => {
            // On Proposal:
            //   - round_change_justification := set of RoundChange messages
            //   - prepare_justification      := set of Prepare messages (the single prepared cert)
            for signed in &consensus_message.round_change_justification {
                let inner = decode_inner_qbft(signed)?;
                if inner.qbft_message_type != QbftMessageType::RoundChange {
                    return Err(ValidationFailure::JustificationWrongType {
                        expected: QbftMessageType::RoundChange,
                        got: inner.qbft_message_type,
                    });
                }
                verify_message_signatures(signed, operator_pub_keys)?;
                // IMPORTANT: do NOT recurse into inner.justifications here.
                // The spec does not require per-RC nested-prepare validation on proposal validation.
            }

            for signed in &consensus_message.prepare_justification {
                let inner = decode_inner_qbft(signed)?;
                if inner.qbft_message_type != QbftMessageType::Prepare {
                    return Err(ValidationFailure::JustificationWrongType {
                        expected: QbftMessageType::Prepare,
                        got: inner.qbft_message_type,
                    });
                }
                verify_message_signatures(signed, operator_pub_keys)?;
                // No recursion; prepares never carry justifications per spec model.
            }
        }

        QbftMessageType::RoundChange => {
            // On RoundChange:
            //   - prepare_justification MUST be empty (already enforced above)
            //   - round_change_justification := set of Prepare messages proving "prepared" (if prepared)
            for signed in &consensus_message.round_change_justification {
                let inner = decode_inner_qbft(signed)?;
                if inner.qbft_message_type != QbftMessageType::Prepare {
                    return Err(ValidationFailure::JustificationWrongType {
                        expected: QbftMessageType::Prepare,
                        got: inner.qbft_message_type,
                    });
                }
                verify_message_signatures(signed, operator_pub_keys)?;
            }
        }

        // For Prepare / Commit / others, both sets must be empty (already enforced above).
        _ => { /* nothing else to validate */ }
    }

    Ok(())
}

/// Decode the embedded QbftMessage out of a SignedMessage’s SSZ bytes.
fn decode_inner_qbft(signed: &SignedMessage) -> Result<QbftMessage, ValidationFailure> {
    QbftMessage::from_ssz_bytes(signed.ssv_message().data())
        .map_err(|_| ValidationFailure::MalformedJustifications)
}
```

Zacholme7: If we are confident on not checking the inner justifications I think that looks nice

diegomrsantos: For a proposal, we don't need to check the prepares inside the round change

diegomrsantos: But for a standalone round change, we actually need to

jking-aus (dismissed):

jking-aus:
I'm good for this to go in. Tag it to merge when you're ready @dknopik

diegomrsantos (commented):
Removing the request

jking-aus (approved):

PR #608, opened 2025-09-12
fix!: Removed operators in committee

dknopik:
## Issue Addressed

Operators can be removed by their owner at any time. Existing committees are supposed to continue operation, treating the removed operator as permanently offline. Messages by removed operators should not be accepted anymore.

## Proposed Changes

- Do not remove the operator from `cluster_members` on removal. This is important in order to keep the committee size correct.
- In the validator, since a recent change, we get public keys for the whole committee. However, while fetching the keys, we must silently accept missing keys, and only error if they are needed (i.e. a signature by that validator must be verified).

## Additional Info

Actually, that first point also caused an inconsistency of on-disk and in-memory representation: in memory, the operator is removed only from the operator state. While on disk, the cluster membership is removed.


diegomrsantos:
> treating the removed operator as permanently offline.

Why is that?

diegomrsantos:
> Do not remove the operator from cluster_members on removal. This is important in order to keep the committee size correct.

Keeping a removed operator, in fact, makes the committee size incorrect.

dknopik:
> Keeping a removed operator, in fact, makes the committee size incorrect.

Citing the description: Existing committees are supposed to continue operation, treating the removed operator as permanently offline.

diegomrsantos:
> > Keeping a removed operator, in fact, makes the committee size incorrect.
> 
> Citing the description: Existing committees are supposed to continue operation, treating the removed operator as permanently offline.

Why would removing an operator stop a committee?

dknopik:
> Why would removing an operator stop a committee?

Because of the bug I am trying to fix in this PR

diegomrsantos:
> > > Keeping a removed operator, in fact, makes the committee size incorrect.
> > 
> > 
> > Citing the description: Existing committees are supposed to continue operation, treating the removed operator as permanently offline.
> 
> Why would removing an operator stop a committee?

So that's the reason?

> Because the validator’s key shares and thresholds were generated for that specific set of operators.

diegomrsantos:
And this is true?

> with an offline member still in the committee, rounds where they’re proposer will time out and trigger round-change

anchor/database/src/update_0_to_1.sql
```diff
@@ -0,0 +1,19 @@
+
+PRAGMA foreign_keys=OFF;
+
+BEGIN;
+UPDATE metadata SET schema_version = 1;
+
+-- we can not remove a foreign key constraint. so we have to recreate the table instead.
```
diegomrsantos: Can't we ask ppl to delete and recreate the DB instead of this?

dknopik: Usually, I'd say that this is exactly what the database migration mechanism is for, but I just realized that we need to tell them to recreate as potentially the database has missing data due to the previous incorrect logic. 

will change

dknopik: thanks!

anchor/database/src/tests/cluster_tests.rs
```diff
@@ -64,25 +64,6 @@ mod cluster_database_tests {
...
-            &fixture.validator.public_key,
-        )];
-        let mut conn = fixture.db.connection().unwrap();
-        let tx = conn.transaction().unwrap();
-        fixture
-            .db
-            .insert_validator(cluster, &metadata, shares, &tx)
-            .expect_err("Insertion should fail");
-    }
-
```
diegomrsantos, line 85: This test should still fail

dknopik: the existence of all operators for a new validator is enforced by `process_validator_added`. This tests `insert_validator`, which no longer implicitly checks existence of operator as the foreign key constraint has been removed.

in your opinion, should we check again for operator existence in `insert_validator`?

diegomrsantos: If currently it's not possible to create a cluster with non-existent validators in the DB and this test was checking only the foreign key, so I agree with the deletion

diegomrsantos: But I think it makes more sense to check operator existence during the cluster creation than validator insertion, doesn't it?

anchor/database/src/operator_operations.rs
```diff
@@ -72,9 +72,8 @@ impl NetworkDatabase {
             )));
         }
 
-        // Remove from db and in memory. This should cascade to delete this operator from all of the
-        // clusters that it is in and all of the shares that it owns
-        tx.prepare_cached(sql_operations::DELETE_OPERATOR)?
+        // Remove from db and in memory.
```
diegomrsantos: nit: we need to update the comment

anchor/database/src/tests/metadata_tests.rs
```diff
@@ -31,8 +31,8 @@ mod tests {
         let metadata = queries::get_metadata(&conn).expect("Failed to get metadata");
 
         assert_eq!(
-            metadata.schema_version, 0,
-            "Initial schema version should be 0"
+            metadata.schema_version, 1,
+            "Initial schema version should be 1"
```
diegomrsantos, line 35: Do we need to change it? Apart from this, I think everything else is backwards compatible?

diegomrsantos: But I guess it's not a big deal to require the DB to be deleted for now

dknopik: If the schema version is not bumped, current users will not be prompted to remove the database and will get confusing errors as the updated sql operations will fail.

diegomrsantos:
should we update `const GET_OPERATOR: &str ` in `database/tests/utils.rs`?

anchor/database/src/schema.rs
```diff
@@ -129,19 +129,22 @@ fn determine_database_type(conn: &Connection, domain: DomainType) -> DatabaseTyp
...
+    let schema_version = conn.query_row(sql_operations::GET_METADATA, [], |row| {
+        row.get("schema_version")
+    })?;
+    Ok(schema_version)
 }
 
 // Register upgrade scripts in this function and mark the current version. Define any versions for
 // which the schema is not upgradable as "Recreate" and all versions after the current version as
 // "Future".
 fn get_upgrade_action(version: Option<SchemaVersion>) -> UpgradeAction {
```
diegomrsantos, line 144: Could you please explain this function? I think the comments are outdated.

dknopik: This function defines what to do when a specific schema version is encountered. 

If there is no schema version (because the database was created before we introduced this mechanism) or the schema version is 0, we advise the user to delete the database and exit.

If the schema version is 1, we are up to date and can proceed with running the application.

If the schema version is 2 or higher, this database is from a future version of anchor and can not work with it.

In future, we can register upgrade scripts to update the database and be able to proceed without user intervention.

diegomrsantos: Ah, so that's why I was confused. Currently, we don't handle the apply migrations case?

dknopik: Currently, there is no migration to apply :)

see this commit to see how it might look like when we add a migration:

https://github.com/sigp/anchor/pull/608/commits/2b83119b7e63e219de3c8ac88d0a451b55b068b0

dknopik: having this be a function that we can constantly and dynamically adjust has the advantage of easily being able to later remove migrations if we are sure no old databases remain, which keeps the codebase lean

anchor/database/src/table_schema.sql
```diff
@@ -70,3 +71,22 @@ BEGIN
...
+    DELETE FROM operators WHERE operator_id = OLD.operator_id AND removed = TRUE;
+END;
+CREATE TRIGGER delete_empty_removed_operators_after_update
+    AFTER UPDATE ON operators
+    WHEN NOT EXISTS (
+        SELECT 1 FROM cluster_members
+        WHERE operator_id = OLD.operator_id
+    )
+BEGIN
+    DELETE FROM operators WHERE operator_id = OLD.operator_id AND removed = TRUE;
```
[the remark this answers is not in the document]
dknopik: we do not update the `operator_id`, so this does not really matter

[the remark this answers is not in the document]
diegomrsantos: Do we need `delete_empty_removed_operators_after_update`?

anchor/database/src/table_schema.sql
```diff
@@ -1,8 +1,8 @@
--- SCHEMA VERSION 0
+-- SCHEMA VERSION 1
 
 -- we should avoid removing columns from this to keep compatibility between anchor Versions
 CREATE TABLE metadata (
-    schema_version INTEGER NOT NULL DEFAULT 0, 
+    schema_version INTEGER NOT NULL DEFAULT 1,
```
diegomrsantos, line 5: Could you please elaborate on that?

anchor/database/src/table_schema.sql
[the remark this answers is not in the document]
dknopik: Yes. This ensures that operators are deleted immediately if they are not members anymore.

If it is omitted, the operator is simply marked as removed, but the `delete_empty_removed_operators_after_delete` can't be triggered anymore, so it is never physically deleted.

There may be a more elegant way to do this :sweat_smile:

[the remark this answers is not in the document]
diegomrsantos: Don't we already have a trigger to delete the operators when the cluster is deleted?

anchor/database/src/table_schema.sql
dknopik: sure!

Current databases have schema_version `0`. On startup with the new anchor version, we check which schema_version is currently deployed. If it is `0`, we ask the user to remove the database for a resync. Then, the database will be recreated and `schema_version` `1` will be inserted, as the database now has the updated schema. Then, the `schema_version` check will pass on subsequent startups.

anchor/database/src/table_schema.sql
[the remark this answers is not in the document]
diegomrsantos: When the last cluster an operator belongs to is deleted, the operator should be deleted as well.

anchor/database/src/table_schema.sql
diegomrsantos: Thanks!

anchor/database/src/table_schema.sql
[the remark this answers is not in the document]
dknopik: yes. but this only works for this order:

```
1. Operator is removed
2. Cluster is removed (<- trigger here)
```

it does not work for this order:

```
1. Cluster removed (<- trigger here, but operator is not deleted as it is not removed yet)
2. Operator is removed
```

With two triggers, the order does not matter:


```
1. Operator is removed (<- trigger here, but operator is not deleted as cluster still exists)
2. Cluster is removed (<- trigger here, operator deleted)
```

```
1. Cluster removed (<- trigger here, but operator is not deleted as it is not removed yet)
2. Operator is removed (<- trigger here, operator deleted)
```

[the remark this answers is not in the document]
diegomrsantos:
```
1. Cluster removed (<- trigger here, but operator is not deleted as it is not removed yet)
2. Operator is removed
```

If the operator doesn't belong to any cluster, it can be immediately removed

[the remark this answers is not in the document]
diegomrsantos: At 2, when the operator's owner requests the removal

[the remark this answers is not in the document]
dknopik: right, and this is what `delete_empty_removed_operators_after_update` does.

I am open to another solution for that though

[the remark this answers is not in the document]
diegomrsantos: We don't need a trigger. We can just delete the operator instead of marking it as removed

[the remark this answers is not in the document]
dknopik: will change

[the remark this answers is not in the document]
dknopik: done, wdyt?

anchor/database/src/table_schema.sql
```diff
@@ -70,3 +71,13 @@ BEGIN
...
+-- Add triggers to clean up removed operators
+CREATE TRIGGER delete_empty_removed_operators_after_delete
+    AFTER DELETE ON cluster_members
+    WHEN NOT EXISTS (
+        SELECT 1 FROM cluster_members
+        WHERE operator_id = OLD.operator_id
+    )
+BEGIN
+    DELETE FROM operators WHERE operator_id = OLD.operator_id AND removed = TRUE;
+END;
```
diegomrsantos: nit: new line

anchor/database/src/operator_operations.rs
```diff
@@ -72,16 +72,15 @@ impl NetworkDatabase {
...
-        if count == 0 {
-            tx.prepare_cached(sql_operations::DELETE_OPERATOR)?
-                .execute(params![*id])?;
-        } else {
+        // Deleting failed, likely because of a foreign key restraint. The operator is still member
+        // of a committee.
+        if result.is_err() {
             // Mark the operator as removed. This will allow cluster membership to remain recorded.
-            // The operator will be removed by triggers if no cluster membership remains.
+            // The operator will be removed by a trigger if no cluster membership remains.
```
diegomrsantos: nit: 

```rust
if let Err(e) = tx.prepare_cached(sql_operations::DELETE_OPERATOR)?
    .execute(params![*id])
{
    // Deleting failed, likely because of a foreign key restraint. The operator is still member
    // of a committee.
    // Mark the operator as removed. This will allow cluster membership to remain recorded.
    // The operator will be removed by a trigger if no cluster membership remains.
    tx.prepare_cached(sql_operations::MARK_OPERATOR_REMOVED)?
        .execute(params![*id])?;
}
```

diegomrsantos (dismissed):
Thanks a lot

diegomrsantos (approved):
LGTM!

Issue #560, opened 2025-09-01
Perform safe arithmetic in `calculate_round_timeout`


dknopik:
While there are no known ways for an attacker to bump the qbft round very high to cause an overflow in `calculate_round_timeout`, we should be vigilant and use checked arithmetic in `calculate_round_timeout`, returning `None` if the round is too high for the timeout to be properly calculated.

Note: we might want to reevaluate other code locations as well w.r.t. checked arithmetic.

PR #609, opened 2025-09-16
fix: safe arithmetic in timeout calculation

petarjuki7:
## Issue Addressed

Addresses issue #560 

## Proposed Changes

Makes sure the round can't be bumped to high for the addition to overflow in `calculate_round_timeout`, returning `None` if the round is too high for the timeout to be properly calculated.

### DIFF
--- a/anchor/qbft_manager/src/instance.rs
+++ b/anchor/qbft_manager/src/instance.rs
@@ -176,10 +176,10 @@ impl<D: QbftData<Hash = Hash256>> Initialized<D> {
         // round to advance
         let round_end = calculate_round_timeout(self.qbft.get_round().into(), &self.start_time);
 
-        let timeout_instant = round_end.unwrap_or_else(|| {
-            error!("Round timeout calculation overflowed, defaulting to maximum");
-            tokio::time::Instant::now() + tokio::time::Duration::MAX
-        });
+        let Some(timeout_instant) = round_end else {
+            error!("Round timeout calculation overflowed, stopping the instance");
+            return RecvResult::Closed;
+        };
 
         let round_timeout_sleep = tokio::time::sleep_until(timeout_instant);
         tokio::pin!(round_timeout_sleep);
--- a/anchor/qbft_manager/src/instance.rs
+++ b/anchor/qbft_manager/src/instance.rs
@@ -175,7 +175,13 @@ impl<D: QbftData<Hash = Hash256>> Initialized<D> {
         // We calculate the sleep dynamically, as both messages and the local timer might cause the
         // round to advance
         let round_end = calculate_round_timeout(self.qbft.get_round().into(), &self.start_time);
-        let round_timeout_sleep = tokio::time::sleep_until(round_end);
+
+        let timeout_instant = round_end.unwrap_or_else(|| {
+            error!("Round timeout calculation overflowed, defaulting to maximum");
+            tokio::time::Instant::now() + tokio::time::Duration::MAX
+        });
+
+        let round_timeout_sleep = tokio::time::sleep_until(timeout_instant);
         tokio::pin!(round_timeout_sleep);
 
         select! {
--- a/anchor/qbft_manager/src/instance.rs
+++ b/anchor/qbft_manager/src/instance.rs
@@ -177,7 +177,13 @@ impl<D: QbftData<Hash = Hash256>> Initialized<D> {
         let round_end = calculate_round_timeout(self.qbft.get_round().into(), &self.start_time);
 
         let Some(timeout_instant) = round_end else {
-            error!("Round timeout calculation overflowed, stopping the instance");
+            error!(
+                "Round timeout calculation overflowed for round {}, stopping the instance. \
+            QBFT identifier: {:?}, instance height: {:?}",
+                self.qbft.get_round(),
+                self.qbft.get_identifier(),
+                self.qbft.get_instance_height()
+            );
             return RecvResult::Closed;
         };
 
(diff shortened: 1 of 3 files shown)


anchor/qbft_manager/src/instance.rs
```diff
@@ -175,7 +175,13 @@ impl<D: QbftData<Hash = Hash256>> Initialized<D> {
         // We calculate the sleep dynamically, as both messages and the local timer might cause the
         // round to advance
         let round_end = calculate_round_timeout(self.qbft.get_round().into(), &self.start_time);
-        let round_timeout_sleep = tokio::time::sleep_until(round_end);
+
+        let timeout_instant = round_end.unwrap_or_else(|| {
+            error!("Round timeout calculation overflowed, defaulting to maximum");
+            tokio::time::Instant::now() + tokio::time::Duration::MAX
```
petarjuki7: This is probably too large as its about 600b years... Should I put a max at an hour?

Zacholme7: imo if we somehow hit this, if its even possible, we should just stop the instance entirely and just return `RecvResult::Closed` so it does not process any more messages. 

Not sure though, curious other thoughts

dknopik: :+1: yeah, that is a good idea! @petarjuki7 sorry didn't come to my mind when we chatted earlier

anchor/qbft_manager/src/timeout.rs
```diff
@@ -22,5 +22,5 @@ pub fn calculate_round_timeout(round: u64, start_time: &Instant) -> Instant {
         quick_portion + slow_portion
     };
 
-    *start_time + additional_timeout
+    start_time.checked_add(additional_timeout)
```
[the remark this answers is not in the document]
diegomrsantos: Don't know what's going on with Copilot. `additional_timeout` is a `Duration`. 

Other operations can also overflow, can't they? Like `quick_portion + slow_portion`

anchor/qbft_manager/src/instance.rs
```diff
@@ -175,7 +175,13 @@ impl<D: QbftData<Hash = Hash256>> Initialized<D> {
         // We calculate the sleep dynamically, as both messages and the local timer might cause the
         // round to advance
         let round_end = calculate_round_timeout(self.qbft.get_round().into(), &self.start_time);
-        let round_timeout_sleep = tokio::time::sleep_until(round_end);
+
+        let Some(timeout_instant) = round_end else {
+            error!("Round timeout calculation overflowed, stopping the instance");
```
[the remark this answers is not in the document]
diegomrsantos: We could also add more info like `identifier`, `instance_height`, `current_round`and `state`.

anchor/qbft_manager/src/timeout.rs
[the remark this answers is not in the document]
dknopik: Correct, all arithmetic operations in here should be checked.

anchor/qbft_manager/src/timeout.rs
```diff
@@ -19,8 +19,12 @@ pub fn calculate_round_timeout(round: u64, start_time: &Instant) -> Instant {
 
         // The slow poritin is (round - threshold) * SLOW_TIMEOUT
         let slow_portion = Duration::from_secs((round - QUICK_TIMEOUT_THRESHOLD) * SLOW_TIMEOUT);
-        quick_portion + slow_portion
+
+        let Some(additional_timeout) = quick_portion.checked_add(slow_portion) else {
+            return None;
+        };
+        additional_timeout
```
Zacholme7, line 26: cant we just return `quick_portion.checked_add(slow_portion)` here

Zacholme7, line 26: cant we just return `quick_portion.checked_add(slow_portion)` here

Zacholme7: I think we can just do `quick_portion.checked_add(slow_portion)?`

will unwrap and return the `None` or keep going with the unwrapped `Some`

petarjuki7: Yeah, just pushed it

anchor/qbft_manager/src/timeout.rs
```diff
@@ -19,8 +19,9 @@ pub fn calculate_round_timeout(round: u64, start_time: &Instant) -> Instant {
 
         // The slow poritin is (round - threshold) * SLOW_TIMEOUT
         let slow_portion = Duration::from_secs((round - QUICK_TIMEOUT_THRESHOLD) * SLOW_TIMEOUT);
```
diegomrsantos: how about this?

petarjuki7: Added few more checks

anchor/qbft_manager/src/timeout.rs
```diff
@@ -6,21 +6,24 @@ const QUICK_TIMEOUT_THRESHOLD: u64 = 8; // Round 8
...
     let additional_timeout = if round <= QUICK_TIMEOUT_THRESHOLD {
         // If we are below the quick timeout threshold the additonal timeout is round *
         // QUICK_TIMEOUT
-        Duration::from_secs(round * QUICK_TIMEOUT)
+        Duration::from_secs(round.checked_mul(QUICK_TIMEOUT)?)
     } else {
         // For higher rounds, use a combination of quick and slow timeouts
 
         // The quick portion is the timeout threshold * QUICK_TIMEOUT
         let quick_portion = Duration::from_secs(QUICK_TIMEOUT_THRESHOLD * QUICK_TIMEOUT);
```
dknopik, line 18: one more here ;)

petarjuki7: Aren't those two constants, so maybe they don't benefit from a checked_mul? Basically a 8 * 2?

dknopik: ah true! good point, sorry

dknopik (approved):
thanks! lets ship it

PR #616, opened 2025-09-18
feat: Allow RUST_LOG to override specific logging targets while preserving defaults

diegomrsantos:
## Issue Addressed

Improve RUST_LOG environment variable support for debugging discv5 and libp2p_gossipsub components.

## Proposed Changes

Replace hardcoded `EnvFilter::try_new()` with combined filter approach that supports selective RUST_LOG overrides while preserving defaults:

- Build filter from defaults + RUST_LOG using tracing-subscriber's directive precedence
- Use `EnvFilter::builder().parse_lossy()` for graceful error handling
- RUST_LOG overrides only specified components, others keep defaults

**Examples:**
- `RUST_LOG=discv5=trace` → only discv5 becomes trace; libp2p stays at default debug
- `RUST_LOG=libp2p_gossipsub::gossip_promises=info` → only that component changes; others remain default
- `RUST_LOG=debug` → sets base level for unmatched targets while preserving explicit settings

## Additional Info

Tested with network runs confirming selective overrides work correctly and unspecified components maintain their hardcoded defaults.

### DIFF
--- a/anchor/src/main.rs
+++ b/anchor/src/main.rs
@@ -208,14 +208,20 @@ pub fn enable_logging(
             // Create filter that reduces external library noise to WARN level while preserving
             // the configured file log level for Anchor crates
 
-            logging_layers.push(
-                libp2p_discv5_layer
-                    .with_filter(
-                        EnvFilter::try_new("warn,libp2p_gossipsub::peer_score=debug,libp2p_gossipsub::gossip_promises=debug")
-                            .unwrap_or_else(|_| EnvFilter::new("debug")),
-                    )
-                    .boxed(),
-            );
+            let base = "discv5=warn,libp2p_gossipsub::peer_score=debug,libp2p_gossipsub::gossip_promises=debug";
+            let env = std::env::var(EnvFilter::DEFAULT_ENV).unwrap_or_default();
+
+            // Defaults first, env last (env overrides only what it mentions)
+            let combined = if env.is_empty() {
+                base.to_string()
+            } else {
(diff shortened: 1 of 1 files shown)


anchor/src/main.rs
```diff
@@ -208,14 +208,20 @@ pub fn enable_logging(
...
+            let combined = if env.is_empty() {
+                base.to_string()
+            } else {
+                format!("{base},{env}")
+            };
+
+            // Use parse_lossy so a bad RUST_LOG doesn't crash; it falls back to the base part.
+            let layer_filter = EnvFilter::builder().parse_lossy(combined);
+
+            logging_layers.push(libp2p_discv5_layer.with_filter(layer_filter).boxed());
```
dknopik, line 224: why not use https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.Builder.html#method.with_default_directive to make the default completely overridable?

diegomrsantos: What do you mean by completely overridable?

dknopik: currently, we can only append to the hardcoded `base`. What is the advantage over being able to replace it completely using `RUST_LOG`? This is also what is done above for the workspace filter.

diegomrsantos: I still don't understand what you mean. Currently, it's possible to override what's defined in the code. Have you tried and weren't able to?

diegomrsantos:
replaced by https://github.com/sigp/anchor/pull/619

PR #617, opened 2025-09-19
feat: Add CLI flag to disable Enr auto update

AgeManning:
## Description

Some complicated networks (and generally bootnodes) can be hindered by discovery's automatic external socket discovery mechanism. It's sometimes preferable to let the user specify the ENR fields and prevent discovery from modifying them. 

Currently Anchor allows users to set the ENR field, but has no option to prevent discovery from changing them as it runs if it decides there are better values for the ENR. 

This PR adds the CLI option to allow users to disable the discovery mechanism, in effect allowing them to keep the values of the ENR they set in the CLI.

### DIFF
--- a/anchor/client/src/cli.rs
+++ b/anchor/client/src/cli.rs
@@ -427,6 +427,16 @@ pub struct Node {
 
     #[clap(
         long,
+        global = true,
+        help = "Discovery can automatically discover external addresses if the node has correctly set up port forwards.\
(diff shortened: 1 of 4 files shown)


dknopik (dismissed):
LGTM, thanks!

dknopik (approved):
rebased on `release-v0.3.0`

dknopik:
@mergify requeue

PR #619, opened 2025-09-19
feat: Add logging flags for dependencies

dknopik:
## Proposed Changes

To allow debugging issues relating to `discv5` and `libp2p_gossipsub`, allow configuring the log level for these dependencies on the command line.

Also, add the possibility to let `RUST_LOG` replace this behaviour.

## Additional Info

I kept the enforced logging of penalty information, but i am not sure about that.

### DIFF
--- a/anchor/logging/src/logging.rs
+++ b/anchor/logging/src/logging.rs
@@ -63,7 +63,7 @@ pub struct FileLoggingFlags {
     #[arg(
         long,
         global = true,
-        default_value_t = Level::WARN,
+        default_value_t = Level::DEBUG,
         value_parser = Level::from_str,
         help = "Specifies the verbosity level used for the discv5 dependency log file")]
     pub discv5_debug_level: Level,
@@ -71,7 +71,7 @@ pub struct FileLoggingFlags {
     #[arg(
         long,
         global = true,
-        default_value_t = Level::WARN,
+        default_value_t = Level::DEBUG,
         value_parser = Level::from_str,
         help = "Specifies the verbosity level used for the libp2p dependency log file. \
                 Certain score penalty information is logged regardless of this setting.")]
--- a/anchor/logging/src/logging.rs
+++ b/anchor/logging/src/logging.rs
@@ -59,6 +59,23 @@ pub struct FileLoggingFlags {
 
     #[arg(long, global = true, help = "Enables colors in logfile.")]
     pub logfile_color: bool,
+
+    #[arg(
+        long,
+        global = true,
+        default_value_t = Level::WARN,
+        value_parser = Level::from_str,
+        help = "Specifies the verbosity level used for the discv5 dependency log file")]
+    pub discv5_debug_level: Level,
+
+    #[arg(
+        long,
+        global = true,
+        default_value_t = Level::WARN,
+        value_parser = Level::from_str,
+        help = "Specifies the verbosity level used for the libp2p dependency log file. \
+                Certain score penalty information is logged regardless of this setting.")]
+    pub libp2p_debug_level: Level,
 }
 
 impl FileLoggingFlags {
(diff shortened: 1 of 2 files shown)


anchor/src/main.rs
```diff
@@ -205,14 +205,19 @@ pub fn enable_logging(
         let file_logging_layer = init_file_logging(&logs_dir, file_logging_flags.clone());
 
         if let Some(libp2p_discv5_layer) = libp2p_discv5_layer {
-            // Create filter that reduces external library noise to WARN level while preserving
-            // the configured file log level for Anchor crates
+            // Create filter that reduces external library noise to separately configured levels
+            // while preserving the configured file log level for Anchor crates
+
+            let default = format!(
+                "discv5={},libp2p_gossipsub={},libp2p_gossipsub::peer_score=debug,libp2p_gossipsub::gossip_promises=debug",
```
dknopik: I am not 100% sure if this is the desired way of keeping the `debug` logging of penalties. Currently, it is only overridable by `RUST_LOG`, but this might be fine?

anchor/logging/src/logging.rs
```diff
@@ -59,6 +59,23 @@ pub struct FileLoggingFlags {
...
+        global = true,
+        default_value_t = Level::WARN,
+        value_parser = Level::from_str,
+        help = "Specifies the verbosity level used for the discv5 dependency log file")]
+    pub discv5_debug_level: Level,
+
+    #[arg(
+        long,
+        global = true,
+        default_value_t = Level::WARN,
```
diegomrsantos: Should be debug, no?

anchor/src/main.rs
diegomrsantos: Cant we use the variables directly in the string?

diegomrsantos: Maybe we can have only `libp2p_gossipsub` and set it as `debug` by default

anchor/logging/src/logging.rs
```diff
@@ -59,6 +59,23 @@ pub struct FileLoggingFlags {
...
+        value_parser = Level::from_str,
+        help = "Specifies the verbosity level used for the libp2p dependency log file. \
```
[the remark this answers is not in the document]
dknopik: This is false.

anchor/logging/src/logging.rs
dknopik: done

anchor/src/main.rs
dknopik: done

dknopik: this only works for plain variables. access to fields as done here does not work within the brackets.

anchor/src/main.rs
```diff
@@ -205,14 +205,19 @@ pub fn enable_logging(
...
 
         if let Some(libp2p_discv5_layer) = libp2p_discv5_layer {
-            // Create filter that reduces external library noise to WARN level while preserving
-            // the configured file log level for Anchor crates
+            // Create filter that reduces external library noise to separately configured levels
+            // while preserving the configured file log level for Anchor crates
+
+            let default = format!(
+                "discv5={},libp2p_gossipsub={}",
+                file_logging_flags.discv5_debug_level, file_logging_flags.libp2p_debug_level
```
diegomrsantos: Should they be named `*_log_level` instead?

dknopik: done

diegomrsantos (approved):
Looks great, thanks a lot!

anchor/logging/src/logging.rs
```diff
@@ -59,6 +59,23 @@ pub struct FileLoggingFlags {
 
     #[arg(long, global = true, help = "Enables colors in logfile.")]
     pub logfile_color: bool,
+
+    #[arg(
+        long,
+        global = true,
+        default_value_t = Level::DEBUG,
```
diegomrsantos, line 66: Actually, do we want to make `debug` the default?

dknopik: personally, I am fine with it. recent issue showed that this may carry valuable info

dknopik: dont care strongly though

diegomrsantos: Me neither, but it can be noisy

diegomrsantos:
Feel free to merge then ;)

PR #623, opened 2025-09-19
chore: Enable mainnet

dknopik:
## Proposed Changes

- Allow passing `mainnet` to `network`.
- Remove `mainnet` safeguard from client startup function.
- Adjust comments and docs.
- Make `mainnet` the default network.

### DIFF
--- a/anchor/common/global_config/src/lib.rs
+++ b/anchor/common/global_config/src/lib.rs
@@ -9,7 +9,7 @@ use tracing::Level;
 use crate::data_dir::DataDir;
 
 /// Default network, used to partition the data storage
-pub const DEFAULT_HARDCODED_NETWORK: &str = "hoodi";
+pub const DEFAULT_HARDCODED_NETWORK: &str = "mainnet";
 
 /// Config that applies to all subcommands: The resolved network and datadir. This avoids repeated
 /// logic matching the datadir from the actual CLI definition.
@@ -49,9 +49,9 @@ pub struct GlobalFlags {
         long,
         global = true,
         value_name = "NETWORK",
-        value_parser = vec!["holesky", "hoodi"],
+        value_parser = vec!["mainnet", "holesky", "hoodi"],
         conflicts_with = "testnet_dir",
-        help = "Name of the chain Anchor will validate. Mainnet is not supported.",
+        help = "Name of the chain Anchor will validate.",
         display_order = 0,
         default_value = DEFAULT_HARDCODED_NETWORK,
     )]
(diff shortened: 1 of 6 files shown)


anchor/common/global_config/src/lib.rs
```diff
@@ -9,7 +9,7 @@ use tracing::Level;
 use crate::data_dir::DataDir;
 
 /// Default network, used to partition the data storage
-pub const DEFAULT_HARDCODED_NETWORK: &str = "hoodi";
+pub const DEFAULT_HARDCODED_NETWORK: &str = "mainnet";
```
diegomrsantos, line 12: Should the default be mainnet?

anchor/common/global_config/src/lib.rs
```diff
@@ -49,9 +49,9 @@ pub struct GlobalFlags {
         long,
         global = true,
         value_name = "NETWORK",
-        value_parser = vec!["holesky", "hoodi"],
+        value_parser = vec!["mainnet", "holesky", "hoodi"],
```
diegomrsantos, line 52: When will `holesky` support be terminated?

anchor/eth/README.md
```diff
@@ -1,5 +1,5 @@
 ## Execution Layer
-This crate implements the execution layer component of the SSV node, responsible for monitoring and processing SSV network events on Ethereum L1 testnet networks.
+This crate implements the execution layer component of the SSV node, responsible for monitoring and processing SSV network events on Ethereum L1 networks (Mainnet and Holesky).
```
diegomrsantos: `(Mainnet and Hoodi)`?

anchor/common/global_config/src/lib.rs
dknopik: Yes. For example, all Ethereum clients that I know of use `mainnet` as default

anchor/common/global_config/src/lib.rs
dknopik: https://blog.ethereum.org/2025/09/01/holesky-shutdown-announcement

IMO we should remove `holesky` on the first release after that

anchor/eth/README.md
dknopik: I removed the list again, because I feel like it does not really provide value here. This slipped in accidentally

anchor/common/global_config/src/lib.rs
diegomrsantos: ok, thanks

diegomrsantos (approved):

dknopik:
LFG! :partying_face:

PR #624, opened 2025-09-22
fix: Remove/Downgrade some logs triggerable by bad contract events in the `validatorExited` code path

dknopik:
## Issue Addressed

Anchor will log `Cluster not found for validator` during initial sync on Hoodi.

This happens due to an event referring to a non-existent validator.

## Proposed Changes

Remove any error log that will be logged instead by the top level `malformed event` log. In one instance, downgrade from `warn` to `trace` in the case of an undeposited validator being requested.

### DIFF
--- a/anchor/eth/src/event_processor.rs
+++ b/anchor/eth/src/event_processor.rs
@@ -682,10 +682,6 @@ impl EventProcessor {
         let validator_metadata = match state.metadata().get_by(validator_pubkey) {
             Some(metadata) => metadata,
             None => {
-                error!(
-                    validator_pubkey = %validator_pubkey,
-                    "Validator metadata not found"
-                );
                 return Err(ExecutionError::InvalidEvent(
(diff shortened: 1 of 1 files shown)


anchor/eth/src/event_processor.rs
```diff
@@ -696,7 +692,7 @@ impl EventProcessor {
         let validator_index = match validator_metadata.index {
             Some(index) => Some(index),
             None => {
-                warn!(
+                trace!(
```
diegomrsantos, line 695: Why, in this case, don't we return an error?

diegomrsantos (approved):

anchor/eth/src/event_processor.rs
dknopik: I guess because in the context of this function (`get_validator_index`) we treat not having a validator index not as an error as it is perfectly valid for a validator pubkey to not have an index yet.

Issue #519, opened 2025-08-20
Loudly warn if logs are not being served


dknopik:
Reth does currently not store old contract events by default. We should check for such behaviour and warn the user, as elsewise sync does not complete correctly and the user gets a "Synced, waiting for operator to appear on chain"

Zacholme7:
Is this because of the potentially missing logs or something else? 

If the former, you can run reth as an archive or just not prune all receipts so we could query block 1 and if receipts returns then we are good to go, otherwise warn

dknopik:
Yeah, thanks. I guess not being able to detect archive nodes is my main issue. 

The thing is that we support multiple RPCs, but we do not know where the request actually ends up - resulting in a partial sync with some time spans missing. This is why I considered manually getting the version info and warning for reth.

diegomrsantos:
> Reth does currently not support getLogs in a sufficient way

@dknopik could you please elaborate more on that?

dknopik:
@diegomrsantos Sorry, that phrasing was misguided - As Zac noted, you can configure Reth to properly store all receipts. Will update the description.

diegomrsantos:
So it seems the real issue is that we need to be able to find out if Anchor has synced correctly?

diegomrsantos:
Additionally, is this an exclusive Reth behaviour? Other clients can do the same, can't they? We need to be able to verify that we have synced correctly, regardless of what current or future execution clients are used. Does it make sense?

PR #625, opened 2025-09-22
fix: warn if OperatorAdded events not in sync

petarjuki7:
## Issue Addressed

Adresses issue #519 

## Proposed Changes

We keep track of all `OperatorAdded` events in a new variable stored in the database (`max_operator_id_seen`).
Then, when we get a new `OperatorAdded` event, we can check does it match the expected `OperatorId` we _should_ have seen next.

Also, during historical sync we check if we saw any `OperatorAdded` event, if not, there is likely a sync error.


anchor/database/src/table_schema.sql
```diff
@@ -4,7 +4,8 @@
 CREATE TABLE metadata (
     schema_version INTEGER NOT NULL DEFAULT 0, 
     domain_type INTEGER NOT NULL,
-    block_number INTEGER NOT NULL DEFAULT 0 CHECK (block_number >= 0)
+    block_number INTEGER NOT NULL DEFAULT 0 CHECK (block_number >= 0),
+    max_operator_id_seen INTEGER NOT NULL DEFAULT 0
```
dknopik: For simplicity, I am not sure if we should persist this. It might be sufficient to keep this as an `Option<OperatorId>` during sync, initialized on the first seen `operatorAdded`, and checked subsequently during this `anchor` instances lifetime

anchor/eth/src/event_processor.rs
```diff
@@ -189,6 +199,19 @@ impl EventProcessor {
             )));
         }
 
+        let max_seen = self.db.state().get_max_operator_id_seen();
+
+        if live && max_seen != operatorId - 1 {
+            warn!(
+                "Missing OperatorAdded events: database has only seen up to id {max_seen}, \
+                but got operator {operator_id}."
```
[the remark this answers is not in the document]
petarjuki7: There is both so it should be fine, `operator_id` is an `OperatorId` which implements `Debug` and `Display`

anchor/eth/src/event_processor.rs
```diff
@@ -144,6 +149,10 @@ impl EventProcessor {
             }
         }
 
+        if (false, false) == (live, operator_added) {
+            warn!("No OperatorAdded events found in historical sync, there is likely a sync error");
```
dknopik: As `process_logs` is called once per batch, this may trigger for networks where the first operator is not immediately added after deployment, or if there are not any additons for a while. I suggest placing this check after historic sync is done, and before live sync is started. Maybe at the end of `historical_sync`?

anchor/eth/src/event_processor.rs
```diff
@@ -189,6 +199,19 @@ impl EventProcessor {
             )));
         }
 
+        let max_seen = self.db.state().get_max_operator_id_seen();
+
+        if live && max_seen != operatorId - 1 {
```
dknopik: This should also trigger if not live.

anchor/eth/src/event_processor.rs
```diff
@@ -189,6 +199,19 @@ impl EventProcessor {
...
+            warn!(
```
dknopik: This should be an `error`, as this is always a faulty condition.

anchor/eth/src/event_processor.rs
```diff
@@ -144,6 +149,10 @@ impl EventProcessor {
             }
         }
 
+        if (false, false) == (live, operator_added) {
```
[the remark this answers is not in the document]
petarjuki7: Yeah, forgot to change... committed

dknopik (commented):
Nice! Here are my thoughts. 

If we decide on not persisting the max operator id seen, I think the implementation might be straight forward enough to warrant inclusion in `v1.0.0`. wdyt?

anchor/database/src/table_schema.sql
petarjuki7: I was thinking about persistence also... I think the problem may be with the historical sync because, if I understood the code correctly, after we do it for the first time we persist the block data, and on the next startup of the `anchor` instance we only sync from the `last_processed_block`, but our `operator_seen` count would start from 0 which could cause inconsistencies.

dknopik: That is correct. This is why I suggest using an `last_operator_add_seen: Option<OperatorId>`, initialized to `None`. To avoid the inconsistency you mentioned, we would not do the check on the first `operatorAdded` we see, but use it to initialize the `last_operator_add_seen`. Then we can do the check when receiving more `operatorAdded`.

The downside is that we do not catch missing `operatorsAdded` between the last run and the first `operatorAdded` we get in the current run. I am on the fence whether that is alright.

petarjuki7: Ah okay, I understand.
The downside might be fine, because it if it's truly not catching the events, we will find out soon enough?

diegomrsantos:
The current approach might not be enough. Does the operator id start at zero or one and then continue increasing by 1? If so, we need to guarantee that we have seen all of them up to the current value in the contract.

petarjuki7:
> we need to guarantee that we have seen all of them up to the current value in the contract.

I think that was the downside discussed about this approach when we don't persist the `max_operator_id_seen` variable. We kind of "trust" the first `OperatorAdded` event we see will initialize the variable with the highest `OperatorId`.

I think the purpose of this PR is more to warn users that something with syncing is wrong so they configure their EL client correctly.

But I could implement a check that goes through our saved `OperatorId`s in the database and checks if they are monotonically increasing by 1 up until the `max_seen`?

diegomrsantos:
What I described is a suggestion about how to check if Anchor has synced correctly starting with an empty DB. Additionally, we also need to persist the highest operator id we have seen to be able to do the check when we start syncing after being offline.

diegomrsantos:
I think we do have to persist the `highest_operator_id_seen`. It will be created initially with the first value used by the contract.

> But I could implement a check that goes through our saved OperatorIds in the database and checks if they are monotonically increasing by 1

This needs to be done during sync, starting with the current value in the DB.

anchor/database/src/table_schema.sql
diegomrsantos: > we would not do the check on the first operatorAdded we see, but use it to initialize the last_operator_add_see

That's the crux of the issue, isn't it? A pruned client could have deleted the state containing some initial operators

diegomrsantos:
And before we start syncing, we could read the last operator id from the ssv contract and check that see all ids up to this value. But not sure it's necessary

petarjuki7:
Yeah, I understand the idea, can do it.
Just tagging @dknopik to confirm, if I remember correctly we decided on the non-persistence path yesterday.

dknopik:
@diegomrsantos In yesterdays meeting we discussed that we can also not persist the `highest_operator_id_seen` and initialize it with the first operator ID seen per sync. This would be sufficient to catch any missing `operatorAdded` for the lifetime of that instance.

In the meeting, I asked if there is agreement to do it this way for now, as it is a tradeoff for not having to modify the database that closely to the release. While it is of course okay to change one's mind, I believe the simpler option is better for a short term change

dknopik:
@petarjuki7 please rebase onto `release-v1.0.0`

anchor/eth/src/event_processor.rs
```diff
@@ -189,6 +189,20 @@ impl EventProcessor {
...
+        let max_operator_id = self.db.state().get_max_operator_id_seen();
+
+        if let Some(max_operator_id) = max_operator_id {
+            if max_operator_id != operatorId - 1 {
+                error!(
+                    "Missing OperatorAdded events: database has only seen up to id {max_operator_id}, \
+                but got operator {operator_id}."
+                );
+            }
+            self.db.bump_max_operator_id_seen();
```
dknopik: I think it would be a bit cleaner to just have `set_max_operator_id_seen`

anchor/eth/src/sync.rs
```diff
@@ -510,6 +510,16 @@ impl SsvEventSyncer {
...
         info!("Historical sync completed");
+
+        if self
+            .event_processor
+            .db
+            .state()
+            .get_max_operator_id_seen()
+            .is_none()
+        {
+            warn!("No OperatorAdded events found in historical sync, there is likely a sync error");
```
dknopik, line 521: if we do not persist `get_max_operator_id_seen`, we need to check the `operators` map instead here, as very short historic syncs (e.g. after a restart) trigger this otherwise.

anchor/database/src/sql_operations.rs
```diff
@@ -2,6 +2,9 @@
 pub const INSERT_METADATA: &str = r#"INSERT INTO metadata (domain_type) VALUES (?1)"#;
 pub const GET_METADATA: &str = r#"SELECT schema_version, domain_type FROM metadata"#;
 pub const GET_LEGACY_BLOCK: &str = r#"SELECT * FROM block"#;
+pub const GET_MAX_OPERATOR_ID_SEEN: &str = r#"SELECT max_operator_id_seen FROM metadata"#;
+pub const BUMP_MAX_OPERATOR_ID_SEEN: &str =
+    r#"UPDATE metadata SET max_operator_id_seen = max_operator_id_seen + 1"#;
```
dknopik: We need this to take the new max operator id as parameter so that we can handle the case of updating existing databases

anchor/database/src/table_schema.sql
```diff
@@ -4,7 +4,8 @@
 CREATE TABLE metadata (
     schema_version INTEGER NOT NULL DEFAULT 1,
     domain_type INTEGER NOT NULL,
-    block_number INTEGER NOT NULL DEFAULT 0 CHECK (block_number >= 0)
+    block_number INTEGER NOT NULL DEFAULT 0 CHECK (block_number >= 0),
+    max_operator_id_seen INTEGER NOT NULL DEFAULT 0
```
dknopik: This needs to be `NULL`-able. `NULL` would represent missing data due to this being a migrated database.

anchor/eth/src/event_processor.rs
```diff
@@ -188,6 +188,19 @@ impl EventProcessor {
             )));
         }
 
+        let max_seen = self.db.state().get_max_operator_id_seen();
+
+        if max_seen != operatorId - 1 {
+            error!(
+                "Missing OperatorAdded events: database has only seen up to id {max_seen}, \
+                but got operator {operator_id}."
+            );
```
dknopik: We should discuss whether we want to keep this a log-only soft error or return an error here to signal that the sync has failed. I think personally, I prefer the latter

anchor/database/src/table_schema.sql
```diff
@@ -4,7 +4,8 @@
 CREATE TABLE metadata (
     schema_version INTEGER NOT NULL DEFAULT 1,
     domain_type INTEGER NOT NULL,
-    block_number INTEGER NOT NULL DEFAULT 0 CHECK (block_number >= 0)
+    block_number INTEGER NOT NULL DEFAULT 0 CHECK (block_number >= 0),
+    max_operator_id_seen INTEGER
```
dknopik: please also add a migration for this :)

anchor/database/src/table_schema.sql
```diff
@@ -1,10 +1,11 @@
...
+-- SCHEMA VERSION 2
 
 -- we should avoid removing columns from this to keep compatibility between anchor Versions
 CREATE TABLE metadata (
-    schema_version INTEGER NOT NULL DEFAULT 1,
+    schema_version INTEGER NOT NULL DEFAULT 2,
     domain_type INTEGER NOT NULL,
-    block_number INTEGER NOT NULL DEFAULT 0 CHECK (block_number >= 0)
+    block_number INTEGER NOT NULL DEFAULT 0 CHECK (block_number >= 0),
+    max_operator_id_seen INTEGER
```
dknopik: This should be `DEFAULT 0` so that newly created databases start with a `max_operator_id_seen` of `0` instead of `NULL`.

dknopik (commented):
looks good!

diegomrsantos:
@claude review this PR. Is it worth moving forward with this approach, given https://eips.ethereum.org/EIPS/eip-4444 and the fact that we are currently investigating a solution similar to Ethereum's checkpoint sync to stop relying on all logs existing in execution clients?

petarjuki7:
> Clarify operator ID starting value - verify the contract starts at 1, not 0

It does, but I agree it's a good thing to point out / double check

> Good placement, but consider: what if operators were added before the contract deployment block being synced from?

By definition they shouldn't be.

For other points I agree but **EIP-4444** doesn't seem to be arriving soon, and there will be a number of users receiving this error (unfortunately) so I think it makes sense to include it into **1.2.0.**

anchor/eth/src/event_processor.rs
petarjuki7: The problem I encountered while testing is that by returning an `Error(...)` type, it gets handled in the `process_logs()` function, and if its a historical sync it only gets logged as a `trace`. So it's practically invisible without running the client with `RUST_LOG = trace` or similar, its not really a "**loud warning**" as stated in the issue description.
Maybe keep the `error!()` log and also return an `Error()` ?

jking-aus:
hey just chiming in to say four fours is realistically not happening in the next two forks so we dont need to worry about that. We are however having a consistent problem where users aren't aware their sync is working because of missing receipts. So we need something like this PR -- good work Petar.

petarjuki7:
Forgot to mention here, I tested the happy case, works fine.
I simulated an error by just adding `if operatorId == 15 { early return; }`, and it showed errors correctly (picture attached)
image

dknopik (approved):
LGTM!

PR #633, opened 2025-09-25
test: add comprehensive operator lifecycle integration test

diegomrsantos:
## Issue Addressed

Missing test coverage for operator lifecycle management in SSV protocol.

## Proposed Changes

- Add comprehensive operator lifecycle integration test
- Test validates soft delete -> hard delete behavior when operators are removed
- Ensures database integrity through SQL triggers and foreign key constraints

## Additional Info

Critical for validating that operator removal maintains cluster data consistency while properly cleaning up unused operator records.


anchor/eth/tests/operator_lifecycle.rs
```diff
@@ -0,0 +1,149 @@
...
+    // Verify all operators exist
+    for &operator_id in &operator_ids {
+        verify_operator_stored(&processor, OperatorId(operator_id));
+    }
+
+    // Create a cluster with all 4 operators using ValidatorAdded event
+    // CRITICAL: Use the exact same values as in VALID_SHARES_DATA for signature verification
+    // The shares data contains a BLS signature over "owner:nonce" that must match exactly
+    let cluster_owner =
+        Address::from_str("0x000000633b68f5d8d3a86593ebb815b4663bcbe0").expect("Invalid address");
```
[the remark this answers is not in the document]
diegomrsantos: done

anchor/eth/tests/common/mod.rs
```diff
@@ -0,0 +1,424 @@
+use std::sync::Arc;
```
dknopik, line 10: This file is `common/mod.rs` while the other modules are just `integration.rs` and `operator_lifecycle.rs`. Unless there are plans to add submodules soon, I'd suggest moving this to `common.rs`

anchor/eth/tests/common.rs
```diff
@@ -0,0 +1,424 @@
...
+        topics,
+        data.into(),
+        Some(12346),
+        Some(FixedBytes::default()),
+        Some(1),
+    )
+}
+
+/// Helper function to create an OperatorRemoved event log
+#[allow(dead_code)]
```
dknopik: Same as copilot noted: remove `#[allow(dead_code)]`

Also several occurrences below.

anchor/eth/tests/common.rs
```diff
@@ -0,0 +1,361 @@
...
+            index_sync_tx,
+            exit_tx,
+            slashing_protection,
+        },
+    );
+    (processor, index_sync_rx)
+}
+
+/// Create a KeySplit mode EventProcessor for testing
+#[allow(dead_code)]
```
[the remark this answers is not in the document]
diegomrsantos: we need this

anchor/eth/tests/common/mod.rs
```diff
@@ -0,0 +1,424 @@
...
+
+    // Verify operator does not exist in database (hard delete removes completely)
+    let mut conn = processor
+        .db
+        .connection()
+        .expect("Failed to get database connection");
+    let tx = conn.transaction().expect("Failed to start transaction");
+
+    assertions::operator::exists_not_in_db(operator_id, &tx);
+}
```
dknopik: These function do the exact same thing, is this your intention?

anchor/eth/tests/common.rs
diegomrsantos: Clippy complains that the fn isn't used

anchor/eth/tests/common.rs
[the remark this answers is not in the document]
dknopik: why? the function appears to be used, the `allow` should not be needed

anchor/eth/tests/common/mod.rs
diegomrsantos: They should be different, thanks. Fixed.

anchor/eth/tests/common/mod.rs
```diff
@@ -0,0 +1,444 @@
...
+        .expect("Failed to get database connection");
+    let tx = conn.transaction().expect("Failed to start transaction");
+
+    let is_soft_deleted = processor
+        .db
+        .is_operator_soft_deleted(operator_id, &tx)
+        .expect("Failed to check if operator exists in database");
+
+    assert!(
+        !is_soft_deleted,
```
diegomrsantos: I'm probably gonna improve this

diegomrsantos: done

anchor/eth/tests/common/mod.rs
diegomrsantos: done

anchor/database/src/operator_operations.rs
```diff
@@ -97,4 +107,46 @@ impl NetworkDatabase {
...
         Ok(())
     }
+
+    /// Get the status of an operator in the database
+    pub fn get_operator_status(
+        &self,
+        id: OperatorId,
+        tx: &Transaction<'_>,
+    ) -> Result<OperatorStatus, DatabaseError> {
+        let query = "SELECT removed FROM operators WHERE operator_id = ?1";
```
dknopik: Move this query to be a constant in `sql_operations.rs` to keep them all in one place.

anchor/eth/tests/common.rs
dknopik: Ah, okay. I guess what's happening is that each file in the `tests` folder directly is being individually compiled as integration test crate. And while it is used by `integration` and `operator_lifecycle`, it is not used by itself. 

Now I found this note in the rust docs on integration tests: "Creating the module as tests/common.rs also works, but is not recommended because the test runner will treat the file as a test crate and try to run tests inside it."

I did not know that - this is the first time where I heard about `xyz.rs` being treated differently from `xyz/mod.rs` :D

In that case, your previous approach of `common/mod.rs` was actually more correct. sorry for the back and forth. I believe maybe then the `allow(dead_code)` can be removed.

diegomrsantos: It was also happening with the previous approach :D

anchor/database/src/operator_operations.rs
diegomrsantos: done

anchor/eth/tests/common.rs
diegomrsantos: pushed a fix

dknopik (approved):
would be good to somehow avoid the `allow(dead_code)`, but we can take a look at that later ;)

PR #634, opened 2025-09-26
fix: Sign validator registration at the same time as Go-SSV

dknopik:
## Issue Addressed

We can not rely on Lighthouse's `preparation_service` to trigger signing the validator MEV registration, as we need to do it at the same time as Go-SSV.

## Proposed Changes

Implement our own `registration_service` and use the `preparation_service` only for proposal preparations.

### DIFF
--- a/anchor/validator_store/src/lib.rs
+++ b/anchor/validator_store/src/lib.rs
@@ -1,5 +1,6 @@
 pub mod metadata_service;
 mod metrics;
+pub mod registration_service;
 
 use std::{
     collections::{HashMap, HashSet},
@@ -1081,7 +1082,6 @@ impl<T: SlotClock, E: EthSpec> ValidatorStore for AnchorValidatorStore<T, E> {
     ) -> Result<SignedValidatorRegistrationData, Error> {
         let future = async {
             let domain_hash = self.spec.get_builder_domain();
-            let signing_root = validator_registration_data.signing_root(domain_hash);
 
             let (validator, cluster) =
                 self.get_validator_and_cluster(validator_registration_data.pubkey)?;
@@ -1090,13 +1090,16 @@ impl<T: SlotClock, E: EthSpec> ValidatorStore for AnchorValidatorStore<T, E> {
             let epoch = self
                 .slot_clock
                 .slot_of(Duration::from_secs(validator_registration_data.timestamp))
-                .unwrap_or(self.spec.genesis_slot)
+                .ok_or(SpecificError::SlotClock)?
                 .epoch(E::slots_per_epoch());
-            let sign_slot = epoch.start_slot(E::slots_per_epoch());
-            let validity_slot = epoch.end_slot(E::slots_per_epoch());
-            if let Some(duration) = self.slot_clock.start_of(sign_slot) {
-                validator_registration_data.timestamp = duration.as_secs();
-            }
+            let slot = epoch.start_slot(E::slots_per_epoch());
+            let duration = self
+                .slot_clock
+                .start_of(slot)
+                .ok_or(SpecificError::SlotClock)?;
+            validator_registration_data.timestamp = duration.as_secs();
+
+            let signing_root = validator_registration_data.signing_root(domain_hash);
 
             let signature = self
                 .collect_signature(
@@ -1106,7 +1109,7 @@ impl<T: SlotClock, E: EthSpec> ValidatorStore for AnchorValidatorStore<T, E> {
                     &validator,
                     &cluster,
                     signing_root,
-                    validity_slot,
+                    slot,
                 )
                 .await?;
 
--- a/anchor/validator_store/src/lib.rs
+++ b/anchor/validator_store/src/lib.rs
@@ -1078,7 +1078,7 @@ impl<T: SlotClock, E: EthSpec> ValidatorStore for AnchorValidatorStore<T, E> {
 
     async fn sign_validator_registration_data(
         &self,
-        mut validator_registration_data: ValidatorRegistrationData,
+        validator_registration_data: ValidatorRegistrationData,
     ) -> Result<SignedValidatorRegistrationData, Error> {
         let future = async {
             let domain_hash = self.spec.get_builder_domain();
@@ -1097,7 +1097,10 @@ impl<T: SlotClock, E: EthSpec> ValidatorStore for AnchorValidatorStore<T, E> {
                 .slot_clock
                 .start_of(slot)
                 .ok_or(SpecificError::SlotClock)?;
-            validator_registration_data.timestamp = duration.as_secs();
+            let validator_registration_data = ValidatorRegistrationData {
+                timestamp: duration.as_secs(),
+                ..validator_registration_data
+            };
 
             let signing_root = validator_registration_data.signing_root(domain_hash);
 
--- a/anchor/validator_store/src/registration_service.rs
+++ b/anchor/validator_store/src/registration_service.rs
@@ -103,7 +103,7 @@ impl<S: ValidatorStore + 'static, T: SlotClock + 'static> Inner<S, T> {
     fn collect_validator_registration_data(
         &self,
         slot: Slot,
-        slots_between_registrations: u64,
+        number_of_slots_between_registrations: u64,
     ) -> Vec<ValidatorRegistrationData> {
         let all_pubkeys: Vec<_> = self
             .validator_store
@@ -125,7 +125,7 @@ impl<S: ValidatorStore + 'static, T: SlotClock + 'static> Inner<S, T> {
                     pubkey,
                     timestamp,
                     slot,
-                    slots_between_registrations,
+                    number_of_slots_between_registrations,
                 )
             })
             .collect()
@@ -136,17 +136,13 @@ impl<S: ValidatorStore + 'static, T: SlotClock + 'static> Inner<S, T> {
         pubkey: PublicKeyBytes,
         timestamp: u64,
         slot: Slot,
-        slots_between_registrations: u64,
+        number_of_slots_between_registrations: u64,
     ) -> Option<ValidatorRegistrationData> {
         let proposal_data = self.validator_store.proposal_data(&pubkey)?;
         // Ignore fee recipients for keys without indices, they are inactive.
         let index = proposal_data.validator_index?;
 
-        // To not sign for all validators at once, select based on the current slot.
-        // Note that it is important that this is the same across client implementations, as else
-        // the nodes broadcast partial signatures for their validators at varying times.
-        // The modulo is applied to both slot and index to balance the load.
-        if slot % slots_between_registrations != index % slots_between_registrations {
+        if is_scheduled_for_slot(slot, index, number_of_slots_between_registrations) {
             return None;
         }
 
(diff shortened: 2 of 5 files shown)


anchor/validator_store/src/registration_service.rs
```diff
@@ -0,0 +1,187 @@
...
+        info!("Validator registration service started");
+
+        let spec = spec.clone();
+        let slot_duration = Duration::from_secs(spec.seconds_per_slot);
+
+        let executor = self.inner.executor.clone();
+
+        let validator_registration_fut = async move {
+            loop {
+                if let Some(slot) = self.inner.slot_clock.now() {
```
diegomrsantos, line 72: What should happen when it's `None`?

anchor/validator_store/src/lib.rs
```diff
@@ -1090,13 +1090,16 @@ impl<T: SlotClock, E: EthSpec> ValidatorStore for AnchorValidatorStore<T, E> {
...
-            let validity_slot = epoch.end_slot(E::slots_per_epoch());
-            if let Some(duration) = self.slot_clock.start_of(sign_slot) {
-                validator_registration_data.timestamp = duration.as_secs();
-            }
+            let slot = epoch.start_slot(E::slots_per_epoch());
+            let duration = self
+                .slot_clock
+                .start_of(slot)
+                .ok_or(SpecificError::SlotClock)?;
+            validator_registration_data.timestamp = duration.as_secs();
```
diegomrsantos: Can we create a new value instead of mutating the input?

anchor/validator_store/src/registration_service.rs
```diff
@@ -0,0 +1,187 @@
+use std::sync::Arc;
```
diegomrsantos, line 15: How about this module docs?

```rust
//! Coordinates periodic MEV/Builder **validator registrations** for Anchor.
//!
//! Why this exists: in DVT the validator key is split, and the **registration payload includes
//! a timestamp**; if different operators construct it at different moments they’ll sign different
//! bytes. Anchor therefore **owns the timing** of registration (instead of Lighthouse’s
//! `preparation_service`) so all operators sign the **same** message and submit it to relays via BN.
//!
//! This module:
//! - derives a **slot-start** timestamp (deterministic) for each window,
//! - builds `ValidatorRegistrationData` from proposer prefs (fee recipient, gas limit) and pubkey,
//! - requests signatures from the `ValidatorStore`,
//! - and submits `SignedValidatorRegistrationData` to connected beacon nodes (Builder API).
//!
//! Per-validator builder on/off and gas-limit are dynamic and may be configured via VC APIs.
```

anchor/validator_store/src/registration_service.rs
```diff
@@ -0,0 +1,187 @@
...
+                if let Some(slot) = self.inner.slot_clock.now() {
+                    let inner = self.inner.clone();
+                    let executor = inner.executor.clone();
+                    let future = async move {
+                        // Poll the endpoint immediately to ensure fee recipients are received.
+                        if let Err(e) = inner.register_validators(slot).await {
+                            error!(error = ?e, "Error during validator registration");
+                        }
+                    };
+                    executor.spawn(future, "validator_registration");
```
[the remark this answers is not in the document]
diegomrsantos: The issue here is that degraded BN/relay could cause piling tasks.

anchor/validator_store/src/registration_service.rs
```diff
@@ -0,0 +1,187 @@
...
+            .map(|duration| duration.as_secs())
+        else {
+            // Try again later.
+            return vec![];
+        };
+
+        all_pubkeys
+            .into_iter()
+            .filter_map(|pubkey| {
+                let proposal_data = self.validator_store.proposal_data(&pubkey)?;
```
diegomrsantos: Can we extract this code block to a function?

anchor/validator_store/src/registration_service.rs
```diff
@@ -0,0 +1,187 @@
...
+                    %err,
+                    "Unable to publish validator registrations to the builder network"
+                ),
+            }
+        }
+    }
+
+    /// Register validators with builders, used in the blinded block proposal flow.
+    async fn register_validators(&self, slot: Slot) -> Result<(), String> {
+        let slots_per_registration =
```
diegomrsantos: Would `number_of_slots_between_registrations` be a more descriptive name?

anchor/validator_store/src/registration_service.rs
```diff
@@ -0,0 +1,187 @@
...
+
+        all_pubkeys
+            .into_iter()
+            .filter_map(|pubkey| {
+                let proposal_data = self.validator_store.proposal_data(&pubkey)?;
+                // Ignore fee recipients for keys without indices, they are inactive.
+                let index = proposal_data.validator_index?;
+
+                // To not sign for all validators at once, select based on the current slot.
+                if slot % slots_per_registration != index % slots_per_registration {
```
diegomrsantos: One downside of this approach is that if we can't do the registration for the validators chosen for a specific slot, we need to wait 10 epochs to try again. Is that how go-ssv does it? Do we need to do the same?

anchor/validator_store/src/lib.rs
dknopik: why?

I think changing the field more clearly conveys the intent of changing a single field

anchor/validator_store/src/lib.rs
```diff
@@ -1090,13 +1090,16 @@ impl<T: SlotClock, E: EthSpec> ValidatorStore for AnchorValidatorStore<T, E> {
             let epoch = self
                 .slot_clock
                 .slot_of(Duration::from_secs(validator_registration_data.timestamp))
-                .unwrap_or(self.spec.genesis_slot)
+                .ok_or(SpecificError::SlotClock)?
```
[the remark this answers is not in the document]
dknopik: IMO its fine, especially as this is an established pattern in the validator store.

anchor/validator_store/src/registration_service.rs
dknopik: If it is `None` (i.e. we are for some reason before genesis), it tries again later :)

anchor/validator_store/src/registration_service.rs
dknopik: Yes, this is how go-ssv does it. Yes, we need to do the same, as else the partial signatures are not broadcast at the same time.

anchor/validator_store/src/registration_service.rs
```diff
@@ -0,0 +1,187 @@
...
+        };
+
+        all_pubkeys
+            .into_iter()
+            .filter_map(|pubkey| {
+                let proposal_data = self.validator_store.proposal_data(&pubkey)?;
+                // Ignore fee recipients for keys without indices, they are inactive.
+                let index = proposal_data.validator_index?;
+
+                // To not sign for all validators at once, select based on the current slot.
```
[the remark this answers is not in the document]
dknopik: pushing something similar

anchor/validator_store/src/registration_service.rs
dknopik: done

anchor/validator_store/src/registration_service.rs
dknopik: pushed `slots_between_registrations`, is that fine?

anchor/validator_store/src/registration_service.rs
dknopik: done, but removed the last sentence (hallucinated from LH VC info)

anchor/validator_store/src/registration_service.rs
diegomrsantos: Should we handle it and log something?

anchor/validator_store/src/lib.rs
diegomrsantos: I believe it's not a good practice to mutate input parameters to a function

anchor/validator_store/src/registration_service.rs
diegomrsantos: Not sure tbh, it was kinda hard for me to understand what was going on

anchor/validator_store/src/registration_service.rs
diegomrsantos: Can we extract it to a function named something like `is_scheduled_for_slot`? We can also add a comment like "Assign each validator to one of K slot buckets by index % K. In slot s, we serve bucket s % K. This ensures each validator refreshes once every K slots.”

anchor/validator_store/src/lib.rs
diegomrsantos: IIRC, it's not even possible in Kotlin

anchor/validator_store/src/registration_service.rs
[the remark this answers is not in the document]
dknopik: The issue with not spawning is that collecting the signatures might take a while. If other enough other operators are not sending partial signatures, it times out after the slot after the current.

Then, below we wait until the next slot for the next iteration.

So, we miss 2 iterations if any signature times out. This is a problem as the next attempt for the validators assigned in these iterations is in 10 epochs. Furthermore, if most nodes behave this way, the problem will become worse: missing iterations will cause other clients to time out, and again miss 2 future iterations, which will cause other nodes to time out, etc...

Finally, existing services also do per-slot spawning: e.g.
https://github.com/sigp/lighthouse/blob/522bd9e9c6ac167f2231525e937c9ebbcb86cf6e/validator_client/validator_services/src/attestation_service.rs#L185
and
https://github.com/sigp/lighthouse/blob/522bd9e9c6ac167f2231525e937c9ebbcb86cf6e/validator_client/validator_services/src/sync_committee_service.rs#L119

anchor/validator_store/src/registration_service.rs
dknopik: done

anchor/validator_store/src/lib.rs
dknopik: done

anchor/validator_store/src/registration_service.rs
dknopik: done, wdyt?

anchor/validator_store/src/registration_service.rs
dknopik: done, wdyt?

anchor/validator_store/src/registration_service.rs
diegomrsantos: perfect, thanks

anchor/validator_store/src/registration_service.rs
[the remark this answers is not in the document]
diegomrsantos: If the downstream services are struggling to process our requests, continuing to request will make things worse. The suggestion is to slow down in this case, some kind of back-pressure.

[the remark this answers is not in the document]
diegomrsantos: There's this on LH [docs](https://lighthouse-book.sigmaprime.io/help_vc.html), for ex:

--validator-registration-batch-size <INTEGER>
          Defines the number of validators per validator/register_validator
          request sent to the BN. This value can be reduced to avoid timeouts
          from builders. [default: 500]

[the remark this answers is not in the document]
dknopik: > --validator-registration-batch-size 

I intentionally did not include this. 

We (soon) have up to 3000 validators, with a registration occurring every 320 slots. This means that we have 9.375 registrations per slot on average. Even if we are unlucky, we are far away from the LH default of 500.

diegomrsantos (approved):
LGTM

anchor/validator_store/src/registration_service.rs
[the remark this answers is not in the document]
diegomrsantos: We agreed to create an issue to evaluate this and other similar parts of the system

Issue #632, opened 2025-09-25
Keysplit takes password as CLI


AgeManning:
The `anchor keysplit` command, currently requires `--password` as an argument. 

This is conventionally bad practice. It forces users to put key material as a CLI parameter. This will show up in the process list (so systems with multiple users, other users can read key material), also many shells record history of commands, and this key material will often get stored in shell history. 

We should disallow any key material in CLI arguments and force the user to enter it via stdin.

PR #639, opened 2025-09-29
fix: Read keystore password from disk or stdin

dknopik:
## Issue Addressed

- #632

## Proposed Changes

Change `--password` to `--password-file` and make it optional. If it is provided, the password is read from the provided path, else it is read from stdin.

### DIFF
--- a/anchor/keysplit/src/lib.rs
+++ b/anchor/keysplit/src/lib.rs
@@ -57,7 +57,7 @@ pub fn run_keysplitter(
     let keys = keystore
         .decrypt_keypair(
             read_password(shared.password_file.as_deref())
-                .map_err(KeysplitError::Keystore)?
+                .map_err(|e| KeysplitError::Keystore(format!("Unable to get password: {e}")))?
                 .as_bytes(),
         )
         .map_err(|e| KeysplitError::Keystore(format!("Failed to decrypt keystore file: {e:?}")))?;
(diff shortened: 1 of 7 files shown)


anchor/keysplit/src/lib.rs
```diff
@@ -54,7 +55,11 @@ pub fn run_keysplitter(
     // 2) Extract the validator keys from the keystore file
     info!("Extracting keys from keystore file...");
     let keys = keystore
-        .decrypt_keypair(shared.password.as_bytes())
+        .decrypt_keypair(
+            read_password(shared.password_file.as_deref())
+                .map_err(KeysplitError::Keystore)?
```
diegomrsantos, line 60: Should we do sth like https://github.com/sigp/anchor/blob/ff56d2c17b1935105755c44cc001ee4386bdaa81/anchor/keysplit/src/lib.rs#L51 ?

diegomrsantos: Should we add more context, like at line 52?

diegomrsantos:
Can we add tests?

dknopik:
> Can we add tests?

what kind of tests do you have in mind? this being a function that involves interactiveness and is otherwise very simple find it hard to test meaningfully.

anchor/keysplit/src/lib.rs
dknopik: done!

diegomrsantos:
> > Can we add tests?
> 
> what kind of tests do you have in mind? this being a function that involves interactiveness and is otherwise very simple find it hard to test meaningfully.

Pushed a commit, let me know what you think

dknopik:
Sure. feels a bit much to me, but i am fine wiht it

diegomrsantos (approved):

diegomrsantos:
> Sure. feels a bit much to me, but i am fine wiht it

imo, using smaller functions makes it easier to understand

PR #641, opened 2025-09-30
fix: Use xxhash64 for gossipsub message ID to match Go-SSV

diegomrsantos:
## Issue Addressed

Fixes message ID mismatch between Anchor and Go-SSV nodes causing peer blocking on mainnet.

## Problem

Anchor was using SHA256 (20 bytes) for gossipsub message IDs while Go-SSV uses xxhash64 (12 bytes). This mismatch caused:
- Different message IDs for the same message between Rust and Go implementations
- Anchor nodes blocking SSV nodes for not responding to IWANT requests (the messages existed but had different IDs)
- Network connectivity issues on mainnet

## Proposed Changes

- Replace SHA256-based message ID with xxhash64 (12-byte output)
- Add `twox-hash` v2.1.2 dependency (more popular and better maintained than alternatives)
- Implement `ssv_message_id()` function matching [Go's implementation](https://github.com/ssvlabs/ssv/blob/main/network/topics/msg_id.go#L125-L138)
- Add comprehensive unit tests

**Why xxhash64?**

The previous implementation used SHA256 as specified in the [SSV spec](https://github.com/ssvlabs/ssv-spec/blob/main/p2p/CONFIG.md#msg-id), but Go-SSV changed to xxhash64 for better performance. After the Alan upgrade, message bodies can be identical across operators, so hashing the full message ensures uniqueness.

## Additional Info

This ensures message ID compatibility and proper peer scoring between Rust and Go SSV implementations.


diegomrsantos:
@nkryuchkov could you please take a look?

dknopik (approved):
LGTM, thanks for implementing!

nkryuchkov (commented):
@diegomrsantos LGTM

PR #646, opened 2025-10-02
chore: Reduce number of log files

AgeManning:
## Issue Addressed

I was looking through some logs on our servers and noticed a large number of compressed log files. 

I see that we have a default of 100 compressed log files. Each are default to 100MB. I think this is a bit excessive. I've reduced the default to 10 log files.


dknopik (dismissed):
LGTM, the old default was back from when we logged even more

dknopik:
I think we even want this in `v1.0.0`. I can handle rebasing if you agree

dknopik (approved):
LGTM

PR #647, opened 2025-10-02
chore: Update libp2p to latest version

AgeManning:
## Issue Addressed

This bumps the libp2p version to latest. It has improved queuing and the ability to cancel messages that have been queued when we receive and IDONTWANT. It should hopefully improve performance.


dknopik:
Closing, will be included by backmerging from v1.0.0

PR #649, opened 2025-10-04
fix: implement periodic peer discovery and connection failure tracking

diegomrsantos:
## Issue Addressed

Fixes peer count plateau at 12-14 peers instead of target 50. Root cause: discovery only executed once at startup, preventing continuous peer growth.

### Core Fix: Periodic Discovery
- Trigger peer discovery every 30s in heartbeat when below target
- Previously: single discovery attempt at startup
- Result: Sustained peer growth to target levels

### Connection Failure Tracking
- Blacklist peers after 5 consecutive connection failures
- 1-hour expiration with automatic retry
- Periodic cleanup every 30s (prevents memory leaks)
- Hard limit of 1000 failure records (defense-in-depth)
- Auto-clear on successful connection

### ENR to Multiaddr Conversion
- Custom conversion returning only TCP/QUIC addresses
- Skip plain UDP addresses (not usable by libp2p)
- Reduces wasted connection attempts

### Logging Improvements
- Connection errors at debug level (normal P2P behavior)
- Subnet bitfield logging in handshakes and metadata updates
- Increased log retention (1→5 files, ~500MB history)

## Testing Results

**2-hour mainnet test:**
- Peer count: 4 → 30 (650% increase, 60% of target)
- Connection errors: 2,358 → ~60 (97% reduction)
- Memory: Bounded at ~48KB for failure tracking
- No memory leaks observed

## Additional Info

**Memory Protection Strategy:**
- Expiration: 1 hour per failure record
- Cleanup: Every 30s in heartbeat
- Hard limit: 1000 records maximum

**Future Enhancements:**
- Consider exponential backoff (similar to go-libp2p BackoffConnector)
- Evaluate optimal failure threshold and expiration duration
- Monitor long-term peer stability and churn rates

### DIFF
--- a/anchor/logging/src/tracing_libp2p_discv5_layer.rs
+++ b/anchor/logging/src/tracing_libp2p_discv5_layer.rs
@@ -81,12 +81,12 @@ pub fn create_libp2p_discv5_tracing_layer(
         let libp2p_writer =
             LogRollerBuilder::new(tracing_log_path.clone(), PathBuf::from("libp2p.log"))
                 .rotation(Rotation::SizeBased(RotationSize::MB(max_log_size)))
-                .max_keep_files(1);
+                .max_keep_files(5);
 
         let discv5_writer =
             LogRollerBuilder::new(tracing_log_path.clone(), PathBuf::from("discv5.log"))
                 .rotation(Rotation::SizeBased(RotationSize::MB(max_log_size)))
-                .max_keep_files(1);
+                .max_keep_files(5);
 
         let libp2p_writer = match libp2p_writer.build() {
             Ok(writer) => writer,
--- a/anchor/network/src/network.rs
+++ b/anchor/network/src/network.rs
@@ -223,6 +223,18 @@ impl<R: MessageReceiver> Network<R> {
                                     self.check_block_and_prune_peers_by_score();
                                 }
 
+                                // Trigger periodic peer discovery if below target
+                                let connected_peers = self.swarm.behaviour().peer_manager.connected_peers();
+                                let target_peers = self.swarm.behaviour().peer_manager.target_peers();
+                                if connected_peers < target_peers {
+                                    debug!(
+                                        connected_peers,
+                                        target_peers,
+                                        "Below target peer count, triggering peer discovery"
+                                    );
+                                    self.swarm.behaviour_mut().discovery.discover_peers(target_peers);
+                                }
+
                                 // Disconnect peers that no longer subscribe to any needed subnets
                                 let to_disconnect = self
                                     .swarm
@@ -248,6 +260,18 @@ impl<R: MessageReceiver> Network<R> {
                                 &mut self.swarm.behaviour_mut().handshake,
                                 peer_id
                             );
+                        },
+                        SwarmEvent::OutgoingConnectionError { peer_id, error, .. } => {
+                            debug!(?peer_id, ?error, "Outgoing connection error");
+                            if let Some(peer_id) = peer_id {
+                                self.peer_manager().record_connection_failure(&peer_id);
+                            }
+                        },
+                        SwarmEvent::IncomingConnectionError { error, .. } => {
+                            debug!(?error, "Incoming connection error");
+                        },
+                        SwarmEvent::ConnectionClosed { peer_id, cause, .. } => {
+                            debug!(?peer_id, ?cause, "Connection closed");
                         },
                         SwarmEvent::NewListenAddr { listener_id, address } => {
                             self.on_new_listen_addr(listener_id, address);
@@ -474,10 +498,20 @@ impl<R: MessageReceiver> Network<R> {
 
         // update enr and metadata to new state
         self.discovery().set_subscribed(subnet, subscribed);
-        if let Some(metadata) = &mut self.node_info.metadata
-            && let Err(err) = metadata.set_subscribed(subnet, subscribed)
-        {
-            error!(?err, "unable to update node info");
+        if let Some(metadata) = &mut self.node_info.metadata {
+            match metadata.set_subscribed(subnet, subscribed) {
+                Ok(()) => {
+                    info!(
+                        subnet = *subnet,
+                        subscribed = subscribed,
+                        subnets_bitfield = %metadata.subnets,
+                        "Updated node_info metadata subnet bitfield"
+                    );
+                }
+                Err(err) => {
+                    error!(?err, "unable to update node info");
+                }
+            }
         }
     }
 
--- a/anchor/network/src/peer_manager/connection.rs
+++ b/anchor/network/src/peer_manager/connection.rs
@@ -134,10 +134,7 @@ impl ConnectionManager {
         }
 
         // Don't dial peers with too many consecutive failures (unless expired)
-        if let Some(record) = self.peer_failures.get(peer_id)
-            && record.last_failure.elapsed() < FAILURE_EXPIRATION
-            && record.consecutive_failures >= MAX_CONSECUTIVE_FAILURES
-        {
+        if self.is_blacklisted_by_failures(peer_id) {
             return false;
         }
 
--- a/anchor/network/src/peer_manager/connection.rs
+++ b/anchor/network/src/peer_manager/connection.rs
@@ -65,6 +65,9 @@ pub struct ConnectionManager {
     observed_peer_subnets: HashMap<PeerId, Bitfield<Fixed<U128>>>,
     // Tracks connection failures for blacklisting persistently failing peers
     peer_failures: HashMap<PeerId, FailureRecord>,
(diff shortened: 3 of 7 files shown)


anchor/network/src/peer_manager/connection.rs
```diff
@@ -115,6 +133,14 @@ impl ConnectionManager {
             return false;
         }
 
+        // Don't dial peers with too many consecutive failures (unless expired)
+        if let Some(record) = self.peer_failures.get(peer_id)
+            && record.last_failure.elapsed() < FAILURE_EXPIRATION
+            && record.consecutive_failures >= MAX_CONSECUTIVE_FAILURES
```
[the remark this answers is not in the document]

jking-aus (dismissed):
lgtm - nice work diego, curious why we're hitting 30 peers not target of 50 though?

jking-aus (dismissed):
I'll come back and shift these counts to metrics in the future -- lgtm

anchor/logging/src/tracing_libp2p_discv5_layer.rs
```diff
@@ -81,12 +81,12 @@ pub fn create_libp2p_discv5_tracing_layer(
...
             LogRollerBuilder::new(tracing_log_path.clone(), PathBuf::from("libp2p.log"))
                 .rotation(Rotation::SizeBased(RotationSize::MB(max_log_size)))
-                .max_keep_files(1);
+                .max_keep_files(5);
 
         let discv5_writer =
             LogRollerBuilder::new(tracing_log_path.clone(), PathBuf::from("discv5.log"))
                 .rotation(Rotation::SizeBased(RotationSize::MB(max_log_size)))
-                .max_keep_files(1);
+                .max_keep_files(5);
```
dknopik, line 89: I think this change slipped in accidentally?

dknopik:
I am not sure if the connection attempt blacklisting is a good idea. 

Is it necessary? What are you trying to solve?

My concern is that it prevents dialing for quite a while if we have a general connection issue.

jking-aus:
> I am not sure if the connection attempt blacklisting is a good idea.
> 
> Is it necessary? What are you trying to solve?
> 
> My concern is that it prevents dialing for quite a while if we have a general connection issue.

I'll let @diegomrsantos confirm, but my understanding is there were some peers in the initial discovered set that were repeatedly failing connection requests.  

I think we need to separate out the important piece in this PR which is the discovery on heartbeat when < threshold. 

If we're going to block peers for repeated connection failures, which is worth considering, this shouldn't be a separate timeout list it should be integrated with the blockingmanager imo. I'm assuming the peers we were blocking before are probably hidden in this new blacklist and I would still like to confirm the why via metrics.

anchor/network/src/peer_manager/discovery.rs
```diff
@@ -40,7 +89,7 @@ impl PeerDiscovery {
     ) -> Option<DialOpts> {
         let id = enr.peer_id();
 
-        let multiaddrs = enr.multiaddr();
+        let multiaddrs = enr_to_multiaddrs(&enr);
```
dknopik, line 92: what is this solving?

anchor/network/src/peer_manager/discovery.rs
```diff
@@ -26,6 +26,55 @@ const PEER_OVERDIAL_FACTOR: usize = 2;
...
+            let mut addr = Multiaddr::empty();
+            addr.push(Protocol::Ip6(ip6));
+            addr.push(Protocol::Udp(quic6_port));
+            addr.push(Protocol::QuicV1);
+            multiaddrs.push(addr);
+        }
+    }
+
+    multiaddrs
+}
```
AgeManning: I don't think any of this is necessary. We have it as an EnrExt trait. @dknopik even moved it out into its own useful importable crate:
https://github.com/sigp/lighthouse/blob/stable/common/network_utils/src/enr_ext.rs#L14-L42

anchor/network/src/network.rs
```diff
@@ -223,6 +223,18 @@ impl<R: MessageReceiver> Network<R> {
...
+                                let connected_peers = self.swarm.behaviour().peer_manager.connected_peers();
+                                let target_peers = self.swarm.behaviour().peer_manager.target_peers();
+                                if connected_peers < target_peers {
+                                    debug!(
+                                        connected_peers,
+                                        target_peers,
+                                        "Below target peer count, triggering peer discovery"
+                                    );
+                                    self.swarm.behaviour_mut().discovery.discover_peers(target_peers);
+                                }
```
AgeManning, line 236: There are quite a few caveats with dialing peers. Some we might run into later. Things like attempting to dial peers, that we are already currently dialing. This can lead to multiple dialing attempts. Notice in Lighthouse we check if we are connected OR dialing this peer before attempting to dial. 

I think we should draw inspiration from the lighthouse code as it has already dealt and solved a lot of the race conditions and bugs we are starting to run into here. i.e:

https://github.com/sigp/lighthouse/blob/stable/beacon_node/lighthouse_network/src/peer_manager/mod.rs#L353-L403

AgeManning (commented):
We went with a similar approach in Lighthouse. We called a temporary ban, but it only happens once we've reached our target peer count. 

We shouldn't really be doing lots of dialing attempts to the same peer if discovery is working correctly. We should fail and move on to new peers. 

In general however, I think we are now starting to add in the complexity that exists in the `PeerDB` in Lighthouse. I know we've had the discussion about moving the peer-db to libp2p, but I think this PR demonstrates that there's more logic and reasoning that might be application specific. 

There is more in the Lighthouse PeerDB that I suspect we might end up re-writting in anchor. Specifically around a lot of race conditions that occur when peers change state. 

I tried to group all this logic into a single match statement in the PeerDB: https://github.com/sigp/lighthouse/blob/stable/beacon_node/lighthouse_network/src/peer_manager/peerdb.rs#L869

Unless we can export this to libp2p, we might end up needing it in Anchor. 

Anyway, for this specific issue: https://github.com/sigp/lighthouse/blob/stable/beacon_node/lighthouse_network/src/peer_manager/mod.rs#L855-L876

this function handles what we do with dial failures. It calls the giant connection state and depending on what the previous state was, we can temporarily ban: https://github.com/sigp/lighthouse/blob/stable/beacon_node/lighthouse_network/src/peer_manager/mod.rs#L308

the peer.  There is also a `disconnect_peer` function associated with discovery, that allows to mark the peer as disconnected in the DHT. This has the added effect that we will not advertise this peer to other peers and it will be removed from a bucket in favour of others. 

IIRC the temporary ban was actually added to mitigate ecplise attacks where an attacker kept trying to reconnect over and over. I don't recall multiple outgoing dial failures needing a temporary ban. 

tldr; I think we should take inspiration from lighthouse which has through time resolved a lot of these issues.

anchor/network/src/peer_manager/discovery.rs
dknopik: I think the idea why this is replicated is that this here does not take the non-quic UDP ports, which are for discovery, not for libp2p connection attempts. Not sure if this is actually impactful though.

anchor/logging/src/tracing_libp2p_discv5_layer.rs
diegomrsantos: No, I wanted to increase the number of logs we have. Useful to identify issues

anchor/network/src/peer_manager/discovery.rs
diegomrsantos: Exactly. Is there a version without UDP?

anchor/logging/src/tracing_libp2p_discv5_layer.rs
diegomrsantos: let's keep it?

anchor/network/src/peer_manager/discovery.rs
diegomrsantos: Pushed a new version. Please let me know what you think. Do you think we should keep it?

anchor/network/src/network.rs
diegomrsantos: > Things like attempting to dial peers, that we are already currently dialing. This can lead to multiple dialing attempts.

I think this can be avoided with https://github.com/sigp/anchor/blob/f30d3f0cc1ce8714637c8aecca848d66ad574214/anchor/network/src/peer_manager/discovery.rs#L193

diegomrsantos:
> I'll let @diegomrsantos confirm, but my understanding is there were some peers in the initial discovered set that were repeatedly failing connection requests.

true

diegomrsantos:
> I'm assuming the peers we were blocking before are probably hidden in this new blacklist and I would still like to confirm the why via metrics.

I believe the blocked were different ones, they were possibly slow as the logs indicate. These ones we couldn't even connect

diegomrsantos:
superseded by https://github.com/sigp/anchor/pull/650

PR #650, opened 2025-10-06
fix: implement periodic peer discovery without connection failure tracking

diegomrsantos:
## Issue Addressed

Fixes peer count plateau and connection denial issues caused by generic discovery attempting to connect to peers without needed subnets.

### Subnet-Aware Periodic Discovery
- Trigger discovery every 30s when peer count is below target
- **Uses subnet filtering** via `start_subnet_query()` instead of generic `discover_peers()`
- Only discovers and dials peers that offer our needed subnets
- Prevents `MissingNeededSubnets` connection denials

### Connection Metrics
- Track inbound vs outbound connection counts
- Provides visibility into connection direction distribution
- Helps monitor peer connectivity patterns

### Logging Improvements
- Connection errors logged at debug level (normal P2P behavior)
- Reduces log noise while maintaining debugging capability
- Increased log retention (5 files vs 1) for better historical tracking

## Root Cause Analysis

The original PR #649 added generic `discover_peers()` which found ANY peers on the network:
1. Discovery finds peer with subnets [1, 2, 3]
2. We need subnets [5, 6, 7]
3. We dial the peer
4. Connection attempt succeeds
5. Connection manager checks if peer offers needed subnets
6. Connection **denied** with `MissingNeededSubnets` error
7. Wasted connection attempt, cycle repeats

## Solution

Use subnet-aware discovery that filters ENRs before dialing:
1. Discovery filters by our needed subnets [5, 6, 7]
2. Only peers offering these subnets are discovered
3. We dial filtered peers
4. Connection manager checks subnets
5. Connection **accepted** - peer offers what we need
6. Peer count grows successfully

## Additional Info

This PR intentionally excludes the connection failure blacklist mechanism from PR #649 to evaluate a simpler approach.

**Two complementary discovery mechanisms:**
1. **Per-subnet discovery** (existing `check_subnet_peers`): Ensures minimum 6 peers per subnet
2. **Overall discovery** (new, subnet-aware): Helps reach target peer count by discovering for ALL needed subnets

**Comparison with PR #649:**
- ✅ Includes: Periodic subnet-aware discovery, connection tracking, logging improvements
- ❌ Excludes: Connection failure tracking, peer blacklisting, memory-bounded failure records
- ✅ Fixes: Connection denial bug from generic discovery

**Advantages:**
- Simpler implementation with less state management
- No memory overhead for failure tracking
- Avoids potential issues with false-positive blacklisting
- Prevents wasted connection attempts to unsuitable peers
- Focus on proactive subnet-filtered discovery

**Trade-offs:**
- No automatic avoidance of persistently failing peers
- Relies on discovery to find working peers with correct subnets

### DIFF
--- a/anchor/logging/src/tracing_libp2p_discv5_layer.rs
+++ b/anchor/logging/src/tracing_libp2p_discv5_layer.rs
@@ -81,12 +81,12 @@ pub fn create_libp2p_discv5_tracing_layer(
         let libp2p_writer =
             LogRollerBuilder::new(tracing_log_path.clone(), PathBuf::from("libp2p.log"))
                 .rotation(Rotation::SizeBased(RotationSize::MB(max_log_size)))
-                .max_keep_files(1);
+                .max_keep_files(5);
 
         let discv5_writer =
             LogRollerBuilder::new(tracing_log_path.clone(), PathBuf::from("discv5.log"))
                 .rotation(Rotation::SizeBased(RotationSize::MB(max_log_size)))
-                .max_keep_files(1);
+                .max_keep_files(5);
 
         let libp2p_writer = match libp2p_writer.build() {
             Ok(writer) => writer,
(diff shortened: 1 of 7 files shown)


AgeManning (approved):
Looks good to me

dknopik (approved):
LGTM, thank you!!!

PR #651, opened 2025-10-06
fix: metrics scraping bug

petarjuki7:
## Issue Addressed

Fixes Prometheus unable to scrape metrics from Anchor when running in Docker.

## Proposed Changes

Fix a trailing comma and change to `host.docker.internal:5154`

## Additional Info

When Prometheus runs in Docker, `localhost` refers to the container, not the host. `host.docker.internal` is Docker's DNS name for accessing the host machine from containers (macOS/Windows).

### DIFF
--- a/metrics/scrape-targets/scrape-targets.json
+++ b/metrics/scrape-targets/scrape-targets.json
@@ -4,7 +4,7 @@
       "job": "nodes"
     },
     "targets": [
-      "localhost:5154",
+      "host.docker.internal:5154"
     ]
   }
 ]


dknopik:
This change breaks things when NOT running in docker though, right?

AgeManning:
Does this still work outside of docker?

petarjuki7:
> Does this still work outside of docker?

Not sure, in our [documentation](https://anchor-book.sigmaprime.io/metrics) it says 

> In order to run a metrics server, docker is required to be installed. 

So I just followed that.

dknopik:
I assume this is configuration for the provided docker compose file.

So as this is a docker-specific file anyway, it should be fine. Users setting up their own Prometheus servers provide their own scraping config anyway

I am more concerned that this fixes macos and windows but breaks on linux. Will test.

dknopik:
I think it should work without this, as `network_mode: host` is used. IIRC, docker does not do any bridging when that is used, so `localhost` should in fact refer to the host.

AgeManning (approved):

PR #654, opened 2025-10-07
chore: Fix libp2p metrics feature

jking-aus:
with_bandwidth_metrics requires feature flag for metrics in libp2p

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,7 @@ docs/docs/dist/
 
 # Claude Code files
 /.claude/settings.local.json
+
+# User directories
+/for_claude/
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -26,3 +26,7 @@ docs/docs/dist/
 
 # Claude Code files
 /.claude/settings.local.json
+
+# User directories
+/for_claude/
+/kingy_scripts/
```
dknopik, line 32: In future, can you please put directories specific to you into `.git/info/exclude` instead?

jking-aus: yep will clean up

Issue #645, opened 2025-10-01
Use actual execution and consensus node versions in handshake metadata


diegomrsantos:
## Problem

Handshake metadata uses hardcoded placeholder versions instead of actual node versions:

```rust
// anchor/network/src/network.rs:126-130
let node_info = NodeInfo::new(
    domain_type,
    Some(NodeMetadata {
        node_version: version_with_platform(),
        execution_node: "geth/v1.10.8".to_string(),      // ❌ Hardcoded
        consensus_node: "lighthouse/v1.5.0".to_string(), // ❌ Hardcoded
        subnets: "00000000000000000000000000000000".to_string(),
    }),
);
```

These values are shared with peers during the SSV handshake protocol for debugging and network monitoring (not used for validation).

## Why Not Done Yet

Getting actual versions requires async HTTP calls:
- Beacon node: `GET /eth/v1/node/version`
- Execution node: `web3_clientVersion` JSON-RPC

Challenges:
- Adds startup latency
- Nodes might not be reachable during network initialization
- Network initialization happens before full client startup

### Option 1: Query During Startup
Query connected nodes during client initialization and pass versions to `Network::try_new()`:

```rust
let execution_version = query_execution_version(&config.execution_nodes[0]).await?;
let consensus_version = query_consensus_version(&config.beacon_nodes[0]).await?;
```

**Pros:** Always accurate, automatic
**Cons:** Adds startup latency, needs error handling

### Option 2: Lazy Update
Start with "unknown", update after successful connection:

```rust
// Start with placeholder
execution_node: "unknown".to_string(),

// Later, after connection:
network.update_node_versions(execution_version, consensus_version);
```

**Pros:** No startup delay, accurate once connected, graceful failure handling
**Cons:** More complex implementation, requires making NodeInfo mutable

## Notes

- Go-SSV also uses placeholder values
- These fields don't affect protocol operation
- Should use timeouts (e.g., 5 seconds) to avoid blocking startup
- Fall back to "unknown" if queries fail

PR #661, opened 2025-10-08
fix: Allow incomming peers

AgeManning:
## Issue Addressed

Currently all new peers that are trial dial us, we reject. In principle, peers should only dial us if we are on a subnet that they need, so optimisitcally they should be useful peers. 

The current logic works as follows:
New peer tries to connect -> We look to see if we know their ENR -> We don't know ENR or subnets as they are new -> We reject them because we don't know anything about them. 

I think we shouldn't blanket reject all incomming peers, rather we should let them connect, and then upon handshake figure out if they are of any use to us and then reject the connection. 

I had a quick look through the code, it looks like we don't fail the handshake if our subnets are not aligned, it also looks like we are hard-coding the subnets field. I think it's related to this: https://github.com/sigp/anchor/issues/645

cc:  @dknopik @diegomrsantos - I think we need to not hard code the subnets in the handshake and use that to disconnect peers. We might get over-run with incoming peers that are no use to us if we don't have that logic.

### DIFF
--- a/anchor/network/src/peer_manager/connection.rs
+++ b/anchor/network/src/peer_manager/connection.rs
@@ -207,9 +207,13 @@ impl ConnectionManager {
         }
 
         let Some(bitfield) = self.get_peer_subnets_with_enr_fallback(peer, peer_store) else {
-            return false;
+            // Most peers that connect to us, that we have never seen, we will not know of their
+            // ENR. We should allow all incoming peers and then later reject them if
+            // they pose no use to us.
+            return true;
         };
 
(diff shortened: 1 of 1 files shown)


jxs (approved):
LGTM

dknopik:
> I think we need to not hard code the subnets in the handshake 

we do not, we update it as we subscribe to subnets

> and use that to disconnect peers

we use gossipsub subscription info instead, as it is more accurate. Else, peers could advertise being subscribed without actutally subscribing.

dknopik (approved):
LGTM

PR #662, opened 2025-10-08
fix: Prevent dialing connected peers

AgeManning:
## Issue Addressed

The logs were indicating that we were dialing already connected peers. 

This PR filters out already connected peers from the list of discovered peers. It improves the logging so that we can more easily identify our peer count, ie if we are just continually discovering already connected peers.

### DIFF
--- a/anchor/network/src/peer_manager/discovery.rs
+++ b/anchor/network/src/peer_manager/discovery.rs
@@ -67,10 +67,13 @@ impl PeerDiscovery {
     /// Track a subnet as needed and return actions to find peers for it
     pub fn track_subnet_peers(
         subnet_id: SubnetId,
+        needed_subnets: &mut HashSet<SubnetId>,
         peer_store: &MemoryStore<Enr>,
         connection_manager: &ConnectionManager,
         blocked_peers: &HashSet<PeerId>,
     ) -> ConnectActions {
+        needed_subnets.insert(subnet_id);
+
         Self::determine_actions_for_subnets(
             &[subnet_id],
             peer_store,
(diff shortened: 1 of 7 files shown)


dknopik:
IMO a better fix would be modifying `ConnectionManager::should_dial_peer`. Your current approach prevents updating the peer store with a potentially updated ENR.

anchor/network/src/peer_manager/discovery.rs
```diff
@@ -67,13 +67,10 @@ impl PeerDiscovery {
...
     pub fn track_subnet_peers(
         subnet_id: SubnetId,
-        needed_subnets: &mut HashSet<SubnetId>,
         peer_store: &MemoryStore<Enr>,
         connection_manager: &ConnectionManager,
         blocked_peers: &HashSet<PeerId>,
     ) -> ConnectActions {
-        needed_subnets.insert(subnet_id);
-
         Self::determine_actions_for_subnets(
```
dknopik, line 77: I think this is unrelated?

anchor/network/src/peer_manager/mod.rs
```diff
@@ -71,9 +71,10 @@ impl PeerManager {
 
     /// Join subnet and dial peers for it
     pub fn join_subnet(&mut self, subnet_id: SubnetId) -> ConnectActions {
+        self.needed_subnets.insert(subnet_id);
+
         PeerDiscovery::track_subnet_peers(
             subnet_id,
-            &mut self.needed_subnets,
             self.peer_store.store(),
```
dknopik: same

AgeManning:
Yeah it's unrelated. 

It's this commit:
> [Saw this change earlier, figured it's cleaner this way](https://github.com/sigp/anchor/pull/662/commits/c91f0b2b422784af4d506ad985f3410deb2f6e83)

I put this in intentionally because I just saw it and figured it didn't deserve its own PR. 

Can remove into its own PR if you prefer

dknopik:
Let's do it separately, I believe the way it is currently was intentional during @diegomrsantos' refactoring of the peer manager, so it might be worthwhile to discuss this

dknopik (approved):
LGTM, good find!

PR #663, opened 2025-10-08
feat: dynamically populate nodeinfo

jking-aus:
## Issue Addressed

Addresses  https://github.com/sigp/anchor/issues/645

## Proposed Changes

Gets required subnets from peermanager and updates the handshake
Gets node versions on startup and errors if not present

## Additional Info

Will add preferences for own subets in peer selection

### DIFF
--- a/anchor/client/src/lib.rs
+++ b/anchor/client/src/lib.rs
@@ -480,6 +480,7 @@ impl Client {
             outcome_rx,
             executor.clone(),
             spec.clone(),
+            database.watch(),
         )
         .await
         .map_err(|e| format!("Unable to start network: {e}"))?;
--- a/anchor/client/src/lib.rs
+++ b/anchor/client/src/lib.rs
@@ -356,6 +356,14 @@ impl Client {
         let index_sync_tx =
             start_validator_index_syncer(beacon_nodes.clone(), database.clone(), executor.clone());
 
+        // Fetch node versions from beacon and execution nodes before moving config fields
+        let (execution_version, consensus_version) = fetch_node_versions(
+            &beacon_nodes,
+            &config.execution_nodes,
+            &config.execution_nodes_websocket,
+        )
+        .await?;
+
         // We create the channel here so that we can pass the receiver to the syncer. But we need to
         // delay starting the voluntary exit processor until we have created the validator store.
         let (exit_tx, exit_rx) = unbounded_channel();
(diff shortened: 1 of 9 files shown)


anchor/client/src/lib.rs
```diff
@@ -776,6 +794,49 @@ async fn wait_for_genesis(genesis_time: u64) -> Result<(), String> {
     Ok(())
 }
 
+/// Fetches node version information from beacon and execution nodes
+async fn fetch_node_versions<T: SlotClock>(
+    beacon_nodes: &BeaconNodeFallback<T>,
+    execution_http_urls: &[SensitiveUrl],
+    execution_ws_url: &SensitiveUrl,
+) -> Result<(String, String), String> {
```
dknopik, line 802: IMO it is unclear if we actually want to do this. e.g. Go-SSV just leaves the node version in the handshake empty

anchor/network/src/metrics.rs
```diff
@@ -5,3 +5,25 @@ use metrics::*;
...
+    )
+});
+
+pub static HANDSHAKE_SUBNET_MATCHES: LazyLock<Result<IntGaugeVec>> = LazyLock::new(|| {
+    try_create_int_gauge_vec(
+        "libp2p_handshake_subnet_matches",
+        "Count of successful handshakes by number of matching subnets",
+        &["match_count"],
+    )
+});
```
dknopik, line 29: Metrics for the handshake is not really what this PR advertises to do, please open a separate PR

anchor/network/src/network.rs
```diff
@@ -106,15 +142,12 @@ impl<R: MessageReceiver> Network<R> {
...
-                node_version: version_with_platform(),
-                execution_node: "geth/v1.10.8".to_string(),
-                consensus_node: "lighthouse/v1.5.0".to_string(),
-                subnets: "00000000000000000000000000000000".to_string(),
-            }),
-        );
+
+        // Get initial subnets from peer_manager and populate metadata
+        let mut node_metadata = node_metadata;
+        node_metadata.subnets = subnets_to_hex(behaviour.peer_manager.needed_subnets());
```
dknopik, line 148: this will effectively set the same, all empty subnet string as before, as the `peer_manager` will have no needed subnets until the subnet tracker sends join events.

dknopik (commented):
I don't see how this mechanism of updating the subnets in our metadata (generating it anytime ti is needed) is superior to our current approach (updating it as we subscribe/unsubscribe to subnets)

jking-aus:
> I don't see how this mechanism of updating the subnets in our metadata (generating it anytime ti is needed) is superior to our current approach (updating it as we subscribe/unsubscribe to subnets)

It’s not - close this if that’s happening already.

PR #666, opened 2025-10-08
fix: only fetch the sync duties once per committee period

dknopik:
## Issue Addressed

Currently, we fetch the sync duties for all validators in the SSV network at the beginning of every slot.

This is needless, we only need to do that once per period and for added validators.

## Proposed Changes

Change our duty map to hold a boolean for every (period, validator) pair. Check if an entry already exists before sending the request, and only send a request for containing unknown validators.

## Additional Info

This is still not ideal - better would be some kind of notification system for added validators, so that we can avoid checking every slot if we already have the data. But better than asking the BN every slot.


petarjuki7 (approved):
Looks good to me, in spite of the PR number.

PR #668, opened 2025-10-08
chore: add handshake metrics

jking-aus:
## Issue Addressed

adding some handshake metrics -- on my machine I'm seeing a lot of outbound failures but they look like appropriate prioritisation by peers due to low subnet overlap

### DIFF
--- a/anchor/network/src/network.rs
+++ b/anchor/network/src/network.rs
@@ -46,6 +46,26 @@ use crate::{
 
 const MAX_TRANSMIT_SIZE_BYTES: usize = 5_000_000;
 
+/// Count the number of matching subnet bits between two hex-encoded subnet strings
+fn count_matching_subnets(our_subnets: &str, their_subnets: &str) -> usize {
+    // Decode both subnet strings
+    let our_bytes = match hex::decode(our_subnets) {
+        Ok(bytes) => bytes,
+        Err(_) => return 0,
+    };
+    let their_bytes = match hex::decode(their_subnets) {
+        Ok(bytes) => bytes,
+        Err(_) => return 0,
+    };
+
+    // Count matching bits using bitwise AND
+    our_bytes
+        .iter()
+        .zip(their_bytes.iter())
+        .map(|(a, b)| (a & b).count_ones() as usize)
+        .sum()
+}
+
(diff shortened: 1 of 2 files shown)


anchor/network/src/network.rs
```diff
@@ -551,11 +571,63 @@ impl<R: MessageReceiver> Network<R> {
...
+                        matching_subnets = matching_count,
+                        "Handshake completed"
+                    );
+
+                    // Record subnet match count: 0, or the actual number, or "5+" for 5 or more
+                    if let Ok(gauge_vec) = crate::metrics::HANDSHAKE_SUBNET_MATCHES.as_ref() {
+                        let label = if matching_count == 0 {
+                            "0"
+                        } else if matching_count >= 5 {
+                            "5+"
```
dknopik: why 5+ instead of just the actual count?

jking-aus: our canaries only have 4 subnets. We can do real count above 5 but it's not exactly a useful metric until we preference peers. It would also result in a lot of buckets if you have a lot of subnets to match on grafana visualisations. Idc though happy to just have count - it's a one line change.

dknopik: Ah makes sense. I see you already changed it, but I don't care strongly either way

anchor/network/src/network.rs
```diff
@@ -591,14 +591,12 @@ impl<R: MessageReceiver> Network<R> {
...
                     if let Ok(gauge_vec) = crate::metrics::HANDSHAKE_SUBNET_MATCHES.as_ref() {
                         let label = if matching_count == 0 {
                             "0"
-                        } else if matching_count >= 5 {
-                            "5+"
                         } else {
-                            // For 1-4, use the actual count
+                            // For 1+ use the actual count
                             &matching_count.to_string()
                         };
```
dknopik: Nitpick: No if/else is needed anymore, you can do just `let label = &matching_count.to_string()`

jking-aus: yeah but that was a two line change! I agreed to one! j/k fixed

dknopik (approved):
LGTM thanks!

Issue #652, opened 2025-10-07
Add version in the metrics


AgeManning:
As we develop the SSV Metrics, one handy feature is to know the version running. 

This allows us to identify from a dash which node is running which version. 

Lighthouse does it here for example: https://github.com/sigp/lighthouse/blob/stable/lighthouse/src/metrics.rs#L14

PR #670, opened 2025-10-08
chore: add client version in metrics

petarjuki7:
## Issue Addressed

Addresses Issue #652 

## Proposed Changes

No longer just save `Enr` in `PeerStore`, but a new `PeerInfo` struct which lets us hold its `Enr` and `ClientType` (`Anchor` or `go-ssv`)
Then, add metric to count how many `go-ssv` vs `Anchor` nodes are connected to us.

## Additional Info

I expose two fields in the `PeerManager` struct as `public` (`peer_store` and `connection_manager`).

### DIFF
--- a/anchor/network/src/network.rs
+++ b/anchor/network/src/network.rs
@@ -31,15 +31,16 @@ use types::{ChainSpec, EthSpec};
 use version::version_with_platform;
 
 use crate::{
-    Config, Enr,
+    ClientType, Config, Enr,
     behaviour::{AnchorBehaviour, AnchorBehaviourEvent, BehaviourError},
     discovery::{DiscoveredPeers, Discovery, DiscoveryError},
-    handshake,
-    handshake::node_info::{NodeInfo, NodeMetadata},
+    handshake::{
+        self,
+        node_info::{NodeInfo, NodeMetadata},
+    },
     keypair_utils::load_private_key,
     network::NetworkError::SwarmConfig,
-    peer_manager,
-    peer_manager::{ConnectActions, PeerManager},
+    peer_manager::{self, ConnectActions, PeerManager},
     scoring::topic_score_config::topic_score_params_for_subnet_with_rate,
     transport::build_transport,
 };
@@ -512,7 +513,42 @@ impl<R: MessageReceiver> Network<R> {
                 their_info,
             }) => {
                 debug!(%peer_id, ?their_info, "Handshake completed");
-                // Update peer store with their_info
+
+                if let Some(metadata) = their_info.metadata {
+                    let client_type = ClientType::from(metadata.node_version);
+
+                    let peer_info_opt = self
+                        .swarm
+                        .behaviour()
+                        .peer_manager
+                        .peer_store
+                        .store()
+                        .get_custom_data(&peer_id)
+                        .cloned();
+
+                    if let Some(mut peer_info) = peer_info_opt {
+                        // Update the client type
+                        peer_info.client_type = Some(client_type);
+
+                        // Insert back into peer store
+                        {
+                            let behaviour = self.swarm.behaviour_mut();
+                            behaviour
+                                .peer_manager
+                                .peer_store
+                                .store_mut()
+                                .insert_custom_data(&peer_id, peer_info);
+                        }
+
+                        // Trigger metric recalculation
+                        let behaviour = self.swarm.behaviour();
+                        let store_ref = behaviour.peer_manager.peer_store.store();
+                        behaviour
+                            .peer_manager
+                            .connection_manager
+                            .update_metrics_if_changed(true, Some(store_ref));
+                    }
+                }
             }
             Err(handshake::Failed { peer_id, error }) => {
                 debug!(%peer_id, ?error, "Handshake failed");
--- a/anchor/network/src/network.rs
+++ b/anchor/network/src/network.rs
@@ -31,7 +31,7 @@ use types::{ChainSpec, EthSpec};
 use version::version_with_platform;
 
 use crate::{
-    ClientType, Config, Enr, PeerInfo,
+    Config, Enr,
     behaviour::{AnchorBehaviour, AnchorBehaviourEvent, BehaviourError},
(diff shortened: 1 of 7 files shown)


AgeManning:
This is good. 

I think with #652 I was hoping to add a single endpoint so that we know the version of our current running anchor node so we can display on our dash. Would be nice to add that in too.

anchor/network/src/network.rs
```diff
@@ -552,7 +513,42 @@ impl<R: MessageReceiver> Network<R> {
...
+                    let peer_info_opt = self
+                        .swarm
+                        .behaviour()
+                        .peer_manager
+                        .peer_store
+                        .store()
+                        .get_custom_data(&peer_id)
+                        .cloned();
+
+                    if let Some(mut peer_info) = peer_info_opt {
```
dknopik: This approach causes us to save nothing if there is no peer info yet. If we get nothing from the store, we should create a new peer info

anchor/network/src/network.rs
```diff
@@ -552,7 +513,42 @@ impl<R: MessageReceiver> Network<R> {
...
+
+                if let Some(metadata) = their_info.metadata {
+                    let client_type = ClientType::from(metadata.node_version);
+
+                    let peer_info_opt = self
+                        .swarm
+                        .behaviour()
+                        .peer_manager
+                        .peer_store
+                        .store()
```
dknopik: create a variable for the store reference, so that we do not have to do this summoning ritual again below 😅

anchor/network/src/peer_manager/types.rs
```diff
@@ -27,3 +29,35 @@ pub enum Event {
...
+        if value.starts_with("Anchor/") {
+            Self::Anchor
+        } else {
+            Self::GoSSV
+        }
+    }
+}
+#[derive(Clone)]
+pub struct PeerInfo {
+    pub enr: Enr,
```
dknopik: Should be `Option<Enr>`

anchor/network/src/peer_manager/connection.rs
```diff
@@ -246,69 +232,68 @@ impl ConnectionManager {
...
-        self.outbound_count
+        self.connected.remove(peer_id)
     }
 
     /// Update metrics if connection state changed
-    pub fn update_metrics_if_changed(&self, changed: bool) {
+    pub fn update_metrics_if_changed(
+        &self,
+        changed: bool,
+        peer_store: Option<&MemoryStore<PeerInfo>>,
```
dknopik: Why is this an `Option`?

dknopik (commented):
Nice!

anchor/network/src/peer_manager/types.rs
petarjuki7: Isn't the first time where we create PeerInfo (in `process_discover_peer()`) we are sure of the enr so we make a new insert into `PeerStore`? Or a better way, I think we are sure that `on_discover_peers()` happens before the handshake event?
Relates also to the comment above.

anchor/network/src/peer_manager/connection.rs
petarjuki7: Was trying something else and forgot to remove, fixed

anchor/network/src/peer_manager/types.rs
dknopik: If someone discovers us before we discover them, they may connect and successfully handshake. We have no ENR in that case.

anchor/network/src/network.rs
petarjuki7: Haha yeah it's a bit wierd, but I'm fighting with the borrowing rules a bit, because I need a shared reference, then an exclusive reference then a new shared reference so I think I need a new (shared) reference each time?

anchor/network/src/peer_manager/types.rs
petarjuki7: ah okay, makes sense

anchor/network/src/network.rs
dknopik: Is something preventing you from just getting an exclusive reference in the beginning?

anchor/network/src/peer_manager/discovery.rs
```diff
@@ -49,13 +49,18 @@ impl PeerDiscovery {
...
         }
-        peer_store.insert_custom_data(
-            &id,
-            PeerInfo {
-                enr,
-                client_type: None,
-            },
-        );
+
+        if let Some(potential) = peer_store.get_custom_data_mut(&id) {
```
dknopik: Nitpick: `potiential` -> `peer_info`

anchor/network/src/peer_manager/discovery.rs
```diff
@@ -114,7 +119,11 @@ impl PeerDiscovery {
                 continue;
             };
 
-            let subnets = discovery::committee_bitfield(&peer_info.enr).unwrap_or_default();
+            let subnets = peer_info
+                .enr
+                .as_ref()
+                .and_then(|enr| discovery::committee_bitfield(enr).ok())
+                .unwrap_or_default();
```
dknopik: If we have no ENR, we should hit the `continue` above. Both approaches work, but IMO the intent is clearer: if we have no ENR, we do not consider the peer.

dknopik (commented):
Two nitpicks:

anchor/network/src/network.rs
```diff
@@ -576,30 +577,63 @@ impl<R: MessageReceiver> Network<R> {
...
+                        debug!(%peer_id, "Handshake completed");
                     }
+
+                    // Trigger metric recalculation
+                    let behaviour = self.swarm.behaviour();
+                    let store_ref = behaviour.peer_manager.peer_store.store();
+                    behaviour
+                        .peer_manager
+                        .connection_manager
+                        .update_metrics_if_changed(true, store_ref);
```
dknopik: Actually, what do you think about moving everything inside this `if` into the peer manager and/or the connection manager within it? This would allow us to avoid making the `peer_store` and `conncection_manager` `pub`, for better encapsulation

dknopik (approved):
LGTM, thank you!

PR #673, opened 2025-10-09
feat: metric for own version and startup timestamp

dknopik:
## Issue Addressed

- #652 

## Proposed Changes

Steal some code from LH for version exposure via metrics. While I am at it I also took some code for startup timestamp, because why not.

### DIFF
--- a/anchor/client/src/metrics.rs
+++ b/anchor/client/src/metrics.rs
@@ -1,0 +1,37 @@
+use std::{
+    sync::LazyLock,
+    time::{SystemTime, UNIX_EPOCH},
+};
+
+use metrics::*;
+use tracing::error;
+use version::VERSION;
+
+pub static PROCESS_START_TIME_SECONDS: LazyLock<Result<IntGauge>> = LazyLock::new(|| {
+    try_create_int_gauge(
+        "process_start_time_seconds",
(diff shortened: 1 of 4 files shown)


anchor/client/src/metrics.rs
```diff
@@ -0,0 +1,37 @@
...
+pub static ANCHOR_VERSION: LazyLock<Result<IntGaugeVec>> = LazyLock::new(|| {
+    try_create_int_gauge_vec(
+        "anchor_info",
+        "The build of Anchor running on the server",
+        &["version"],
+    )
+});
+
+pub fn expose_process_start_time() {
+    match SystemTime::now().duration_since(UNIX_EPOCH) {
```
petarjuki7, line 26: So we just expose the process start time (Unix timestamp) and then Prometheus / Grafana calculates the actual uptime / other metrics?

petarjuki7 (approved):

anchor/client/src/metrics.rs
dknopik: I assume grafana has some way to integrate current time into queries, so yeah! 

As mentioned, this is how LH does it, so I assume it is possible to do something useful with this :D

PR #674, opened 2025-10-09
feat: add panic hook for logging

dknopik:
## Issue Addressed

When going through LH's and Anchor's startup procedures, I noticed that Anchor does not have a panic hook, causing panics to only be printed on stderr, but not in logs. This is bad in cases where the user only retains logs, but not stderr.

## Proposed Changes

Steal panic hook code from LH, logging the panic properly.


jking-aus (approved):
lgtm!

PR #692, opened 2025-10-15
feat: operator doppelgänger protection with slot-based detection

### LINKED ISSUE #627: Operator Doppelgänger Protection
A operator doppelgänger is the duplicate use of a single operator key in multiple SSV client instances.

This causes equivocation in QBFT instances, which will be punished by message validation and can cause general mayhem in the P2P layer.

In the worst case, when there are multiple doppelgängers in a committee, this can even lead to slashing of the validators associated with it.

We should introduce a feature that lets Anchor listen to the P2P layer for a configurable while (e.g. for one epoch) before commencing duties, and quit if it finds a signature that originates from its own private key. 

This can e.g. implemented by `await`-ing something during startup, or by extending the `is_synced` mechanism: https://github.com/sigp/anchor/blob/544dece7b9b54beef82a6dc98faec98d4b464b4f/anchor/client/src/lib.rs#L560-L566

diegomrsantos:
## Issue Addressed

Closes #627

## Proposed Changes

Implements **Operator Doppelgänger Protection** to detect when multiple instances of the same operator are running, preventing QBFT protocol violations.

### Problem

If two operator instances run simultaneously:
- Both participate in QBFT consensus  
- Both send messages with the same operator ID
- QBFT receives conflicting votes from one operator ID
- Protocol correctness breaks (QBFT expects one vote per operator per round)

### Solution

**Slot-Based Detection**: Records the startup slot and monitors the network for messages with our operator ID that reference slots after startup.

| Phase      | Outgoing Messages | Detection Logic                                          |
|------------|-------------------|----------------------------------------------------------|
| Monitoring | ❌ BLOCKED         | Messages with slot > startup_slot → Twin detected        |
| Completed  | ✅ SENT            | No longer checking                                       |

**Why Slot Comparison Works:**
- All outgoing messages blocked during monitoring
- Messages with `slot ≤ startup_slot`: Ignored (our own old messages)
- Messages with `slot > startup_slot`: Twin detected (we didn't send them)
- No race conditions possible since we never compete with twins

**State Management:**
- Simple AtomicBool for lock-free monitoring state
- No grace period needed (slot comparison handles message filtering)
- Detection active immediately on startup

**Twin Detection:**
- Monitors single-signer QBFT and partial signature messages
- Twin detected → logs detailed context → graceful shutdown
- Handles edge cases: restart in same slot, network delays, clock skew

### Benefits Over Time-Based Approaches

- **Faster Startup**: No 411-second grace period wait
- **More Reliable**: Slot comparison is deterministic, unaffected by network delays
- **Simpler**: Fewer states, cleaner logic with lock-free operations

### Configuration

This feature is **opt-in** to avoid imposing a monitoring delay on all operators:

- `--operator-dg`: Enable protection (default: false)  
- `--operator-dg-wait-epochs`: Monitoring duration in epochs (default: 2, ~768 seconds)

**Rationale for Opt-In:**
- QBFT's Byzantine tolerance already handles some operator duplication scenarios
- No clear path from operator duplication to validator slashing (operators don't directly sign beacon chain messages)
- Provides operational discipline rather than security guarantees
- ~12 minute startup delay may not be acceptable for all deployment scenarios
- Operators can enable when operational safety is prioritized

### Testing

8 comprehensive tests covering slot-based detection, state management, and edge cases. Manual testing verified all configuration scenarios work correctly.


anchor/client/src/lib.rs
```diff
@@ -466,6 +470,57 @@ impl Client {
 
         let (outcome_tx, outcome_rx) = mpsc::channel::<message_receiver::Outcome>(9000);
 
+        // Initialize operator doppelgänger protection if enabled
+        let doppelganger_config = if config.operator_dg && config.impostor.is_none() {
+            let own_operator_id = operator_id
+                .get()
+                .ok_or_else(|| "Operator ID not yet available".to_string())?;
```
dknopik: This will crash all nodes on first startup, as they are not synced yet

anchor/client/src/cli.rs
```diff
@@ -484,6 +484,43 @@ pub struct Node {
...
+        long,
+        value_name = "HEIGHTS",
+        help = "The freshness threshold for detecting operator twins. Only messages within \
+                this many consensus heights from the maximum observed height are considered \
+                fresh evidence of a twin. This prevents false positives from replayed old messages.",
+        display_order = 0,
+        default_value_t = 3,
+        requires = "operator_dg"
+    )]
+    pub operator_dg_fresh_k: u64,
```
dknopik: I do not think this is necessary. Have you tested this and confirmed there to be an issue without this feature?

anchor/client/src/lib.rs
```diff
@@ -466,6 +470,57 @@ impl Client {
...
+
+            let doppelganger_service = Arc::new(OperatorDoppelgangerService::<E, _>::new(
+                own_operator_id,
+                slot_clock.clone(),
+                config.operator_dg_wait_epochs,
+                config.operator_dg_fresh_k,
+                true, // enabled
```
dknopik: If we only ever pass `true` here, we might as well remove the parameter

anchor/message_receiver/src/manager.rs
```diff
@@ -161,6 +178,26 @@ impl<S: SlotClock + 'static, D: DutiesProvider> MessageReceiver
...
+
+                            // Trigger shutdown - we'll only do this once
+                            if let Ok(mut guard) = config.shutdown_tx.lock()
+                                && let Some(tx) = guard.take()
+                            {
+                                let _ = tx.send(());
+                            }
+
+                            return;
+                        }
```
dknopik: what about partial signature messages?

anchor/client/src/cli.rs
```diff
@@ -484,6 +484,43 @@ pub struct Node {
...
+        long,
+        value_name = "EPOCHS",
+        help = "Number of epochs to wait in monitor mode before starting normal operation. \
+                During this period, the node listens for messages from its own operator ID \
+                to detect if another instance is running.",
+        display_order = 0,
+        default_value_t = 2,
+        requires = "operator_dg"
+    )]
+    pub operator_dg_wait_epochs: u64,
```
dknopik, line 518: we are not actually waiting anywhere

diegomrsantos:
It's still a heavily drafted PR; it is not ready for review

anchor/client/src/cli.rs
diegomrsantos: Could you elaborate more on why you think it's not necessary?

anchor/message_receiver/src/manager.rs
diegomrsantos: I'm wondering if it's really necessary, as we need to send a qbft message first, don't we?

anchor/operator_doppelganger/src/service.rs
```diff
@@ -0,0 +1,531 @@
...
+        // Only check in monitor mode (background task handles transition)
+        if !state.is_monitoring() {
+            return false;
+        }
+
+        // Extract committee ID from message
+        let committee_id = match signed_message.ssv_message().msg_id().duty_executor() {
+            Some(DutyExecutor::Committee(committee_id)) => committee_id,
+            _ => return false, // Not a committee message
+        };
```
dknopik: how is the committee_id relevant at all?

anchor/operator_doppelganger/src/service.rs
```diff
@@ -0,0 +1,531 @@
...
+                operator_id = *self.own_operator_id,
+                committee = ?committee_id,
+                height = qbft_message.height,
+                "Received stale message with our operator ID (likely replay), ignoring"
+            );
+            return false;
+        }
+
+        // Update height tracking for this fresh message
+        state.update_max_height(committee_id, qbft_message.height);
```
dknopik: This is only called when we are about to go down

anchor/operator_doppelganger/src/state.rs
```diff
@@ -0,0 +1,151 @@
...
+    Active,
+}
+
+/// State for operator doppelgänger detection
+#[derive(Debug, Clone)]
+pub struct DoppelgangerState {
+    /// Current operating mode
+    mode: DoppelgangerMode,
+    /// Maximum consensus height observed per committee
+    recent_max_height: HashMap<CommitteeId, u64>,
```
dknopik: Why bother with this? Height is slot number, we can just check against the slot clock to check "freshness"

anchor/client/src/cli.rs
```diff
@@ -484,6 +484,43 @@ pub struct Node {
...
+
+    #[clap(
+        long,
+        value_name = "EPOCHS",
+        help = "Number of epochs to wait in monitor mode before starting normal operation. \
+                During this period, the node listens for messages from its own operator ID \
+                to detect if another instance is running.",
+        display_order = 0,
+        default_value_t = 2,
+        requires = "operator_dg"
```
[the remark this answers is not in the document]
dknopik: to claude: `requires` does not prevent an option from being listed

anchor/operator_doppelganger/src/service.rs
```diff
@@ -0,0 +1,531 @@
...
+                    tokio::time::sleep(self.slot_duration).await;
+
+                    if let Some(slot) = self.slot_clock.now() {
+                        let current_epoch = slot.epoch(E::slots_per_epoch());
+                        if current_epoch >= self.monitor_end_epoch {
+                            self.transition_to_active();
+                            break; // Done monitoring
+                        }
+                    }
+                }
```
dknopik: why not just sleep until the target epoch starts?

I think we do not even need a service for this, a function called by the startup code suffices.

anchor/operator_doppelganger/src/state.rs
```diff
@@ -0,0 +1,151 @@
...
+use ssv_types::CommitteeId;
+
+/// Operating mode for doppelgänger protection
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum DoppelgangerMode {
+    /// Monitor mode: listen for messages with our operator ID
+    Monitor,
+    /// Active mode: normal operation
+    Active,
+}
```
dknopik: I do not think we need separate modes.

anchor/client/src/lib.rs
```diff
@@ -83,6 +87,90 @@ const HTTP_DEFAULT_TIMEOUT_QUOTIENT: u32 = 4;
...
+    ),
+    String,
+> {
+    if !operator_dg {
+        return Ok((None, None));
+    }
+
+    let own_operator_id = operator_id
+        .get()
+        .ok_or_else(|| "Operator ID not yet available".to_string())?;
```
dknopik: As previously noted: The code needs to be able to cope with not having an operator ID yet. Even if sync completes, crashing here is unacceptable.

anchor/client/src/lib.rs
```diff
@@ -83,6 +87,90 @@ const HTTP_DEFAULT_TIMEOUT_QUOTIENT: u32 = 4;
...
+                // Trigger executor shutdown with failure reason
+                let _ = executor_clone
+                    .shutdown_sender()
+                    .try_send(task_executor::ShutdownReason::Failure(
+                        "Operator doppelgänger detected",
+                    ));
+            }
+        },
+        "doppelganger-shutdown",
+    );
```
dknopik: This thing simply waits on one channel to post to another channel. Is something preventing us from using the `shutdown_sender` directly?

anchor/message_receiver/src/manager.rs
dknopik: There are duties where there is signature collection before the qbft instance (Proposals) and even duties where there is no qbft instance at all (validator registration, validator exit).

So focussing on Qbft message feels arbitrary and has no advantage at all.

anchor/client/src/cli.rs
dknopik: I understand now that this is about preventing replay attacks, not necessarily about this occurring naturally.

dknopik (commented):
I feel like a lot of this is needlessly complex

anchor/operator_doppelganger/src/service.rs
diegomrsantos: Can't an operator be part of many committees?

anchor/operator_doppelganger/src/state.rs
diegomrsantos: I recommend checking the proposal https://github.com/sigp/anchor/issues/627#issuecomment-3406787984. In the future, please let's clarify questions before the implementation starts.

anchor/client/src/lib.rs
diegomrsantos: not relevant anymore

anchor/client/src/lib.rs
diegomrsantos: done

anchor/client/src/lib.rs
```diff
@@ -466,6 +554,17 @@ impl Client {
...
+        // Initialize operator doppelgänger protection if enabled
+        let (doppelganger_config, is_monitoring) = initialize_operator_doppelganger::<E>(
+            config.operator_dg && config.impostor.is_none(),
+            config.operator_dg_wait_epochs,
+            config.operator_dg_fresh_k,
+            &operator_id,
+            &slot_clock,
+            Duration::from_secs(spec.seconds_per_slot),
+            &executor,
+        )?;
```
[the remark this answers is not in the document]
diegomrsantos: There's a new design

anchor/client/src/lib.rs
diegomrsantos: not relevant anymore

anchor/client/src/cli.rs
diegomrsantos: This isn't about replay attacks. What makes you think so?

anchor/client/src/lib.rs
diegomrsantos: There's a new design, let me know what you think

anchor/message_receiver/src/manager.rs
diegomrsantos: I'm using partoal sig msgs now

anchor/operator_doppelganger/src/state.rs
diegomrsantos: not relevant anymore

anchor/operator_doppelganger/src/service.rs
diegomrsantos: not relevant anymore

anchor/operator_doppelganger/src/service.rs
diegomrsantos: not relevant anymore

anchor/message_receiver/src/manager.rs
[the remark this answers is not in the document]
diegomrsantos: not relevant anymore

anchor/operator_doppelganger/src/state.rs
diegomrsantos: removed

anchor/operator_doppelganger/src/service.rs
diegomrsantos: Applied first suggestion

diegomrsantos:
> I feel like a lot of this is needlessly complex

It has been simplified substantially

anchor/operator_doppelganger/src/service.rs
```diff
@@ -0,0 +1,531 @@
...
+    /// Check if a message indicates a potential doppelgänger
+    ///
+    /// Returns true if a twin is detected (should trigger shutdown)
+    #[must_use]
+    pub fn check_message(
+        &self,
+        signed_message: &SignedSSVMessage,
+        qbft_message: &QbftMessage,
+    ) -> bool {
+        let mut state = self.state.lock();
```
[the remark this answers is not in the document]
diegomrsantos: done

anchor/operator_doppelganger/src/service.rs
```diff
@@ -0,0 +1,535 @@
...
+    /// This method performs pure detection logic without side effects (except logging).
+    ///
+    /// Checks all single-signer messages (QBFT consensus and partial signatures) signed
+    /// with our operator ID.
+    pub fn is_doppelganger(
+        &self,
+        signed_message: &SignedSSVMessage,
+        qbft_message: Option<&QbftMessage>,
+    ) -> bool {
+        let state = self.state.lock();
```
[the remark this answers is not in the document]
diegomrsantos: done

anchor/client/src/cli.rs
dknopik: > What makes you think so?

The sentence "This prevents false positives from replayed old messages."

If another node maliciously holds onto one of our messages, they can use it to crash us once we do not have it in the duplicate cache anymore (due to cache timeout or node restart).

dknopik:
Btw, thinking about it more, doesn't doing it after validation possibly mask detection of some messages? e.g. if we and a doppelganger both send a qbft propose message, validation will only consider our message as valid, as we only accept one propose per round. So I think we only need to make sure that the signature is valid and that the message is fresh for doppelganger protection

anchor/client/src/cli.rs
diegomrsantos: > > What makes you think so?
> 
> The sentence "This prevents false positives from replayed old messages."
> 
> If another node maliciously holds onto one of our messages, they can use it to crash us once we do not have it in the duplicate cache anymore (due to cache timeout or node restart).

The sentence is misleading; the idea is a protection from a protocol behavior. I thought about this other malicious case, but I don't know a solution to it yet. Not only is our gossipsub seen cache empty, but our message counts in validation as well.

anchor/message_sender/src/network.rs
```diff
@@ -39,6 +53,14 @@ impl<S: SlotClock + 'static, D: DutiesProvider> MessageSender for Arc<NetworkMes
         committee_id: CommitteeId,
         additional_message_callback: Option<Box<MessageCallback>>,
     ) -> Result<(), Error> {
+        // Check if in doppelgänger monitoring period - silently drop
+        if let Some(dg) = &self.doppelganger_service
+            && dg.is_monitoring()
```
diegomrsantos: > https://github.com/sigp/anchor/pull/692#issuecomment-3436215337

@dknopik reallly good point. Made me realize this line is wrong, we shouldn't send any message for the whole dg process (including the grace period). Changing this line should avoid the issue you mentioned, right?

diegomrsantos:
@dknopik can we get this tested on `hoodi`? I'd be good to let it run for a few epochs and then restart it from time to time and see if there'll be false positives.

anchor/client/src/cli.rs
```diff
@@ -484,6 +484,31 @@ pub struct Node {
     #[clap(long, help = "Disables gossipsub topic scoring.", hide = true)]
     pub disable_gossipsub_topic_scoring: bool,
 
+    // Operator Doppelgänger Protection
+    #[clap(
+        long,
+        help = "Enable operator doppelgänger protection. When enabled, the node will monitor \
+                for messages signed by its operator ID on startup and shut down if a twin \
+                (duplicate operator) is detected. Enabled by default.",
```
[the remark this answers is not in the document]
diegomrsantos: changed it with sth else

diegomrsantos:
@claude are you sure "Critical Issue: Validation Masking Detection" is an issue? We won't send any message and we won't check for a twin during the grace period. Also we won't send any message for the whole time the doppelganger protection is active

anchor/message_sender/src/network.rs
dknopik: @diegomrsantos not fully. Yes, it solves my example, but the underlying issue still persists: we do not catch messages sent by a doppelgänger that we consider to be invalid. this can also occur due to other issues.

diegomrsantos: Could you please elaborate more? I still don't get it

dknopik:
An idea I just had: I think we can get rid of the grace period if we only consider messages created after we start. So basically, we get the current slot at the beginning, and then only crash if we get a message referencing a later slot

diegomrsantos:
> An idea I just had: I think we can get rid of the grace period if we only consider messages created after we start. So basically, we get the current slot at the beginning, and then only crash if we get a message referencing a later slot

That's actually a great idea! I don't know why we didn't think about this before 😅

anchor/message_receiver/src/manager.rs
```diff
@@ -159,6 +164,21 @@ impl<S: SlotClock + 'static, D: DutiesProvider> MessageReceiver
...
+                // Check for operator doppelgänger before processing any message
+                if let Some(service) = &receiver.doppelganger_service {
+                    // If in monitoring mode, check for twin and drop message
+                    if service.is_monitoring() {
+                        // Extract QBFT message for detailed logging if twin detected
+                        let qbft_msg = match &ssv_message {
+                            ValidatedSSVMessage::QbftMessage(msg) => Some(msg),
+                            ValidatedSSVMessage::PartialSignatureMessages(_) => None,
+                        };
+                        service.check_message(&signed_ssv_message, qbft_msg);
```
petarjuki7: Could you also send just the `signed_ssv_message` because you are sure you can extract the slot from it? Or are we sure most of the messages will be QBFT so it's more optimised like this?

diegomrsantos: Your comment made me realize the API was overly complicated, so I simplified it. Let me know what you think. Thanks for the review.

anchor/client/src/lib.rs
```diff
@@ -419,15 +424,38 @@ impl Client {
...
         );
 
+        // Create operator doppelgänger protection if enabled (will be started after sync)
+        let doppelganger_service = if config.operator_dg && config.impostor.is_none() {
+            // Get current slot for slot-based detection baseline
+            let current_slot = slot_clock.now().ok_or_else(|| {
+                "Failed to get current slot for doppelgänger protection".to_string()
+            })?;
+            // Convert types::Slot to ssv_types::Slot
+            let startup_slot = ssv_types::Slot::new(current_slot.as_u64());
```
dknopik: `ssv_types::Slot` is just a reexport, no conversion is necessary

anchor/client/src/notifier.rs
```diff
@@ -71,16 +125,51 @@ async fn notify<E: EthSpec, T: SlotClock + 'static>(
...
+            count,
+        ) if count > 0 => {
+            let operator_id = operator_id.expect("operator_id is Some in this branch");
             info!(%operator_id, cluster_count, "Operator active");
+            // Only call Lighthouse's notifier when we're actually performing duties
             validator_services::notifier_service::notify(duties_service).await;
         }
-        (Some(operator_id), true) => info!(%operator_id, "Operator ready, no validators assigned"),
+        (SyncState::Synced, OperatorState::OperatorPresent { .. }, _, _) => {
+            let operator_id = operator_id.expect("operator_id is Some in this branch");
```
dknopik: lets store the operator_id in `OperatorState::OperatorPresent` to avoid this `expect`.

diegomrsantos: done

anchor/client/src/lib.rs
diegomrsantos: done

anchor/client/src/cli.rs
diegomrsantos: it's a bit weird that when we use this, we need to set `--operator-dg=true`

anchor/src/main.rs
```diff
@@ -205,9 +205,15 @@ fn start_anchor(anchor_config: &Node, global_config: GlobalConfig, mut environme
...
-            error!(reason = msg.to_string(), "Failed to shutdown gracefully");
+        ShutdownReason::Success(_) => {
+            info!("Shutdown complete");
+            // Exit code 0 (success)
+        }
+        ShutdownReason::Failure(_) => {
+            info!("Shutdown complete");
+            // Exit with code 1 to indicate failure to process monitors, systemd, orchestrators,
+            // etc.
+            std::process::exit(1);
```
diegomrsantos: @dknopik this is pretty important, just letting you know this has been done in this PR

dknopik: can we keep logging the message contained in `Failure`?

diegomrsantos: This has already been logged in line 200

dknopik: ah, missed that, thanks

dknopik: from the documentation of `std::process:exit`: 

> Note that because this function never returns, and that it terminates the process, no destructors on the current stack or any other thread’s stack will be run. If a clean shutdown is needed it is recommended to only call this function at a known point where there are no more destructors left to run; or, preferably, simply return a type implementing [Termination](https://doc.rust-lang.org/std/process/trait.Termination.html) (such as [ExitCode](https://doc.rust-lang.org/std/process/struct.ExitCode.html) or Result) from the main function and avoid this function altogether

We hold a few guards whose `drop` code should be run, such as the lockfile and logfile buffers, so can we return `ExitCode` from `main` instead?

diegomrsantos: oh I didn't realize it would actually exit right away

diegomrsantos: But I think there's nothing left to do at this point, is there? I'll change it anyway

diegomrsantos: extracted it to https://github.com/sigp/anchor/pull/727

dknopik: > But I think there's nothing left to do at this point, is there? I'll change it anyway

We are holding a bunch of `WorkerGuard` in `main`. From it's documentation:

```rust
/// Writing to a [`NonBlocking`] writer will **not** immediately write a span or event to the underlying
/// output. Instead, the span or event will be written by a dedicated logging thread at some later point.
/// To increase throughput, the non-blocking writer will flush to the underlying output on
/// a periodic basis rather than every time a span or event is written. This means that if the program
/// terminates abruptly (such as through an uncaught `panic` or a `std::process::exit`), some spans
/// or events may not be written.
```

dknopik (approved):
LGTM, thanks!

PR #693, opened 2025-10-16
chore: improve claude workflow

jking-aus:
Improving the code actions workflow: 

- giving claude more turns 
- adding a recheck flag
- adding a dont-review flag
- adding some more basic tools like read and grep


jking-aus:
this is why I need a dont review flag sir

dknopik (approved):
LGTM

Issue #628, opened 2025-09-23
Majority Fork Protection


dknopik:
see https://github.com/ssvlabs/ssv-spec/issues/555

This would be implemented in Anchor by modifying the `BeaconVoteValidator` to compare the `value` with `our_value`.

PR #695, opened 2025-10-16
feat: Majority Fork Protection

dknopik:
## Issue Addressed

- #628 

## Proposed Changes

In the `BeaconVoteValidator`, which validates `BeaconVote`s proposed by other nodes, add a step to compare the source, and a step to compare the target, including epoch and checkpoing for each.

## Additional Info

Implements https://github.com/ssvlabs/ssv-spec/issues/555.

See also the Go-SSV implementation: https://github.com/ssvlabs/ssv/pull/2358

### DIFF
--- a/anchor/common/ssv_types/src/consensus.rs
+++ b/anchor/common/ssv_types/src/consensus.rs
@@ -758,3 +758,223 @@ pub enum BeaconVoteValidationError {
     #[error("Attestation would be slashable: {0}")]
     SlashableAttestation(NotSafe),
 }
+
+#[cfg(test)]
+mod tests {
+    use std::collections::HashMap;
+
+    use types::{Checkpoint, Epoch, FixedBytesExtended, MainnetEthSpec};
+
+    use super::*;
+
+    /// Helper function to create a BeaconVoteValidator for testing.
+    /// This validator has slashing protection disabled for simpler testing.
+    fn create_test_validator() -> BeaconVoteValidator<MainnetEthSpec> {
+        let spec = Arc::new(ChainSpec::mainnet());
+        let validator_attestation_committees = HashMap::new();
+        let genesis_validators_root = Hash256::zero();
+        let slot = Slot::new(100);
+
+        BeaconVoteValidator::new(
+            slot,
+            None,
+            spec,
+            validator_attestation_committees,
+            genesis_validators_root,
+        )
+    }
+
+    #[test]
+    fn test_mismatched_source_different_epochs() {
+        let validator = create_test_validator();
+
+        let our_source = Checkpoint {
+            epoch: Epoch::new(2),
+            root: Hash256::from_low_u64_be(1),
+        };
+        let our_target = Checkpoint {
+            epoch: Epoch::new(3),
+            root: Hash256::from_low_u64_be(2),
+        };
+        let our_vote = BeaconVote {
+            block_root: Hash256::random(),
+            source: our_source,
+            target: our_target,
+        };
+
+        // Create a proposed vote with different source epoch
+        let proposed_source = Checkpoint {
+            epoch: Epoch::new(1), // Different epoch
+            root: Hash256::from_low_u64_be(1),
+        };
+        let proposed_vote = BeaconVote {
+            block_root: Hash256::random(),
+            source: proposed_source,
+            target: our_target,
+        };
+
+        let result = validator.do_validation(&proposed_vote, &our_vote);
+        assert!(result.is_err());
+        match result.unwrap_err() {
+            BeaconVoteValidationError::DifferentSource { our, proposed } => {
+                assert_eq!(our, our_source);
+                assert_eq!(proposed, proposed_source);
+            }
+            err => panic!("Expected DifferentSource error, got: {:?}", err),
+        }
+    }
+
+    #[test]
+    fn test_mismatched_source_same_epoch_different_roots() {
+        let validator = create_test_validator();
+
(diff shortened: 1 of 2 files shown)


anchor/common/ssv_types/src/consensus.rs
```diff
@@ -679,6 +679,19 @@ impl<E: EthSpec> BeaconVoteValidator<E> {
...
+            return Err(BeaconVoteValidationError::DifferentSource {
+                our: our_value.source,
+                proposed: value.source,
+            });
+        }
+        if value.target != our_value.target {
+            return Err(BeaconVoteValidationError::DifferentTarget {
+                our: our_value.source,
+                proposed: value.source,
+            });
```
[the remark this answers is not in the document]
dknopik: oops. Of course this was intentional to test claude :upside_down_face::upside_down_face::upside_down_face:

anchor/common/ssv_types/src/consensus.rs
```diff
@@ -679,6 +679,19 @@ impl<E: EthSpec> BeaconVoteValidator<E> {
             )));
         }
 
+        if value.source != our_value.source {
+            return Err(BeaconVoteValidationError::DifferentSource {
+                our: our_value.source,
+                proposed: value.source,
+            });
+        }
```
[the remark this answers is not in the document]
dknopik: added a better comment

diegomrsantos:
@claude review this PR

diegomrsantos:
@claude review this PR

anchor/common/ssv_types/src/consensus.rs
```diff
@@ -679,6 +679,19 @@ impl<E: EthSpec> BeaconVoteValidator<E> {
...
+        if value.target != our_value.target {
+            return Err(BeaconVoteValidationError::DifferentTarget {
+                our: our_value.source,
+                proposed: value.source,
+            });
+        }
```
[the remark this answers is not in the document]
dknopik: done

petarjuki7 (dismissed):
makes sense, looks good to me!

anchor/common/ssv_types/src/consensus.rs
```diff
@@ -615,8 +615,8 @@ impl QbftData for BeaconVote {
 
 pub struct BeaconVoteValidator<E: EthSpec> {
     slot: Slot,
-    slashing_database: Arc<SlashingDatabase>,
-    disable_slashing_protection: bool,
+    // `None` if slashing protection is disabled via CLI.
+    slashing_database: Option<Arc<SlashingDatabase>>,
```
diegomrsantos, line 619: Is this necessary for the implementation? If not, what do you think about reverting it?

anchor/common/ssv_types/src/consensus.rs
```diff
@@ -679,10 +677,25 @@ impl<E: EthSpec> BeaconVoteValidator<E> {
...
+            return Err(BeaconVoteValidationError::DifferentSource {
+                our: our_value.source,
+                proposed: value.source,
+            });
         }
+        if value.target != our_value.target {
+            return Err(BeaconVoteValidationError::DifferentTarget {
+                our: our_value.target,
+                proposed: value.target,
+            });
```
diegomrsantos: nit: what do you think about always validating both and returning source and target in the error if they diverge?

anchor/common/ssv_types/src/consensus.rs
dknopik: While it is not strictly necessary, it made implementing the tests more ergonomic.

I decided to switch to an option here as it is in fact optional: we do not need it at all if slashing protection is disable.

I can revert it, but then we need to create an in-memory database for the tests below. What do you prefer?

diegomrsantos: Sorry, didn't realize the test part. So it makes sense

diegomrsantos: It's even better than before imo

anchor/common/ssv_types/src/consensus.rs
dknopik: Good idea, done. wdyt?

diegomrsantos: Looks good, thanks

anchor/common/ssv_types/src/consensus.rs
```diff
@@ -679,18 +677,35 @@ impl<E: EthSpec> BeaconVoteValidator<E> {
...
+        self.check_attestation_slashing(value)?;
+
         Ok(())
     }
 
     fn check_attestation_slashing(
         &self,
         value: &BeaconVote,
     ) -> Result<(), BeaconVoteValidationError> {
+        let Some(slashing_database) = &self.slashing_database else {
```
diegomrsantos, line 705: nit: How about making this check before calling the function?

diegomrsantos (approved):
LGTM

anchor/common/ssv_types/src/consensus.rs
dknopik: Personally, I prefer it this way, because otherwise we have to pass the `slashing_database` into this function which feels redundant and obscures the origin of the value for the reader of this function. But I am not feeling strongly, so if you still prefer it this way, I can change it :)

diegomrsantos: Ah, ok, I see what you mean. It's not that important.

PR #697, opened 2025-10-17
fix: Re-add node version to Handshake completed

dknopik:
## Issue Addressed

In `rc.1`, we (accidentally?) removed logging the node version after a completed handshake.

## Proposed Changes

This info is very useful for debugging, e.g. if we see repeated invalid messages from a certain peer and we want to know what software sent those messages. This PR readds the field.


jking-aus (approved):
yep

PR #698, opened 2025-10-17
chore: change default network to Mainnet

dknopik:
## Issue Addressed

We decided undo the default network change previously, rolling back to Hoodi. 

## Proposed Changes

As we are now releasing the actual `v1.0.0`, we change the default network to Mainnet, mirroring many Ethereum clients as well as go-ssv.


jking-aus (approved):
lgtm

Issue #702, opened 2025-10-20
Add CLI arg to set target peer count


dknopik:
While the `target_peers` is already in the network config struct, there seems to not exist a CLI arg to overwrite the default - we should add that.

PR #704, opened 2025-10-22
feat(cli): add target peers config

petarjuki7:
## Issue Addressed

Addresses issue #702 

## Proposed Changes

Add CLI flag to customize target peers count


dknopik (approved):
LGTM, thanks

PR #709, opened 2025-10-23
chore: Update Lighthouse for Fulu and other deps

dknopik:
## Issue Addressed

I realized that the last bump of the used Lighthouse version does not include the Fulu epoch definitions for the testnets, so we need to bump it.

## Proposed Changes

Update LH to `v8.0.0-rc.2` - which is not stable, but at least a release instead of a random commit! I also as usual used the opportunity to bump all semver compatible dependency version upgrades via `cargo update`.

## Additional Info

`safe_arith` was spun out of the LH repo as an independent crate, so I added it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1408,7 +1408,7 @@ checksum = "55248b47b0caf0546f7988906588779981c43bb1bc9d0c44087278f80cdb44ba"
 [[package]]
 name = "beacon_node_fallback"
 version = "0.1.0"
-source = "git+https://github.com/sigp/lighthouse?rev=ffa7b2b2#ffa7b2b2b9e3b4e70678e2c749b8bc45234febd7"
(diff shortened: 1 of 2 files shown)


diegomrsantos (approved):

diegomrsantos:
@claude can we use v8.0.0-rc.2 instead of a commit?

diegomrsantos:
@jking-aus is it intentional that Claude skipped the PR review?

dknopik:
@diegomrsantos Personally, I dislike using tag, as tags (theoretically) can be changed. Commit hashes are immutable. 

But I might be a little paranoid :sweat_smile:

diegomrsantos:
> @diegomrsantos Personally, I dislike using tag, as tags (theoretically) can be changed. Commit hashes are immutable.
> 
> But I might be a little paranoid 😅

@claude analyse daniel's point from a security perspective. Also, what's the difference between using a tag and just a version for a cargo dep?

diegomrsantos:
Not sure there's a tag replacement attack :sweat_smile: 

Feel free to merge it

PR #710, opened 2025-10-23
fix: Sync committee partial signature limit

dknopik:
## Issue Addressed

Message validation would limit the number of partial signatures for `SyncCommittee` duties to 1, even though it is supposed to be 13. This happens as we do not enter the `Role::SyncCommittee` branch due to its `if` guard. Next, the `_` branch below is checked, which enforces a limit of `1`. 

## Proposed Changes

Move the `if` into the `Role::SyncCommittee` branch to avoid falling through to the `_ if message_count > 1`  below if we have `Role::SyncCommittee`.

## Additional Info

This is very unlikely to happen in large networks such as Mainnet or Hoodi, as a validator needs to have multiple sync committee duties at the same time to send messages with more than one signature. However, in smaller networks, such as devnets and Sepolia, this happens easily.

### DIFF
--- a/anchor/message_validator/src/partial_signature.rs
+++ b/anchor/message_validator/src/partial_signature.rs
@@ -783,10 +783,10 @@ mod tests {
     }
 
     #[test]
-    fn test_passes_sync_committee_messages() {
+    fn test_sync_committee_accepts_multiple_signatures_within_limit() {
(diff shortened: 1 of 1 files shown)


diegomrsantos:
@claude how could we make the test clearer? Is the coverage enough?

diegomrsantos (approved):
looks good, thanks

Issue #712, opened 2025-10-24
Allow `pubkey` and `pubKey` in encrypted_private_key.json


dknopik:
A while ago, Go-SSV changed their private key format. The key of the property containing the public key was changed from `pubkey` to `pubKey`. We should support both, and emit the new variant (`pubkey`) when generating a new key.

PR #714, opened 2025-10-24
docs: Add warning for key format issue

dknopik:
## Issue Addressed

- Quick and dirty warning for #712

## Proposed Changes

To avoid user confusion and to provide a workaround, give a short explanation in the docs until a fix is released in a stable version.


dknopik:
> the PR is good to merge as-is if you prefer to keep it minimal.

:+1: , wdyt?

PR #723, opened 2025-11-05
chore: Update to Lighthouse v8.0.0

dknopik:
## Proposed Changes

Update to Lighthouse's stable release `v8.0.0`, which sets the Mainnet Pectra fork epoch.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1384,7 +1384,7 @@ checksum = "55248b47b0caf0546f7988906588779981c43bb1bc9d0c44087278f80cdb44ba"
 [[package]]
 name = "beacon_node_fallback"
 version = "0.1.0"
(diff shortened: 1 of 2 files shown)


diegomrsantos:
@dknopik wdyt about the docs suggestions?

dknopik:
> 1. Add a note about Pectra fork epoch to release notes or changelog

We have no `CHANGELOG.md` or similar to add something like this. We will of course mention this in the release page notes for 1.1.0.

> 2. Document the Lighthouse version update in project documentation
> 3. Consider linking to [Lighthouse v8.0.0 release notes](https://github.com/sigp/lighthouse/releases/tag/v8.0.0) for reference

I don't think the users should care about the Lighthouse version used within Anchor. It is essentially an implementation detail. Mentioning this explicitly might spark confusion that Anchor supports Lighthouse only.

diegomrsantos (approved):

PR #725, opened 2025-11-05
chore: Remove old `blsful` lagrange impl

dknopik:
## Proposed Changes

Remove the `blsful` feature in `bls_lagrange` and any associated dependencies and code.

## Additional Info

We have been using our custom `blst` implementation for a while now. The `blsful` variant has several downsides:

- We have to use a custom fork of a dependency due to conflicting transient dependencies.
- It has worse performance.
- It is less tested.
- It inconvenienced us, as it forced `bls_lagrange` to depend on a legacy `rand` version. (see #461)

Advantages might include keeping it around for future testing efforts.


dknopik:
@claude `blst_single_thread` still influences the behaviour in `blst.rs`

diegomrsantos:
@claude doesn't the text below go against "don't roll your own crypto"? Or is it a different context?

> We have been using our custom blst implementation for a while now. The blsful variant has several downsides:
We have to use a custom fork of a dependency due to conflicting transient dependencies.
It has worse performance.
It is less tested.
It inconvenienced us, as it forced bls_lagrange to depend on a legacy rand version. (see https://github.com/sigp/anchor/issues/461)
Advantages might include keeping it around for future testing efforts.

anchor/common/bls_lagrange/src/lib.rs
```diff
@@ -1,14 +1,9 @@
 use bls::SecretKey;
 use rand::prelude::*;
 
-#[cfg(feature = "blst")]
 pub mod blst;
-#[cfg(all(not(feature = "blsful"), feature = "blst"))]
```
shane-moore, line 6: nice that this simplifies features flags a bit

shane-moore (approved):

PR #726, opened 2025-11-05
feat(network): add dynamic peer count based on active subnets

petarjuki7:
## Proposed Changes

Implements dynamic peer target calculation similar to go-ssv:
- Base: 60 peers
- Add 3 peers per unique subnet
- Cap at 150 peers maximum

### Breaking Change
- `target_peers` config field changed from `usize` to `Option<usize>`
- **Previous default: Static 50 peers**
- **New default: Dynamic calculation (60-150 based on active subnets)**
- **Impact**: Users who relied on the implicit default of 50 peers will now get dynamic peer counts starting at 60 base + scaling with subnet duties

### Behavior
- **If user provides `--target-peers <N>`**: That static value is always used, regardless of subnet count
- **If user does NOT provide `--target-peers`**: Target peers are calculated dynamically based on active subnets (60 + subnets × 3, capped at 150)

### Implementation Notes
The dynamic calculation formula is copied from the go-ssv implementation for compatibility with the existing SSV network. The values are now defined as named constants (`BASE_PEER_COUNT`, `PEERS_PER_SUBNET`, `MAX_PEER_COUNT`) for maintainability.


dknopik (changes requested):
This does not dynamically adjust the target peers if we subscribe or unsubscribe to a subnet while the node is running.

diegomrsantos:
> If the user provides more than the calculated target_peers, then that value is used.

Shouldn't the value provided by the user be used no matter what?

petarjuki7:
@diegomrsantos @dknopik 
I followed the go-ssv code as the reference implementation for now, this is how they do it. I will now update with these changes

anchor/network/src/network.rs
```diff
@@ -100,6 +101,8 @@ pub struct Network<R: MessageReceiver> {
     domain_type: DomainType,
     metrics_registry: Option<Registry>,
     spec: Arc<ChainSpec>,
+    user_set_target_peers: bool,
+    active_subnets: HashSet<SubnetId>,
```
dknopik: there is already `needed_subnets` in the behaviour

anchor/network/src/network.rs
```diff
@@ -542,6 +555,13 @@ impl<R: MessageReceiver> Network<R> {
         }
     }
 
+    fn update_target_peers(&mut self) {
+        if !self.user_set_target_peers {
+            let new_target = min(60 + self.active_subnets.len() * 3, 150);
+            self.peer_manager().connection_manager.target_peers = new_target;
+        }
```
dknopik: We also need to update the limits within `ConnectionManager::connection_limits`

anchor/network/src/network.rs
```diff
@@ -500,11 +505,19 @@ impl<R: MessageReceiver> Network<R> {
 
                 let actions = self.peer_manager().join_subnet(subnet);
                 self.handle_connect_actions(actions);
+
+                self.active_subnets.insert(subnet);
+                self.update_target_peers();
```
dknopik: maybe make `update_target_peers` a function in the `ConnectionManager` and have it be called in `join_subnet`?

anchor/network/src/config.rs
```diff
@@ -100,7 +103,8 @@ impl Config {
             enr_udp6_port: None,
             enr_tcp6_port: None,
             enr_quic6_port: None,
-            target_peers: 50,
+            target_peers: 60,
```
dknopik: I think this default is not meaningful anymore as it will be overwritten by either the user or the dynamic limit. Let's maybe change the `target_peers` field to be `Option` here for clarity?

This would then also make `user_set_target_peers` redundant, as `None` signals that the user did not set something.

anchor/client/src/lib.rs
```diff
@@ -476,6 +478,21 @@ impl Client {
...
+            for cluster_id in state.get_own_clusters() {
+                if let Some(cluster) = state.clusters().get_by(cluster_id) {
+                    let subnet = SubnetId::from_committee(cluster.committee_id(), SUBNET_COUNT);
+                    unique_subnets.insert(subnet);
+                }
+            }
+            let dyn_peers = min(60 + unique_subnets.len() * 3, 150);
+            if config.network.target_peers < dyn_peers {
+                config.network.target_peers = dyn_peers;
+            }
```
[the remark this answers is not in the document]
petarjuki7: changed now

anchor/network/src/config.rs
```diff
@@ -100,7 +100,7 @@ impl Config {
...
```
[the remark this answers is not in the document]
petarjuki7: Changed to be the same as the go-ssv implementation, I don't think users _**rely**_ on the default value?

dknopik (commented):
👍

anchor/client/src/lib.rs
```diff
@@ -476,6 +478,21 @@ impl Client {
             message_validator,
         );
 
+        let dynamic_target_peers = config.network.target_peers.is_none();
+        if dynamic_target_peers {
```
jxs: nitpick:
```suggestion
       let target_peers = match config.network.target_peers {
           Some(target_peers) => target_peers,
           None => {...}
```

anchor/client/src/lib.rs
```diff
@@ -476,6 +478,21 @@ impl Client {
...
+            let state = database.state();
+            let mut unique_subnets = HashSet::new();
+            for cluster_id in state.get_own_clusters() {
+                if let Some(cluster) = state.clusters().get_by(cluster_id) {
+                    let subnet = SubnetId::from_committee(cluster.committee_id(), SUBNET_COUNT);
+                    unique_subnets.insert(subnet);
+                }
+            }
+            let dyn_peers = min(60 + unique_subnets.len() * 3, 150);
```
jxs: we can simplify this by counting the unique subnets using `itertools` [`unique`](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.unique) (which we already depend transitively:
 ```suggestion
           let unique_subnets = state
                .get_own_clusters()
                .iter()
                .filter_map(|id| state.clusters().get_by(id))
                .map(|s| SubnetId::from_committee(s.committee_id(), SUBNET_COUNT))
                .unique()
                .count();

            let dyn_peers = min(60 + unique_subnets.len() * 3, 150);
```

File: anchor/network/src/network.rs [hunk 9]
```diff
@@ -113,6 +115,7 @@ impl<R: MessageReceiver> Network<R> {
         outcome_rx: mpsc::Receiver<Outcome>,
         executor: TaskExecutor,
         spec: Arc<ChainSpec>,
+        dynamic_target_peers: bool,
```
jxs:
```suggestion
        target_peers: bool,
```
they can also be static right? If they are user provided

anchor/client/src/lib.rs
```diff
@@ -476,6 +478,21 @@ impl Client {
             message_validator,
         );
 
+        let dynamic_target_peers = config.network.target_peers.is_none();
```
diegomrsantos: `is_dynamic_target_peers`

anchor/client/src/lib.rs
```diff
@@ -476,6 +478,21 @@ impl Client {
...
+            for cluster_id in state.get_own_clusters() {
+                if let Some(cluster) = state.clusters().get_by(cluster_id) {
+                    let subnet = SubnetId::from_committee(cluster.committee_id(), SUBNET_COUNT);
+                    unique_subnets.insert(subnet);
+                }
+            }
+            let dyn_peers = min(60 + unique_subnets.len() * 3, 150);
+
+            config.network.target_peers = Some(dyn_peers);
+        }
```
diegomrsantos: Can we calculate this in the network or in another place where we have the number of subnets?

anchor/network/src/config.rs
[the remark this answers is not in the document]
diegomrsantos: I believe it means that users assumed the default value was 50, and now it has been changed

anchor/client/src/lib.rs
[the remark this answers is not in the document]
diegomrsantos: @claude can we calculate this in the network crate or in another place where the number of subnets is known?

diegomrsantos:
@claude reply to this https://github.com/sigp/anchor/pull/726#discussion_r2495448640

anchor/client/src/lib.rs
petarjuki7: It could probably be done in the network, my logic was to keep it here for all the startup / config code to stay in the `client` crate. `ConnectionManager` needs the `target_peers` in its constructor.

anchor/network/src/peer_manager/connection.rs
```diff
@@ -55,41 +55,77 @@ pub struct ConnectionManager {
...
+            ))
+            .with_max_established_outgoing(Some(
+                (new_target as f32 * (1.0 + PEER_EXCESS_FACTOR)).ceil() as u32,
+            ))
+            .with_max_established(Some(
+                (new_target as f32 * (1.0 + PEER_EXCESS_FACTOR)).ceil() as u32,
+            ))
+            .with_max_established_per_peer(Some(1));
+
+        self.connection_limits = connection_limits::Behaviour::new(limits);
```
dknopik: I think it would be cleaner to update the limits using `limits_mut` on the existing behaviour

diegomrsantos:
@claude reply to this https://github.com/sigp/anchor/pull/726#discussion_r2495448640

anchor/client/src/lib.rs
dknopik: can you e.g. `update_dynamic_target_peers` on network creation instead of having the duplicated logic here?

diegomrsantos: Claude gave a great reply https://github.com/sigp/anchor/pull/726#issuecomment-3492741172

anchor/client/src/lib.rs
```diff
@@ -476,6 +478,21 @@ impl Client {
...
+            let mut unique_subnets = HashSet::new();
+            for cluster_id in state.get_own_clusters() {
+                if let Some(cluster) = state.clusters().get_by(cluster_id) {
+                    let subnet = SubnetId::from_committee(cluster.committee_id(), SUBNET_COUNT);
+                    unique_subnets.insert(subnet);
+                }
+            }
+            let dyn_peers = min(60 + unique_subnets.len() * 3, 150);
+
+            config.network.target_peers = Some(dyn_peers);
```
jxs: are we using the value in `dyn_peers` anywhere else? If not we can avoid declaring `config` as `mut` in the input parameters

jxs (commented):
Hi Petar, overall looks good to me, left some comments

anchor/network/src/peer_manager/connection.rs
petarjuki7: As far as I can tell, `limits_mut` gives me a `&mut ConnectionLimits`, but all the methods for updating values (`.with_max_established_incoming()` etc.) take `self` (`ConnectionLimits`). And also the fields of the `ConnectionLimits` struct are private so I can't just update the fields through `&mut ConnectionLimits`.

Or did you have in mind to just do something like:
```
*self.connection_limits.limits_mut() =
 ConnectionLimits::default()
      .with_max_pending_incoming(Some(5))
      .with_max_pending_outgoing(Some(16))
      .with_max_established_incoming(Some(...))
```

anchor/client/src/lib.rs
```diff
@@ -476,6 +477,26 @@ impl Client {
...
+        let initial_subnet_count = {
+            let state = database.state();
+            state
+                .get_own_clusters()
+                .iter()
+                .filter_map(|id| state.clusters().get_by(id))
+                .map(|cluster| SubnetId::from_committee(cluster.committee_id(), SUBNET_COUNT))
+                .collect::<HashSet<_>>()
+                .len()
+        };
```
dknopik: is this necessary?

as soon as the first subnet join event comes in, the target will be changed again. So why not start with a target of 60 to avoid adding this here?

anchor/client/src/lib.rs
```diff
@@ -476,6 +477,26 @@ impl Client {
...
+
+        // Determine if we should dynamically adjust target_peers when subnets change.
+        // If the user specified a target_peers value (Some), we keep it static throughout runtime.
+        // If not specified (None), we calculate and adjust target_peers dynamically based on the
+        // number of active subnets as validators join/leave subnet duties.
+        let is_dynamic_target_peers = config.network.target_peers.is_none();
```
dknopik: we pass the whole network config below. why do we pass this flag additionally seperately?

diegomrsantos: true, removed

anchor/client/src/lib.rs
diegomrsantos: true, done

anchor/network/src/network.rs
```diff
@@ -100,11 +100,13 @@ pub struct Network<R: MessageReceiver> {
     domain_type: DomainType,
     metrics_registry: Option<Registry>,
     spec: Arc<ChainSpec>,
+    is_dynamic_target_peers: bool,
 }
 
 impl<R: MessageReceiver> Network<R> {
     // Creates an instance of the Network struct to start sending and receiving information on the
     // p2p network.
+    #[allow(clippy::too_many_arguments)]
```
dknopik: This is no longer required

anchor/network/src/peer_manager/connection.rs
```diff
@@ -54,42 +54,62 @@ pub struct ConnectionManager {
...
             inbound_count: 0,
             outbound_count: 0,
         }
     }
 
+    /// Update the target peer count and recalculate connection limits.
+    ///
+    /// This is called by PeerManager when dynamic peer calculation is enabled
+    /// and the number of active subnets changes.
+    pub fn set_target_peers(&mut self, new_target: usize) {
```
dknopik, line 100: Can we add a log in this function? `debug` I guess?

dknopik (commented):
Two nits :) 

Can we also adjust the `--target-peers` help text to describe the behaviour if it is omitted?

diegomrsantos:
> Two nits :)
> 
> Can we also adjust the `--target-peers` help text to describe the behaviour if it is omitted?

All done

dknopik (approved):
LGTM, thanks both!

PR #727, opened 2025-11-05
refactor: use Result for error handling instead of of early returns

diegomrsantos:
## Issue Addressed

Fixes incorrect exit codes on startup errors.

## Proposed Changes

Refactor `main()` and `start_anchor()` to return `Result<(), String>` instead of using early returns from void functions. This fixes a bug where errors would exit with code 0 (success) instead of code 1 (failure), causing issues with process managers like systemd, Docker, and Kubernetes.

**Before:**
```rust
fn main() {
    let config = match parse_config() {
        Ok(config) => config,
        Err(e) => {
            eprintln!("Error: {e}");
            return;  // ❌ exits with code 0 (success)
        }
    };
}
```

**After:**
```rust
fn main() -> Result<(), String> {
    let config = parse_config()
        .map_err(|e| format!("Error: {e}"))?;  // ✅ exits with code 1 (failure)
    Ok(())
}
```

### Benefits
- **Correct exit codes**: Process managers can now detect startup failures
- **Cleaner code**: Using `?` operator instead of verbose match statements
- **Standard Rust idiom**: Termination trait handles Ok/Err conversion
- **Better error propagation**: Errors flow naturally up the call stack

### Changes
- `main()` and `start_anchor()` return `Result<(), String>`
- All error handling uses `?` operator with `map_err` for context
- Keysplit and Keygen subcommands use `?` operator
- Exit code determined by Result return value

### Testing
Added comprehensive integration tests in `anchor/tests/cli_exit_codes.rs`:
- 12 tests covering success and error scenarios
- Uses `assert_cmd` for CLI testing
- Verifies exit codes, stdout, and stderr
- All tests passing ✅

## Additional Info

This is critical for production deployments where orchestrators rely on exit codes for restart policies and health checks. The previous behavior would report success even when configuration failed, causing silent failures in automation.


dknopik (approved):
LGTM, thanks!

Issue #722, opened 2025-10-30
Restore in-memory database from disk on critical sync error


dknopik:
Basically, there are two classes of errors in the event processor:

1. Incorrect event: The event is faulty due to a mistake by the transaction sender, e.g. bad signature in validatorAdded
2. Internal fault: if e.g. disk space runs out, we wont be able to modify the database.

For the first kind, we should just continue the sync with the next event. 

But for the second kind, we should retry, because we must not miss events. But this is tricky: while we can roll back the database transaction, we can not roll back the internal memory cache. But we rely on the internal cache being in the same state as the database.

We need to cope with this. As internal faults should be exceedingly rare, it should be fine to just reload the in-memory database.

PR #735, opened 2025-11-10
feat: restore in-memory state from disk on internal errors

petarjuki7:
## Issue Addressed

Addresses issue #722 

 ## Proposed Changes

Added error categorisation and automatic state recovery for critical database errors during event processing.

- Refactored error types to use `thiserror` for cleaner error handling
- Implemented `is_critical()` method to distinguish system faults (database/sync errors) from recoverable errors (malformed events)
- Added `reload_state_from_disk()` to restore in-memory state after transaction rollback
- Critical errors now trigger transaction rollback and state reload to maintain consistency
- Recoverable errors are logged and skipped without affecting the batch

## Additional Info

The sync loop automatically retries from `last_processed_block` after critical errors, so retry logic operates at the batch level rather than per-event.


anchor/eth/src/event_processor.rs
```diff
@@ -134,23 +129,29 @@ impl EventProcessor {
...
                     .unwrap_or_else(|| "unknown".to_string());
-                if live {
-                    warn!(tx_hash, "Malformed event: {e}");
+
+                if e.is_critical() {
+                    // Critical error - rollback transaction and reload state from disk
+                    warn!(tx_hash, "Critical error during event processing: {e}");
+                    tx.rollback()?;
+                    self.db.reload_state_from_disk()?;
+                    return Err(e);
```
dknopik: Not sure if this is a good location for this. If e.g. the `processed_block` or `tx.commit()` below errors, it is not caught.

anchor/database/src/error.rs
```diff
@@ -1,40 +1,20 @@
...
-}
+    #[error("SQL error")]
+    SQLError(#[from] rusqlite::Error),
 
-impl Display for DatabaseError {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(f, "{self:?}")
-    }
+    #[error("connection pool error")]
+    SQLPoolError(#[from] r2d2::Error),
```
dknopik: add `{0}` to these format strings

anchor/eth/src/error.rs
```diff
@@ -1,20 +1,51 @@
...
+
+    #[error("{0}")]
     Misc(String),
+
+    #[error("duplicate entry: {0}")]
     Duplicate(String),
-    Database(String),
+
+    #[error("database error")]
+    Database(#[from] DatabaseError),
```
dknopik: add `{0}`

dknopik (commented):
Nice!

diegomrsantos:
> - Added reload_state_from_disk() to restore in-memory state after transaction rollback
> - Critical errors now trigger transaction rollback and state reload to maintain consistency

I'm not sure this is a good idea, @claude what do you think?

dknopik:
I'm sure this is a good idea, @claude what do you think?

diegomrsantos:
We have a constant issue with claude setup getting stuck, like above. 

Trying again to test how claude is influenced by the prompt: I'm sure this is a good idea, @claude what do you think?

dknopik:
Okay. 

As we are considering to refactor the code base to broadcast certain events to other parts of the codebase, and having them handle potential database reloads might be more bothersome than its worth, lets crash instead of reloading the DB as suggested by Diego.

Additionally, this can help fixing issues that might not be caught by reloading like a broken db connection in the connection pool, etc.

Thanks @diegomrsantos for the suggestion

CC @petarjuki7

diegomrsantos:
@petarjuki7 @dknopik Does anyone know the status of this PR?

diegomrsantos:
> @petarjuki7 @dknopik Does anyone know the status of this PR?

Hi guys, any news on this?

diegomrsantos:
@dknopik please describe what the current issues this PR is gonna fix, the risk being introduced, and mitigations. Otherwise let's keep it close, please.

dknopik:
> @dknopik please describe what the current issues this PR is gonna fix

- https://github.com/sigp/anchor/issues/722

let me know what exactly is unclear

diegomrsantos:
Linking #722 does not answer the questions that were asked.

The request was for the current problem statement for this PR, the risk it introduces, and the mitigations. The issue body was written on October 30, 2025, and point 2 there is still too vague to serve as a design basis for a reliability-sensitive change. "Internal fault" currently lumps together very different failure modes like disk-full conditions, SQLite failures, pool failures, and other runtime faults, and those do not all justify the same handling. Industry-standard reliability work starts with a precise failure model and clear operational assumptions, not a broad critical bucket.

There is also important missing context here:
- On December 17, 2025, the thread explicitly moved away from "reload state from disk" toward "crash and restart" as the safer approach.
- This PR was closed on March 4, 2026 and reopened on March 9, 2026 without a proper updated rationale.
- The PR description still says this change adds `reload_state_from_disk()`, while the current review discussion is about crashing on critical errors.

So before asking for more review, please update the PR description with:
1. the concrete scenario this PR fixes today
2. the exact error classes that should crash vs not crash, and why
3. why restart is the correct recovery model in our deployment environment
4. the risks introduced by the new classification/exit behavior and the mitigations
5. the tests that cover those cases

Also, please respond directly to the questions being asked. "Let me know what exactly is unclear" is not constructive when the requested information was already spelled out.

Once that context is in place, happy to review the implementation. Until then, I do not think it makes sense to keep this PR open.

dknopik:
Let me start by giving a specific example of misbehavior I want us to solve with this PR: 

1. Let's say a user stores their Anchor data dir on an external hard drive, and accidentally disconnects it while Anchor is running. The disconnection occurs in the middle of a event batch being processed.
2. This causes all SQLite operations to fail as SQLite can not write uncommitted data to the disk anymore. 
3. When an event is processed, an `Err` is returned by the corresponding `process_*` method of the `EventProcessor`. This error is caught by this part: 
     ```rust
    // Handle any errors from the event processing
    if let Err(e) = result {
        let tx_hash = log
            .transaction_hash
            .map(|hash| hash.to_string())
            .unwrap_or_else(|| "unknown".to_string());
        if live {
            warn!(tx_hash, "Malformed event: {e}");
        } else {
            trace!(tx_hash, "Malformed event: {e}");
        }
        continue;
    }
     ```
    This not only hides the error (unless in live sync), it also just ignores the error and continues.
4. When the database transaction is committed, it fails. However, the in-memory database still has the changes of successfully processed events (before the disconnection).
5. This is problematic, as this inconsistency might cause more inconsistencies once event processing functions again: In `insert_operator`, the database layer checks the in-memory storage to see if the operator already exists, so it will not re-attempt to store that operator into the SQLite database - causing an inconsistent state.

Therefore, we must ensure that when SQLite operations fail, the in memory store is rolled back as well. Whether we do this via a shutdown or an ad-hoc reload is not that important.

Note that the example above uses disconnection of external media as an example, but there are many other errors that can occur, e.g. a failing harddrive, out of disk space, a bug in SQLite, an inconsistent database schema, etc. - However, all of these can cause the same inconsistent state, which is why I lumped these together.

To answer your other questions:

> 2. the exact error classes that should crash vs not crash, and why

I believe we should stick to crash or not crash regardless of the exact error. This of course excludes actually malformed events, which should be handled as before.

> 3. why restart is the correct recovery model in our deployment environment

I will not defend restarting here, as I was previously for not restarting, but reloading the in memory database. Please refer to your Claude invocation above for the arguments that caused us to switch to restarting.

> 4. the risks introduced by the new classification/exit behavior and the mitigations

The main risk that come to my mind with the restarting behaviour is that restarting is more involved: We have to find peers again, we have to populate our duty tracking, etc. 

A risk with the new classification is obviously misclassification, but as there are only two classes, this risk is rather low.

anchor/eth/src/event_processor.rs
```diff
@@ -71,87 +71,99 @@ impl EventProcessor {
...
-                    .map(|hash| hash.to_string())
-                    .unwrap_or_else(|| "unknown".to_string());
-                if live {
-                    warn!(tx_hash, "Malformed event: {e}");
-                } else {
-                    trace!(tx_hash, "Malformed event: {e}");
-                }
-                continue;
+        if let Err(e) = result {
+            if e.is_critical() {
```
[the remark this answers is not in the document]
dknopik: While I agree that we should exit differently - I just want to point out that 

> Database file corruption if SQLite has pending WAL entries

is incorrect - sqlite databases can not corrupt if the process exits while a transaction is ongoing, unless the process exits in the middle of *committing* a transaction, which is not the case here

diegomrsantos:
I think this PR is still treating the symptom, not the design bug.

The core problem is that event processing does not currently have a single transactional boundary. During a batch we mutate shared in-memory state before the SQLite transaction is durably committed, and we also emit side effects before commit (for example slashing-protection registration, validator index-sync requests, and exit-processing requests). That means a database/I/O failure is not just "disk and memory diverged"; it can also leave other subsystems having observed or cached partially applied batch state.

Because of that, I do not think `reload vs restart` is a minor implementation detail. An in-process reload can maybe reconstruct `NetworkState`, but it cannot reliably rewind already-emitted side effects or arbitrary derived caches held elsewhere in the process. Restart is the only sane short-term recovery model because it reinitializes the whole process from a known-good state.

More importantly, I do not think `is_critical()` / `DatabaseOperation` is the right abstraction here. It is papering over the lack of a real batch boundary. The simple rule should be:

- malformed / semantically invalid event: log and skip
- infrastructure / database / internal processing failure during event handling: fatal, terminate process

Longer term, I think the design needs to be tightened so that event processing is:

1. parse/validate into a batch delta
2. commit durable state
3. publish the new in-memory snapshot exactly once
4. emit side effects only after commit

Until we have that separation, any attempt to "recover in place" is going to be fragile.

dknopik:
> That means a database/I/O failure is not just "disk and memory diverged"; it can also leave other subsystems having observed or cached partially applied batch state.

good point 

> The simple rule should be:
>  - malformed / semantically invalid event: log and skip
>  - infrastructure / database / internal processing failure during event handling: fatal, terminate process

Isn't this what this PR is doing?

Feel free to further discuss the actual implementation with @petarjuki7, the author of the PR. As we have shifted now to discussing the precise approach to take, hopefully we have established by now that this is a valid issue/PR that should be handled instead of closed. Thank you.

diegomrsantos:
@dknopik thanks, I agree this is a real issue and should be fixed.

That said, I am not convinced it is worth merging the current mitigation as-is.

My concern is that it still leaves us with a system that is fragile and hard to reason about:
- shared in-memory state is mutated before durable commit
- side effects are emitted before commit
- correctness depends on a coarse crash vs non-crash classification layer
- that layer already loses information in places like `DatabaseOperation(String)`

So even if this improves one concrete failure mode, it still keeps the underlying design in a state that is non-intuitive and easy to regress. In that sense, it feels more like moving the fragility around than actually removing it.

My preference would be to fix the design bug properly rather than commit a mitigation that we already know is not a robust model:
1. define the actual invariant we want
   - batch-local processing
   - durable commit first
   - publish in-memory state after commit
   - emit side effects after commit
2. clearly define which failures are invalid-event skips vs fatal process failures
3. implement against that model directly

If there is urgent operational pressure for a short-term safeguard, then I could understand doing a narrowly scoped interim fail-fast change. But absent that, I think the saner option is to close this PR and reopen the work from a fresh issue with the design bug and target model stated explicitly.

diegomrsantos:
I think the deeper issue is that the cache/coherence boundary is in the wrong place.

More specifically: the in-memory state should be owned by the database abstraction together with its transactional semantics. Right now `NetworkDatabase` already physically owns both the durable store and the in-memory cache, but the consistency model is effectively orchestrated from outside of it by `EventProcessor`. That is what makes this hard to reason about.

Today the flow is roughly:
1. `EventProcessor` opens a SQL transaction
2. it calls database helper methods while processing each event
3. those helper methods mutate the shared in-memory state during the batch
4. some side effects are emitted before the batch is durably committed
5. only at the end do we try to commit

That means the database layer owns the cache structurally, but does not actually own cache coherence.

The model I think we should aim for is this:

1. `EventProcessor` parses and validates logs into a batch-local set of domain changes
   - no global cache mutation
   - no externally visible side effects
2. `NetworkDatabase` owns a single “apply batch” API
   - it opens the SQL transaction internally
   - it applies the durable writes
   - it stages the corresponding in-memory changes privately
3. if the transaction commits successfully
   - publish the new in-memory snapshot exactly once
   - then emit any side effects derived from the committed batch
4. if anything fails before commit
   - discard the staged changes
   - nothing has been published
   - no rollback/reload logic is needed for cache consistency

In other words, the invariant should be:

- before commit: changes are batch-local only
- after commit: durable state and published in-memory state advance together
- side effects happen only from committed state

That is much easier to reason about than the current model, and it removes the need to classify a lot of subtle intermediate failures in order to preserve correctness.

So from my perspective, the right fix is not to keep refining this mitigation, but to move toward a design where the database abstraction truly owns both persistence and cache publication as one unit.

If the team agrees with that direction, I would be happy to work on it.

diegomrsantos:
Closing — the underlying problem (in-memory state diverging from DB after transaction rollback) has been structurally eliminated by the tx-scoped design landed in #898, #904, and #909.

In-memory `NetworkState` is now only updated via `publish_pending_state_updates` after a successful commit, so a rolled-back transaction can never leave the cache inconsistent. The reload-from-disk approach is no longer needed.

PR #738, opened 2025-11-11
fix: only store one node info and update that

dknopik:
## Issue Addressed

We recently introduced a regression where we fail all handshakes. The root cause for this is that we started storing the `NodeInfo` in a new place but did not adjust all places to use this.

## Proposed Changes

- Remove the field `Network::node_info`
- Add accessors for the node metadata
- Adjust old references of `Network::node_info` to use new accessors instead

### DIFF
--- a/anchor/network/src/network.rs
+++ b/anchor/network/src/network.rs
@@ -33,7 +33,7 @@ use crate::{
     Config, Enr,
     behaviour::{AnchorBehaviour, AnchorBehaviourEvent, BehaviourError},
     discovery::{DiscoveredPeers, Discovery, DiscoveryError},
-    handshake::{self, node_info::NodeMetadata},
+    handshake,
     keypair_utils::load_private_key,
     network::NetworkError::SwarmConfig,
     peer_manager::{self, ConnectActions, PeerManager},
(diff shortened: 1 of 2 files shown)


anchor/network/src/network.rs
```diff
@@ -571,7 +563,7 @@ impl<R: MessageReceiver> Network<R> {
         peer_id: PeerId,
         their_metadata: &NodeMetadata,
     ) {
-        if let Some(our_metadata) = &self.node_info.metadata {
+        if let Some(our_metadata) = self.node_metadata() {
```
diegomrsantos: Should this function be moved to the handshake module?

diegomrsantos:
@claude review this PR

diegomrsantos:
@claude suggest how a test like the one described below could be implemented:

> While the existing handshake tests in handshake/mod.rs (lines 347-440) provide good coverage of the protocol, consider adding an integration test that:
Subscribes to a subnet
Verifies the metadata is correctly updated via the accessor methods
Confirms the updated metadata is sent during subsequent handshakes
This would verify the end-to-end flow of the regression fix.

anchor/network/src/network.rs
dknopik: there is a `node_metadata` function on the handshake module, I just added another one in `network.rs` as a shortcut, calling the one in the handshake module.

diegomrsantos: I meant `record_handshake_subnet_match_metrics`

diegomrsantos (approved):

PR #748, opened 2025-11-26
chore!: Remove builder proposals flag

### LINKED ISSUE #746: Deprecate `--builder-proposals` and always create validator registrations.
The `--builder-proposals` flag only controls whether the node should participate in creating validator registrations. However, to correctly participate in the SSV protocol, we should always do so.

The local user can still chose not to propose MEV blocks by simply not configuring a relay in the local BN. Additionally, we should log errors indicating that the BN has no builder configured as `DEBUG` only - it is simply an artifact of that choice.

shane-moore:
## Issue Addressed
Closes #746

## Proposed Changes
- Remove the `--builder-proposals` CLI flag
- Always create validator registrations
- Change the log level for builder registration failures from `warn` to `debug`

### Why this change: 
In DVT, validator registrations require threshold signatures from multiple operators. If some operators had `--builder-proposals` disabled, they wouldn't participate in signing registrations, potentially causing the registration to fail for the entire committee

### User impact: 
Users who previously used `--builder-proposals` will see a CLI error when trying to use the flag

### DIFF
--- a/anchor/client/src/cli.rs
+++ b/anchor/client/src/cli.rs
@@ -442,6 +442,16 @@ pub struct Node {
 
     #[clap(
         long,
+        alias = "private-tx-proposals",
+        help = "Deprecated and ignored. Validator registrations are now always created.",
+        display_order = 0,
+        help_heading = FLAG_HEADER,
+        hide = true
+    )]
+    pub builder_proposals: bool,
+
+    #[clap(
+        long,
         value_name = "UINT64",
         help = "Defines the boost factor, \
                 a percentage multiplier to apply to the builder's payload value \
--- a/anchor/client/src/cli.rs
+++ b/anchor/client/src/cli.rs
@@ -433,24 +433,12 @@ pub struct Node {
         long,
         value_name = "INTEGER",
         default_value_t = 36_000_000,
-        requires = "builder_proposals",
         help = "The gas limit to be used in all builder proposals for all validators managed. \
                 Note this will not necessarily be used if the gas limit \
                 set here moves too far from the previous block's gas limit.",
         display_order = 0
     )]
     pub gas_limit: u64,
-
-    #[clap(
-        long,
-        alias = "private-tx-proposals",
-        help = "If this flag is set, Anchor will query the Beacon Node for only block \
-                headers during proposals and will sign over headers. Useful for outsourcing \
-                execution payload construction during proposals.",
-        display_order = 0,
-        help_heading = FLAG_HEADER
-    )]
-    pub builder_proposals: bool,
 
     #[clap(
         long,
--- a/anchor/client/src/config.rs
+++ b/anchor/client/src/config.rs
@@ -201,7 +201,8 @@ pub fn from_cli(cli_args: &Node, global_config: GlobalConfig) -> Result<Config, 
 
     if cli_args.builder_proposals {
         warn!(
-            "The --builder-proposals flag is deprecated and ignored. Validator registrations are now always created."
+            "The --builder-proposals flag is deprecated and ignored. Validator registrations are \
+             now always created. This flag will be removed in a future release and will error then."
         );
     }
 
--- a/anchor/client/src/config.rs
+++ b/anchor/client/src/config.rs
@@ -64,8 +64,6 @@ pub struct Config {
     pub impostor: Option<OperatorId>,
     /// Gas limit on blocks
     pub gas_limit: u64,
-    /// Should payload construction be outsourced
-    pub builder_proposals: bool,
     /// Block boost factor
     pub builder_boost_factor: Option<u64>,
     /// Should external payloads always be preferred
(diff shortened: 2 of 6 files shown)


anchor/client/src/cli.rs
```diff
@@ -433,25 +433,13 @@ pub struct Node {
...
-    #[clap(
-        long,
-        alias = "private-tx-proposals",
-        help = "If this flag is set, Anchor will query the Beacon Node for only block \
-                headers during proposals and will sign over headers. Useful for outsourcing \
-                execution payload construction during proposals.",
-        display_order = 0,
-        help_heading = FLAG_HEADER
-    )]
-    pub builder_proposals: bool,
```
dknopik, line 453: To ensure that we do not break the CLI, we should (as a first step) only hide this option, and emit a warning on startup if it is supplied, notifying the user that it is deprecated.

anchor/validator_store/src/registration_service.rs
```diff
@@ -199,9 +197,10 @@ impl<S: ValidatorStore + 'static, T: SlotClock + 'static> Inner<S, T> {
                     count = signed.len(),
                     "Published validator registrations to the builder network"
                 ),
-                Err(err) => warn!(
+                Err(err) => debug!(
                     %err,
-                    "Unable to publish validator registrations to the builder network"
+                    "Unable to publish validator registrations to the builder network. \
+                     This is expected if no relay is configured in your Beacon Node."
```
dknopik: Ideally, we only debug log this if the error is in fact caused by this (and keep the error log if it is not). I think it should be possible to check this. For reference, here is the log emitted currently:

```
2025-12-03T13:39:11.456556Z  WARN Unable to publish validator registrations to the builder network err=Some endpoints failed, num_failed: 1 http://localhost:5052/ => RequestFailed(ServerMessage(ErrorMessage { code: 500, message: "UNHANDLED_ERROR: BuilderMissing", stacktraces: [] }))
```

dknopik (commented):
Hey Shane, thanks for the PR! Looks good, some notes below.

anchor/client/src/cli.rs
shane-moore: @dknopik, good call! hiding for now and emiting warning if the flag is supplied in https://github.com/sigp/anchor/pull/748/commits/00c8fb5fa35af5a7bf9dab41c1e13306a465d1df

anchor/validator_store/src/registration_service.rs
shane-moore: @dknopik, yep, makes sense. I found in lighthouse  they emit the `BuilderMissing` error type during `post_validator_register_validator` at:
https://github.com/sigp/lighthouse/blob/4e35e9d5875679855fa2e9587bcc95cf905bc59d/beacon_node/http_api/src/validator/mod.rs#L549-L558

In commit https://github.com/sigp/anchor/pull/748/commits/00c8fb5fa35af5a7bf9dab41c1e13306a465d1df, I made it `debug` log if the builder isn't configured by BN and therefore emits `BuilderMissing`. Else, an error is logged.

however, only lighthouse likely emits `BuilderMissing` error type, so if other BN's are configured, the error will fall through to the error log. not sure if anchor is meant to be used with other BN's besides lighthouse. just flagging this as a limitation of the approach

anchor/client/src/config.rs
```diff
@@ -199,10 +196,15 @@ pub fn from_cli(cli_args: &Node, global_config: GlobalConfig) -> Result<Config,
...
 
     // MEV options
-    config.builder_proposals = cli_args.builder_proposals;
     config.builder_boost_factor = cli_args.builder_boost_factor;
     config.prefer_builder_proposals = cli_args.prefer_builder_proposals;
 
+    if cli_args.builder_proposals {
+        warn!(
+            "The --builder-proposals flag is deprecated and ignored. Validator registrations are now always created."
+        );
```
dknopik, line 206: nitpick: explicitly mention that this flag will be removed in a future release and will error then.

anchor/validator_store/src/registration_service.rs
dknopik: good point! Anchor is supposed to support all BNs. I did not consider that the other clients will have different error messages.

Hmm. I think false positive warnings should be avoided. But I am also not particularly keen on handling all BN's possible error messages and keeping that up to date... Maybe we should revert to your previous impl. What do you think?

shane-moore: @dknopik, gotcha, yah, since we're supporting all BN's, definitely agree on keeping things simple and reverting to `debug!`

The con is that if the default log level is set to `warn` and the issue is something other than mev relay not set up, it will be missed

however, since this fn runs every slot, i dont think setting this to `warn` is justified because that would be a lot of useless messages when mev relay isn't configured.  And I would think they'd set the log level to `debug` in that scenario to troubleshoot more deeply, in which case the `debug` message will surface

updated in commit https://github.com/sigp/anchor/pull/748/commits/750073ca378837de53fc4dbccc89b1280b0d7082

anchor/client/src/config.rs
shane-moore: certainly! updated in https://github.com/sigp/anchor/pull/748/commits/750073ca378837de53fc4dbccc89b1280b0d7082

dknopik (approved):
LGTM, thanks!

PR #749, opened 2025-11-26
refactor: relax BeaconVote validation to epoch-only comparison

diegomrsantos:
## Issue Addressed

Addresses liveness issues during reorgs caused by strict checkpoint validation, as discussed in the SSV spec:
- https://github.com/ssvlabs/ssv-spec/issues/555
- https://github.com/ssvlabs/ssv-spec/pull/589

## Proposed Changes

Modified `BeaconVoteValidator` to compare only checkpoint epochs instead of full checkpoints (epoch + root):

- **Validation logic**: Now accepts attestations with matching epochs but different roots, allowing DVs to remain live during benign reorgs when operators' beacon nodes temporarily see different chain heads
- **Error handling**: Renamed `CheckpointMismatch` to `EpochMismatch` and simplified to store only epoch values
- **Test coverage**: Removed tests for same-epoch different-root failures, added test verifying this scenario now succeeds

## Additional Info

This change prioritizes validator liveness over fork protection. Slashing protection remains active and prevents actual slashable attestations.

The broader design question of whether DVs should actively prevent justifying potentially wrong forks (vs. focusing solely on slashing protection) remains open. This implementation keeps the SIP focused on core slashing protection, while cluster-level fork heuristics may be explored separately in the future.

### DIFF
--- a/anchor/client/src/cli.rs
+++ b/anchor/client/src/cli.rs
@@ -518,6 +518,18 @@ pub struct Node {
     )]
     pub operator_dg_wait_epochs: u64,
 
+    // Majority fork protection
+    #[clap(
+        long,
+        help = "Enable strict majority fork protection. When enabled, the node will not \
+                participate in attestation production if the checkpoint roots mismatch. \
+                Using this flag might reduce validator performance if cluster operators have \
+                struggling nodes, but can help to avoid finalization of a faulty majority fork.",
+        display_order = 0,
+        help_heading = FLAG_HEADER,
+    )]
+    pub strict_mfp: bool,
+
     #[clap(flatten)]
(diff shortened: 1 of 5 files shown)


dknopik:
TODOs:
- add CLI flag to use old logic
- ~~rebase on `release-1.1.0`, as this will be in a hot fix release~~

diegomrsantos:
> TODOs:
> 
> * add CLI flag to use old logic
> * rebase on `release-1.1.0`, as this will be in a hot fix release

Why flag to use old logic?

petarjuki7 (dismissed):
Looks good to me, tested with both flags locally, they were both publishing attestations.

petarjuki7 (approved):

PR #751, opened 2025-12-05
feat: query multiple beacon nodes in parallel

petarjuki7:
## Issue Addressed

WAD - Weighted Attestation Data

## Proposed Changes

- Added parallel fetching from all configured beacon nodes instead of sequential fallback
- Implemented scoring algorithm based on checkpoint epochs + head slot proximity
- Added dual timeout strategy: soft timeout at 500ms (return if we have responses), hard timeout at 1s (return unconditionally)

The implementation uses `FuturesUnordered` to query all beacon nodes concurrently and selects the highest-scoring attestation data. Scoring works by summing source and target epochs (base score) plus a proximity bonus based on how close the head slot is to the attestation slot.


anchor/validator_store/src/metadata_service.rs
```diff
@@ -78,6 +88,12 @@ impl<E: EthSpec, T: SlotClock + 'static> MetadataService<E, T> {
     async fn update_metadata(&self) -> Result<(), String> {
         let slot = self.slot_clock.now().ok_or("Failed to read slot clock")?;
 
+        let weighted = false;
```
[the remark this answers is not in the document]
petarjuki7: Resolving this one and the two below, that is WIP for now

anchor/validator_store/src/metadata_service.rs
```diff
@@ -157,4 +173,287 @@ impl<E: EthSpec, T: SlotClock + 'static> MetadataService<E, T> {
...
+                .get_beacon_blocks::<E>(BlockId::Root(block_root))
+                .await
+                .ok()
+                .flatten()
+                .map(|resp| resp.data().slot())
+        })
+        .await
+        .ok()
+        .flatten()
+    }
```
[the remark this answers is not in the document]
petarjuki7: Will do!

anchor/validator_store/src/metadata_service.rs
```diff
@@ -157,4 +173,287 @@ impl<E: EthSpec, T: SlotClock + 'static> MetadataService<E, T> {
...
+        debug!(num_clients, "Starting weighted attestation data fetch");
+
+        // Spawn all fetch requests in parallel
+        let mut futures: FuturesUnordered<_> = clients
+            .into_iter()
+            .map(|(addr, client)| async move {
+                let result = self.fetch_and_score(&client, slot).await;
+                (addr, result)
+            })
+            .collect();
```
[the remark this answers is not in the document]
petarjuki7: Not true, this compiles and works fine, futures can have multiple references to `&self`. This would only fail if `fetch_and_score` required `&mut self`.

anchor/validator_store/src/metadata_service.rs
```diff
@@ -157,4 +173,273 @@ impl<E: EthSpec, T: SlotClock + 'static> MetadataService<E, T> {
...
+                                error = %e,
+                                succeeded,
+                                failed,
+                                "Failed to fetch attestation data"
+                            );
+                        }
+                    }
+
+                    // All responses received, exit early
+                    if succeeded + failed == num_clients {
```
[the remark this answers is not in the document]
petarjuki7: 99% sure this isn't correct, this code runs sequentially once the future is ready. Futures only make progress when polled, and select! only polls them one at a time.

anchor/validator_store/src/metadata_service.rs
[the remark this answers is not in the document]
petarjuki7: Thinking more about this, I think it's not needed. It's going to be a different block almost every time so it's not that helpful, and adds overhead and clutter in the MetadataService struct, and also dealing with Mutexes...

If it turns out it's a performance bottleneck, could be added later

anchor/validator_store/src/metadata_service.rs
```diff
@@ -1,23 +1,34 @@
...
-use types::{ChainSpec, EthSpec};
+use tracing::{debug, error, info, trace, warn};
+use types::{AttestationData, ChainSpec, EthSpec, Hash256, Slot};
 use validator_services::duties_service::DutiesService;
 
 use crate::{AnchorValidatorStore, ContributionWaiter, SlotMetadata};
 
+const SOFT_TIMEOUT: Duration = Duration::from_millis(500);
+const HARD_TIMEOUT: Duration = Duration::from_secs(1);
+const BLOCK_SLOT_LOOKUP_TIMEOUT: Duration = Duration::from_millis(125);
```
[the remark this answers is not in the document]
petarjuki7: Timeouts taken inspiration from the same parts of code in the Go implementation of SSV.
They use:
Soft timeout = 2 seconds
Hard timeout = 5 seconds
Block slot lookup timeout = 1.25 seconds

I found that a bit too long so adjusted to 1 second, 3 seconds and 0.5 seconds respectively

anchor/validator_store/src/metadata_service.rs
```diff
@@ -157,4 +173,290 @@ impl<E: EthSpec, T: SlotClock + 'static> MetadataService<E, T> {
...
+        let mut failed = 0;
+        let mut best_data: Option<ScoredAttestationData> = None;
+        let mut soft_timeout_hit = false;
+
+        // We have two timeouts: a soft timeout and a hard timeout.
+        // At the soft timeout, we return if we have any responses so far.
+        // At the hard timeout, we return unconditionally.
+        // The soft timeout is half the duration of the hard timeout.
+
+        // Collect responses until soft timeout (500ms)
```
dknopik: This comment says 500ms, but the timeout is 1s

anchor/validator_store/src/metadata_service.rs
```diff
@@ -157,4 +173,290 @@ impl<E: EthSpec, T: SlotClock + 'static> MetadataService<E, T> {
...
+        // If no responses yet, wait until hard timeout (1s)
+        if best_data.is_none() && succeeded + failed < num_clients {
+            let remaining = HARD_TIMEOUT.saturating_sub(started.elapsed());
+            let hard_timeout = sleep(remaining);
+            tokio::pin!(hard_timeout);
+
+            loop {
+                tokio::select! {
+                    biased;
+
```
dknopik: I am unsure if the separated loops is the cleanest approach. I think we can combine them, for a cleaner and less error prone approach.

anchor/validator_store/src/metadata_service.rs
```diff
@@ -157,4 +173,290 @@ impl<E: EthSpec, T: SlotClock + 'static> MetadataService<E, T> {
...
+    /// Get the slot number for a given block root with timeout
+    async fn get_block_slot(
+        &self,
+        client: &BeaconNodeHttpClient,
+        block_root: Hash256,
+    ) -> Option<Slot> {
+        tokio::time::timeout(BLOCK_SLOT_LOOKUP_TIMEOUT, async {
+            client
+                .get_beacon_blocks::<E>(BlockId::Root(block_root))
```
dknopik: I think it would be better to only get the block header

dknopik (commented):
Thanks for the PR! I did a first pass, see below.

Additionally, I think we need to integrate this into `sign_attestation` to have it use the attestation from the metadata instead of the passed attestation if the weighting is enabled.

anchor/client/src/cli.rs
```diff
@@ -571,4 +571,15 @@ pub struct Node {
...
+    #[clap(
+        long,
+        help = "Enable parallel querying and scoring of attestation data across multiple beacon nodes. \
+            When enabled, Anchor queries all configured beacon nodes simultaneously and selects \
+            the attestation data with the highest score based on checkpoint epochs and head block \
+            proximity. Only useful when multiple beacon nodes are configured via --beacon-nodes. \
+            Disabled by default.",
+        display_order = 0
+    )]
+    pub with_weighted_attestation_data: bool,
```
diegomrsantos, line 584: Is parallel querying with weighted selection strictly better when we have multiple beacon nodes? If so, then it should just be the implementation, not a user choice. We already have 54 parameters.

anchor/validator_store/src/metadata_service/tests.rs
```diff
@@ -0,0 +1,150 @@
+use types::{AttestationData, Checkpoint, Epoch, FixedBytesExtended, Hash256, Slot};
+
+/// Calculate the score for attestation data based on checkpoint epochs and head slot proximity.
+/// Extracted from fetch_and_score for testing.
+fn calculate_attestation_score(
```
diegomrsantos: Could you please elaborate on the point of this function?

anchor/client/src/cli.rs
petarjuki7: Fair point about the number of parameters... I wouldn't say it's strictly better. SSV describes it as:

> Improves attestation accuracy by scoring responses from multiple Beacon nodes based on epoch and slot proximity. Adds slight latency to duties but includes safeguards (timeouts, retries).

So it's a tradeoff between accuracy and latency so it should probably stay as an option.
More about trade offs [here](https://github.com/ssvlabs/ssv/blob/v2.3.0/beacon/goclient/WAD.md)

anchor/validator_store/src/metadata_service/tests.rs
petarjuki7: It's a more stripped down version of `fetch_and_score` from the `metadata_service.rs`. It's to check the correctness and output of the scoring mechanism, but omits the parts about fetching attestations and block_slots, since they are provided in the test context.

anchor/client/src/cli.rs
diegomrsantos: Thanks, that makes sense. But then the question isn't whether a tradeoff exists; it's whether operators should be required to understand it and make a choice.

1. The latency concern is already addressed by design; the timeouts enforce bounded latency regardless of beacon node response times.
2. The tradeoff isn't operator-dependent: Unlike some configuration choices (e.g., "how much disk space to allocate"), there's no operator-specific context that changes the optimal decision here:
    - 1 beacon node → WAD provides no benefit (nothing to compare)
    - Multiple beacon nodes → WAD provides better accuracy with bounded latency
3. Operators shouldn't need to read SSV WAD documentation: To make an informed choice about this flag, an operator would need to understand attestation scoring algorithms, epoch proximity, timeout implications, etc. That's an unreasonable cognitive load for what should be "run my validator correctly."

diegomrsantos: What we could do is to collect data on mainnet that will help us to make an informed decision.

A practical measure first plan (no new operator config required) could look like this:

We ship it as “shadow mode” for one release:
	•	When --beacon-nodes has 2+ entries, run the parallel fetch/scoring in the background, but still use the current behavior for the actual duty result.
	•	Record metrics about what would have been selected and how long it took.

That gives us mainnet distributions with near-zero functional risk, and we can decide later whether it should become the default implementation. What do you think?

diegomrsantos: cc @dknopik

petarjuki7: I like the idea of collecting metrics first and maybe hiding the flag for the first release, but I would still leave the functionality as accessible if the users (staking pools) which requested it want to use it right away. And for the "general" public we track metrics and see if it should become the default implementation.

anchor/validator_store/src/metadata_service/tests.rs
diegomrsantos: I get why you extracted this, but if the tests use `calculate_attestation_score` (or an identical copy of the formula) to compute the expected result, they’re tautological: the expected value is derived from the implementation, so a bug in the scoring logic won’t be caught. It's also code duplication.  ￼

IMO, it's better to add tests with hard-coded fixtures and hand-computed expected outcomes (epoch dominates, proximity tie-break, missing head_slot behavior).

anchor/client/src/cli.rs
diegomrsantos: Makes sense that some pools want to try this ASAP - but after thinking more about it, I’m worried we’re framing it as a trade-off when it’s really an unvalidated hypothesis. So the real claim here is: waiting/doing more work to pick "better" attestation data will improve outcomes (head correctness / inclusion) more than it harms them via added delay / load. That needs evidence. Adding a new public flag effectively ships a production-path behavior change and asks operators to run the experiment for us.

I’d strongly prefer we measure first: ship instrumentation + run the weighted selection in shadow mode when 2+ beacon nodes are configured (compute scores/timings and export metrics, but keep the current selection for duties). Then we can decide default/kill based on real mainnet distributions - without permanently growing the CLI surface.

If we must unblock specific pools immediately, I’d rather keep it clearly experimental/temporary (e.g. hidden-from-help / config-only) + mandatory metrics, with an explicit revisit and remove/make-default after N releases plan. Also, making it very clear that they are using it at their own risk.

anchor/validator_store/src/metadata_service.rs
```diff
@@ -157,4 +173,248 @@ impl<E: EthSpec, T: SlotClock + 'static> MetadataService<E, T> {
...
+
+        Ok(ScoredAttestationData {
+            client_addr,
+            attestation_data,
+            score,
+        })
+    }
+
+    /// Get the slot number for a given block root with timeout
+    async fn get_block_slot(
```
diegomrsantos, line 426: In Go SSV, they keep a `blockRootToSlotCache` populated from the CL event stream (so most lookups are zero-latency), and when scoring, they retry on cache miss every ~100ms up to a short timeout. The motivation seems to be a race they observed in practice: `attestation_data` can reference a block produced milliseconds ago, and the corresponding event/cache entry may arrive just after the first scoring attempt; plus the `BeaconBlockHeader` lookup can be slow/unreliable, so they wrap it in tight timeouts and retry to avoid hurting duty latency.

diegomrsantos: cc @petarjuki7

anchor/validator_store/src/metadata_service/tests.rs
diegomrsantos: cc @petarjuki7

petarjuki7: Did some refactoring so I can have a proper extracted function in production code to test, let me know what do you think!

anchor/validator_store/src/metadata_service.rs
petarjuki7: I had that comment from Claude Code, had a small discussion somewhere up there in the comments. 

Their cache is populated from HTTP responses as far as I can tell and managing cache state adds complexity. We'd have to add state management to our `MetadataService` struct as far as I can tell with something like a `Arc<Mutex<Hashmap<...>>>`.  With our current setup (parallel queries, 1s soft / 3s hard timeout), we have plenty of opportunities to catch the block somewhere.

I think the simpler approach is fine for now, worst case we miss the proximity bonus on one node but get it from another. I think for now it could be fine without an explicit cache or retrys, but let me know what you think.

diegomrsantos:
@petarjuki7 @dknopik Do we still want to merge this PR?

petarjuki7:
@diegomrsantos since the `metadata_service` in which this PR makes most of it's changes has changed quite a bit since this has been started, I'm probably going to close this PR and open a new one that changes the updated `metadata_service` if that sound ok. I still think this is a good feature to have and something that has been requested of us

diegomrsantos:
> @diegomrsantos since the `metadata_service` in which this PR makes most of it's changes has changed quite a bit since this has been started, I'm probably going to close this PR and open a new one that changes the updated `metadata_service` if that sound ok. I still think this is a good feature to have and something that has been requested of us

I'm cool with that. Let me know what you decide.

petarjuki7:
Closing in favour of #875

PR #753, opened 2025-12-15
feat: add CLI options for BeaconNodeFallback configuration

### LINKED ISSUE #248: Configure `BeaconNodeFallback` via CLI
The `BeaconNodeFallback` struct from LH provides a way to connect to multiple Beacon Nodes. It has two parameters that can be configured beyond the BN URLs: 
- **`broadcast_topics`**: Which requests should be broadcasted to all BNs instead of sending sequentially, stopping with the first successful request.
- **`sync_tolerances`**: Which nodes do we consider synced, based on distance between current slot and current head?

We currently use the default values only. Power users might want to set these values via CLI, as is possible with Lighthouses CLI. Implement this, reusing CLI parameter names from lighthouse.

zemse:
## Issue Addressed

Closes #248

## Proposed Changes

- Add `--broadcast` flag to configure which API topics are sent to all beacon nodes
- Add `--beacon-nodes-sync-tolerances` flag to configure sync distance thresholds

## Additional Info

None

### DIFF
--- a/anchor/client/src/cli.rs
+++ b/anchor/client/src/cli.rs
@@ -4,6 +4,7 @@ use std::{
     path::PathBuf,
 };
 
+use beacon_node_fallback::ApiTopic;
 use clap::{
     Parser,
     builder::{ArgAction, ArgPredicate},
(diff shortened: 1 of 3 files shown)


dknopik (approved):
Thank you for the PR zemse!

LGTM!

PR #754, opened 2025-12-16
feat: implement subnet topology fork transition (SIP-43)

diegomrsantos:
## Issue Addressed

Implements SIP-43 subnet topology fork transition (committee-id based -> operator-id MinHash) and ensures slot-based fork domain/topic selection across sender/validator/QBFT.

### Fork Management (anchor/common/fork)
- ForkSchedule + ForkMonitor with domain types per fork
- Startup grace window handling (if node starts during grace)

### Subnet Service (anchor/subnet_service)
- Fork-aware subnet calculation + subscription orchestration
- Emits TopicEvents; network only subscribes/unsubscribes and updates ENR domain
- Dual-subscribe during warm-up (new fork topics) and drop old topics after grace
- Removed preparation-topic tagging from network

### Network + Topic Handling
- ParsedTopic supports Alan (`/ssv/ssv/<subnet>`) and Boole (`/ssv/<network>/boole/<subnet>`)
- Message sender routes by *message slot* (slot-based subnet/topic prefix)
- Message validator checks subnet/domain by message slot
- IncorrectTopic still maps to Ignore; no prep-topic Accept upgrade

### Signatures + QBFT
- Signature collector uses slot-based domain type per message
- QBFT manager derives MessageId domain from instance slot to avoid stale domains across fork

## Timeline

| Window | Topics Subscribed | Publish |
| --- | --- | --- |
| Warm-up (pre-activation) | current + next fork | current fork |
| Activation | next fork | next fork |
| Grace (post-activation) | current + next fork | next fork |

## Notes

- Based on SIP-43 spec and fork schedule
- Warm-up/grace windows used for gossipsub mesh continuity


anchor/common/ssv_network_config/src/lib.rs
```diff
@@ -38,6 +39,9 @@ pub struct SsvNetworkConfig {
     pub ssv_contract: Address,
     pub ssv_contract_block: u64,
     pub ssv_domain_type: DomainType,
+    /// The epoch at which the subnet topology fork activates.
+    /// `None` means pre-fork operation (no fork configured).
+    pub subnet_topology_fork_epoch: Option<Epoch>,
```
dknopik: This should be `boole_fork_epoch`.

anchor/eth/src/event_processor.rs
```diff
@@ -190,9 +190,10 @@ impl EventProcessor {
 
         let max_seen = self.db.state().get_max_operator_id_seen();
 
-        // Only check for missing operators if we have a previous max (not a migrated database)
+        // Only check for missing operators if we have a previous max (not a migrated database).
+        // Use checked_sub to avoid underflow if operatorId is 0.
         if let Some(max_seen) = max_seen
-            && max_seen != operatorId - 1
+            && operatorId.checked_sub(1) != Some(max_seen)
```
dknopik: This appears to be a unrelated change

anchor/subnet_service/src/routing.rs
```diff
@@ -0,0 +1,736 @@
...
+/// Fork schedule configuration.
+///
+/// Defines when the fork happens and the transition window.
+/// Uses hardcoded transition period: 2 epochs pre-subscribe (PRIOR_WINDOW per SIP-43).
+#[derive(Debug, Clone)]
+pub struct ForkSchedule {
+    /// The epoch at which the fork activates (publishing switches from PreFork to PostFork).
+    /// `None` means pre-fork forever (system operates in PreFork mode indefinitely).
+    fork_epoch: Option<Epoch>,
+}
```
dknopik: IMO, this sort of helper should live elsewhere as other parts of the codebase will also be interested in the fork schedule of Boole and other future forks

anchor/subnet_service/src/routing.rs
```diff
@@ -0,0 +1,736 @@
...
+
+/// Fixed-size schema list with an explicit active length.
+///
+/// This keeps routing decisions allocation-free while still allowing dual-schema
+/// transitions during fork windows.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct SchemaSet {
+    len: usize,
+    schemas: [SubnetSchema; 2],
+}
```
dknopik: This is overkill IMO. Just use a vec.

anchor/validator_store/src/lib.rs
```diff
@@ -116,6 +116,7 @@ pub struct AnchorValidatorStore<T: SlotClock + 'static, E: EthSpec> {
 }
 
 impl<T: SlotClock, E: EthSpec> AnchorValidatorStore<T, E> {
+    // TODO: Refactor to use config struct to reduce parameter count (see #755)
```
dknopik: unrelated change, see also below in the file

anchor/message_receiver/src/manager.rs
```diff
@@ -120,6 +126,18 @@ impl<S: SlotClock + 'static, D: DutiesProvider> MessageReceiver
...
+                // During pre-subscribe window (F-2, F-1), PostFork messages are accepted for
+                // gossipsub propagation but not processed for committee work.
+                if !receiver.router.should_process(matched_schema) {
+                    trace!(
+                        gossipsub_message_id = ?message_id,
+                        ?matched_schema,
+                        "Message accepted for propagation but not processed (pre-subscribe window)"
+                    );
+                    return;
+                }
```
dknopik: This idea does not work.

As validation is stateful, validating but discarding a message received on a post fork topic may cause the message to fail validation if it is then received on a pre fork topic.

anchor/subnet_service/src/subnet.rs
```diff
@@ -90,13 +99,32 @@ impl SubnetId {
...
-        // Safe: x % subnet_count is always < subnet_count, which is a u64.
         let subnet_id: u64 = (id % modulus)
             .try_into()
             .map_err(|_| SubnetCalculationError::SubnetIdOutOfRange)?;
 
         Ok(SubnetId(subnet_id))
     }
+
+    /// Topic prefix for SSV gossipsub topics.
+    pub const TOPIC_PREFIX: &'static str = "ssv.v2.";
```
dknopik: This will also need to be configurable via ssv_network_config

anchor/subnet_service/src/routing.rs
```diff
@@ -0,0 +1,736 @@
...
+use crate::{SubnetCalculationError, SubnetId, calculate_subnet_for_committee};
+
+/// Subnet calculation algorithm schema.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum SubnetSchema {
+    /// Pre-fork schema: subnet derived from committee ID (alan algorithm)
+    PreFork,
+    /// Post-fork schema: subnet derived from operator IDs (MinHash topology algorithm)
+    PostFork,
+}
```
dknopik: This will get very confusing as soon as we underwent multiple forks.

We should either: Rename these variants to e.g. `CommitteeHash` and `MinHash`, and/or add a `enum NetworkUpgrade { Alan, Boole }` somewhere and use that directly in the code below.

dknopik (commented):
Here some remarks from my first pass

anchor/eth/src/event_processor.rs
diegomrsantos: Yes, this is a fix for a potential underflow. Would you like to create a new PR for that?

anchor/subnet_service/src/routing.rs
diegomrsantos: This has been changed

anchor/message_receiver/src/manager.rs
diegomrsantos: Honest nodes won't publish in the post-fork topics before the fork, and after that, we unsubscribe from the pre-fork topics.

anchor/subnet_service/src/subnet.rs
diegomrsantos: done

anchor/subnet_service/src/routing.rs
diegomrsantos: done

anchor/common/ssv_network_config/src/lib.rs
diegomrsantos: N/A anymore

dknopik:
Am I missing something or have most changes to the subnet service been removed?

diegomrsantos:
> Am I missing something or have most changes to the subnet service been removed?

Yes, I'm implementing a simpler design

diegomrsantos:
@claude 1. Topic Format Documentation Incorrect in fork.rs:31 seems correct to me

dknopik (dismissed):
This makes a couple of changes to the common fork related code, so please review that @shane-moore and @petarjuki7 

I think generally it looks good to me, have not tested it though.

I am not sure if a notification based architecture is warranted, but I do not feel strongly. I feel like the network crate will end up being the only subscriber, so I still think concepts like the preparation epoch are more appropriate to live there (or in the subnet service).

I am also on the fence about having the subnet transition in the network crate - I feel like that is more the responsibility of the subnet service. Do not feel too strongly about both of these topics though, so approving for now

diegomrsantos:
> This makes a couple of changes to the common fork related code, so please review that @shane-moore and @petarjuki7
> 
> I think generally it looks good to me, have not tested it though.
> 
> I am not sure if a notification based architecture is warranted, but I do not feel strongly. I feel like the network crate will end up being the only subscriber, so I still think concepts like the preparation epoch are more appropriate to live there (or in the subnet service).
> 
> I am also on the fence about having the subnet transition in the network crate - I feel like that is more the responsibility of the subnet service. Do not feel too strongly about both of these topics though, so approving for now

Thanks for the review!

You're right that the network crate is likely the only subscriber, but I think the main value of this design isn't about multiple subscribers - it's about modularization and boundaries.

With this approach, all fork timing logic (preparation windows, epoch calculations, FORK_PREPARATION_EPOCHS) is encapsulated in the fork crate. The network crate doesn't need to understand any of it - it just reacts to explicit Preparing and Activated events.

The alternative would mean:
  - Network needs to poll the slot clock and check in_preparation_window(current_epoch)
  - Fork-related constants and timing logic spread across modules
  - Changes to fork behavior (e.g., different preparation window duration) would touch multiple crates

With the current design, the fork crate decides when transitions happen, and network decides how to react. Each module has one job, and changes to fork timing are isolated to one place.

That said, I'm open to revisiting if this proves to be over-engineered in practice.

diegomrsantos:
Regarding putting this in subnet_service - the existing architecture already has subnet_service emitting events (SubnetEvent::Join/Leave) while network handles the actual subscriptions.  ForkPhase follows that same pattern, so it's consistent with how we already do things.

petarjuki7:
I went through the code and it makes sense, the logic looks correct to me.

I was focusing on exploring how the ENRs work with these changes and discovered that we change the `DomainType` between Alan and Boole (each fork has its own domain type in `ForkConfig`), but it seems we don't actually update the ENR's `domaintype` field when the fork activates.

This is important because in [start_query()](https://github.com/sigp/anchor/blob/unstable/anchor/network/src/discovery.rs#L431-L438) we filter peers based on domain type.

So after the fork, nodes with stale ENRs (still advertising Alan's domain type) won't be able to discover nodes that have updated.

This can be added to this PR, or after merge I can open a PR with the fix, either way works fine by me.

shane-moore:
should we consider validating `msgId` when new messages are received over wire to enforce proper fork is in the `msgId`?
https://github.com/sigp/anchor/blob/2b39ae41bb380f67727aef3d897d8c4358919736/anchor/message_validator/src/lib.rs#L321-L330

we could use this unused error type: https://github.com/sigp/anchor/blob/2b39ae41bb380f67727aef3d897d8c4358919736/anchor/message_validator/src/lib.rs#L84

seems defensive to validate `msg_id.domain()` against the expected fork domain for incoming messages

diegomrsantos:
@claude do a general code review for the whole implementation

diegomrsantos:
> should we consider validating `msgId` when new messages are received over wire to enforce proper fork is in the `msgId`?
> 
> https://github.com/sigp/anchor/blob/2b39ae41bb380f67727aef3d897d8c4358919736/anchor/message_validator/src/lib.rs#L321-L330
> 
> we could use this unused error type:
> 
> https://github.com/sigp/anchor/blob/2b39ae41bb380f67727aef3d897d8c4358919736/anchor/message_validator/src/lib.rs#L84
> 
> seems defensive to validate `msg_id.domain()` against the expected fork domain for incoming messages

Great catch, thanks! Done.

diegomrsantos:
> I went through the code and it makes sense, the logic looks correct to me.
> 
> I was focusing on exploring how the ENRs work with these changes and discovered that we change the `DomainType` between Alan and Boole (each fork has its own domain type in `ForkConfig`), but it seems we don't actually update the ENR's `domaintype` field when the fork activates.
> 
> This is important because in [start_query()](https://github.com/sigp/anchor/blob/unstable/anchor/network/src/discovery.rs#L431-L438) we filter peers based on domain type.
> 
> So after the fork, nodes with stale ENRs (still advertising Alan's domain type) won't be able to discover nodes that have updated.
> 
> This can be added to this PR, or after merge I can open a PR with the fix, either way works fine by me.

Great catch, thanks! Done.

diegomrsantos:
I ran it locally and it works fine

diegomrsantos:
I tested it with ssv-mini. This PR correctly handles topic transitions and ENR updates during the fork, but the `domain_type` used for signing messages in `SignatureCollectorManager` and `QbftManager` is never updated. This needs to be addressed before the PR can work correctly for fork transitions.

anchor/subnet_service/src/topic.rs
```diff
@@ -0,0 +1,314 @@
...
+///
+/// This is a convenience function for cases where you only care about
+/// the subnet ID, not the fork (e.g., tracking peer subscriptions).
+///
+/// # Warning
+///
+/// This function discards fork information. For message validation or routing
+/// decisions that depend on the fork, use [`parse_topic`] instead to get both
+/// the subnet ID and the fork.
+#[must_use = "Fork information is discarded. Use parse_topic() if fork context matters for validation or routing"]
```
dknopik, line 93: The must use attribute is for indicating that a return value must be used. This is used to guard against inappropriate API usage, e.g. to avoid having the user expect a side effect on pure functions, or to avoid skipped error checks. 

While `must_use` is kinda appropriate here, the info string should indicate why the return value must be used instead of suggesting alternative functions.

anchor/subnet_service/src/topic.rs
```diff
@@ -0,0 +1,314 @@
...
+pub fn create_topic(prefix: &str, subnet: SubnetId) -> IdentTopic {
+    IdentTopic::new(format!("{}{}", prefix, *subnet))
+}
+
+/// Result of parsing a topic hash.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct ParsedTopic {
+    pub subnet_id: SubnetId,
+    pub fork: Fork,
+}
```
dknopik: what do you think about also parsing the network name?

anchor/subnet_service/src/topic.rs
```diff
@@ -0,0 +1,314 @@
...
+            // Parse fork name dynamically to support future forks
+            let fork: Fork = parts[3].parse().ok()?;
+            // Alan uses legacy format, not this path
+            if fork == Fork::Alan {
+                return None;
+            }
+            let subnet_num: u64 = parts[4].parse().ok()?;
+            let subnet_id = parse_and_validate_subnet(subnet_num)?;
+            return Some(ParsedTopic { subnet_id, fork });
+        }
```
dknopik: style suggestion: I like pattern matching for stuff like this:

```rust
if let ["", "ssv", network, fork, subnet_id] = parts.as_slice() {
```

anchor/src/main.rs
```diff
@@ -132,7 +133,12 @@ fn start_anchor(
         e
     })?;
 
-    config.network.domain_type = config.global_config.ssv_network.ssv_domain_type;
+    config.network.domain_type = config
+        .global_config
+        .ssv_network
+        .fork_schedule
+        .domain_type(Fork::Alan)
+        .expect("Alan fork must have domain type in schedule");
```
dknopik: Return an error instead of panicing

anchor/common/fork/src/schedule.rs
```diff
@@ -3,65 +3,159 @@
...
-        Self { activations }
+    pub fn new(baseline_domain_type: DomainType, network_name: &str) -> Self {
+        let mut configs = BTreeMap::new();
+        configs.insert(
+            Fork::Alan,
+            ForkConfig::new(
+                Fork::Alan,
+                Epoch::new(0),
+                baseline_domain_type,
+                network_name,
```
dknopik: What do you think about putting this not into the `BTreeMap`? This has the advantage that the `domain_type` function can return `DomainType` instead of `Option<DomainType>`.

The network name is already next to the `BTreeMap`, I dont think it needs to be overridable through forks

anchor/common/ssv_network_config/src/lib.rs
```diff
@@ -9,21 +9,54 @@ use alloy::primitives::Address;
...
+    /// Create a new SSV network name.
+    pub fn new(name: impl Into<String>) -> Self {
+        Self { name: name.into() }
+    }
+
+    /// Get the network name (e.g., "mainnet", "holesky").
+    pub fn name(&self) -> &str {
+        &self.name
+    }
+}
```
dknopik: Not sure if this wrapper is necessary, feels a little overkill

anchor/client/src/lib.rs
```diff
@@ -359,13 +341,48 @@ impl Client {
...
+                )
+            } else {
+                NetworkDatabase::new(
+                    &config.global_config.data_dir.database_file(),
+                    &pubkey,
+                    domain_type,
+                )
+            }
+            .map_err(|e| format!("Unable to open Anchor database: {e}"))?,
+        );
```
dknopik: Hmmmh, this is problematic. 

The domain type is used to make sure the database is not used across networks - we did not have the network name before. So I think we have to either migrate to the network name, or keep using the alan domain type here.

Like this, the database will refuse to open after the fork-

anchor/common/fork/src/schedule.rs
```diff
@@ -73,38 +167,57 @@ impl ForkSchedule {
...
+
+    /// Get the domain type for a fork, if scheduled.
+    pub fn domain_type(&self, fork: Fork) -> Option<DomainType> {
+        self.configs.get(&fork).map(|config| config.domain_type)
+    }
+
+    /// Get the full configuration for a fork, if scheduled.
+    pub fn config(&self, fork: Fork) -> Option<&ForkConfig> {
+        self.configs.get(&fork)
     }
```
dknopik, line 192: It would be nice to also have `active_fork_config` to avoid `expect` elsewhere.

anchor/message_receiver/src/manager.rs
```diff
@@ -69,17 +70,37 @@ impl<S: SlotClock + 'static, D: DutiesProvider> MessageReceiver
...
+                        ValidationResult::PostDecodeFailure(ValidationFailure::IncorrectTopic, _)
+                    );
+
+                    if is_topology_failure {
+                        // IncorrectTopic normally results in Ignore, upgrade to Accept
+                        // because the message might be valid under new fork rules
+                        debug!("Converting Ignore to Accept for topology mismatch on preparation topic");
+                        action = MessageAcceptance::Accept;
+                    }
+                }
```
dknopik: As most of the validation was not executed, we should not process these messages.

EDIT:

Wait, this can not happen anyway as the validation is skipped here, right? https://github.com/sigp/anchor/pull/754/changes#diff-e78a617e82f41db85b09de526249f65676c24a328fe1ac1c2475f33cd8449aafR518-R524

anchor/client/src/lib.rs
diegomrsantos: Should we use the network name instead?

diegomrsantos: let me know what you think [21c9dde22ee5f362b9ab44e3c81851a6c00a8b1e](https://github.com/sigp/anchor/pull/754/changes/21c9dde22ee5f362b9ab44e3c81851a6c00a8b1e)

dknopik: looks good :)

dknopik: thanks!

anchor/common/fork/src/schedule.rs
diegomrsantos: I believe the `expect` in this case is reasonable; we can't recover from this. Wdyt?

anchor/src/main.rs
diegomrsantos: I believe the `expect` in this case is reasonable; we can't recover from this. Wdyt?

anchor/subnet_service/src/topic.rs
diegomrsantos: done, thanks!

anchor/common/ssv_network_config/src/lib.rs
diegomrsantos: used `pub type SsvNetworkName = String` instead

anchor/src/main.rs
dknopik: In general, we should be able to handle bad user input gracefully - e.g. by returning an error from the main function instead of panicing, which might show a backtrace and make the user think there is a programming error.

diegomrsantos: This isn't user input

dknopik: the fork schedule can be user input through the  `--testnet-dir` CLI flag

diegomrsantos: The Alan fork isn't a user input

anchor/network/src/network.rs
```diff
@@ -483,7 +476,7 @@ impl<R: MessageReceiver> Network<R> {
             .set_topic_params(topic.clone(), topic_score_params)
         {
             Ok(_) => {
-                debug!(
```
shane-moore, line 427: downgrading to `trace`  and removing the extra debugs should make it easier to differentiate between signal and noise when going through logs

anchor/common/fork/src/fork.rs
```diff
@@ -45,6 +48,17 @@ impl Fork {
...
     }
+
+    /// Get the topic prefix for a given fork and subnet.
+    ///
+    /// - Alan fork: returns the legacy prefix `ssv.v2.`
+    /// - Post-Alan forks: returns `/ssv/{network}/{fork}/` format
+    pub fn topic_prefix(&self, network_name: &str) -> String {
+        match self {
+            Fork::Alan => ALAN_TOPIC_PREFIX.to_string(),
+            _ => format!("/ssv/{}/{}/", network_name, self.name()),
```
shane-moore, line 59: I normally dislike `_` in match statements because it can make it harder in the future to find parts of the codebase that need to be refactored, i.e. if they were to update the topic prefix again in the future. however, I find it unlikely that will happen anytime soon, if ever, so I'm cool with this

anchor/subnet_service/src/subscriptions.rs
```diff
@@ -1,124 +1,163 @@
...
 use slot_clock::SlotClock;
+use ssv_types::OperatorId;
 use tokio::{sync::mpsc, time::sleep};
 use tracing::{debug, error, warn};
-use types::{EthSpec, Slot};
+use types::EthSpec;
+
+use crate::{SUBNET_COUNT, SubnetId, SubnetServiceError, TopicEvent, service::SubnetService};
+
+pub(crate) struct ServiceState {
```
shane-moore, line 20: makes sense to remove all the `current_*` and `previous_*` logic and have it fork based. easier to reason about

anchor/client/src/lib.rs
```diff
@@ -375,7 +375,7 @@ impl Client {
         );
 
         // Create fork phase channel for fork transition events
-        let (fork_phase_tx, mut fork_phase_rx) = mpsc::channel(16);
+        let (fork_phase_tx, fork_phase_rx) = async_broadcast::broadcast(16);
```
shane-moore, line 378: agree that `broadcast` makes sense for multi-consumer scenarios

anchor/subnet_service/src/service.rs
```diff
@@ -141,48 +141,6 @@ impl<S: SlotClock> SubnetService<S> {
     pub fn router(&self) -> &TopicRouter {
         &self.router
     }
-
-    /// Get the active fork for a given epoch.
-    ///
-    /// This exposes the fork schedule's active fork lookup for use by other components
-    /// (e.g., message validator) that need to determine which fork rules apply.
-    pub fn active_fork(&self, epoch: Epoch) -> Fork {
```
shane-moore: agree with removing these wrappers

.github/custom/local-testnet/params.yaml
```diff
@@ -46,6 +46,8 @@ nodes:
   anchor:
     count: 4
 
+boole_epoch: 3
```
shane-moore, line 49: i'm seeing high sync and attestation % by epoch 3 when running ssv-mini so I think this is fine

shane-moore:
overall, the latest changes to this pr are simplifications, and don't impact the core functionality, i.e. topic renaming and dual subscription, all good

dknopik (approved):
Merging this for now. As always, anything herein can be changed and improved upon later :)

diegomrsantos:
@claude review this PR

PR #756, opened 2025-12-17
feat: add CORS support for HTTP API and metrics endpoints

### LINKED ISSUE #249: Set CORS headers via CLI
[CORS headers](https://en.wikipedia.org/wiki/Cross-origin_resource_sharing) are important for users that want to access API or metrics data via a browser. 

Add CLI options to add CORS headers to metrics and API responses.

zemse:
## Issue Addressed

Closes #249

## Proposed Changes

- Add `--metrics-allow-origin` CLI flag
- Implement CORS for HTTP API, the `--http-allow-origin` flag existed but wasn't wired up.

## Additional Info

None

### DIFF
--- a/anchor/client/src/config.rs
+++ b/anchor/client/src/config.rs
@@ -135,7 +135,7 @@ impl Config {
 
 /// Returns a `Default` implementation of `Self` with some parameters modified by the supplied
 /// `cli_args`.
-pub fn from_cli(cli_args: &Node, global_config: GlobalConfig) -> Result<Config, String> {
+pub fn from_cli(mut cli_args: Node, global_config: GlobalConfig) -> Result<Config, String> {
     let mut config = Config::new(global_config);
 
     config.key_file = cli_args.key_file.clone();
@@ -157,7 +157,7 @@ pub fn from_cli(cli_args: &Node, global_config: GlobalConfig) -> Result<Config, 
     config.disable_slashing_protection = cli_args.disable_slashing_protection;
 
     // Network related
-    config.network.listen_addresses = parse_listening_addresses(cli_args)?;
+    config.network.listen_addresses = parse_listening_addresses(&cli_args)?;
 
     for addr in cli_args.boot_nodes.clone() {
         match addr.parse() {
@@ -249,11 +249,8 @@ pub fn from_cli(cli_args: &Node, global_config: GlobalConfig) -> Result<Config, 
         config.http_api.listen_port = port;
     }
 
-    if let Some(allow_origin) = &cli_args.http_allow_origin {
-        let header_value = allow_origin
-            .parse()
-            .map_err(|_| "Invalid allow-origin value")?;
-        config.http_api.allow_origin = Some(AllowOrigin::exact(header_value));
+    if let Some(allow_origin) = cli_args.http_allow_origin.take() {
+        config.http_api.allow_origin = Some(AllowOrigin::exact(allow_origin));
     }
 
     // Prometheus metrics HTTP server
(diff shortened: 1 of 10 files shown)


anchor/http_metrics/src/lib.rs
```diff
@@ -60,12 +60,19 @@ impl Default for Config {
...
 }
 
-fn create_router<E: EthSpec>(shared_state: Arc<RwLock<Shared<E>>>) -> Router {
+fn create_router<E: EthSpec>(
+    shared_state: Arc<RwLock<Shared<E>>>,
+    allow_origin: Option<String>,
+) -> Router {
+    let origin = allow_origin
+        .map(|o| AllowOrigin::exact(o.parse().expect("validated in config")))
+        .unwrap_or(AllowOrigin::any());
```
jxs: see comment above, this seems un-required, we should be able to receive `allow_origin` as  `AllowOrigin` to not parsed it again here

anchor/client/src/config.rs
```diff
@@ -271,6 +271,14 @@ pub fn from_cli(cli_args: &Node, global_config: GlobalConfig) -> Result<Config,
...
     }
 
+    if let Some(allow_origin) = &cli_args.metrics_allow_origin {
+        // Pre-validate the config value to give feedback to the user on node startup.
+        hyper::header::HeaderValue::from_str(allow_origin)
+            .map_err(|_| "Invalid metrics-allow-origin value")?;
+
+        config.http_metrics.allow_origin = Some(allow_origin.to_string());
+    }
+
```
jxs: we should be able to parse `cli_args.allow_origin` and `cli_args.metrics.allow_origin` as `tower_http::cors::AllowOrigin`  to store both as `AllowOrigin` in the config and not have the need to parse them again later

anchor/http_api/src/lib.rs
```diff
@@ -38,11 +40,19 @@ pub async fn run(config: Config, shared_state: Arc<RwLock<Shared>>) -> Result<()
         return Ok(());
     }
 
+    let origin = config
+        .allow_origin
+        .map(|o| AllowOrigin::exact(o.parse().expect("validated in config")))
+        .unwrap_or(AllowOrigin::any());
```
jxs: see comment above, this seems unrequired, we can have `config.allow_origin`'s type to be `AllowOrigin` to not have to be parsed again here.

jxs (commented):
Hi, thanks for getting into this!
Left some comments

jxs (dismissed):
LGTM thanks!

anchor/http_api/src/config.rs
```diff
@@ -19,7 +19,7 @@ impl Default for Config {
             enabled: false,
             listen_addr: IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1)),
             listen_port: 5062,
-            allow_origin: None,
+            allow_origin: AllowOrigin::any(),
```
dknopik: I am not sure if `any` is a good default.

E.g. Lighthouse defaults to the listen address and port: https://github.com/sigp/lighthouse/blob/f3fd1f210b2f4ed7d208f81f9a09e1edced3bb3d/beacon_node/http_api/src/lib.rs#L340-L344

dknopik (changes requested):
This bumps all dependencies, so probably `cargo update` was run. Usually we do those in separate PRs - please restore the `Cargo.lock` from unstable

anchor/http_api/src/config.rs
jxs: yeah makes sense thanks Daniel.
I suggested `any` based on following the default behavior of `None` but it makes more sense to use the listen address and port as lighthouse does

jxs (dismissed):
LGTM

dknopik (approved):
LGTM, thanks @zemse and @jxs!

PR #759, opened 2025-12-17
feat: add gossipsub topic validation

diegomrsantos:
## Issue Addressed

N/A - Feature parity with go-ssv.

## Proposed Changes

Adds validation that SSV messages are published to the correct gossipsub topic.

**Why this matters:** Each SSV committee has a deterministic subnet assignment based on its committee ID. A message published to the wrong topic indicates either a bug or misconfigured node. Without validation, these messages would be processed normally despite being incorrectly routed.

**How it works:** When a message arrives, we extract the committee ID from the message content and calculate the expected subnet using `SubnetId::from_committee_alan()`. If the topic the message arrived on doesn't match, we return `Ignore` (not `Reject`) to avoid harshly penalizing nodes that may simply be misconfigured.

**Changes:**
- Add `topic_to_subnet()` helper in subnet_service for parsing topic strings
- Thread `TopicHash` through the validation chain (network → message_receiver → validator)
- Add `validate_topic()` helper that compares expected vs received subnet
- Validate outgoing messages before sending (defense in depth)

## Additional Info

This matches [go-ssv's implementation](https://github.com/ssvlabs/ssv/blob/main/message/validation/validation.go#L218-L234) which performs the same validation in `committeeChecks()`.

### DIFF
--- a/anchor/message_sender/src/network.rs
+++ b/anchor/message_sender/src/network.rs
@@ -132,23 +132,27 @@ impl<S: SlotClock + 'static, D: DutiesProvider> NetworkMessageSender<S, D> {
 
     fn do_send(&self, message: SignedSSVMessage, committee_id: CommitteeId) {
         let message_bytes = message.as_ssz_bytes();
+        let subnet = SubnetId::from_committee_alan(committee_id, self.subnet_count);
 
-        if let Some(validator) = self.validator.as_ref()
-            && let Err(err) = validator.validate(&message_bytes).as_result()
-        {
-            // `Reject` is more severe and can be punished by other peers. We should not have
-            // created this message ever, while `Ignore` can be triggered simply because the message
-            // is irrelevant by now.
-            if let MessageAcceptance::Reject = MessageAcceptance::from(err) {
-                warn!(?err, "Validation of outgoing message failed (Reject)");
-                debug!(msg = %message, "Failing message");
-            } else {
-                debug!(?err, "Validation of outgoing message failed (Ignore)");
+        if let Some(validator) = self.validator.as_ref() {
+            // Create topic for validation
+            let topic_string = format!("ssv.v2.{}", *subnet);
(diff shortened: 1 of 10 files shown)


shane-moore (approved):
def could use some more eyes on it but lgtm man, pretty straight forward and clean!

diegomrsantos:
> def could use some more eyes on it but lgtm man, pretty straight forward and clean!

Thank you!

diegomrsantos:
Closing as superseded by #754 (subnet topology fork transition).

PR #754 includes all the topic validation functionality from this PR:
- `TopicHash` threaded through validation chain
- `IncorrectTopic` validation failure
- Outgoing message validation before sending

Additionally, #754 provides **fork-aware** topic validation that handles both PreFork and PostFork subnet calculation schemas during the fork transition, using the centralized `SubnetRouter`.

anchor/message_sender/src/network.rs
```diff
@@ -132,23 +132,27 @@ impl<S: SlotClock + 'static, D: DutiesProvider> NetworkMessageSender<S, D> {
 
     fn do_send(&self, message: SignedSSVMessage, committee_id: CommitteeId) {
         let message_bytes = message.as_ssz_bytes();
+        let subnet = SubnetId::from_committee_alan(committee_id, self.subnet_count);
+
+        if let Some(validator) = self.validator.as_ref() {
+            // Create topic for validation
+            let topic_string = format!("ssv.v2.{}", *subnet);
+            let topic = gossipsub::TopicHash::from_raw(topic_string);
```
dknopik, line 140: Can we move this to a common location in order to not hardcode the topic string pattern multiple times? Maybe as a method on `SubnetId`?

dknopik (commented):
nice, just one suggestion

anchor/message_sender/src/network.rs
diegomrsantos: This PR has been closed

Issue #758, opened 2025-12-17
QBFT: PROPOSER round timeout uses slot-based timing instead of relative timing


moshe-blox:
## Summary

Anchor's QBFT round timeout for PROPOSER duties is calculated from **slot start time**, while Go SSV uses a **relative timeout from QBFT start**. This caused a missed block proposal on mainnet.

## The Difference

**Go SSV** (`protocol/v2/qbft/roundtimer/timer.go:110-114`):
```go
default:  // PROPOSER
    if round <= t.timeoutOptions.quickThreshold {
        return t.timeoutOptions.quick  // Returns 2s directly, no slot-based calc
    }
```

**Anchor** (`qbft_manager/src/timeout.rs` + `validator_store/src/lib.rs:307`):
```rust
// PROPOSER start_time = slot_start + 0
let start_time = self.get_instant_in_slot(slot, Duration::ZERO)?;

// Timeout = start_time + round * 2s = slot_start + 2s
start_time.checked_add(Duration::from_secs(round * QUICK_TIMEOUT))
```

| Implementation | PROPOSER Round 1 Timeout |
|----------------|--------------------------|
| Go SSV | 2 seconds from QBFT start |
| Anchor | slot_start + 2 seconds (absolute) |

## Impact

For PROPOSER duties, QBFT starts ~1.5s into the slot (after pre-consensus + block fetch). With Anchor's slot-based timing:
- Round 1 timeout fires at `slot_start + 2s`
- But QBFT started at `slot_start + 1.5s`
- Only ~500ms available for round 1 instead of 2s

This caused a missed proposal on mainnet slot 13263047 - Anchor timed out and moved to round 2 before collecting commit quorum, rejecting late-arriving commits as `PastRound`.

## Recommendation

Align with Go SSV behavior for PROPOSER duties - use relative timeout from QBFT start rather than slot-based. Go's approach is battle-tested with most operators running it today.

However, I'm open to discussion if you think that'd be wrong.

## Future

Personally, I think `slot + 2s` is a bit too short for proposals because it sometimes takes over 1.5s just to fetch a block, and we'd be giving up on an nearly-completed duty just 500ms later, and delay the time-sensitive proposal to round 2.

However, I also don't like the Go SSV approach which misaligns round timeouts between the operators.

[...]

PR #762, opened 2025-12-19
fix: Do not use fixed start time for proposals

dknopik:
## Issue Addressed

- #758

## Proposed Changes

Instead of using the slot start as start time (which is later used for determining timeouts), use the current time.

### DIFF
--- a/anchor/validator_store/src/lib.rs
+++ b/anchor/validator_store/src/lib.rs
@@ -304,7 +304,8 @@ impl<T: SlotClock, E: EthSpec> AnchorValidatorStore<T, E> {
 
         // first, we have to get to consensus
         let timer = metrics::start_timer_vec(&metrics::CONSENSUS_TIMES, &[metrics::BLOCK]);
-        let start_time = self.get_instant_in_slot(slot, Duration::ZERO)?;
+        // proposals do not use a fixed start time: https://github.com/sigp/anchor/issues/758
+        let start_time = Instant::now();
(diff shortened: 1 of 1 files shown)


anchor/validator_store/src/lib.rs
```diff
@@ -304,7 +304,8 @@ impl<T: SlotClock, E: EthSpec> AnchorValidatorStore<T, E> {
 
         // first, we have to get to consensus
         let timer = metrics::start_timer_vec(&metrics::CONSENSUS_TIMES, &[metrics::BLOCK]);
-        let start_time = self.get_instant_in_slot(slot, Duration::ZERO)?;
+        // proposals do not use a fixed start time: https://github.com/sigp/anchor/issues/758
+        let start_time = Instant::now();
```
diegomrsantos, line 308: Is this equivalent to "a relative timeout from QBFT start"?

dknopik: Thanks, I realized now that the Go-SSV code appears to actually also use timeouts relative to round changes, e.g. induced by justified proposals: https://github.com/ssvlabs/ssv/blob/8808f67d0660821aa3991017272655fbecfbfa04/protocol/v2/qbft/instance/proposal.go#L35-L38

I believe we therefore need two timer modes - the existing one and one to mirror the behaviour above.

Therefore this fix is not sufficient. Closing it. Thanks!

PR #766, opened 2026-01-02
chore(deps): update alloy to 1.2.1 and fix ruint security vulnerability

diegomrsantos:
## Issue Addressed

Fixes CI failure: https://github.com/sigp/anchor/actions/runs/20666620673/job/59340149125

## Proposed Changes

- Update `alloy` from 1.0.22 to 1.2.1
- Pin `libp2p-core` and `libp2p-swarm` to exact versions to resolve patch ambiguity
- Updates `ruint` from 1.17.0 to 1.17.2, fixing RUSTSEC-2025-0137 (unsoundness in `reciprocal_mg10`)

## Additional Info

The `ruint` vulnerability (RUSTSEC-2025-0137) was causing `cargo audit` to fail in CI. The fix requires `ruint >= 1.17.1`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -15,7 +15,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d122413f284cf2d62fb1b7db97e02edb8cda96d769b16e443a4f6195e35662b0"
 dependencies = [
  "crypto-common",
- "generic-array 0.14.9",
+ "generic-array 0.14.7",
 ]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


jking-aus (approved):
lgtm

shane-moore (commented):
@diegomrsantos, i was able to resolve this over in lighthouse a few days ago in [PR](https://github.com/sigp/lighthouse/pull/8617) by bumping the `ruint` version. here, did you bump the alloy version, and then got the updated `ruint` as a side effect? jc. it looks like the build passes so don't see an issue with this approach

I'm unfamiliar with the `libp2p` dependency source overrides so hard for me to comment on whether pinning the versions is necessary or otherwise

diegomrsantos:
> @diegomrsantos, i was able to resolve this over in lighthouse a few days ago in [PR](https://github.com/sigp/lighthouse/pull/8617) by bumping the `ruint` version. here, did you bump the alloy version, and then got the updated `ruint` as a side effect? jc. it looks like the build passes so don't see an issue with this approach
> 
> I'm unfamiliar with the `libp2p` dependency source overrides so hard for me to comment on whether pinning the versions is necessary or otherwise

I understand that `ruint` is a transitive dependency from `alloy`. I'd appreciate a double check

shane-moore:
@diegomrsantos, yep, `ruint` is transitive dependency of alloy.  so makes sense that bumping `alloy` resolved the issue here.  i went with the `cargo update ruint` approach just to fix the vulnerability at the lowest dependency possible as to keep blast radius small

PR #774, opened 2026-01-08
feat: add generic Fork infrastructure for SSV protocol upgrades

diegomrsantos:
## Issue Addressed

Preparation for Boole fork support. Provides complete fork infrastructure that other developers can use immediately for any Boole feature (topic naming, message formats, subnet topology, etc.) without additional configuration.

### Motivation

PR #754 mixes generic fork infrastructure with Boole-specific subnet changes. This coupling makes the code harder to review and maintain. As Daniel noted, temporal names like "PreFork/PostFork" become confusing after multiple upgrades.

### Mental Model

A **fork** is a protocol-wide upgrade that may change multiple subsystems simultaneously (subnet topology, topic naming, message formats, etc.). We separate two concerns:

- **"When"** - A schedule manages activation epochs and transition timing
- **"What"** - Each subsystem queries the active fork to determine its behavior

This separation allows:
- Adding new forks by extending a single enum
- Each subsystem to implement fork-specific logic independently  
- Clean transition handling (dual-subscribe windows before activation)

### Design Decisions

**Named forks over temporal names**: We use `Genesis → Alan → Boole` as stable identifiers. This avoids the confusion of `PreFork/PostFork` which become ambiguous after multiple upgrades.

**Dedicated fork crate**: All fork-related code is consolidated in `anchor/common/fork/`:
- `fork.rs`: Fork enum (Genesis, Alan, Boole)
- `schedule.rs`: ForkSchedule for managing activation epochs
- `monitor.rs`: Standalone task for logging fork transitions

**Automatic predecessor activation**: When scheduling Boole at epoch N, all previous forks (Genesis, Alan) are automatically marked as active from epoch 0.

### Configuration

- **Built-in networks** (mainnet, holesky, hoodi): Default to Alan fork at epoch 0
- **Custom networks**: Specify Boole epoch via `ssv_boole_fork_epoch.txt` file in the network config directory

### Fork Monitor

A standalone task monitors and logs fork transitions:
- Logs current fork at startup
- Logs when entering the preparation window (1 epoch before fork)
- Logs when a fork activates

### Usage

Components that need fork awareness can query:

```rust
let current_fork = fork_schedule.active_fork(current_epoch);
match current_fork {
    Fork::Alan => { /* current behavior */ }
    Fork::Boole => { /* new behavior */ }
    _ => {}
}
```

## Additional Info

- Full test coverage
- No breaking changes to existing behavior
- Ready for other Boole features to build on this infrastructure

### DIFF
--- a/anchor/common/fork/src/fork.rs
+++ b/anchor/common/fork/src/fork.rs
@@ -17,10 +17,7 @@ use serde::{Deserialize, Serialize};
 #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize, Deserialize)]
 #[serde(rename_all = "lowercase")]
 pub enum Fork {
-    /// The initial SSV protocol version.
-    Genesis,
-
-    /// The Alan fork.
+    /// The Alan fork (initial SSV mainnet version).
     ///
     /// Characteristics:
     /// - Subnet topology: `committee_id % 128`
@@ -38,18 +35,12 @@ pub enum Fork {
 impl Fork {
     /// Returns all known forks in chronological order.
     pub const fn all() -> &'static [Fork] {
-        &[Fork::Genesis, Fork::Alan, Fork::Boole]
-    }
-
-    /// Returns the genesis fork (the first fork).
-    pub const fn genesis() -> Self {
-        Fork::Genesis
+        &[Fork::Alan, Fork::Boole]
     }
 
     /// Returns the name of this fork as a string.
     pub const fn name(&self) -> &'static str {
         match self {
-            Fork::Genesis => "genesis",
             Fork::Alan => "alan",
             Fork::Boole => "boole",
         }
@@ -67,7 +58,6 @@ impl std::str::FromStr for Fork {
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         match s.to_lowercase().as_str() {
-            "genesis" => Ok(Fork::Genesis),
             "alan" => Ok(Fork::Alan),
             "boole" => Ok(Fork::Boole),
             _ => Err(format!("Unknown fork: {s}")),
@@ -81,28 +71,23 @@ mod tests {
 
     #[test]
     fn test_fork_ordering() {
-        assert!(Fork::Genesis < Fork::Alan);
         assert!(Fork::Alan < Fork::Boole);
-        assert!(Fork::Genesis < Fork::Boole);
     }
 
     #[test]
     fn test_fork_names() {
-        assert_eq!(Fork::Genesis.name(), "genesis");
         assert_eq!(Fork::Alan.name(), "alan");
         assert_eq!(Fork::Boole.name(), "boole");
     }
 
     #[test]
     fn test_fork_display() {
-        assert_eq!(format!("{}", Fork::Genesis), "genesis");
         assert_eq!(format!("{}", Fork::Alan), "alan");
         assert_eq!(format!("{}", Fork::Boole), "boole");
     }
 
     #[test]
     fn test_fork_parse() {
-        assert_eq!("genesis".parse::<Fork>().unwrap(), Fork::Genesis);
         assert_eq!("alan".parse::<Fork>().unwrap(), Fork::Alan);
         assert_eq!("Boole".parse::<Fork>().unwrap(), Fork::Boole);
         assert_eq!("ALAN".parse::<Fork>().unwrap(), Fork::Alan);
@@ -112,14 +97,8 @@ mod tests {
     #[test]
     fn test_all_forks() {
         let all = Fork::all();
-        assert_eq!(all.len(), 3);
-        assert_eq!(all[0], Fork::Genesis);
-        assert_eq!(all[1], Fork::Alan);
-        assert_eq!(all[2], Fork::Boole);
-    }
-
-    #[test]
-    fn test_genesis() {
-        assert_eq!(Fork::genesis(), Fork::Genesis);
+        assert_eq!(all.len(), 2);
+        assert_eq!(all[0], Fork::Alan);
+        assert_eq!(all[1], Fork::Boole);
     }
 }
--- a/anchor/common/fork/src/fork.rs
+++ b/anchor/common/fork/src/fork.rs
@@ -1,7 +1,4 @@
 //! SSV protocol fork definitions.
-//!
-//! This module defines the SSV protocol forks and provides utilities for
-//! determining which fork is active at a given epoch.
 
 use std::fmt;
 
@@ -124,15 +121,5 @@ mod tests {
     #[test]
     fn test_genesis() {
         assert_eq!(Fork::genesis(), Fork::Genesis);
-    }
-
-    #[test]
-    fn test_serde() {
-        let fork = Fork::Alan;
-        let json = serde_json::to_string(&fork).unwrap();
-        assert_eq!(json, "\"alan\"");
-
-        let parsed: Fork = serde_json::from_str(&json).unwrap();
-        assert_eq!(parsed, Fork::Alan);
     }
 }
--- a/anchor/common/fork/src/monitor.rs
+++ b/anchor/common/fork/src/monitor.rs
@@ -5,6 +5,8 @@
 //! - Current active fork at startup
 //! - Entering the preparation window before a fork
 //! - Fork activation when it occurs
+//!
+//! The monitor exits automatically when all scheduled forks have activated.
 
 use std::{sync::Arc, time::Duration};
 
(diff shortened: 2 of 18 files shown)


anchor/common/fork/src/fork.rs
```diff
@@ -0,0 +1,104 @@
...
+/// across the SSV network, such as:
+/// - Subnet topology calculation
+/// - Topic naming conventions
+/// - Message formats
+/// - Consensus rules
+///
+/// Forks are ordered chronologically - earlier variants represent older forks.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize, Deserialize)]
+#[serde(rename_all = "lowercase")]
+pub enum Fork {
```
shane-moore, line 19: what do you think about renaming this to `SSVForkName` to signal it's an ssv fork? may help prevent confusion in the future since we also use `ForkName` from lighthouse for eth forks

anchor/common/fork/src/schedule.rs
```diff
@@ -0,0 +1,167 @@
...
+/// During this window, nodes prepare for the upcoming fork by subscribing to
+/// new topics while still operating on the current fork's rules.
+pub const FORK_PREPARATION_EPOCHS: u64 = 1;
+
+/// Manages fork activation epochs and provides utilities for fork transitions.
+///
+/// The schedule maps each fork to its activation epoch. Forks without an
+/// activation epoch are not scheduled (the network hasn't reached them yet).
+#[derive(Debug, Clone)]
+pub struct ForkSchedule {
```
shane-moore, line 23: same idea to have this as `SSVForkSchedule`

anchor/common/ssv_network_config/src/lib.rs
```diff
@@ -65,6 +69,9 @@ impl SsvNetworkConfig {
             ssv_domain_type: domain_type
                 .parse()
                 .map_err(|e| format!("Unable to parse built-in domain type: {e}"))?,
+            // All built-in networks are currently on the Alan fork.
+            // Boole fork epoch will be added when scheduled.
+            fork_schedule: ForkSchedule::new(),
```
shane-moore: it could make sense to have network specific fork activation epochs, i.e. a `ssv_fork_schedule.yaml` file for each `built_in_network_configs` network of shape:
```
forks:
  alan:  0
  boole: 12500
  charlie: 25000
  ```
  these should be able to load at compile time
  
  and then for local testing, you could have a runtime override `ssv_fork_schedule.yaml` that you load in like the proposed `ssv_boole_fork_epoch.txt`. wdyt? open to other ideas as well

File: anchor/common/fork/src/fork.rs [hunk 4]
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    /// Returns the name of this fork as a string.
+    pub const fn name(&self) -> &'static str {
+        match self {
+            Fork::Alan => "alan",
+            Fork::Boole => "boole",
+        }
+    }
+}
```
shane-moore, line 48: could we add a couple of helpers here for `alan` and `boole` like in lighthouse:
```
   pub fn altair_enabled(self) -> bool {
        self >= ForkName::Altair
    }

    pub fn bellatrix_enabled(self) -> bool {
        self >= ForkName::Bellatrix
    }

    pub fn capella_enabled(self) -> bool {
        self >= ForkName::Capella
    }
 ```
 
 reason is that we'll likely want to have fork specific logic branches, and it's nice to just do `fork.is_boole_enabled`

File: anchor/common/fork/src/fork.rs (hunk 1 above)
Comment (diegomrsantos) [reply anchor/common/fork/src/fork.rs:19]: `ForkName` comes from Lighthouse. I'd argue the namespace is enough to make the distinction. Wdyt?

Comment (shane-moore) [reply anchor/common/fork/src/fork.rs:19]: gotcha, `Fork` for ssv forks and `ForkName` for eth forks when doing conditional logic

File: anchor/common/ssv_network_config/src/lib.rs (hunk 3 above)
Comment (dknopik) [reply anchor/common/ssv_network_config/src/lib.rs]: I think a flat yaml (only top level key-value pairs) comparable to ethereum's `config.yaml` would be a bit nicer, as it is easier to generate in e.g. kurtosis

File: anchor/common/fork/src/schedule.rs (hunk 2 above)
Comment (diegomrsantos) [reply anchor/common/fork/src/schedule.rs:23]: same as in the previous comment

File: anchor/common/fork/src/monitor.rs [hunk 5]
```diff
@@ -0,0 +1,587 @@
...
+    // Exit early if no forks to monitor
+    if state.is_complete() {
+        return MonitorResult::Completed(all_events);
+    }
+
+    // Check once per epoch for state changes
+    let epoch_duration = Duration::from_secs(slots_per_epoch * seconds_per_slot);
+    let mut check_interval = interval(epoch_duration);
+
+    loop {
```
dknopik: It would be better to sleep until the beginning of the next epoch instead of sleeping for the duration of an epoch - or better yet, just sleep until the fork happens. We know all the appropriate times in advance

anchor/common/ssv_network_config/src/lib.rs
diegomrsantos: It's a good suggestion, I'll do it. Thanks.

anchor/common/fork/src/fork.rs
diegomrsantos: Since Fork derives Ord, we can already write:

```rust
  if fork_schedule.fork_at_epoch(epoch) >= Fork::Boole {
      // Boole or later behavior
  }
```

The helper is slightly more readable but adds maintenance burden (new method per fork). I'd suggest deferring this until we see patterns where it would genuinely help - we can always add them later.

anchor/common/fork/src/schedule.rs
```diff
@@ -0,0 +1,167 @@
...
+
+use types::Epoch;
+
+use crate::Fork;
+
+/// Number of epochs before a fork to start preparing (dual-subscribing, etc.).
+///
+/// During this window, nodes prepare for the upcoming fork by subscribing to
+/// new topics while still operating on the current fork's rules.
+pub const FORK_PREPARATION_EPOCHS: u64 = 1;
```
dknopik, line 16: Double subscribing is the only preparation needed for Boole. What do you think about keeping the notion of a preparation phase out of the common code for now?

anchor/common/fork/src/schedule.rs
```diff
@@ -0,0 +1,167 @@
...
+
+/// Manages fork activation epochs and provides utilities for fork transitions.
+///
+/// The schedule maps each fork to its activation epoch. Forks without an
+/// activation epoch are not scheduled (the network hasn't reached them yet).
+#[derive(Debug, Clone)]
+pub struct ForkSchedule {
+    /// Maps forks to their activation epochs.
+    activations: BTreeMap<Fork, Epoch>,
```
dknopik, line 25: i'd just have a flat struct of fields, e.g.

```
struct ForkSchedule {
    boole_activation_epoch: Epoch,
    // add future forks directly
}
```

Also, right now info like the domain type are missing. We could change this struct to e.g. `ForkConfig` and add fields like `boole_domain_type`

dknopik (commented):
Some notes below.

Also, no super strong feelings about this, but I dislike the term "fork" in this context, as we are not forking a chain. It is more like a "network upgrade", which is also the term used in SIPs

diegomrsantos:
> Some notes below.
> 
> Also, no super strong feelings about this, but I dislike the term "fork" in this context, as we are not forking a chain. It is more like a "network upgrade", which is also the term used in SIPs

Is the term fork exclusively about forking a chain? It could also be seen as changing a behaviour. Moreover, it doesn't necessarily need to be only about the network; this is a generic concept.

anchor/common/ssv_network_config/src/lib.rs
diegomrsantos: done

anchor/common/ssv_network_config/built_in_network_configs/holesky/ssv_fork_schedule.yaml
```diff
@@ -0,0 +1,4 @@
+# SSV Fork Schedule for Holesky
+# Note: Alan fork (genesis) is always at epoch 0
+forks: {}
```
shane-moore: @diegomrsantos, legit! just noting that the comment https://github.com/sigp/anchor/pull/774#discussion_r2682226177 suggests the yaml should be flat, similar to https://github.com/sigp/lighthouse/blob/stable/common/eth2_network_config/built_in_network_configs/hoodi/config.yaml

anchor/common/fork/src/monitor.rs
diegomrsantos: done the latter

anchor/common/ssv_network_config/src/lib.rs
diegomrsantos: I'm not sure the flat file change is worth it. Here's why:

  1. The benefit is marginal - generating forks:\n  boole: X vs boole: X is trivial
  2. The current BTreeMap approach enables generic algorithms (next_fork_after, active_fork) without hardcoding fork names
  3. If we later want to add domain types per fork, the nested format actually scales better:
```
  forks:
    boole:
      epoch: 12500
      domain_type: 0x00000002
```

  The architectural question (BTreeMap vs flat struct) is worth discussing separately, but that's a bigger change with real trade-offs. For now, the current approach works well imo.

anchor/common/fork/src/schedule.rs
```diff
@@ -0,0 +1,241 @@
...
+    /// Create a fork schedule from a map of forks to epoch values.
+    ///
+    /// This is primarily used when loading fork schedules from configuration files.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if:
+    /// - Alan fork is specified (it's always epoch 0 and shouldn't be in config)
+    /// - Fork epochs are not in chronological order
+    pub fn from_fork_epochs(epochs: std::collections::HashMap<Fork, u64>) -> Result<Self, String> {
```
petarjuki7: nit: maybe import `HashMap` at the top together with `BTreeMap` so it looks a bit cleaner?

anchor/common/fork/src/schedule.rs
diegomrsantos: I'd leave it like this for now. If the next fork doesn't need it, we can remove it or make it configurable. Wdyt?

anchor/common/fork/src/schedule.rs
diegomrsantos: done

anchor/common/fork/src/schedule.rs
diegomrsantos: I added domain tyoe

anchor/common/ssv_network_config/src/lib.rs
diegomrsantos: Rust code (ssv_network_config/src/lib.rs):
  - Changed ForkScheduleFile from a struct with forks field to a direct type alias: type ForkScheduleFile = HashMap<Fork, ForkConfig>
  - Updated parsing to use the map directly instead of .forks
  - Updated tests to use flat YAML format

  YAML files (mainnet, holesky, hoodi):
  # Before:
```
  forks: {}
    # boole: <epoch>
```

  # After:
```
  boole:
    epoch: 12500
    domain_type: "00000002"
```

  The file is named ssv_fork_schedule.yaml so it directly contains fork entries without the redundant wrapper.

dknopik (approved):
LGTM

If anyone finds anything to change, feel free to open another PR

Issue #769, opened 2026-01-05
Epoch-aware round-robin proposer


jking-aus:
Update leader function to include Ethereum epoch - makes proposer selection slightly more random. 

https://github.com/ssvlabs/SIPs/pull/74/

PR #776, opened 2026-01-13
chore: Update leader election function to include Eth epoch

petarjuki7:
## Issue Addressed

Addresses issue #769


shane-moore (dismissed):
looks clean and per spec. only thing is that we're going to have to gate all this logic so that it doesn't go live until after we hit the boole fork epoch, using the helpers in this [pr](https://github.com/sigp/anchor/pull/776) after its merged

dknopik:
This only updates the message validation, qbft code needs to be adjusted as well: https://github.com/sigp/anchor/blob/b4660eb8ef194124e62474975c83c987e0ed1261/anchor/common/qbft/src/qbft_types.rs#L29-L47

dknopik:
LGTM, only pending formatting. I might also leave this open until #754 is merged.

dknopik (dismissed):
LGTM

dknopik:
@mergify queue

anchor/common/qbft/src/qbft_types.rs
```diff
@@ -26,8 +26,30 @@ pub trait LeaderFunction {
     ) -> bool;
 }
 
-#[derive(Debug, Clone, Default)]
-pub struct DefaultLeaderFunction {}
+/// Default leader function implementation with configurable epoch-based rotation.
+#[derive(Debug, Clone)]
+pub struct DefaultLeaderFunction {
+    slots_per_epoch: u64,
```
shane-moore: is possible to make this `NonZeroU64` as to defend against divide by 0 errors, i.e. below:

` let eth_epoch = if self.include_epoch_shift {
            height / self.slots_per_epoch as usize`
            
 just an idea. otherwise, lgtm!

petarjuki7: makes sense, thanks, updated!

dknopik (approved):

PR #777, opened 2026-01-15
feat: implement deferred state updates for message validation

diegomrsantos:
## Issue Addressed

Improves message validation architecture by separating validation logic from state mutation.

## Proposed Changes

1. **Add StateUpdate types** - New types to capture deferred state changes:
   - `StateUpdate` enum with `None`, `Consensus`, and `PartialSignature` variants
   - `ConsensusStateUpdate` and `PartialSignatureStateUpdate` structs

2. **Add apply methods** - New methods to apply deferred state updates:
   - `DutyState::apply_consensus_update()`
   - `DutyState::apply_partial_signature_update()`

3. **Refactor validation** - Validation functions now return `(ValidatedSSVMessage, StateUpdate)` instead of mutating state directly

4. **Explicit commit** - `Validator::commit()` method to apply state updates when processing

## Motivation

Validation should be a pure operation that doesn't produce side effects. By returning a `StateUpdate` instead of mutating state during validation, callers have explicit control over when state changes are applied. This makes the code easier to reason about and test.

## Additional Info

- All existing tests pass
- The old `update_*` methods are kept for test compatibility
- Added `#[must_use]` to `StateUpdate` to prevent accidental omission of commit

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7686,6 +7686,7 @@ dependencies = [
  "eth2_network_config",
  "serde_yaml",
  "ssv_types",
+ "types",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7698,8 +7698,10 @@ dependencies = [
  "enr",
  "eth2_network_config",
  "fork",
+ "serde",
  "serde_yaml",
  "ssv_types",
+ "tempfile",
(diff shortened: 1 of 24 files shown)


diegomrsantos:
Closing it for now, if necessary, we can open it again

Issue #770, opened 2026-01-05
Aggregator committee duty


jking-aus:
Merge aggregator and sync committee duties 

https://github.com/ssvlabs/SIPs/pull/61

shane-moore:
@petarjuki7, when you make pr's similar to #834, for the attestation aggregation and sync committee `sign_*` fn's, tag this issue in the description. i think this issue will be able to be closed after those PR's are merged

diegomrsantos:
@shane-moore has this been completed?

shane-moore:
#834 still WIP then PR's for the other `sign*` functions. would say done after that. then, the metadata service refactor PR https://github.com/sigp/anchor/pull/836 I could consider as non-blocking to close out this ticket

diegomrsantos:
From the current unstable code, this looks implemented.

- sign_aggregate_and_proofs uses committee-based batching.
- sign_sync_committee_signatures uses committee-based batching.
- sign_sync_committee_contributions uses committee-based batching.

So I think #770 is no longer tracking missing implementation on unstable; it is effectively tracking promotion to stable and/or any explicitly remaining follow-up work.

PR #779, opened 2026-01-19
chore: update anchor to latest lighthouse package versions

shane-moore:
## Issue Addressed

The agg committee feature is going to require a small update to lighthouse's validator client service, so this is a pre-requisite PR to update anchor to use latest lighthouse `unstable` commit and resolves conflicts that have emerged since the lighthouse commit that anchor is currently targeting.

## Proposed Changes

Update anchor's lighthouse packages to latest

## Additional Info

This PR is a dependency of https://github.com/sigp/anchor/issues/770

### DIFF
--- a/anchor/common/qbft/src/tests.rs
+++ b/anchor/common/qbft/src/tests.rs
@@ -11,7 +11,7 @@ use std::{
 use qbft_types::DefaultLeaderFunction;
 use sha2::{Digest, Sha256};
 use ssv_types::{
-    OperatorId, RSA_SIGNATURE_SIZE,
+    OperatorId, RSA_SIGNATURE_SIZE, VariableList,
     consensus::{NoDataValidation, QbftMessageType},
     message::SignedSSVMessage,
 };
@@ -281,11 +281,12 @@ fn test_round_change_validation_skips_round_one_prepared_values() {
         qbft_message_type: QbftMessageType::RoundChange,
         height: 0,
         round: 2,
-        identifier: [0; 56].to_vec().into(),
+        identifier: VariableList::new([0; 56].to_vec()).unwrap(),
         root: test_data.hash(),
         data_round: 1, // Claims preparation in round 1 - this is the bug trigger!
-        round_change_justification: vec![].into(),
-        prepare_justification: vec![].into(), // INVALID: No justifications for claimed preparation!
+        round_change_justification: VariableList::new(vec![]).unwrap(),
+        prepare_justification: VariableList::new(vec![]).unwrap(), /* INVALID: No justifications
+                                                                    * for claimed preparation! */
     };
 
     // Create signed round change messages (need quorum of 3 for 3-node committee)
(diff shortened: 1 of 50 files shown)


shane-moore:
@claude, please take a look

shane-moore:
@claude, for point 1, these are test fixtures where `.unwrap()` is idiomatic. If the test setup fails, we want it to panic and fail the test.
The rest should be resolved. please check again

anchor/common/ssv_types/src/message.rs
```diff
@@ -445,9 +442,11 @@ impl SignedSSVMessage {
         // Convert Vec<[u8; 256]> to VariableList<VariableList<u8, U256>, U13>
         // First convert each [u8; 256] to VariableList<u8, U256>
         // This will always succeed since sig is [u8; 256] and U256 = 256
-        let signature_variable_lists: Vec<_> = signatures
+        let signature_variable_lists: Vec<VariableList<u8, U256>> = signatures
             .into_iter()
-            .map(|sig| VariableList::from(sig.to_vec()))
+            .map(|sig| {
+                VariableList::new(sig.to_vec()).expect("256-byte array always fits in U256 bound")
```
diegomrsantos: What do you think about returning an error instead?

shane-moore: I agree it's a safer approach, updated in https://github.com/sigp/anchor/pull/779/commits/1c800b694639a14d69fe2cdaa954e07bd9d11f81, thanks!

anchor/common/qbft/src/tests.rs
```diff
@@ -314,15 +315,18 @@ fn test_round_change_validation_skips_round_one_prepared_values() {
...
-        prepare_justification: vec![].into(), // Proposals don't need prepare justifications
+        round_change_justification: VariableList::new(
+            signed_round_changes
+                .into_iter()
+                .map(|msg| VariableList::new(msg.as_ssz_bytes()).unwrap())
+                .collect::<Vec<_>>(),
+        )
+        .unwrap(),
+        prepare_justification: VariableList::new(vec![]).unwrap(), /* Proposals don't need
+                                                                    * prepare justifications */
```
dknopik: For these, use `VariableList::empty()` instead

anchor/common/qbft/src/tests.rs
```diff
@@ -314,15 +315,18 @@ fn test_round_change_validation_skips_round_one_prepared_values() {
         qbft_message_type: QbftMessageType::Proposal,
         height: 0,
         round: 2,
-        identifier: [0; 56].to_vec().into(),
+        identifier: VariableList::new([0; 56].to_vec()).unwrap(),
```
dknopik: `VariableList::repeat_full(0)` might be a nicer alternative here.

dknopik (commented):
Love it! While the unwraps in testing code are fine, some are still avoidable, I've commented on the first instance each.

Otherwise LGTM

anchor/common/qbft/src/tests.rs
shane-moore: makes sense! updated these in last commit https://github.com/sigp/anchor/pull/779/commits/e0e07689ae510d636d3735b59d8136b1293520d6

anchor/common/qbft/src/tests.rs
shane-moore: same for these via https://github.com/sigp/anchor/pull/779/commits/e0e07689ae510d636d3735b59d8136b1293520d6

dknopik:
LGTM, just needs a format. I'll take the liberty of doing that so that the others can merge their PR's if necessary

dknopik (approved):

dknopik:
@mergify queue

PR #781, opened 2026-01-20
refactor: phase metadata service across full slot duration

shane-moore:
**Dependencies**:
  - Lighthouse dependencies will switch to upstream in a subsequent PR once https://github.com/sigp/lighthouse/pull/8684 is merged
  - This PR is a dependency for https://github.com/sigp/anchor/issues/770

**Scope**: This PR focuses on the validator store changes only. Downstream changes to QBFT consensus and P2P message verification for committee-based partial signatures will be completed in a separate PR.

## Issue Addressed
Required for committee-based partial signature collection. The full list of attesting and sync committee validators must be available at slot start before produce_selection_proof is called, which was not possible with the previous single-phase metadata approach.

## Proposed Changes

- Refactor MetadataService to use a phased approach that spans the full slot duration, rather than computing all metadata at a single point
  - Split metadata into three distinct phases: VotingAssignments at slot start, VotingContext at 1/3 slot, and AggregationAssignments at 2/3 slot
  - Add event-driven coordination between MetadataService and DutiesService using watch channels to ensure duties are polled before reading them
  - Update lighthouse dependencies to a fork that exposes the new duties service polling events
  - Update produce_selection_proof and produce_sync_selection_proof to use committee-based partial signature collection after the Boole fork
    - These methods now request signatures from all committee operators rather than just the validator's operators
    - VotingAssignments provides the expected message count for committee-based batching
  - Update lighthouse dependencies to a fork that exposes the new duties service polling events

## Additional Info

The previous approach computed all metadata at 1/3 into the slot. This created a race condition where duties might not be available yet, and aggregation information wasn't accurate until after selection proofs were computed.

  The new phased approach:
  - Phase 1 (slot start): Caches voting assignments immediately after DutiesService completes polling
  - Phase 2 (1/3 slot): Fetches beacon vote and builds voting context using cached assignments
  - Phase 3 (2/3 slot): Reads aggregator status after selection proofs are computed

  This ensures each piece of metadata is available at the earliest point it can be accurately determined.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -68,25 +68,22 @@ ssv_types = { path = "anchor/common/ssv_types" }
 subnet_service = { path = "anchor/subnet_service" }
 version = { path = "anchor/common/version" }
 
-# TODO: Switch back to sigp/lighthouse after https://github.com/sigp/lighthouse/pull/8684 is merged.
-# These changes are expected in lighthouse 8.1 release per https://github.com/sigp/lighthouse/issues/8681
-# Lighthouse fork with watch-attest-sync-duties branch `8725b85c` as of 1/20/2026
-beacon_node_fallback = { git = "https://github.com/shane-moore/lighthouse", rev = "35fb9763" }
-bls = { git = "https://github.com/shane-moore/lighthouse", rev = "35fb9763" }
-eth2 = { git = "https://github.com/shane-moore/lighthouse", rev = "35fb9763" }
-eth2_keystore = { git = "https://github.com/shane-moore/lighthouse", rev = "35fb9763" }
-eth2_network_config = { git = "https://github.com/shane-moore/lighthouse", rev = "35fb9763" }
-health_metrics = { git = "https://github.com/shane-moore/lighthouse", rev = "35fb9763" }
-metrics = { git = "https://github.com/shane-moore/lighthouse", rev = "35fb9763" }
-network_utils = { git = "https://github.com/shane-moore/lighthouse", rev = "35fb9763" }
-slashing_protection = { git = "https://github.com/shane-moore/lighthouse", rev = "35fb9763" }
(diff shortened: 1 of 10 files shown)


shane-moore:
@claude, could you review the changes just for these two files: `anchor/validator_store/src/lib.rs` and `anchor/validator_store/src/metadata_service.rs`

shane-moore:
@claude, could you review the changes just for these two files again: anchor/validator_store/src/lib.rs and anchor/validator_store/src/metadata_service.rs

for your points:
1. updated
2. updated
3. ignore
4. ignore
5. updated
6. ignore
7. ignore
8. ignore
9. ignore
10. ignore
11. ignore

additionally, i modified produce_selection_proof and produce_sync_selection_proof to be fork gated and use these new services so please take a look at those changes in last 3 commits as well

shane-moore:
@claude,
1. good, done
2. already resolved, maybe you looked at wrong commit or something

produce_sync_selection_proof uses wrong PartialSignatureKind in Boole fork
- this is correct actually, should use `PartialSignatureKind::AggregatorCommitteePartialSig` for both types of selection proof

i'm ignoring all the minor suggestions

shane-moore:
@claude, can you take another look?

shane-moore:
@dknopik, @diegomrsantos, @petarjuki7, would say this PR is in good shape for review 🤙

Cargo.toml
```diff
@@ -68,7 +68,6 @@ ssv_types = { path = "anchor/common/ssv_types" }
 subnet_service = { path = "anchor/subnet_service" }
 version = { path = "anchor/common/version" }
 
-# Lighthouse commit is unstable branch `58b153cac` as of 1/19/2026
```
dknopik, line 71: restore this line

shane-moore: restored!

dknopik (approved):
LGTM!

Issue #764, opened 2026-01-02
Allow beacon votes where `source == target && target == 0`


dknopik:
```
2026-01-02T16:51:06.003053Z  WARN Operator proposed invalid beacon vote             err=Invalid epoch order: source 0 >= target 0 instance_id=CommitteeInstanceId { committee: cf97adeedb59e05bfd73a2b4c2a8885708c4f4f70c84c64b27120e72ab733b72, instance_height: InstanceHeight(17) }
2026-01-02T16:51:06.003064Z DEBUG Qbft error: DataValidationFailed instance_id=CommitteeInstanceId { committee: cf97adeedb59e05bfd73a2b4c2a8885708c4f4f70c84c64b27120e72ab733b72, instance_height: InstanceHeight(17) }
```

Anchor should accept this vote as target epoch == source epoch is acceptable for epoch 0

This is relevant for local devnets

PR #783, opened 2026-01-21
fix: allow beacon votes when source and target equal and 0

petarjuki7:
## Issue Addressed

Addresses issue #764


dknopik (approved):
LGTM

PR #787, opened 2026-01-29
fix: do not use fixed start time for proposal duties

petarjuki7:
## Issue Addressed

Closes issue #758


shane-moore:
great job with this! I've taken some time to understand the different scenarios that trigger a round change:
1. receive `PROPOSE` for round > current
2. natural timeout -> round timer expires
3. receive 2f + 1 `ROUND_CHANGE` messages
4. receive f +1 `ROUND_CHANGE` messages

This PR currently handles scenarios 1 and 2 well. We could extend it to cover 3 and 4 by adding the timer reset signal in the round change handlers                               
  https://github.com/sigp/anchor/blob/3b61266fa332b28c870763636f91c039ff7fdf26/anchor/common/qbft/src/lib.rs#L1156 and                                                                 
  https://github.com/sigp/anchor/blob/3b61266fa332b28c870763636f91c039ff7fdf26/anchor/common/qbft/src/lib.rs#L1167                                                                   

but i think a more future proof option would actually be to keep your current solution for scenario 2 and handle scenarios 1, 3, and 4 via adding logic at https://github.com/sigp/anchor/blob/3b61266fa332b28c870763636f91c039ff7fdf26/anchor/qbft_manager/src/instance.rs#L303-L311 to:              
    a. Capture `old_round` before processing the message                                                                                                               
    b. Process the message, which may advance the round                                                                                                                                      
    c. Check if `new_round` > `old_round` and reset timer if in `Relative` mode
    
we'd handle 3 scenarios in one place, which is some nice future proofing. and seems less brittle since we won't need to remember setting signals in multiple qbft code paths

lmk if I'm missing something or if you see any issues with this approach!

shane-moore (changes requested):
great job with this! I've taken some time to understand the different scenarios that trigger a round change:
1. receive `PROPOSE` for round > current
2. natural timeout -> round timer expires
3. receive 2f + 1 `ROUND_CHANGE` messages
4. receive f +1 `ROUND_CHANGE` messages

This PR currently handles scenarios 1 and 2 well. We could extend it to cover 3 and 4 by adding the timer reset signal in the round change handlers                               
  https://github.com/sigp/anchor/blob/3b61266fa332b28c870763636f91c039ff7fdf26/anchor/common/qbft/src/lib.rs#L1156 and                                                                 
  https://github.com/sigp/anchor/blob/3b61266fa332b28c870763636f91c039ff7fdf26/anchor/common/qbft/src/lib.rs#L1167                                                                   

but i think a more future proof option would actually be to keep your current solution for scenario 2 and handle scenarios 1, 3, and 4 via adding logic at https://github.com/sigp/anchor/blob/3b61266fa332b28c870763636f91c039ff7fdf26/anchor/qbft_manager/src/instance.rs#L303-L311 to:              
    a. Capture `old_round` before processing the message                                                                                                               
    b. Process the message, which may advance the round                                                                                                                                      
    c. Check if `new_round` > `old_round` and reset timer if in `Relative` mode
    
we'd handle 3 scenarios in one place, which is some nice future proofing. and seems less brittle since we won't need to remember setting signals in multiple qbft code paths

lmk if I'm missing something or if you see any issues with this approach!

petarjuki7:
Thanks @shane-moore! I think it makes sense because either way that's the thing we are checking in the [`received_propose()`](https://github.com/sigp/anchor/blob/2b39ae41bb380f67727aef3d897d8c4358919736/anchor/common/qbft/src/lib.rs#L551) function, right? The only thing that we need to check is if we are in the `TimeoutMode::Relative`. Fixed!

shane-moore (dismissed):
lgtm!

anchor/qbft_manager/src/lib.rs
```diff
@@ -102,8 +113,10 @@ pub struct QbftInitialization<D: QbftData> {
     validator: Box<dyn QbftDataValidator<D>>,
     /// The message id to be embedded into outgoing messages.
     message_id: MessageId,
-    /// The time when the first round is supposed to start. Rounds will be advanced based on this.
+    /// The time reference for timeout calculations.
     start_time: Instant,
```
dknopik: `start_time` now has multiple possible meaning based on the `TimeoutMode`. This makes it a bit harder to understand and might be a source of bugs in future.

What do you think about e.g. making the `TimeoutMode` instead contain this:

```rust
pub enum TimeoutMode {
    SlotTime {
        instance_start_time: Instant,
    },
    Relative {
        current_round_start_time: Instant,
    },
}
```

This forces using sites to check which mode we are in

petarjuki7: Makes sense, I was considering both approaches, but I agree this makes it more explicit. Updated, thanks!

anchor/qbft_manager/src/instance.rs
```diff
@@ -114,7 +116,25 @@ impl Uninitialized {
...
+        };
+        tokio::time::sleep_until(start_time).await;
+
+        // For Relative: sleep until start_time, but use Instant::now() as reference
+        let timeout_mode = match init.timeout_mode {
+            TimeoutMode::SlotTime { .. } => init.timeout_mode,
+            TimeoutMode::Relative { .. } => TimeoutMode::Relative {
+                current_round_start_time: Instant::now(),
+            },
+        };
```
dknopik: Something like 

```rust
if let TimoutMode::Relative { current_round_start_time } = &mut init.timout_mode {
    *current_round_start_time = Instant::now();
}
```

would be a bit cleaner in this case imo

anchor/qbft_manager/src/instance.rs
```diff
@@ -285,7 +305,32 @@ pub async fn qbft_instance<D: QbftData<Hash = Hash256>>(
...
+                            _ => None,
+                        };
+
                         instance.receive(message);
+
+                        // Reset timer if round advanced (Relative mode only)
+                        if let QbftInstance::Initialized(initialized) = &mut instance
+                            && let Some(old) = old_round
+                            && initialized.qbft.get_round() > old
+                            && matches!(initialized.timeout_mode, TimeoutMode::Relative { .. })
```
dknopik: again, use a let binding

anchor/qbft_manager/src/instance.rs
```diff
@@ -295,6 +340,12 @@ pub async fn qbft_instance<D: QbftData<Hash = Hash256>>(
                 if let QbftInstance::Initialized(initialized) = &mut instance {
                     warn!("Round timer elapsed");
                     initialized.qbft.end_round();
+                    // Reset timer for new round in Relative mode
+                    if matches!(initialized.timeout_mode, TimeoutMode::Relative { .. }) {
+                        initialized.timeout_mode = TimeoutMode::Relative {
+                            current_round_start_time: Instant::now(),
+                        };
+                    }
```
dknopik: Same here

dknopik (approved):
LGTM

PR #788, opened 2026-01-30
chore: Edition in rustfmt.toml

dknopik:
## Issue Addressed

Invoking `rustfmt` directly requires edition information to be noted in the `rustfmt.toml`. Some workflows prefer this over calling `cargo fmt`.


petarjuki7 (approved):
Nice, lgtm

PR #790, opened 2026-01-30
chore: switch to ssv-mini fork for Boole testing

dknopik:
## Issue Addressed

We need to test the Boole fork

## Proposed Changes

Switch to commit supporting Boole and configure epoch 3 to be the switch epoch

## Additional Information

Epoch 3 means that the preparation epoch is epoch 2. Nodes are expected to be online and stable by the start of epoch two (indicated by assertoor sleeping for 5 minutes, which is below 2 epochs (6.4 minutes))

### DIFF
--- a/.github/custom/local-testnet/params.yaml
+++ b/.github/custom/local-testnet/params.yaml
@@ -46,6 +46,8 @@ nodes:
   anchor:
     count: 4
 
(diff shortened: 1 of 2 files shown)


dknopik:
actually let me integrate this into the other pr

Issue #510, opened 2025-08-16
Integer underflow in follow distance calculation


dknopik:
https://github.com/sigp/anchor/blob/44d21ce97f34dce4d0d96e81e1a12b5f80b7fbb0/anchor/eth/src/sync.rs#L385

This can be triggered e.g. in Kurtosis, where we might launch as early as in block 1 to 7. This causes an underflow (if compiled in release mode), and might cause an error response by connected EL client.

petarjuki7:
Is this check not preventing that calcualtion from ever being less than 0?

https://github.com/sigp/anchor/blob/44d21ce97f34dce4d0d96e81e1a12b5f80b7fbb0/anchor/eth/src/sync.rs#L381

And then this check makes sure its a positive number (it has to be bigger than `start_block`):

https://github.com/sigp/anchor/blob/44d21ce97f34dce4d0d96e81e1a12b5f80b7fbb0/anchor/eth/src/sync.rs#L386

dknopik:
huh, weird

I swear I wasn't imagining an underflow though 🤔 maybe it was at some other place. Will keep open as reminder for now

dknopik:
@petarjuki7 

Ah! Found it. It happens in the live sync, so I linked the wrong line above:

https://github.com/sigp/anchor/blob/733ae8702b9dd0b7d21628e15cbce6d437a0d598/anchor/eth/src/sync.rs#L690

dknopik:
Kind of related: we should not attempt to sync a block that is below the contract deployment block

PR #792, opened 2026-02-02
fix: integer underflow in follow distance calculation

petarjuki7:
## Issue Addressed

Addresses issue #510 

## Proposed Changes

Add `if statement` to guard against syncing before contract deployment.

### DIFF
--- a/anchor/eth/src/sync.rs
+++ b/anchor/eth/src/sync.rs
@@ -686,6 +686,20 @@ impl SsvEventSyncer {
 
             // If we have a connection, continuously stream in blocks
             while let Some(block_header) = stream.next().await {
+                // Guard against integer underflow when calculating relevant_block.
+                if block_header.number < self.network.ssv_contract_block + FOLLOW_DISTANCE {
+                    debug!(
+                        block_number = block_header.number,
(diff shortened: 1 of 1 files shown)


anchor/eth/src/sync.rs
```diff
@@ -686,6 +686,20 @@ impl SsvEventSyncer {
...
+                    debug!(
+                        block_number = block_header.number,
+                        contract_block = self.network.ssv_contract_block,
+                        follow_distance = FOLLOW_DISTANCE,
+                        "Received block below minimum valid block number"
+                    );
+                    return Err(ExecutionError::InvalidEvent(format!(
+                        "Block {} is before contract deployment + follow distance",
+                        block_header.number
+                    )));
```
petarjuki7: Should we `Error` here or continue with just a `debug!()` log?
cc: @dknopik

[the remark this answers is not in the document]
petarjuki7: The first one probably makes the most sense

petarjuki7: Updated

dknopik (approved):
LGTM, thanks!

PR #795, opened 2026-02-03
chore: Less strict testnet in CI

dknopik:
## Issue Addressed

CI often fails due as the newly introduced local testnet is checked too strictly. 

## Proposed Changes

Replace the parameters with the less strict parameters found in `ethereum-package`

## Additional Info

Additionally, load the correct parameter file from the repo instead of a gist URL.

### DIFF
--- a/.github/custom/local-testnet/test_anchor_proposal.yaml
+++ b/.github/custom/local-testnet/test_anchor_proposal.yaml
@@ -1,6 +1,6 @@
 id: test_anchor_proposal
 name: "Check for Anchor Proposal"
-timeout: 30m
+timeout: 20m
 tasks:
   - name: sleep
     config:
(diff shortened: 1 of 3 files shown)


.github/custom/local-testnet/test_anchor_proposal.yaml
```diff
@@ -15,9 +15,8 @@ tasks:
             minAttestationCount: 1
         - name: check_consensus_attestation_stats
           config:
-            minTargetPercent: 100
-            minHeadPercent: 100
-            minTotalPercent: 100
+            minTargetPercent: 98
+            minHeadPercent: 80
```
diegomrsantos, line 23: It'd be good to add some reasoning for these values

dknopik: done, wdyt?

diegomrsantos: It's great, thanks!

diegomrsantos (approved):

PR #796, opened 2026-02-03
fix: guard fork phase channel handling

diegomrsantos:
## Why

Fork-phase events are delivered via an async broadcast channel. Today, if the receiver closes, `recv()` can return errors immediately and repeatedly, which risks spinning the network loop or starving other branches. Fork monitor sends also silently drop errors, which can hide missed ENR domain updates.

## What

- Network loop now isolates event handling into helpers and explicitly handles fork-phase `recv()` errors.
  - On channel close, the fork-phase branch is disabled to prevent hot loops.
  - Overflow is logged once for visibility.
- Fork monitor now checks `broadcast_direct` results and exits early if the channel is closed.

## Notes

- No behavior change for normal operation; this only affects shutdown/edge conditions.
- Keeps ENR updates and fork-phase handling observable under failure.

### DIFF
--- a/anchor/common/fork/src/monitor.rs
+++ b/anchor/common/fork/src/monitor.rs
@@ -19,7 +19,7 @@ use std::{ops::ControlFlow, sync::Arc, time::Duration};
 
 use slot_clock::SlotClock;
 use task_executor::TaskExecutor;
-use tracing::{info, warn};
+use tracing::{debug, info, warn};
 use types::{Epoch, Slot};
 
 use crate::{FORK_PREPARATION_EPOCHS, Fork, ForkConfig, ForkSchedule, SUBSEQUENT_WINDOW_SLOTS};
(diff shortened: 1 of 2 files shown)


diegomrsantos:
@claude review this PR again. I wouldn’t add tests. If you meant the earlier network refactor, the only new behavior is handling RecvError::Closed → disable the branch. That’s testable, but it requires a unit test around handle_fork_phase_result, which is more work than it’s worth unless we’ve had regressions.

dknopik:
> Today, if the receiver closes, recv() can return errors immediately and repeatedly, which risks spinning the network loop or starving other branches.

If a future in a select branch completes, but the returned value does not match the pattern, `select` continues waiting on the the other branches until one completes with a value matching corresponding pattern. So starvation can not occur here. See the lifecycle of a `select` here: https://docs.rs/tokio/latest/tokio/macro.select.html#:~:text=Evaluate,panic

diegomrsantos:
> > Today, if the receiver closes, recv() can return errors immediately and repeatedly, which risks spinning the network loop or starving other branches.
> 
> If a future in a select branch completes, but the returned value does not match the pattern, `select` continues waiting on the the other branches until one completes with a value matching corresponding pattern. So starvation can not occur here. See the lifecycle of a `select` here: https://docs.rs/tokio/latest/tokio/macro.select.html#:~:text=Evaluate,panic

You're right, thanks for the explanation. I would still keep the refactor tho. Wdyt?

PR #797, opened 2026-02-03
fix: update bytes to 1.11.1 to fix RUSTSEC-2026-0007

diegomrsantos:
## Issue Addressed

Fixes cargo audit failure due to RUSTSEC-2026-0007 (integer overflow vulnerability in `BytesMut::reserve`).

## Proposed Changes

- Update `bytes` crate from 1.11.0 to 1.11.1 via Cargo.lock

## Additional Info

This is a lockfile-only update. The `bytes` 1.11.1 release patches the integer overflow vulnerability.


shane-moore (approved):
lgtm!

PR #798, opened 2026-02-04
fix: do not exit subnet service if there are no more fork changes

dknopik:
## Issue Addressed

The subnet service must not quit if there are no more changes, as DB changes might still trigger subnet changes.

## Proposed Changes

Remove code block that returns if the channel returns an error.


anchor/subnet_service/src/subscriptions.rs
```diff
@@ -85,12 +85,7 @@ impl<S: SlotClock> SubnetService<S> {
                 _ = sleep(delay), if !self.disable_gossipsub_topic_scoring => {
                     self.send_scoring_rate_updates::<E>(&service_state).await;
                 }
-                phase = fork_phase_rx.recv() => {
-                    let Ok(phase) = phase else {
-                        warn!("Fork phase channel closed");
-                        return;
-                    };
-
+                Ok(phase) = fork_phase_rx.recv() => {
```
diegomrsantos, line 88: Don't we want to log?

dknopik: I added some logging at the sender, what do you think? IMO this is preferable to individual logs by the receivers.

Also recall that panics are logged anyway.

diegomrsantos (approved):

shane-moore (approved):

PR #799, opened 2026-02-04
refactor(network): split run loop handlers

diegomrsantos:
## Why

The network run loop had grown large and hard to follow. Splitting it into helpers improves readability and makes future changes safer.

## What

- Extracted swarm event, gossipsub, peer-manager heartbeat, and channel handling into focused helper methods.
- Kept the run loop structure and behavior the same.

## Notes

- No functional changes intended.


petarjuki7:
Nice, looks good, well needed refactor!
Only question, would it make sense to use `Result<(), ()>` (or a more descriptive `Error`) instead of the `LoopControl` enum, and then you could return `Ok(())` or `Err(())` where needed and match on them like:
```
event = self.message_rx.recv() => {
    if self.handle_outbound_message(event).is_err() {
        return;
    }
}
```

Would eliminate the need for a custom enum, but works fine either way, no strong opinion.

diegomrsantos:
> Nice, looks good, well needed refactor! Only question, would it make sense to use `Result<(), ()>` (or a more descriptive `Error`) instead of the `LoopControl` enum, and then you could return `Ok(())` or `Err(())` where needed and match on them like:
> 
> ```
> event = self.message_rx.recv() => {
>     if self.handle_outbound_message(event).is_err() {
>         return;
>     }
> }
> ```
> 
> Would eliminate the need for a custom enum, but works fine either way, no strong opinion.

Thanks! I see the appeal of Result, but it reads like an error path, whereas this is just control‑flow. If we want to avoid a custom enum, `std::ops::ControlFlow<()>` feels more idiomatic here (continue/break without implying failure). Wdyt?

dknopik:
I think if you make `run` return `ControlFlow`, `?` might be able to work as `ControlFlow` implements `Try`

diegomrsantos:
> I think if you make `run` return `ControlFlow`, `?` might be able to work as `ControlFlow` implements `Try`

Not sure it's worth it. Only two helpers need this, and it wouldn't be a refactor PR anymore.

dknopik (approved):
LGTM

PR #801, opened 2026-02-04
refactor: remove `DutyState` to use `OperatorState` for better parallelism

### LINKED ISSUE #745: Reduce lock contention by flattening message validator state
Currently, we have a `duty_state_map: DashMap<MessageId, DutyState>` on the top level of the message validator. `DutyState` is 

```rust
/// DutyState manages the state for duty validation across operators and slots
pub(crate) struct DutyState {
    /// Tracks the duty state for an operator
    operators: HashMap<OperatorId, OperatorState>,
    /// The number of slots for which state is stored (defines the size of the circular buffer)
    stored_slot_count: usize,
}
```

Not that `stored_slot_count` is effectively constant.

This implies that if the duty state is write locked for a message id, we can not process messages from multiple operators in parallel (w.r.t. that duty).

We should consider moving to `duty_state_map: DashMap<(MessageId, OperatorId), OperatorState>` for better parallelism through avoided lock contention.

PoulavBhowmick03:
## Issue Addressed

Fixes #745 

## Proposed Changes

- Changed `duty_state_map` from `DashMap<MessageId, DutyState>` to `DashMap<(MessageId, OperatorId), OperatorState>` to enable parallel processing of messages
- Removed `DutyState` wrapper struct.
- Updated validation functions

### DIFF
--- a/anchor/message_validator/src/consensus_message.rs
+++ b/anchor/message_validator/src/consensus_message.rs
@@ -16,13 +16,13 @@ use types::Epoch;
 
 use crate::{
     FIRST_ROUND, ValidatedSSVMessage, ValidationContext, ValidationFailure, compute_quorum_size,
-    duty_state::DutyState, hash_data, slot_start_time, validate_beacon_duty, validate_duty_count,
-    validate_slot_time, verify_message_signatures,
+    duty_state::OperatorState, hash_data, slot_start_time, validate_beacon_duty,
+    validate_duty_count, validate_slot_time, verify_message_signatures,
 };
 
 pub(crate) fn validate_consensus_message(
     validation_context: ValidationContext<impl SlotClock>,
-    duty_state: &mut DutyState,
+    operator_state: &mut OperatorState,
     duty_provider: Arc<impl DutiesProvider>,
 ) -> Result<ValidatedSSVMessage, ValidationFailure> {
     // Decode message to QbftMessage
@@ -54,12 +54,12 @@ pub(crate) fn validate_consensus_message(
         validation_context.operator_pub_keys,
     )?;
 
-    validate_qbft_logic(&validation_context, &consensus_message, duty_state)?;
+    validate_qbft_logic(&validation_context, &consensus_message, operator_state)?;
 
     validate_qbft_message_by_duty_logic(
         &validation_context,
         &consensus_message,
-        duty_state,
+        operator_state,
         duty_provider,
     )?;
 
@@ -68,10 +68,13 @@ pub(crate) fn validate_consensus_message(
         validation_context.operator_pub_keys,
     )?;
 
-    duty_state.update_for_consensus_message(
+    let msg_slot = Slot::from(consensus_message.height);
+    let msg_epoch = msg_slot.epoch(validation_context.slots_per_epoch);
+    operator_state.update(
         validation_context.signed_ssv_message,
         &consensus_message,
-        validation_context.slots_per_epoch,
+        &msg_slot,
+        &msg_epoch,
     );
 
     // Return the validated message
@@ -232,7 +235,7 @@ fn validate_justification_list<N: Unsigned>(
 pub(crate) fn validate_qbft_logic(
     validation_context: &ValidationContext<impl SlotClock>,
     consensus_message: &QbftMessage,
-    duty_state: &mut DutyState,
+    operator_state: &mut OperatorState,
 ) -> Result<(), ValidationFailure> {
     let signed_ssv_message = validation_context.signed_ssv_message;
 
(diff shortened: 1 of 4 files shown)


anchor/message_validator/src/consensus_message.rs
```diff
@@ -260,55 +263,49 @@ pub(crate) fn validate_qbft_logic(
...
-                // If they've sent a future message due to clock error,
-                // they'd have to wait for the next slot/round to be accepted.
-                return Err(ValidationFailure::RoundAlreadyAdvanced {
-                    got: consensus_message.round,
-                    want: signer_state.round,
-                });
-            }
+    // Get or create the operator state first, then check if there's a signer state
+    let Some(signer_state) = operator_state.get_signer_state(&msg_slot) else {
+        return Ok(());
```
shane-moore, line 278: thanks for the pr! This `return Ok()` causes `validate_round_in_allowed_spread`  to be skipped for single signer messages when there's no prior state. The old `continue` fell through to the round 
  spread check after the loop
  
  perhaps move the round spread check before the early return:
  ```
  if signers.len() == 1 {
      validate_round_in_allowed_spread(consensus_message, validation_context)?;
  }
  ```

shane-moore (commented):
in `anchor/message_validator/src/lib.rs`, line 440-443, lets do `.ok_or(ValidationFailure::NoSigners)?;` instead of `.expect`

shane-moore:
Two loops over signers were removed in this PR                                               
                                                                                                                                                                                        
   ### 1. `validate_qbft_logic` - Valid Optimization (but deviates from Go)                                                                                                             
                                                                                                                                                                                        
   The Go implementation [loops over all signers](https://github.com/ssvlabs/ssv/blob/cb97aff9c43e9b087031bb3643b4b4fb2b0261f5/message/validation/consensus_validation.go#L197-L202)   for the SeenSigners duplicate check. However, this loop is redundant because:                                                                                                        
   - Operator IDs are always sorted
   - Same quorum → same sorted order → same first operator
   - Checking only the first operator's `seen_signers` is sufficient

While technically redundant, the savings are negligible (a few extra hash      
   lookups) and matching Go makes the code easier to audit and maintain. so I'm in favor of restoring the loop

   ### 2. `validate_qbft_message_by_duty_logic` - Bug

   The Go implementation [validates duty count for ALL
   signers](https://github.com/ssvlabs/ssv/blob/cb97aff9c43e9b087031bb3643b4b4fb2b0261f5/message/validation/consensus_validation.go#L293-L300):

   ```go
   for _, signer := range signedSSVMessage.OperatorIDs {
       signerStateBySlot := state.Signer(committeeInfo.signerIndex(signer))
       if err := mv.validateDutyCount(..., signerStateBySlot); err != nil {
           return err
       }
   }
   ```

   This loop needs to be restored. Each operator has their own per-epoch duty limit, so all signers must be validated. The new `DashMap<(MessageId, OperatorId), OperatorState>`
   structure still supports this, just look up each operator's state individually.

anchor/message_validator/src/consensus_message.rs
PoulavBhowmick03: thanks, i missed that. will make the change

dknopik:
@PoulavBhowmick03 can you please take a look at the above comments and merge from unstable to fix CI?

PoulavBhowmick03:
> Two loops over signers were removed in this PR
> 
> ### 1. `validate_qbft_logic` - Valid Optimization (but deviates from Go)
> The Go implementation [loops over all signers](https://github.com/ssvlabs/ssv/blob/cb97aff9c43e9b087031bb3643b4b4fb2b0261f5/message/validation/consensus_validation.go#L197-L202) for the SeenSigners duplicate check. However, this loop is redundant because:
> 
> * Operator IDs are always sorted
> * Same quorum → same sorted order → same first operator
> * Checking only the first operator's `seen_signers` is sufficient
> 
> While technically redundant, the savings are negligible (a few extra hash lookups) and matching Go makes the code easier to audit and maintain. so I'm in favor of restoring the loop
> 
> ### 2. `validate_qbft_message_by_duty_logic` - Bug
> The Go implementation [validates duty count for ALL signers](https://github.com/ssvlabs/ssv/blob/cb97aff9c43e9b087031bb3643b4b4fb2b0261f5/message/validation/consensus_validation.go#L293-L300):
> 
> ```go
> for _, signer := range signedSSVMessage.OperatorIDs {
>     signerStateBySlot := state.Signer(committeeInfo.signerIndex(signer))
>     if err := mv.validateDutyCount(..., signerStateBySlot); err != nil {
>         return err
>     }
> }
> ```
> 
> This loop needs to be restored. Each operator has their own per-epoch duty limit, so all signers must be validated. The new `DashMap<(MessageId, OperatorId), OperatorState>` structure still supports this, just look up each operator's state individually.

The Go code receives a state object that can look up any signer's state via `state.Signer(committeeInfo.signerIndex(signer))`
in that case, should I pass `DashMap<(MessageId, OperatorId), OperatorState>` to the function for the state lookup?

PoulavBhowmick03:
> @PoulavBhowmick03 can you please take a look at the above comments and merge from unstable to fix CI?

Yes sure, working on it rn

dknopik:
Hey @PoulavBhowmick03!

Again, thanks for your PR.

A problem that I did not think about when creating the original issue is that "Decided messages", i.e. messages with multiple operators as sender, would require holding a write lock on multiple entries of the map after the refactor. This is not possible without risking a deadlock.

Decided messages are likely to be removed in future: https://github.com/ssvlabs/SIPs/pull/63
The current approach is more likely to work then.

That being said, we might still be able to make this work: `SignerState::seen_signers` is only used for decided messages. We might be able to create a second `DashMap`, used exclusively for tracking which decided messages we have seen. Something like: `seen_decideds: DashMap<MessageId, SeenDecideds>` 
This will add additional complexity though.

Another idea might be a more involved general refactor of message validation, which was previously ported from go-ssv in very straightforward way. I am sure there is a better way to structure the existing code. 

Apologies that I did not realize that this isn't as straightforward as I thought when creating the issue. Let me know your thoughts!

PoulavBhowmick03:
> Hey @PoulavBhowmick03!
> 
> Again, thanks for your PR.
> 
> A problem that I did not think about when creating the original issue is that "Decided messages", i.e. messages with multiple operators as sender, would require holding a write lock on multiple entries of the map after the refactor. This is not possible without risking a deadlock.
> 
> Decided messages are likely to be removed in future: [ssvlabs/SIPs#63](https://github.com/ssvlabs/SIPs/pull/63) The current approach is more likely to work then.
> 
> That being said, we might still be able to make this work: `SignerState::seen_signers` is only used for decided messages. We might be able to create a second `DashMap`, used exclusively for tracking which decided messages we have seen. Something like: `seen_decideds: DashMap<MessageId, SeenDecideds>` This will add additional complexity though.
> 
> Another idea might be a more involved general refactor of message validation, which was previously ported from go-ssv in very straightforward way. I am sure there is a better way to structure the existing code.
> 
> Apologies that I did not realize that this isn't as straightforward as I thought when creating the issue. Let me know your thoughts!

Hey! I would need a little time to look into this as well, and the different ways to work around it for now. I will get back to you on this

diegomrsantos:
Closing this PR after review — the flattened `DashMap<(MessageId, OperatorId), OperatorState>` approach has a fundamental issue:

**Deadlock**: `validate_decoded_message` acquires a `RefMut` (write lock on a DashMap shard) for the first signer's entry, then passes both `&mut OperatorState` (borrowed from the RefMut) and `&DashMap` into the validation pipeline. Inside `validate_qbft_message_by_duty_logic`, the code calls `duty_state_map.entry((message_id, signer))` for each signer in a loop — including the first signer, whose entry is already write-locked. Since DashMap's `RwLock` is not reentrant, this deadlocks on every consensus message.

This isn't a fixable detail — it's inherent to the flattened design. The validation logic needs cross-operator state access (slot advancement checks, duty counts), which requires locking multiple `(MessageId, OperatorId)` entries simultaneously. Even skipping the first signer wouldn't help for decided messages, where other signers could hash to the same DashMap shard.

See #745 for a proposed alternative approach that solves the same contention problem without changing the map structure.

dknopik:
> See https://github.com/sigp/anchor/issues/745 for a proposed alternative approach that solves the same contention problem without changing the map structure.

This PR was attempting to implement the issue linked. #745 is also affected by this inherent issue.

As noted above, once messages have only one signer (with the removal of Decided messages), this becomes possible.

EDIT: ah, I guess you meant to link your comment there?

PR #803, opened 2026-02-05
feat: Allow conversion between key files

dknopik:
## Proposed Changes

Allow users to encrypt/decrypt existing keys. This is done by adding additional parameters to the `keygen` util.

## Additional Info

Also moves/refactors some functions to allow their reuse.

### DIFF
--- a/anchor/keygen/src/lib.rs
+++ b/anchor/keygen/src/lib.rs
@@ -7,6 +7,7 @@ use operator_key::{
     ConversionError,
     encrypted::{EncryptedKey, EncryptionError},
     public, unencrypted,
+    util::FileReadError,
 };
 use thiserror::Error;
 use tracing::{error, info};
@@ -34,6 +35,9 @@ pub enum KeygenError {
 
     #[error("Key file(s) already exist in {0}")]
     Exists(String),
+
+    #[error("Failed read existing key for conversion: {0}")]
+    ReadError(#[from] FileReadError),
 }
 
 #[derive(Parser, Clone, Debug)]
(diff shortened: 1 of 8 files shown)


anchor/keygen/src/lib.rs
```diff
@@ -34,6 +35,9 @@ pub enum KeygenError {
 
     #[error("Key file(s) already exist in {0}")]
     Exists(String),
+
+    #[error("Failed read existing key for conversion: {0}")]
```
petarjuki7: typo nitpick: `Failed to read` or `Failed reading`

anchor/keygen/src/lib.rs
```diff
@@ -58,16 +62,38 @@ pub struct Keygen {
...
+        help = "Path to a file containing the password to use for the new key file",
+        requires = "encrypt",
+        value_name = "PATH"
     )]
     pub password_file: Option<PathBuf>,
+
+    #[clap(
+        long,
+        help = "Path to an existing key file to use instead of generating a new one. If encrypted, \
+            the password is read from terminal or via --password-file-for existing. The existing \
```
petarjuki7:
```suggestion
            the password is read from terminal or via --password-file-for-existing. The existing \
```

petarjuki7 (dismissed):
Looks good to me! 2 small typo nitpicks

shane-moore (approved):
lgtm! reviewed to get a feel of how we handle the rsa keys. makes sense to allow conversion b/w key formats w/o generating new key. probably still some key related features that we're missing compared to the ssv guys. i'll make a note to look into those, and the team can prioritize them over time, like the dkg functionality for example

PR #805, opened 2026-02-05
chore: Get rid of more unused dependencies

dknopik:
## Issue Addressed

While `cargo udeps` spots unused dependencies, it has some false negatives.

## Proposed Changes

Remove unused dependencies I found.

## Additional Info

This was done by manually going through each `Cargo.toml` and removing the unused deps as highlighted by RustRover. There may be further false negatives.


petarjuki7 (approved):
Confirmed it on my machine, works fine!

PR #820, opened 2026-02-10
fix: make observed_peer_subnets fork-aware

### LINKED ISSUE #818: Make `observed_peer_subnets` Fork aware
`observed_peer_subnets` is currently only a bitmap of size 128, and is updated on subscription/unsubscription of any fork subnet.

This might cause issues: If a peer is to subscribed to both `ssv.v2.42` and `/ssv/mainnet/boole/42`, it will mark `42` as not subscribed when the peer unsubscribes from `ssv.v2.42` at the end of the grace period, even though it is still subscribed to `/ssv/mainnet/boole/42`.

diegomrsantos:
## Issue Addressed

Closes #818

## Proposed Changes

During fork transitions (e.g., Alan → Boole), peers may be subscribed to the same subnet via topics from different forks simultaneously — for example, `ssv.v2.42` (Alan) and `/ssv/mainnet/boole/42` (Boole). The peer manager's `observed_peer_subnets` previously stored a single flat bitmap per peer, so unsubscribing from one fork's topic would incorrectly clear the subnet bit even though the peer was still subscribed via the other fork.

This PR makes the subscription tracking fork-aware by maintaining a separate bitmap per fork per peer. Query methods (used for peer counting, disconnect decisions, and priority dialing) aggregate across forks transparently, so the rest of the peer management logic is unaffected.

## Additional Info

- The gossipsub subscription/unsubscription event handlers now use `topic::parse_topic()` (which returns fork info) instead of `topic::parse_subnet_id()` (which discarded it), threading the fork identity through to the connection manager.
- This also lays groundwork for future fork-aware peer selection, e.g., preferring peers that are on the current fork's topics after a grace period ends.

### DIFF
--- a/anchor/network/src/peer_manager/connection.rs
+++ b/anchor/network/src/peer_manager/connection.rs
@@ -138,6 +138,12 @@ impl ConnectionManager {
 
         if idx < bitfield.len() {
             let _ = bitfield.set(idx, subscribed);
+        } else {
+            tracing::warn!(
+                subnet = idx,
+                max = bitfield.len(),
+                "Subnet ID exceeds bitfield capacity"
+            );
         }
 
         // Clean up empty entries to keep maps small
@@ -814,6 +820,27 @@ mod tests {
         assert!(
             mgr.peer_offers_needed_subnets_observed_only(&peer, &needed),
             "Peer should still offer subnet 42 after unsubscribing only from Alan"
+        );
+    }
+
+    // ==================== Idempotency ====================
+
+    #[test]
+    fn test_duplicate_subscribe_then_single_unsubscribe_clears_bit() {
+        // Arrange
+        let mut mgr = create_test_manager();
+        let peer = connect_random_peer(&mut mgr);
+        let subnet = SubnetId::new(5);
+
+        // Act: subscribe twice on the same fork+subnet, then unsubscribe once
+        mgr.set_peer_subscribed(peer, Fork::Alan, subnet, true);
+        mgr.set_peer_subscribed(peer, Fork::Alan, subnet, true);
+        mgr.set_peer_subscribed(peer, Fork::Alan, subnet, false);
+
+        // Assert: bit should be cleared — bitfield tracks presence, not a count
+        assert!(
+            !is_subnet_set(&mgr, &peer, 5),
+            "A single unsubscribe should clear the bit regardless of duplicate subscribes"
         );
     }
 }
--- a/anchor/network/src/peer_manager/connection.rs
+++ b/anchor/network/src/peer_manager/connection.rs
@@ -4,6 +4,7 @@ use std::{
 };
 
 use discv5::libp2p_identity::PeerId;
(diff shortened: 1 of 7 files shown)


anchor/network/src/peer_manager/connection.rs
```diff
@@ -116,20 +120,40 @@ impl ConnectionManager {
...
+        fork: Fork,
+        subnet: SubnetId,
+        subscribed: bool,
+    ) {
         let idx = *subnet.deref() as usize;
-        if idx < entry.len() {
-            // Safe to ignore the result of `set` because we have already checked that `idx <
-            // entry.len()`
-            let _ = entry.set(idx, subscribed);
+        let fork_map = self.observed_peer_subnets.entry(peer).or_default();
```
shane-moore, line 136: wdyt of using `.or_insert_with(|| HashMap::with_capacity(Fork::all().len()))` in lieu of `.or_default` to avoid potential reallocation

anchor/network/src/peer_manager/connection.rs
```diff
@@ -116,20 +120,40 @@ impl ConnectionManager {
...
+        let bitfield = fork_map.entry(fork).or_default();
+
+        if idx < bitfield.len() {
+            let _ = bitfield.set(idx, subscribed);
+        } else {
+            tracing::warn!(
+                subnet = idx,
```
shane-moore, line 144: think useful to have `peer` in this `warn`?

anchor/network/src/peer_manager/connection.rs
```diff
@@ -263,9 +287,22 @@ impl ConnectionManager {
...
+        let fork_map = self.observed_peer_subnets.get(peer)?;
+        Some(Self::aggregate_fork_bitmaps(fork_map))
+    }
+
+    /// OR all per-fork bitmaps together into a single aggregate bitfield.
+    fn aggregate_fork_bitmaps(
+        fork_map: &HashMap<Fork, Bitfield<Fixed<U128>>>,
+    ) -> Bitfield<Fixed<U128>> {
+        let mut result = Bitfield::default();
+        for bitfield in fork_map.values() {
```
shane-moore: wdyt of functional approach instead:
```
fork_map.values().fold(Bitfield::default(), |acc, bf|
   acc.union(bf))
```

anchor/network/src/peer_manager/connection.rs
shane-moore: thought more about it, and this `with_capacity` probably overkill since we shouldn't have more than 2 forks to worry about at a time in `set_peer_subscribed`, if i understand correctly. so I'll just close this

dknopik (approved):
LGTM, thanks!

anchor/network/src/peer_manager/connection.rs

anchor/network/src/peer_manager/connection.rs
diegomrsantos: done

PR #824, opened 2026-02-11
feat: replace SharedDomainType with SharedForkLifecycle for fork-aware peer selection

diegomrsantos:
## Issue Addressed

Follows up on PR #820 (fork-aware `observed_peer_subnets`). While #820 correctly stores per-fork bitmaps, all query methods still aggregate across forks with a blind `union()`. After the Boole grace period ends, peers only subscribed to Alan topics appear useful for Boole subnets — leading to incorrect peer selection.

## Proposed Changes

- Introduce `ForkLifecycle` enum (`Normal`, `WarmUp`, `GracePeriod`) and `SharedForkLifecycle` shared state in the fork crate, replacing `SharedDomainType`
- `ForkMonitor` updates the shared lifecycle state before broadcasting each `ForkPhase` event
- Peer selection queries in `ConnectionManager` now consult the lifecycle: in `Normal` state only the current fork's bitmap is considered; during `WarmUp`/`GracePeriod` both forks are aggregated
- All components that previously read `SharedDomainType` (Discovery, Handshake) now read `SharedForkLifecycle::domain_type()` instead
- `Network::on_fork_phase` simplified — no longer sets domain type (monitor handles it), only updates ENR

## Additional Info

- Depends on #814, #815, and #820
- `SharedForkLifecycle` follows the same `Arc<RwLock<T>>` single-writer pattern as the `SharedDomainType` it replaces
- The enum makes invalid states unrepresentable — a single value captures fork identity, domain type, and transition phase
- ENR updates remain in `Network::on_fork_phase` because they require explicit discv5 mutation
- `count_observed_peers_for_subnets` hoists the lifecycle read outside the per-peer loop to avoid repeated lock acquisition


diegomrsantos:
@claude review this PR

anchor/common/fork/src/lifecycle.rs
```diff
@@ -0,0 +1,200 @@
...
+    Normal {
+        current: Fork,
+        domain_type: DomainType,
+    },
+
+    /// Preparing for an upcoming fork. Dual-subscribing to new topics.
+    ///
+    /// Both forks' contexts are relevant — peers subscribed to either
+    /// the current or upcoming fork are useful.
+    WarmUp {
```
shane-moore, line 25: What about collapsing `WarmUp` and `GracePeriod` variants into something like
```
Transitioning {
   current,
   domain_type
}
```
just an idea since consumers don't seem to care whether they're in `WarmUp` or `GracePeriod`, just that `current` and `domain_type` are correct. and `upcoming`/`previous` fields seem unused

anchor/network/src/peer_manager/connection.rs
```diff
@@ -288,11 +294,34 @@ impl ConnectionManager {
...
+    /// forks are aggregated (union). This prevents peers subscribed only to a
+    /// defunct fork from appearing useful for subnet coverage.
+    fn get_peer_subnets_for_lifecycle(
+        &self,
+        peer: &PeerId,
+        lifecycle: &ForkLifecycle,
+    ) -> Option<Bitfield<Fixed<U128>>> {
         let fork_map = self.observed_peer_subnets.get(peer)?;
-        Some(Self::aggregate_fork_bitmaps(fork_map))
+        match lifecycle {
```
shane-moore: when fork_map is empty (newly connected peer, no gossipsub yet), `Normal` returns `None` (triggering ENR fallback) but `WarmUp`/`GracePeriod` returns  `Some(all-zeros)` (skipping ENR fallback). Adding an  early `if fork_map.is_empty() { return None }` would make the behavior consistent

anchor/common/fork/src/lifecycle.rs
```diff
@@ -0,0 +1,200 @@
...
+use std::sync::Arc;
+
+use parking_lot::RwLock;
+use ssv_types::domain_type::DomainType;
+
+use crate::Fork;
+
+/// Fork lifecycle state. Updated only by ForkMonitor.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum ForkLifecycle {
```
shane-moore, line 12: thinking of ways we could simplify this further @dknopik. I think one way would be to convert this enum to a flat struct:
```
pub struct ForkState {
    pub domain_type: DomainType,
    pub current_fork: Fork,
    pub in_transition: bool,
}
```
since network components seem to just care about what is the current domain type and whether the network is in a fork transition period (warm up/grace period) vs if normal operation

another idea would just be to extend `ForkSchedule` with a `current` field:
```
  pub struct ForkSchedule {
      configs: BTreeMap<Fork, ForkConfig>,
      network_name: String,
      // Runtime: updated by ForkMonitor, read by
  network components
      current: RwLock<ActiveFork>,
  }

  struct ActiveFork {
      fork: Fork,
      in_transition: bool,
  }
```

benefit of this is killing need for new `ForkState` module. con is that it mixes config (current `ForkSchedule` struct) with runtime characteristics (new `current` field)

or perhaps you wanted to take a look and come up with something. if so, feel free to lmk if you want me to try out whatever you're thinking

dknopik:
I think we can simplify this - I am just not sure how exactly. Here are a few possibilities:

1. Roll back #814. Considering the original motivation was a bugfix for keeping the current domain type updated, this has grown quite complex - hundreds of lines and a lot of code to test. Instead, I think the original approach is way simpler and at this point less error prone.
2. Kind of this approach, but less complex: I think having a `RwLock` for the current `ForkConfig` is simple enough. Any parts of the code interested in the lifecycle can use the existing broadcast channel.
3. Finally, avoid the shared state, but instead pass around the `ForkSchedule` to parts of the code. Or maybe a new wrapper that brings along a `SlotClock` for easier querying of the current fork.

Right now, this is the rough order of my preference. Happy to discuss all of these approaches!

diegomrsantos:
Thanks for the feedback, I'll take a look at it next week. Did you consider this is also adding a fork-aware peer selection and isn't just a refactor? The idea was to have only one source of truth for fork-dependent data.

shane-moore:
@dknopik @diegomrsantos, made a POC of how I interpret option 1 against this pr just to see what it could look like https://github.com/diegomrsantos/anchor/compare/feat/shared-fork-lifecycle...shane-moore:anchor:option1-simplification

def kills a lot of code

idea is that `ForkPhase` broadcast is source of truth for fork transitions. so we replace `SharedForkLifecycle` and have `Network::on_fork_phase` dispatch directly to the connection manager / discovery / handshake

and we keep the fork aware peer selection logic from this PR via having `ConnectionManager` track `current_fork` and `in_transition`

Not saying what i'm suggesting is necessarily the right call, just showing what it could look like

diegomrsantos:
> idea is that ForkPhase broadcast is source of truth for fork transitions. so we replace SharedForkLifecycle and have Network::on_fork_phase dispatch directly to the connection manager / discovery / handshake

They serve two different goals. ForkPhase is for components that need to act on fork events. SharedForkLifecycle is for components that only need the current value. 
Less code is good, but not at the expense of clarity. The priority is code that is easy to understand and reason about; cleverness that obscures behavior is a net negative.

One nuance: fewer lines often help clarity, so it’s not less code vs clarity, it’s clarity first.

shane-moore:
agreed with clarity first! i think it's also nice to think about these type of decisions from the perspective of "what's easier for someone new to this part of the codebase to debug?". i would think if you were trying to debug something related to this, you'd just trace through the call paths and find in `Network::on_fork_phase` where domain type propogates to the three components and go from there.  With `SharedForkLifecycle`, it has the benefit of self-contained state, so components don't need to know who updates them. But it might be more daunting to debug since you're tracing through `Arc<RwLock<>>` shared state that gets mutated from the monitor task, vs just following fn calls in `Network::on_fork_phase`

If we find we need this shared state across more components in the future, we can always introduce something like `SharedForkLifecycle` then since it'd for sure be more justified with more consumers

Just my thoughts!

diegomrsantos:
I’m still struggling to see where the additional complexity is coming from here.

We already need a background task (ForkMonitor) to track fork schedule and broadcast ForkPhase transitions. That’s true regardless of whether we propagate state via events or via a shared handle. In this PR, the monitor simply becomes the single writer of a single value (SharedForkLifecycle), and everyone else reads it when they need the current domain type / lifecycle. This is the same pattern we already had with SharedDomainType, just with a better invariant model (ForkLifecycle makes invalid states unrepresentable).  ￼

What this change buys us is removing duplicated “cache derived value on event” logic from multiple components and avoiding desync between “what phase are we in?” vs “what domain type are we using?”  ￼

If the concern is specifically:
	•	traceability (following explicit call paths from Network::on_fork_phase), or
	•	locks (hot-path reads),

can we call that out directly? From what I can see, reads are cheap (and we already hoist outside loops where it matters), and the writer is centralized.  ￼

Happy to simplify further, but I’d like to keep the decision rule explicit:
	•	single source of truth,
	•	minimal duplicated derived state,
	•	easiest to debug “current fork / domain type” at any point in time.

If there’s a concrete code path where reading SharedForkLifecycle is harder to reason about/debug than propagating via Network::on_fork_phase, point me at it and we’ll address it.

anchor/common/fork/src/lifecycle.rs
diegomrsantos: Good call on simplification. I’d keep runtime state separate from `ForkSchedule` (it’s config-only today), so I don’t think adding mutable current there is the right direction.
I’m fine simplifying the consumer API with an `is_transition()` helper, but I’d keep `ForkLifecycle` as an enum for now so we retain `WarmUp` vs `GracePeriod` semantics if/when we need them

dknopik:
Hey, looking through this, I realized that my above feedback was misdirected and too imprecise. 

I worked on some changes that I want to discuss with you guys. However it is pretty late and I want to spend an appropriate amount of time explaining why I think that variant is better, so I will do so tomorrow.

diegomrsantos:
I’ve shared my reasoning, and I don’t think more back-and-forth here will add signal. If you strongly prefer the alternative approach, please take ownership and push a patch; I’m happy to review. Otherwise, I suggest we proceed with the current approach.

dknopik:
I pushed my idea, please take a look. The rough idea is to have one channel, both for lookup of the current fork and for updates, so I used `tokio::sync::watch`. Let me know what you think :)

anchor/common/fork/src/monitor.rs
```diff
@@ -6,377 +6,133 @@
...
-                }),
-                _ => None,
+        for &fork in Fork::all() {
+            // Only consider scheduled forks.
+            let Some(fork_config) = schedule.config(fork) else {
+                continue;
+            };
+            let fork_epoch = fork_config.epoch.as_u64();
+            // Only consider future forks. This check also ensures that `fork_epoch` is non-zero.
+            if fork_epoch <= current_epoch.as_u64() {
```
diegomrsantos: I really like this refactor overall. Using a single `watch` lifecycle makes the flow much easier to follow and reason about.

I think there is one edge case on restart. If the node restarts during the grace window we can come up as `Normal` too early. The `fork_epoch <= current_epoch` check is epoch based, but grace is slot based inside that same epoch. So restarting at `activation_slot + 1` still means grace should be active, but we skip that fork and act like post grace.

Could we add a startup check for `activation_slot <= now < grace_end_slot` and initialize `GracePeriod { current, previous }` in that case, then queue `Normal` at `grace_end_slot`? That should keep the simpler model while covering the boundary case.

diegomrsantos:
Thanks for pushing the refactor @dknopik — I looked through it carefully and I think there are ideas worth keeping. Here's where I land.

**What I like about the `watch` channel approach:**

The core insight is sound: `tokio::sync::watch` naturally provides both "read current state" (`rx.borrow()`) and "react to changes" (`rx.changed()`) in a single primitive. That collapses `SharedForkLifecycle` + `async_broadcast<ForkPhase>` into one channel and eliminates the `ForkPhase` enum entirely. Less surface area, fewer things to keep in sync.

**Proposed middle ground**

Keep the `watch` channel while preserving explicit transition semantics and restart correctness. Concretely:

1. **`watch<ForkLifecycle>` as the single source of truth.** ForkMonitor is the sole sender; all components hold receivers.

2. **One explicit transition handler with (prev, new).** Instead of each component reacting independently to `changed()`, Network's event loop tracks the previous state and dispatches through a single function:

   ```rust
   let mut prev = self.lifecycle_rx.borrow().clone();
   loop {
       tokio::select! {
           // ... other branches ...
           Ok(()) = self.lifecycle_rx.changed() => {
               let new = self.lifecycle_rx.borrow_and_update().clone();
               self.apply_transition(&prev, &new);
               prev = new;
           }
       }
   }
   ```

   The transition handler makes side effects explicit and domain-type-driven:

   ```rust
   fn apply_transition(&mut self, prev: &ForkLifecycle, new: &ForkLifecycle) {
       info!(
           prev_state = ?std::mem::discriminant(prev),
           new_state = ?std::mem::discriminant(new),
           prev_fork = %prev.current_fork_config().fork,
           new_fork = %new.current_fork_config().fork,
           domain_type = ?new.current_fork_config().domain_type,
           "Fork lifecycle transition"
       );

       // ENR update: only when domain type actually changes
       let old_domain = prev.current_fork_config().domain_type;
       let new_domain = new.current_fork_config().domain_type;
       if old_domain != new_domain {
           if let Err(e) = self.discovery().update_enr_domain_type(new_domain) {
               error!(?e, "Failed to update ENR domain type");
           }
       }

       // Subnet cleanup: when grace period ends
       if let (
           ForkLifecycle::GracePeriod { previous, .. },
           ForkLifecycle::Normal { .. },
       ) = (prev, new) {
           // unsubscribe from previous fork's topics
       }
   }
   ```

3. **Structured transition logging.** Every state change logs slot, prev/new state, fork, and domain type — restoring the debugging context that `ForkPhase` events provided.

4. **Restart boundary tests.** At minimum:
   - Restart at `activation_slot + 1` → initial state is `GracePeriod`
   - Restart at `grace_end_slot` → initial state is `Normal`
   - Restart in prep window → initial state is `WarmUp`
   - No transition emitted before target slot

**What stays the same:** `ForkLifecycle` enum, fork-aware peer selection in `ConnectionManager`, all the peer subnet tracking logic. The monitor's internal approach (stateful vs pre-computed) is an independent decision.

**What this avoids:** Rewriting every consumer's constructor signature, changing `ForkLifecycle` variant shapes, and rewriting all monitor tests in one commit.

What do you think about this direction?

anchor/subnet_service/src/subscriptions.rs
```diff
@@ -85,80 +95,53 @@ impl<S: SlotClock> SubnetService<S> {
...
-                                    "Failed to unsubscribe from forks of older fork"
-                                )
-                            }
-                        },
-                    };
-
-                    self.handle_subnet_changes::<E>(&mut service_state).await;
+                Ok(()) = lifecycle_rx.changed() => {
+                    let new_lifecycle = lifecycle_rx.borrow_and_update().clone();
+                    self.on_lifecycle_transition(new_lifecycle, &mut service_state);
```
diegomrsantos: Bug: `on_lifecycle_transition` updates the internal `service_state` (which forks to track), but `handle_subnet_changes::<E>()` is never called afterward. That means no `TopicEvent::Subscribe` / `TopicEvent::Unsubscribe` events are actually emitted to the network on fork transitions.

The previous code always called `handle_subnet_changes` after every `ForkPhase` event:

```rust
// before refactor
Ok(phase) = fork_phase_rx.recv() => {
    match phase { ... };
    self.handle_subnet_changes::<E>(&mut service_state).await;
}
```

Fix — add the missing call:

```rust
Ok(()) = lifecycle_rx.changed() => {
    let new_lifecycle = lifecycle_rx.borrow_and_update().clone();
    self.on_lifecycle_transition(new_lifecycle, &mut service_state);
    self.handle_subnet_changes::<E>(&mut service_state).await;
}
```

Without this, the node won't dual-subscribe to upcoming fork topics on WarmUp and won't unsubscribe old fork topics when the grace period ends.

diegomrsantos: Follow-up: even with `handle_subnet_changes` added back, there's a second issue. When transitioning from `GracePeriod` to `Normal`, `set_subscribed_forks([current])` calls `retain` which **drops the previous fork's `ForkSubscriptions` entry** — including its `currently_subscribed` set — before anyone can emit unsubscribe events for those topics.

The old code handled this explicitly by extracting the subscription state before removing it:

```rust
ForkPhase::GracePeriodEnded { previous, .. } => {
    if let Some(fork) = service_state.forks.remove(&previous.fork)
        && let Err(err) = self.send_unsubscribes(
            &fork.config,
            fork.currently_subscribed   // uses the set before dropping it
        ).await
    { ... }
}
```

With the current refactor, `set_subscribed_forks` silently discards that state via `retain`, so there's no way to know which topics to unsubscribe from afterward.

Fix: either capture and unsubscribe removed forks before `retain` drops them, or have `set_subscribed_forks` return the removed entries so the caller can emit unsubscribes.

diegomrsantos:
@claude review the current PR

diegomrsantos:
> @claude review the current PR

@shane-moore it seems claude review stopped working when mentioned

shane-moore:
> > @claude review the current PR
> 
> @shane-moore it seems claude review stopped working when mentioned

interesting! will look into this

diegomrsantos:
**Test coverage gap in subnet subscription flow**

The monitor internals (`ForkMonitor::new`, `run`) have solid test coverage — the pre-computed transition model is well verified.

However, the subnet subscription flow and network-side ENR update path have zero test coverage. To be fair, these paths weren't tested before the refactor either. But the refactor changed their behavior (see inline comments on `subscriptions.rs:100` about the missing `handle_subnet_changes` call and the `retain` data loss), which makes the lack of coverage riskier now — there's no safety net to catch the regressions.

At minimum, a test that sends a lifecycle transition through the `watch` channel and asserts on the `TopicEvent`s coming out of the `mpsc` receiver would catch both issues. The setup is straightforward since all inputs (`db`, `lifecycle_rx`, `slot_clock`) are already injectable.

anchor/subnet_service/src/subscriptions.rs
dknopik: Good catches, thanks!

anchor/common/fork/src/monitor.rs
diegomrsantos: Done

dknopik (approved):
LGTM, thanks!

PR #829, opened 2026-02-11
fix(ci): fix claude-pr-review failing on fork PRs

diegomrsantos:
## Issue Addressed

`claude-pr-review` fails on all fork PRs since Feb 11 with:
```
fatal: couldn't find remote ref pull/N/head
```

`pull_request_target` reads workflow files from the default branch (`stable`), so this fix must land here.

## Proposed Changes

- Remove fork-specific `repository`/`ref` overrides from checkout (origin must point to base repo where PR refs live)
- Generate the GitHub App token **before** checkout and pass it via `token:`

Root cause: `claude-code-action@v1` merged [anthropics/claude-code-action#851](https://github.com/anthropics/claude-code-action/pull/851) on Feb 10, changing branch fetch from `git fetch origin <branchName>` to `git fetch origin pull/N/head:<branchName>`. The action calls `setupBranch()` before `configureGitAuth()`, so it uses the token embedded by `actions/checkout`. The default `github.token` cannot fetch fork PR refs.

## Additional Info

- Same fix was merged to `unstable` via #827 and #828 but had no effect since `pull_request_target` reads from `stable`
- Same-repo runs (merge queue) are unaffected since `github.token` can access those refs

PR #830, opened 2026-02-12
refactor: claude-mentions

shane-moore:
## Issue Addressed
N/A — Improvement based on team discussion.

## Proposed Changes
 Following our team call discussion about the utility of Claude GitHub Actions, I have some ideas:
  - Remove `claude-pr-review.yml` will removes Claude as a CI check on PRs. We shouldn't be blocked from merging by Claude failures, as we saw with the brittleness in #829 (and the fixes needed in #827, #828). All the review value is still available on-demand via @claude mentions.
  - Improve claude-mentions.yml review prompt via restructuring to prioritize security and correctness over style nits, added explicit "do NOT comment on" rules to reduce noise, and added Rust-specific focus areas (idioms, ownership patterns, design).

## Additional Info
Anyone who wants a Claude review on a PR can still get one by commenting @claude review this. This gives us the same capability without the CI gate or the noise of auto-reviewing every PR.


.github/workflows/claude-mentions.yml
```diff
@@ -82,6 +82,33 @@ jobs:
...
+                      - TODOs, FIXMEs, or documentation formatting
+                      - Code with lint suppression comments (already acknowledged)
+                      - Pre-existing issues not introduced by this PR
+                      - Suggestions that are nice-to-have but not required for correctness
+
+                      Use inline comments for specific issues with concrete suggestions.
+                      Use a single top-level comment for summary only if there are findings.
                   claude_args: |
                       --max-turns 50
                       --model us.anthropic.claude-sonnet-4-5-20250929-v1:0
```
dknopik: Do we want to upgrade the model used?

diegomrsantos:
What's the problem with current claude review? I believe it adds a lot of value.

The problem wasn't caused by the claude review. The problem was that claude action wasn't configured with a fixed version.

diegomrsantos:
@claude wdy5t about https://github.com/sigp/anchor/pull/830#issuecomment-3897392460

diegomrsantos:
@claude I disagree with your arguments.

> Eliminates CI gate that can block merges on infrastructure issues

This can happen with any other task, and we won't remove all of them.

> Reduces notification noise from automated reviews on every PR update

The automatic review happens only once. More noise can be produced with on demand. People who aren't interested in the reviews can ignore if they believe it's noise.

> Maintains same capability via @claude review this (on-demand)

I believe having bigger reviews when it adds value is more common than the opposite.

What could be interesting is if there's a way to configure the automatic review to be less noisy.  Maybe adding only comments about issues it found. @shane-moore @dknopik

.github/workflows/claude-mentions.yml
diegomrsantos: Yes, Opus 4.6 is waaay better

diegomrsantos: Great catch

shane-moore: agreed, updated in https://github.com/sigp/anchor/pull/830/changes/f07886cd3e1def8432a5067fe3048fe61c3ecdf6

shane-moore:
> > Maintains same capability via @claude review this (on-demand)
> 
> I believe having bigger reviews when it adds value is more common than the opposite.
> 
> What could be interesting is if there's a way to configure the automatic review to be less noisy. Maybe adding only comments about issues it found. @shane-moore @dknopik

@diegomrsantos, i like the idea of having the automatic review only commenting when it finds an actual issue. And we use the claude mentions tool separately as to investigate something more deeply.
I updated the pr to fix the version to `claude-code-action@v1.0.52`, which is from yesterday

feel free to lmk what you think about the prompts for `claude-mentions` and `claude-pr-review`. hoping explicitly including in the `claude-pr-review` to only leave comment if it finds a potential issue will help. If it finds no issues, it should just reply with "No actionable issues found." now 🤞

shane-moore:
@claude

regarding the CI failures:
- `target-branch-check` would have to be bypassed since we dont want to merge everything from `unstable` to `stable` for some time
- `check-code` failure was resolved with https://github.com/sigp/anchor/pull/797 to `unstable`, so those changes will get into `stable` eventually

shane-moore:
@dknopik @diegomrsantos, realized these changes need to be merged to `stable` since `stable` is the default branch and both `pull_request_target` and `issue_comment` triggers read workflow files from there. so I updated pr to point at `stable` and cherry picked over recent commits related to the workflow files

ran this flow locally w/o issue to prove to myself we wouldn't see merge issues later:
1. merge this PR → stable: fast-forward (no merge needed, stable
   was behind)
2. merge stable → unstable: auto-merged cleanly (Git       
  resolved overlapping changes automatically since the 
  diffs were compatible)
3. merge unstable → stable: fast-forward again



also, i commented on the CI failures here: https://github.com/sigp/anchor/pull/830#issuecomment-3916924940

.github/workflows/claude-pr-review.yml
```diff
@@ -88,50 +92,47 @@ jobs:
...
+
+                      Only comment on issues you are CONFIDENT are real problems:
+
+                      1. **Security** — vulnerabilities, unsafe code, input validation, auth logic
+                      2. **Correctness** — logic errors, race conditions, edge cases, off-by-one errors
+                      3. **Performance** — bottlenecks, unnecessary allocations, resource leaks
+                      4. **Error Handling** — missing error paths, unwrap/expect in non-test code, silent failures
+                      5. **Rust Idioms** — non-idiomatic patterns, unnecessary clones, misuse of ownership/borrowing
+                      6. **Design** — incorrect abstractions, module boundary violations, missing trait bounds
+                      7. **Testing** — missing coverage for new code paths, untested edge cases
```
diegomrsantos, line 119: Why do we have this duplication?

shane-moore: when I made the PR, I wasn't sure if we wanted the claude-mentions and CI review to have the same instructions, since the previous `claude-mentions` didn't have any guidelines previously. and I wasn't sure if during PR review, folks might want different instructions for claude-mentions vs CI

seems like we're fine with same guidelines between the bots. wdyt of something like a shared `.github/claude-review-instructions.md` that contains this shared prompt?

PR #831, opened 2026-02-13
chore: Change lint `allow`s to `expect`s

dknopik:
## Issue Addressed

`#[allow(lint)]` suppresses a certain lint or compiler warning. While this is useful in some cases, these attributes tend to remain for forever, even if a lint or warning no longer occurs. This is bad as e.g. `#[allow(dead_code)]` can be misleading, or lints like `#[allow(clippy::too_many_arguments)]` that may cause developer to reconsider their design are suppressed.

## Proposed Changes

Use `expect` instead. This fails if the lint is actually not suppressed, so we are forced to remove stale attributes. Also, ban `allow` in the future through the `clippy::allow_attributes` lint.

## Additional Info

Especially forbidding `allow` completely is up for discussion :)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -131,7 +131,8 @@ test-full: cargo-fmt test-release test-debug
 # Clippy lints are opt-in per-crate for now. By default, everything is allowed except for performance and correctness lints.
 lint:
 	cargo clippy --workspace --tests $(EXTRA_CLIPPY_OPTS) --features "$(TEST_FEATURES)" -- \
-		-D warnings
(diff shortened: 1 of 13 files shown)


shane-moore (approved):
I like that we'll get a warning from `expect` when the lint suppression is no longer needed. I'm fine with the `clippy::allow_attributes` preventing use of `allow` so that we favor `expect`, good for consistency across the codebase. if we find some edge case where we think `allow` is needed, we can address it then

PR #833, opened 2026-02-16
test: add spec_tests crate foundation with BeaconVote encoding test

shane-moore:
## Issue Addressed

Begins the work started in #581 to add [ssv-spec type spec tests](https://github.com/ssvlabs/ssv-spec/tree/main/types/spectest). This PR lays the foundation and includes one working test type as a proof of concept, with the goal of breaking the full implementation into smaller PRs for easier review.

## Proposed Changes

 - Add the [ssv-spec](https://github.com/ssvlabs/ssv-spec) repository as a git submodule, pinned at commit `4e2f944d` (101 JSON test fixtures)
   - Create the `spec_tests` workspace crate with a simple test runner that dispatches JSON fixtures to test implementations based on filename prefix matching
   - Implement `BeaconVoteEncodingTest` — SSZ decode, encode roundtrip, and tree hash root verification against the fixture's expected values
   - Add serde deserializer utilities for base64-encoded data and byte-array hashroots

Compared to #581, the orchestration is simplified: a single `match` on the exact filename prefix (mirroring the Go [run_test.go](https://github.com/ssvlabs/ssv-spec/blob/main/types/spectest/run_test.go) switch statement) replaces the enum registry, macro, and walkdir-based discovery. The test logic and adapters are largely the same.

## Additional Info

 This is the first in a series of PRs to build out the remaining type spec tests incrementally. Unimplemented prefixes are currently skipped with a warning during test execution.

Run with: `cargo test -p spec_tests --release -- types_spec_tests`

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "anchor/spec_tests/ssv-spec"]
+	path = anchor/spec_tests/ssv-spec
+	url = https://github.com/ssvlabs/ssv-spec
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7194,6 +7194,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "spec_tests"
+version = "0.1.0"
+dependencies = [
+ "base64",
+ "ethereum_ssz",
+ "ethereum_ssz_derive",
+ "hex",
+ "serde",
+ "serde_json",
+ "ssv_types",
+ "tree_hash",
+ "types",
+]
+
+[[package]]
 name = "spin"
 version = "0.9.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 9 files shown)


shane-moore:
@claude,
1. .unwrap() in lib.rs:48,50 ->           
  (path.file_name().unwrap()) is  safe because fs::read_dir entries always have filenames. We also already use unwrap_or_else with a contextual panic message. Plus this is all #![cfg(test)] code, not production 
  2. Missing Documentation -> Both SpecTest and run_test already have doc comments
  3. Error Message Quality ->The filename in the failure output already tells you which test type failed. Adding the Rust type name is noise.
  4. Negative test cases -> Doesn't apply. Spec tests validate against whatever fixtures ssv-spec provides. We don't write our own negative cases.

dknopik (approved):
LGTM

diegomrsantos:
@shane-moore does this work supersede https://github.com/sigp/anchor/pull/581? If so, can we close that PR?

shane-moore:
> @shane-moore does this work supersede #581? If so, can we close that PR?

great point, closed the previous spec tests PR's since their functionality will be fully expressed in this and subsequent PR's

Issue #819, opened 2026-02-10
handle the new `sign_attestations` more efficiently


dknopik:
In #813, we updated to Lighthouse 8.1.0.

In there is batched signing of attestations via `sign_attestations`. We currently just delegate to `sign_attestation` in its previous form. However, as the actual steps (deciding on a vote and collecting the signatures) are preformed on a *per committee* basis, we can optimize it by performing those steps once per committee. This avoids redundant futures.

Issue #768, opened 2026-01-05
Boole SSV fork meta-issue


jking-aus:
This is a tracking issue for implementing Boole fork changes: 

- [x]  #769 
- [x]  #770 
- [x]  #771 
- [x]  #772 
- [x]  #773 
- [x] #834
- [x] #878

shane-moore:
general update that the issue tracker above is up-to-date with remaining work left for the fork. i would consider #773 non-blocking, but #770 and #834 are necessary. After #834 is merged, similar PR's will be made, which should close out #770

diegomrsantos:
`sign_sync_committee_signatures` still follows the same per-validator calling pattern that [#834](https://github.com/sigp/anchor/pull/834) is addressing for attestations: it fans out to `sign_single_sync_committee_signature`, and each branch calls `decide_instance` with the same `CommitteeInstanceId` for validators in the same committee.

To be clear, this is not about preventing duplicate QBFT instance creation in the normal in-flight case: the current manager already reuses the same live instance for the same `CommitteeInstanceId`. Applying the same group-by-committee pattern here would instead make that sharing explicit at the caller level and avoid repeated `decide_instance` calls / waiter registration for validators in the same committee.

So I think this is still a good follow-up to #834, but as a cleanup / clarity / coordination improvement.

diegomrsantos:
Current Boole status after a fresh repo/test audit:

- `#770`: implementation is largely present on `unstable`; this looks more like pending `stable` promotion / tracker cleanup than missing code.
- `#773`: still needs an audit for what remains beyond the already-merged spec-test work.
- `#878`: targeted attestation committee-batching regression coverage now exists on `unstable` in `committee_attestation.rs`.
- `#886`: aggregate committee-batching regression coverage is ready to review on PR #926, which closes `#886`.
- Remaining dedicated `validator_store` batching test gaps are the two sync paths, now tracked in `#928` and `#929`.

So this tracker is a bit stale as written: after `#926`, the remaining batching-test follow-ups are not just `#878`; they are the sync committee message/contribution test issues in `#928` and `#929`, plus whatever is still truly open under `#773` / `#770`.

PR #834, opened 2026-02-17
fix: optimize the signing of attestations

petarjuki7:
## Issue Addressed

Addresses issue #819 

## Proposed Changes

- Batch attestation signing by SSV committee - runs QBFT consensus once per committee instead of once per validator
- Process committees in parallel with join_all
- Added AttestationData<E> type alias for cleaner code

## Additional Info

Follow-up to #813. The sign_attestations batch signing was delegating to single sign_attestation calls which is inefficient since validators in the same committee share the same QBFT decision.

## Blockers / Dependencies

Depends on PR #855 - currently in review, passing Anchor CI
PR #855 adds changes from sigp/lighthouse#8880 - currently in review, passing LH CI
Merge #855 before this
CI passes for this PR


anchor/validator_store/src/lib.rs
```diff
@@ -1282,46 +1290,44 @@ impl<T: SlotClock, E: EthSpec> AnchorValidatorStore<T, E> {
...
-        attestation: &mut Attestation<E>,
-        current_epoch: Epoch,
-    ) -> Result<(), Error> {
-        if !*self.is_synced.borrow() {
-            return Err(Error::SpecificError(SpecificError::NotSynced));
-        }
+        committee_id: CommitteeId,
+        attestations: Vec<AttestationData<E>>,
+    ) -> Result<Vec<(u64, Attestation<E>, PublicKeyBytes)>, Error> {
+        let slot = attestations[0].3.data().slot;
```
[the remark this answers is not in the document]
petarjuki7: I don't think `attestations` can be empty, that would assume the original `sign_attestation` (from the `ValidatorStore` trait) function got an empty `attestations` field.

anchor/validator_store/src/lib.rs
```diff
@@ -198,6 +206,19 @@ impl<T: SlotClock, E: EthSpec> AnchorValidatorStore<T, E> {
         ))
     }
 
+    /// Get validator metadata by public key.
+    fn get_validator_metadata(
```
shane-moore: let's use the existing `get_validator_and_cluster_instead`

anchor/validator_store/src/lib.rs
```diff
@@ -1334,42 +1353,49 @@ impl<T: SlotClock, E: EthSpec> AnchorValidatorStore<T, E> {
...
+        // Sign for each validator in the committee
+        let mut results = Vec::with_capacity(attestations.len());
+        for (validator_index, pubkey, validator_committee_position, mut attestation) in attestations
+        {
+            let validator = self.get_validator_metadata(pubkey)?;
+
+            // Apply consensus result to this attestation
+            attestation.data_mut().beacon_block_root = data.block_root;
+            attestation.data_mut().source = data.source;
+            attestation.data_mut().target = data.target;
```
shane-moore, line 1620: lets move this:
`let signing_root  = first_att_data.signing_root(domain_hash);`

to be below these 3 mutations
and update it to:
`let signing_root = attestation.data().signing_root(domain_hash);`

since it's possible another operator is proposer and their attestation is different than ours, so we'd want to use the `decided` attestation to derive `signing_root`

additionally, in pre-electra world, it's possible for the `AttestationData.index` to differ between validators in same ssv committee. doubt we're running ssv on any pre-electra chains, but regardless, let's preserve the `Attestation.index` field per validator

anchor/validator_store/src/lib.rs
```diff
@@ -1334,42 +1353,49 @@ impl<T: SlotClock, E: EthSpec> AnchorValidatorStore<T, E> {
...
-            )
-            .await?;
-        attestation
-            .add_signature(&signature, validator_committee_position)
-            .map_err(Error::UnableToSignAttestation)?;
+        // Sign for each validator in the committee
+        let mut results = Vec::with_capacity(attestations.len());
+        for (validator_index, pubkey, validator_committee_position, mut attestation) in attestations
+        {
+            let validator = self.get_validator_metadata(pubkey)?;
```
shane-moore: if this errors for whatever reason, it'll break out of the loop and not process the rest of the attestations. let's consider if it errors, we can `warn!` if unknown pubkey error type and then continuing. if any other error type, `error!` and then continue. wdyt?

anchor/validator_store/src/lib.rs
```diff
@@ -1334,42 +1353,49 @@ impl<T: SlotClock, E: EthSpec> AnchorValidatorStore<T, E> {
...
+                    CollectionMode::Committee {
+                        num_signatures_to_collect,
+                        base_hash: data_hash,
+                    },
+                    &validator,
+                    &cluster,
+                    signing_root,
+                    slot,
+                )
+                .await?;
```
shane-moore: same type of concern here of breaking out of the loop if we hit the `?` here

anchor/validator_store/src/lib.rs
```diff
@@ -1334,42 +1353,49 @@ impl<T: SlotClock, E: EthSpec> AnchorValidatorStore<T, E> {
...
+
+            attestation
+                .add_signature(&signature, validator_committee_position)
+                .map_err(Error::UnableToSignAttestation)?;
```
shane-moore: here as well

anchor/validator_store/src/lib.rs
shane-moore: attestation validation is very time sensitive. in this loop over `attestations`, we `.await` each signature before moving onto the next, i.e. sequential processing, whereas it was being done in parallel via `join_all` before. we could turn this back into parallel futures with `join_all` as a fix. 

I think at that point though, we'd just be recreating the pre-PR code, but with some extra steps and complexity. 

other thoughts:
The gains from committee grouping are fewer calls to `get_validator_and_cluster`, `get_voting_context`, and similar, but these are in-memory lookups. And `decide_instance` is already engineered to handle multiple calls to the same instance gracefully.

Given that, I'm not sure the added complexity
  (`sign_committee_attestations`, committee grouping HashMap, `AttestationData` type alias) justifies itself. What do you think about keeping the per-validator approach? And def lmk if i'm overlooking some solid gains we'd get from this PR

anchor/validator_store/src/lib.rs
```diff
@@ -89,6 +89,14 @@ const SYNC_SELECTION_PROOF_LOG_NAME: &str = "sync selection proof";
 const SYNC_COMMITTEE_SIGNATURE_LOG_NAME: &str = "sync committee signature";
 const SYNC_COMMITTEE_CONTRIBUTION_LOG_NAME: &str = "sync committee contribution";
 
+/// Input data for a single validator's attestation within a committee batch.
+type AttestationData<E> = (
```
shane-moore: `AttestationData` is an ethereum consensus specs container type, so better we don't repurpose the name in anchor as to avoid confusion. we could do something like `AttestationEntry`, or any other ideas are good as well

anchor/validator_store/src/lib.rs
petarjuki7: Thanks for the comment! The original idea was that operations like deciding on a vote and collecting the signatures are preformed on a per committee basis, it would make sense to group them by committee. Also you wouldn't need the QBFT instance per validator as its the same for everyone in the committee. But I agree `decide_instance` is engineered to handle multiple calls to the same instance.
Although some overhead is avoided like fewer DashMap lookups, fewer oneshot channels... etc. but these are probably cheap operations.

Would like to hear @dknopik's input, if we agree the issue wasn't that significant I am okay with closing the PR.

anchor/validator_store/src/lib.rs
```diff
@@ -1334,42 +1340,68 @@ impl<T: SlotClock, E: EthSpec> AnchorValidatorStore<T, E> {
             Completed::Success(data) => data,
         };
```
dknopik, line 1585: We can move this outside the loop

anchor/validator_store/src/lib.rs
```diff
@@ -1282,46 +1290,44 @@ impl<T: SlotClock, E: EthSpec> AnchorValidatorStore<T, E> {
...
+        committee_id: CommitteeId,
+        attestations: Vec<AttestationData<E>>,
+    ) -> Result<Vec<(u64, Attestation<E>, PublicKeyBytes)>, Error> {
+        let slot = attestations[0].3.data().slot;
+        let first_att_data = attestations[0].3.data();
 
-        let (validator, cluster) = self.get_validator_and_cluster(validator_pubkey)?;
-        let voting_context_tx = self.get_voting_context(attestation.data().slot).await?;
+        // Look up cluster for QBFT consensus (all validators share the same cluster)
+        let (_, cluster) = self.get_validator_and_cluster(attestations[0].1)?;
```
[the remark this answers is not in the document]
dknopik: `// All validators in the same committee share the same cluster` is in fact incorrect.

But the cluster is actually not needed, if you modify `decide_instance`. Decide instance actually only needs the cluster members - which are in fact equal to the committee members. :)

anchor/validator_store/src/lib.rs
dknopik: Actually this made me realize that the Pre-PR code is also flawed: if even one involved SSV committee is not able to sign (e.g. due to offline operators), the other signing futures will complete but the results will not be processed until those other signings time out. :/

anchor/validator_store/src/lib.rs
```diff
@@ -1334,42 +1340,68 @@ impl<T: SlotClock, E: EthSpec> AnchorValidatorStore<T, E> {
...
+                    CollectionMode::Committee {
+                        num_signatures_to_collect,
+                        base_hash: data_hash,
+                    },
+                    &validator,
+                    &cluster,
+                    signing_root,
+                    slot,
+                )
+                .await
```
dknopik: This will block as it will wait for `collect_signature` calls for the other validators - which will happen in the next iterations, so this deadlocks. 

Therefore, we need a variant that can register for multiple validators.

Likely, this is why the testnet fails.

anchor/validator_store/src/lib.rs
```diff
@@ -312,6 +320,75 @@ impl<T: SlotClock, E: EthSpec> AnchorValidatorStore<T, E> {
         Ok((*collector.await.map_err(SpecificError::from)?).clone())
     }
 
+    /// Collect signatures for multiple validators in a batch.
+    /// This is more efficient than calling collect_signature multiple times because:
+    /// 1. All partial signatures are sent in a single network message
+    /// 2. No risk of deadlock from sequential processing in Committee mode
+    async fn collect_committee_signatures(
```
diegomrsantos, line 326: There might be a potential reliability regression. This new batch-prep loop is fail-fast on per-validator data (`?` on share lookup, decrypt, and index extraction).

That means one bad validator record can abort signing for the entire committee batch, even if other validators in the same committee are healthy.

Could we switch this to best-effort per validator (`warn!/error!` and `continue`) and only fail the committee if no validators can be prepared? That keeps the batching performance win while preserving failure isolation.

diegomrsantos, line 326: Readability suggestion: `collect_committee_signatures` currently mixes 3 responsibilities in one loop:
1) share lookup
2) share decrypt/cache resolution
3) `ValidatorSigningData` construction (`index` extraction)

Could we extract a single helper like `prepare_validator_signing_data(...)` (and optionally `resolve_decrypted_share(...)`)?

That would make the batching flow easier to scan and keep failure policy (fail-fast vs skip-and-continue) centralized in one place.

Example refactor shape:

```rust
  async fn collect_committee_signatures(
      &self,
      signature_kind: PartialSignatureKind,
      role: Role,
      slot: Slot,
      committee_id: CommitteeId,
      cluster: &Cluster,
      validators: Vec<(ValidatorMetadata, Hash256)>,
  ) -> Result<HashMap<ValidatorIndex, Signature>, Error> {
      let metadata = SignatureMetadata {
          kind: signature_kind,
          role,
          threshold: cluster
              .get_f()
              .safe_mul(2)
              .and_then(|x| x.safe_add(1))
              .map_err(SpecificError::from)?,
          slot,
          committee_id,
      };

      let mut validator_data = Vec::with_capacity(validators.len());
      for (validator, signing_root) in validators {
          match self.prepare_validator_signing_data(&validator, signing_root) {
              Ok(data) => validator_data.push(data),
              Err(e) => {
                  warn!(?e, pubkey = ?validator.public_key, "Skipping validator in batch signing");
                  continue;
              }
          }
      }

      if validator_data.is_empty() {
          return Ok(HashMap::new());
      }

      let signatures = self
          .signature_collector
          .sign_and_collect_batch(metadata, validator_data)
          .await

      Ok(signatures.into_iter().map(|(k, v)| (k, (*v).clone())).collect())
  }

  fn prepare_validator_signing_data(
      &self,
      validator: &ValidatorMetadata,
      signing_root: Hash256,
  ) -> Result<ValidatorSigningData, Error> {
      let index = validator.index.ok_or(SpecificError::MissingIndex)?;
      let share = self.resolve_decrypted_share(validator)?;
      Ok(ValidatorSigningData {
          root: signing_root,
          index,
          share,
      })
  }

  fn resolve_decrypted_share(
      &self,
      validator: &ValidatorMetadata,
  ) -> Result<Option<SecretKey>, Error> {
      let Some(operator_key) = &self.private_key else {
          // impostor mode
          return Ok(None);
      };

      let encrypted_private_key = self
          .database
          .state()
          .shares()
          .get_by(&validator.public_key)
          .ok_or(Error::UnknownPubkey(validator.public_key))?
          .encrypted_private_key;

      let key = self
          .decrypted_keys
          .lock()
          .try_get_or_insert(encrypted_private_key, || {
              decrypt_key_share(operator_key, encrypted_private_key, validator.public_key)
                  .map_err(|_| SpecificError::KeyShareDecryptionFailed)
          })
          .cloned()?;

      Ok(Some(key))
  }
```

anchor/validator_store/src/lib.rs
```diff
@@ -2561,46 +2721,55 @@ impl<T: SlotClock, E: EthSpec> ValidatorStore for AnchorValidatorStore<T, E> {
 
     async fn sign_attestations(
```
diegomrsantos: Could we add a regression test for committee batch failure isolation?

 Suggested scenario:
  - one committee with 2+ validators
  - inject one validator prep failure (e.g. missing index/share/decrypt failure)
  - assert healthy validator(s) in the same committee are still signed and returned

This would lock in the intended behavior for the new batch path.

diegomrsantos:
Non-blocking: `validator_store/src/lib.rs` is carrying too many responsibilities now (attestation batching, QBFT orchestration, assignment counting, slashing glue, error mapping, etc.). Splitting into focused modules would improve readability and testability.

Suggested follow-up split:
  - `attestation_signing.rs` (`sign_attestations`, `sign_committee_attestations`, `collect_committee_signatures`)
  - `consensus.rs` (QBFT helpers/validators)
  - `assignments.rs` (`VotingContext`, `VotingAssignments`, counting helpers)
  - `errors.rs` (`SpecificError` + conversions)

I’d prefer this as a separate refactor PR so this optimization PR stays easy to review.

anchor/validator_store/src/lib.rs
diegomrsantos: Do you know if it has been resolved?

anchor/validator_store/src/lib.rs
diegomrsantos: Do you know if it has been resolved?

anchor/qbft_manager/src/lib.rs
```diff
@@ -213,15 +214,20 @@ impl<E: EthSpec, S: SlotClock + Clone + 'static> QbftManager<E, S> {
         let include_epoch_shift = self.fork_schedule.active_fork(epoch) >= Fork::Boole;
         let leader_fn = DefaultLeaderFunction::new(self.slots_per_epoch, include_epoch_shift);
 
-        // General the qbft configuration
+        // Calculate fault tolerance: f = (n - 1) / 3, quorum = n - f
+        let n = committee_members.len();
+        let f = (n.saturating_sub(1) / 3) as u64;
+        let quorum_size = n - f as usize;
```
diegomrsantos: I might be missing context, but I think this duplicates builder defaults.

`ConfigBuilder::new_with_leader_fn(...)` already initializes quorum as `committee_size - f` (with `f = (committee_size - 1) / 3`), which is the same value computed here.

So unless we want to explicitly pin quorum at this call site, we could simplify by removing:
  - the local `n/f/quorum_size` calculation
  - `.with_quorum_size(quorum_size)`

Is there a reason to keep the explicit override?

shane-moore:
we should perhaps mark this PR as draft until changes from https://github.com/sigp/lighthouse/pull/8880 are incorporated. To be unblocked, would recommend updating the lighthouse packages in `Cargo.toml` in this PR to point at the last commit in my branch in https://github.com/sigp/lighthouse/pull/8880 and then refactor accordingly, since I dont anticipate too much request for changes on the lh PR

diegomrsantos:
@shane-moore thanks for the heads-up. Marked it as a draft. @petarjuki7 could you please address @shane-moore suggestion?

petarjuki7:
@diegomrsantos @shane-moore So, I changed the lighthouse version to point to the newest commit on this new PR (https://github.com/sigp/lighthouse/pull/8880). The main change is that I incorporated streams into the `sign_attestations()` function. The trait change for the `ValidatorStore` in the LH pr also introduced some other changes, which I'm not going to do in this PR to not mix responsibilities. So this PR will stay as a draft and will not compile for now, but it can be reviewed. I'm going to open another PR to include all of the changes from the new `ValidatorStore` trait, and after that, this can be merged.

diegomrsantos:
> @diegomrsantos @shane-moore So, I changed the lighthouse version to point to the newest commit on this new PR ([sigp/lighthouse#8880](https://github.com/sigp/lighthouse/pull/8880)). The main change is that I incorporated streams into the `sign_attestations()` function. The trait change for the `ValidatorStore` in the LH pr also introduced some other changes, which I'm not going to do in this PR to not mix responsibilities. So this PR will stay as a draft and will not compile for now, but it can be reviewed. I'm going to open another PR to include all of the changes from the new `ValidatorStore` trait, and after that, this can be merged.

CI is failing right now. Can you get the workflow jobs passing first? Let’s align that PRs are ready for review only when all CI checks are green.

shane-moore:
> CI is failing right now. Can you get the workflow jobs passing first? Let’s align that PRs are ready for review only when all CI checks are green.

yah, this PR has evolved a lot since inception, since it later inherited a dependency on upstream lh changes after some investigation here. To keep this PR focused  on refactoring `sign_attestations` and to also keep CI green, I propose we first review and merge PR https://github.com/sigp/anchor/pull/855 to `unstable` and then rebase this PR on top of it. 

To stay unblocked in the meantime, feel free to rebase this PR on top of it for now as well and continue building out the committee based batched signing for `sign_attestations`

anchor/validator_store/src/lib.rs
```diff
@@ -1282,48 +1393,56 @@ impl<T: SlotClock, E: EthSpec> AnchorValidatorStore<T, E> {
...
-        current_epoch: Epoch,
-    ) -> Result<(), Error> {
-        if !*self.is_synced.borrow() {
-            return Err(Error::SpecificError(SpecificError::NotSynced));
-        }
-
-        let (validator, cluster) = self.get_validator_and_cluster(validator_pubkey)?;
-        let voting_context_tx = self.get_voting_context(attestation.data().slot).await?;
+        committee_id: CommitteeId,
+        attestations: Vec<AttestationEntry<E>>,
```
shane-moore: lets pass `AttestationToSign` instead and kill `AttestationEntry` since don't need it anymore

anchor/validator_store/src/lib.rs
```diff
@@ -2559,67 +2753,88 @@ impl<T: SlotClock, E: EthSpec> ValidatorStore for AnchorValidatorStore<T, E> {
...
+    fn sign_attestations(
         self: &Arc<Self>,
-        mut attestations: Vec<(u64, PublicKeyBytes, usize, Attestation<E>)>,
-    ) -> Result<Vec<(u64, Attestation<E>)>, Error> {
-        if !*self.is_synced.borrow() {
-            return Err(Error::SpecificError(SpecificError::NotSynced));
-        }
+        attestations: Vec<AttestationToSign<E>>,
+    ) -> impl Stream<Item = Result<Vec<(u64, Attestation<Self::E>)>, Error>> + Send {
+        async_stream::stream! {
```
shane-moore: lets consider something like this instead:
```
 if !*self.is_synced.borrow() {
            return futures::future::Either::Left(futures::stream::once(async {
                Err(Error::SpecificError(SpecificError::NotSynced))
            }));
        }
```

so that we can remove the slight overhead of adding `async_stream` dependency, and i think the more vanilla `Either::Left` style is easier to reason about than what's happening under the hood with the `stream!` macro

petarjuki7:
This PR is rebased on top of #855 (refactor/validator-store-streaming). Once that merges into unstable, the diff here will only show the committee batching + streaming changes.

diegomrsantos:
Process suggestion: can we make blockers/dependencies explicit in the PR description?

While reviewing this PR, I had trouble understanding the full context because key dependencies were distributed across comments and other PRs.

Concrete ask:
Please add a `Blockers / Dependencies` section to the PR description with the current status of:
- `#855`
- `sigp/lighthouse#8880`
- expected merge order
- CI expectation for review (green vs draft)

This would make reviews much easier now and for future readers.

If the team agrees, I’m happy to open a small PR to add this section to the PR template.

petarjuki7:
> Process suggestion: can we make blockers/dependencies explicit in the PR description?
> 
> While reviewing this PR, I had trouble understanding the full context because key dependencies were distributed across comments and other PRs.
> 
> Concrete ask: Please add a `Blockers / Dependencies` section to the PR description with the current status of:
> 
> * `#855`
> * `sigp/lighthouse#8880`
> * expected merge order
> * CI expectation for review (green vs draft)
> 
> This would make reviews much easier now and for future readers.
> 
> If the team agrees, I’m happy to open a small PR to add this section to the PR template.

Makes sense, edited!
I would be okay with the new PR section in the PR template also

petarjuki7:
@claude can you re-review this?

Cargo.toml
```diff
@@ -151,6 +151,7 @@ tokio = { version = "1.39.2", features = [
     "sync",
     "macros",
 ] }
+tokio-stream = "0.1.18"
```
shane-moore: think we can remove this dep now

anchor/validator_store/src/lib.rs
```diff
@@ -1927,6 +2108,10 @@ pub enum SpecificError {
     KeyShareDecryptionFailed,
     DataTooLarge(String),
     ClusterLiquidated,
+    /// Empty attestations list provided to sign_committee_attestations
+    NoAttestationsProvided,
+    /// No cluster found for the given committee ID
+    UnknownCommittee(CommitteeId),
```
shane-moore: seems unused. can remove? could add a todo if need in future pr for some reason

anchor/database/src/state.rs
```diff
@@ -270,7 +270,7 @@ impl NetworkState {
         nonces.collect()
     }
 
-    fn get_cluster_members(&self, committee_id: &CommitteeId) -> Option<IndexSet<OperatorId>> {
```
shane-moore, line 273: can this go back to being private fn?

anchor/qbft_manager/Cargo.toml
```diff
@@ -10,6 +10,7 @@ dashmap = { workspace = true }
 database = { workspace = true }
 ethereum_ssz = { workspace = true }
 fork = { workspace = true }
+indexmap = { workspace = true }
```
shane-moore: i think we can remove this dependency and do `use ssv_types::IndexSet` instead

anchor/validator_store/src/lib.rs
```diff
@@ -1556,6 +1576,167 @@ impl<T: SlotClock, E: EthSpec> AnchorValidatorStore<T, E> {
...
+    /// Runs QBFT consensus once for the committee, then collects signatures for each validator.
+    async fn sign_committee_attestations(
+        &self,
+        committee_id: CommitteeId,
+        attestations: Vec<AttestationToSign<E>>,
+    ) -> Result<Vec<(u64, Attestation<E>, PublicKeyBytes)>, Error> {
+        // Early return for empty attestations to avoid index out of bounds
+        let first_attestation = attestations
+            .first()
+            .ok_or(Error::SpecificError(SpecificError::NoAttestationsProvided))?;
```
shane-moore: there's only one caller of `sign_committee_attestations`, and it's already called in a loop, so we should never hit this `SpecificError::NoAttestationsProvided`. i think we can kill this error variant by downgrading to:
```
  let Some(first_attestation) =  attestations.first() else {                  
      warn!("sign_committee_attestations called with empty attestations");
      return Ok(vec![]);
  };
```

shane-moore:
I'm trying to think of what `sign_and_collect_batch` would look like for `sign_attestations` and `sign_sync_committee_signatures` to be able to combine signature on a per committee basis, like they do today, using `num_signatures_to_collect` to indicate to the `committee_signatures` dashmap when it's time to fire the single partial sig message to the network. since both the attestation and sync messages have same `base_hash`, they get keyed to same bucked in the `committee_signatures` dashmap

we would need to extend `sign_and_collect_batch` to add its validators to the `committee_signatures` dashmap. and once `num_signatures_to_collect` is hit, proceed to sending the network message

but this is getting rather close to re-implementing the same functionality as already done in `sign_and_collect`. I'd favor removing `sign_and_collect_batch` and  reverting to having `collect_committee_signatures` use `join_all(collect_signature(Committee))` with correct `num_signatures_to_collect` (which avoids the sequential deadlock that motivated the `sign_and_collect_batch` approach). Reasons:
- less code to maintain
- `join_all` solves the deadlock -> the original  concern was that sequential `collect_signature` calls deadlock because each `.await` waits before the next validator is registered. `join_all` enqueues all signing closures concurrently, so all validators are registered before any result is awaited
- the dashMap accumulation path already handles the attestation + sync combination correctly

I think similar to the concerns in this [issue](https://github.com/sigp/anchor/issues/745#issuecomment-3986977283), we want to avoid refactors for performance without having metrics supporting they are legit bottlenecks.  I think it would be good to add a gh issue to add metrics around this signing flow so we can do proper profiling.  Then, if justified, we can consider a `sign_and_collect_batch`

just my thoughts. open to other suggestions as well, perhaps i'm missing some reason why `sign_and_collect_batch` is required to prevent slow committees blocking faster committees, but i think the stream approach resolves that

anchor/signature_collector/src/lib.rs
```diff
@@ -282,6 +282,132 @@ impl<S: SlotClock + Clone + 'static> SignatureCollectorManager<S> {
...
+
+                // Send ONE message with ALL partial signatures
+                let msg = match manager.create_message(
+                    &metadata,
+                    partial_signatures.clone(),
+                    &DutyExecutor::Committee(metadata.committee_id),
+                ) {
+                    Ok(msg) => msg,
+                    Err(err) => {
+                        error!(%err, "Failed to create batch partial signature message");
```
diegomrsantos: If `create_message` fails here, we return without notifying the registered oneshot waiters. That later surfaces as per-validator `Failed to receive signature` warnings, which obscures the root cause. Could we propagate this batch-creation error to each waiter (or at least add explicit log linkage)?

anchor/validator_store/src/lib.rs
```diff
@@ -1556,6 +1576,167 @@ impl<T: SlotClock, E: EthSpec> AnchorValidatorStore<T, E> {
...
+            Hash256,
+        )> = Vec::with_capacity(attestations.len());
+        for att in attestations {
+            let (validator_index, pubkey, validator_committee_position, mut attestation) = (
+                att.validator_index,
+                att.pubkey,
+                att.validator_committee_index,
+                att.attestation,
+            );
+            let validator = match self.get_validator_and_cluster(pubkey) {
```
diegomrsantos: We already resolve the committee `cluster` above and we grouped by committee before calling this method. Could we avoid `get_validator_and_cluster` per validator here and fetch only `ValidatorMetadata`? That trims repeated state lookups and avoids first/inner lookup drift.

anchor/database/src/state.rs
```diff
@@ -270,7 +270,7 @@ impl NetworkState {
...
+    pub fn get_cluster_members(&self, committee_id: &CommitteeId) -> Option<IndexSet<OperatorId>> {
```
diegomrsantos: Is `pub` needed here? In this PR, `get_cluster_members` still appears to be consumed only internally (via `get_cluster_members_for_validator`). If there is no external caller, keeping it private (`fn`) preserves encapsulation.

anchor/signature_collector/src/lib.rs
```diff
@@ -282,6 +282,132 @@ impl<S: SlotClock + Clone + 'static> SignatureCollectorManager<S> {
         Ok(result_rx.await?)
     }
 
+    /// Sign messages for multiple validators and wait until all signatures are reconstructed.
+    /// This is more efficient than calling sign_and_collect multiple times because:
+    /// 1. All partial signatures are sent in a single network message
+    /// 2. No risk of deadlock from sequential processing in Committee mode
+    pub async fn sign_and_collect_batch(
```
diegomrsantos: This path adds a large new production flow. Can we add dedicated tests for: (1) multi-validator batch success and (2) failure isolation where one validator prep failure does not block other validators in the same committee?

anchor/validator_store/src/lib.rs
```diff
@@ -2623,81 +2808,67 @@ impl<T: SlotClock, E: EthSpec> ValidatorStore for AnchorValidatorStore<T, E> {
...
-                        );
-                    }
+        // Group attestations by SSV committee
+        let mut committee_mapping: HashMap<CommitteeId, Vec<AttestationToSign<E>>> = HashMap::new();
+        for att in attestations {
+            let pubkey = att.pubkey;
+            match self.get_validator_and_cluster(pubkey) {
+                Ok((_, cluster)) => {
+                    committee_mapping
+                        .entry(cluster.committee_id())
```
diegomrsantos, line 2781: Safety check: grouping key is only `committee_id`, while `sign_committee_attestations` derives slot/data from the first item. Current caller appears single-slot, but should we either key by `(committee_id, slot)` or assert slot consistency within each group to prevent future misuse?

anchor/signature_collector/src/lib.rs
diegomrsantos: This method has become quite dense (register waiters, sign/build/send, then await/aggregate). Could we split it into small helpers (e.g. `register_batch_notifiers`, `send_batch_partials`, `collect_batch_results`)? It would make the control flow and error handling (especially early returns in the signing closure) much easier to reason about.

anchor/signature_collector/src/lib.rs
shane-moore: I am proposing in https://github.com/sigp/anchor/pull/834#issuecomment-4000968444 that we don't modify signature collector for reasons described in the comment. If I'm missing something and signature collector mods are necessary, I'm certainly open to it

anchor/signature_collector/src/lib.rs
shane-moore: same here as https://github.com/sigp/anchor/pull/834#discussion_r2899995433

shane-moore: same here as https://github.com/sigp/anchor/pull/834#discussion_r2899995433

anchor/signature_collector/src/lib.rs
diegomrsantos: > I am proposing in [#834 (comment)](https://github.com/sigp/anchor/pull/834#issuecomment-4000968444) that we don't modify signature collector for reasons described in the comment. If I'm missing something and signature collector mods are necessary, I'm certainly open to it

Tbh, I reviewed it before fully understanding your comment. It's hard to keep the full context in mind when someone isn't too familiar with this part of the code. But I'll do it.

diegomrsantos:
I'm a bit confused, is this PR a necessary change for the boole fork, or is it an optimization?

anchor/signature_collector/src/lib.rs
shane-moore: I get that! I spent some hours myself building context in order to formulate that comment because i was less familiar with signature collector. In a nut shell, the validator store changes are something we definitely want to get in for boole fork because it will enable us to send committee based batches of attestations to the beacon node.  Currently, we don't have the batches, and a slow committee's post-consensus sig collection will block faster committee's.  I've convinced myself we can accomplish the batching we need w/o changes to the sig collector logic, but I could of course be wrong. On your end, no rush, and feel free to wait for the next round of changes before reviewing further

diegomrsantos:
> I'd favor removing `sign_and_collect_batch` and reverting to having `collect_committee_signatures` use `join_all(collect_signature(Committee))` with correct `num_signatures_to_collect` (which avoids the sequential deadlock that motivated the `sign_and_collect_batch` approach).

Thanks for the detailed write-up, this helped clarify the tradeoffs.

  After re-reading the flow, I think we’re aligned on direction:

- This PR is not Boole-fork blocking.
- Current pre-PR behavior is already correct:
  - `sign_attestations` fans out per-validator concurrently (`join_all`)
  - QBFT dedups by committee/slot instance internally
- So “one consensus per committee” is already happening implicitly today.

Where I share the concern is `sign_and_collect_batch`:
  - it introduces a second batching path in `signature_collector`
  - it does not go through the existing `CollectionMode::Committee` / `committee_signatures` accumulation contract
  - that increases risk of semantic drift from existing committee counting behavior (`num_signatures_to_collect`, `base_hash`) used by current committee collection flows

I’d favor the lower-risk path for now:
1. keep committee-level streaming in `sign_attestations` (`FuturesUnordered`) so slow committees don’t block fast ones
2. remove `sign_and_collect_batch`
3. inside each committee, use concurrent `join_all(collect_signature(CollectionMode::Committee))` with correct `num_signatures_to_collect` and existing `base_hash` semantics

This keeps behavior aligned with the current collector contract while still delivering the resilience improvement from streaming.

shane-moore:
> I'm a bit confused, is this PR a necessary change for the boole fork, or is it an optimization?

this PR is a fix to our existing blocking flow.  If an operator's committee based attempt to acquire attestation partial signatures fails due to any number of reasons, it blocks all other committtee's signed attestations from being submitted to the BN.  hence the stream of committee based batches approached introduced here.  we want to fix it and test in boole devnet

anchor/validator_store/src/lib.rs
petarjuki7: I added a check for that. Afaik Lighthouse calls `sign_attestations` once per slot with attestations for that slot only, but worth checking if that changes in the future

shane-moore: @diegomrsantos, @petarjuki7  i think a grouping like `(committee_id, slot)` only makes sense if there were scenarios where the attestations' slots received from lh would be different, which is not the case today, since mismatched slots in attestations could lead to incorrect attestation signing

I'd also support removing the new check:
```
     // All attestations in a committee batch must be for the same slot
        if let Some(mismatch) = attestations
            .iter()
            .find(|att| att.attestation.data().slot != slot)
        {
            error!(
                expected = %slot,
                found = %mismatch.attestation.data().slot,
                pubkey = ?mismatch.pubkey,
                "Slot mismatch in committee attestation batch"
            );
        }
```

because it only logs an error but allows for continued processing with wrong data.  Instead, we should take advantage of lh already having various guards in place to prevent this scenario from occuring. additionally, keeping the check makes you feel like this scenario could occur, which is confusing.  if anything, we could add a comment instead of the check explaining why the slot mismatch is a non-credible scenario, i.e.:
```
  // Slot consistency is guaranteed by Lighthouse's `AttestationService`. Duties are
  // filtered to a single slot before reaching `sign_attestations`. No check here
  // because no corrective action is possible once QBFT context is committed.
  ```

I considered other options. Returning `Err` kills the committee. Logging and continuing corrupts one attestation. Simpler to remove and take advantage of the upstream lh invariants

File: anchor/validator_store/src/lib.rs [hunk 26]
```diff
@@ -314,6 +319,63 @@ impl<T: SlotClock, E: EthSpec> AnchorValidatorStore<T, E> {
...
+        let results = join_all(futures).await;
+
+        let mut signatures = HashMap::with_capacity(results.len());
+        for (index, result) in results {
+            match result {
+                Ok(sig) => {
+                    signatures.insert(index, sig);
+                }
+                Err(e) => {
+                    warn!(?index, error = ?e, "Failed to collect signature for validator");
```
shane-moore: feels more like an `error!` instead of warn since missed signatures mean missed attestations, i.e. lost money for the operator

anchor/validator_store/src/lib.rs
petarjuki7: Agree, will update

anchor/validator_store/src/lib.rs
```diff
@@ -2623,81 +2777,67 @@ impl<T: SlotClock, E: EthSpec> ValidatorStore for AnchorValidatorStore<T, E> {
...
+                async move {
+                    let signed = match this
+                        .sign_committee_attestations(committee_id, attestations)
+                        .await
+                    {
+                        Ok(signed) if signed.is_empty() => return Ok(Vec::new()),
+                        Ok(signed) => signed,
+                        Err(e) => {
+                            error!(?committee_id, error = ?e, "Failed to sign committee attestations");
+                            return Err(e);
```
shane-moore: lets log the error then return `Ok(Vec::new())` because otherwise the error propogate up and does a `crit!("Failed to sign attestations")` on the lh side. which is not entirely accurate. it's more that this individual operator failed, but perhaps quorum was reached by other operators

shane-moore:
lets kill this helper that's no longer used
```
 pub fn with_quorum_size(mut self, quorum_size: usize) -> Self {
        self.quorum_size = quorum_size;
        self
    }
```
in `anchor/common/qbft/src/config.rs`

anchor/validator_store/src/lib.rs
shane-moore: @petarjuki7, this would be good to include in the PR

diegomrsantos:
Non-blocking, but I think we should add direct tests for the new committee-batching path before or immediately after merge.

Most of the behavioral change in this PR lives in `anchor/validator_store/src/lib.rs` (`sign_attestations`, `sign_committee_attestations`, `collect_committee_signatures`), but the diff doesnt add targeted coverage there. The current tests dont really lock in the new streaming / per-committee semantics.

The cases I’d want covered are:

- multiple SSV committees in one `sign_attestations()` call  
  Assert that the stream yields separate batches per committee, and that the test does not rely on completion order.

- mixed success/failure across committees  
  Assert that one failing committee does not prevent another committee from producing a successful batch.

- committee attestation batching with the existing sync-committee post-consensus collection contract  
  Assert that the attestation path still uses the same `CollectionMode::Committee` / `num_signatures_to_collect` / `base_hash` semantics expected by the shared collector path.

I’m not calling this a correctness bug in the current code, but without these tests the next cleanup/refactor in `validator_store` could easily regress the new behavior without us noticing.

shane-moore (approved):
I think this PR is in good shape to merge. The CI [failure](https://github.com/sigp/anchor/actions/runs/22922951177/job/66525942236?pr=834) will be resolved once this [PR](https://github.com/sigp/anchor/pull/874#pullrequestreview-3925968603) is merged.

plz make a gh issue to add the follow-up unit tests

diegomrsantos:
> I think this PR is in good shape to merge. The CI [failure](https://github.com/sigp/anchor/actions/runs/22922951177/job/66525942236?pr=834) will be resolved once this [PR](https://github.com/sigp/anchor/pull/874#pullrequestreview-3925968603) is merged.
> 
> plz make a gh issue to add the follow-up unit tests

Will the tests be worked out before we release? It's important.

shane-moore:
> > I think this PR is in good shape to merge. The CI [failure](https://github.com/sigp/anchor/actions/runs/22922951177/job/66525942236?pr=834) will be resolved once this [PR](https://github.com/sigp/anchor/pull/874#pullrequestreview-3925968603) is merged.
> > plz make a gh issue to add the follow-up unit tests
> 
> Will the tests be worked out before we release? It's important.

good point. added the test creation task to the meta boole issue https://github.com/sigp/anchor/issues/768. @petarjuki7, perhaps make some issues for the remaining `sign_*` refactors that we need and add to  https://github.com/sigp/anchor/issues/768. alternatively, add them as sub-issues directly in https://github.com/sigp/anchor/issues/768

PR #837, opened 2026-02-18
test: add message type validation spec tests

shane-moore:
## Issue Addressed
Continuation of spec test PR's

## Proposed Changes
- Adds spec test coverage for SSV message type serialization and validation, covering `SSVMessage`, `SignedSSVMessage`, and `PartialSignatureMessages`.

## Additional Info
Blocked by https://github.com/sigp/anchor/pull/843 since it's dependent on its changes


shane-moore:
@claude, fixed your point 4. 
  Not legit / noise:
  - 1 — .unwrap_or() in deserializers — The review contradicts itself. It flags .unwrap_or() as "critical" then
  immediately says "The code actually handles these correctly!" strip_prefix("0x").unwrap_or(&hex_str) is the
  standard Rust idiom for "strip if present, otherwise keep original." No issue here.
  - 2 - The .unwrap() can never panic — the length was already validated to be exactly 8 on lines 88-93. The try_into()
  (converting &[u8] to [u8; 8]) only fails on length mismatch, which is impossible at this point. The PR's change
  to map_err adds 6 lines of redundant defensive code producing the exact same error that was already returned 4
  lines earlier.  - 3 — .expect() message in test utils — This is test-only code with a hardcoded constant. The .expect() message
  is already fine. Over-engineering the message adds nothing.
  - 5 — "Intermediate structs lack rationale" — The structs already have doc comments explaining exactly why they
  exist (lines 21-26 of partial_sig_message.rs, lines 25-28 of signed_ssv_msg.rs). The reviewer didn't read them.
  - 6 — "Unnecessary allocation" in prepare_signatures — It's test code, and the silent truncation/padding is
  intentional — Go fixtures have variable-length sigs that get padded to [u8; 256] for Anchor's type. Adding a
  length warning would be noise.
  - 7 — Missing JSON format examples in deserializers — Over-engineering. The module doc already says "Go JSON
  fixture format." These are internal deserialize_with functions, not public API.
  - 8 — Submodule not documented — The submodule already existed before this PR. Not in scope.
  - 9 — No unit tests for deserializers — They're exercised by 26 spec test fixtures. Separate unit tests for each
   deserializer would be redundant.

shane-moore:
@claude, please take another look with fresh eyes and comment if have any findings

.claude/agents/code-reviewer-subagent.md
```diff
@@ -1,51 +0,0 @@
----
```
diegomrsantos, line 1: Has this been deleted by accident?

anchor/common/ssv_types/src/deserializers.rs
```diff
@@ -0,0 +1,93 @@
+//! Custom serde deserializers for Go JSON fixture format.
```
diegomrsantos: I have a design concern here, we’re mixing fixture-adapter logic into `ssv_types`.

The new `serde` deserializers in `ssv_types` are primarily handling **Go fixture JSON shape** differences (hex/base64/string-number fields, PascalCase names), e.g.:
  - `MsgID` hex string -> `MessageId`
  - `Data` base64 string -> `VariableList<u8, ...>`
  - `Slot` / `ValidatorIndex` string -> numeric types
  - `PartialSignature` hex -> `bls::Signature`

I think we should separate concerns:

  1. Keep **domain validation logic** in production crates (e.g. `PartialSignatureMessages::validate()` is production-relevant and belongs in `ssv_types`).
  2. Move **fixture parsing adapters** out of `ssv_types` and into`spec_tests` (DTOs + `TryFrom` into `ssv_types` types).

This keeps `ssv_types` focused on runtime/domain behavior and keeps fixture-format compatibility where it belongs (test harness).

diegomrsantos:
## Scope mismatch: `test:` PR modifies production code

The PR is titled `test:` but makes substantive changes to two production crates:

- **`anchor/common/ssv_types/`**: adds `validate()` method, `PartialSignatureMessagesError` enum, `thiserror` dep, `serde` feature, `deserializers` module, and serde derives on `SSVMessage`, `PartialSignatureMessages`, `PartialSignatureMessage`, `OperatorId`, `MsgType`
- **`anchor/message_validator/src/partial_signature.rs`**: refactors validation logic, introduces a new `ZeroSigner` error path, changes error ordering

A `test:` PR should only contain test changes. Production behavior changes belong in their own `refactor:` or `feat:` PR so reviewers can evaluate them independently.

Suggested split:
1. `refactor: extract PartialSignatureMessages::validate() to ssv_types` — production changes
2. `test: add message type validation spec tests` — test-only, depends on (1)

---

**General note:** smaller, focused PRs (one concern per PR) are easier to review, less risky, and get merged faster. When production and test changes land together it's harder to reason about behavioral impact and rollback if needed.

anchor/message_validator/src/partial_signature.rs
```diff
@@ -105,18 +105,29 @@ fn validate_partial_signature_message_semantics(
...
+        }
+        PartialSignatureMessagesError::ZeroSigner => ValidationFailure::ZeroSigner,
+    })?;
+
+    // Rule: Partial signature signer must match the signed message's signer.
+    // validate() ensures all inner signers are the same, so checking one suffices.
+    if partial_signature_messages
+        .messages
+        .first()
+        .expect("validate() ensures non-empty")
```
diegomrsantos: `.expect()` in production code — this function is not behind `#[cfg(test)]`.

While the invariant holds right after the `validate()` call above, a future refactor that reorders or removes that call turns this into a production panic. Prefer fallible handling:

```rust
let first_signer = partial_signature_messages
    .messages
    .first()
    .ok_or(ValidationFailure::NoPartialSignatureMessages)?
    .signer;
if first_signer != signer {
    return Err(ValidationFailure::InconsistentSigners);
}
```

Same safety, zero panic risk, and no coupling to the `validate()` call ordering.

anchor/common/ssv_types/src/partial_sig.rs
```diff
@@ -118,20 +121,92 @@ impl TreeHash for PartialSignatureKind {
...
+    /// 2. All message signers must be the same
+    /// 3. No signer may have ID 0
+    pub fn validate(&self) -> Result<(), PartialSignatureMessagesError> {
+        let first = self
+            .messages
+            .first()
+            .ok_or(PartialSignatureMessagesError::Empty)?;
+
+        for m in self.messages.iter() {
+            if m.signer != first.signer {
```
diegomrsantos, line 161: **Error-precedence concern:** `InconsistentSigners` is checked before `ZeroSigner` here. If the first signer is valid (e.g. `OperatorId(1)`) and a later signer is `OperatorId(0)`, the returned error is `InconsistentSigners` — the `ZeroSigner` branch is never reached for that element.

Worth verifying this matches Go's `PartialSignatureMessages.Validate()` error ordering. If Go checks zero-signer first (or uses a different loop structure), the error codes will diverge on mixed-invalid inputs, which can cause spec test failures for edge cases not yet covered.

anchor/spec_tests/src/types/signed_ssv_msg.rs
```diff
@@ -0,0 +1,164 @@
...
+pub struct SignedSSVMessageTest {
+    messages: Vec<TestSignedSSVMessage>,
+    expected_error_code: i64,
+    #[serde(rename = "RSAPublicKey", default)]
+    rsa_public_key: Option<Vec<String>>,
+}
+
+impl SpecTest for SignedSSVMessageTest {
+    fn run(&self) -> Result<(), String> {
+        for msg in &self.messages {
```
diegomrsantos, line 45: **Inconsistent loop semantics with `PartialSigMsgSpecTest`.**

This loop uses fail-fast: it returns on the first message where the error code doesn't match `expected_error_code`.

In contrast, `PartialSigMsgSpecTest::check_validation()` uses last-error-wins: it iterates all messages and only compares the final error code.

```rust
// Here (fail-fast):
for msg in &self.messages {
    let actual_code = ...;
    if actual_code != self.expected_error_code {
        return Err(...); // stops at first mismatch
    }
}

// PartialSigMsgSpecTest (last-error-wins):
let mut last_error_code = NO_ERROR;
for msg in &self.messages {
    if let Err(code) = Self::validate_message(msg) {
        last_error_code = code;
    }
}
// compare last_error_code at the end
```

If Go uses the same runner pattern for both test types, one of these is wrong. Worth verifying against the upstream Go test runners.

anchor/spec_tests/src/utils/deserializers.rs
```diff
@@ -14,9 +13,54 @@ pub fn deserialize_base64<'de, D: serde::Deserializer<'de>>(
...
+        .map(|s| {
+            STANDARD
+                .decode(&s)
+                .map_err(|e| serde::de::Error::custom(format!("Failed to decode base64: {e}")))
+        })
+        .collect()
+}
+
+/// Deserializes an optional vector of base64 strings into `Option<Vec<Vec<u8>>>`.
+#[expect(dead_code)]
```
diegomrsantos: `#[expect(dead_code)]` on a newly added function — if nothing uses it yet, prefer removing it entirely and adding it when a test type actually needs it. Avoids carrying dead code that suppresses the compiler's signal.

anchor/spec_tests/src/utils/mod.rs
```diff
@@ -1,4 +1,16 @@
...
+use bls::PublicKeyBytes;
+
 pub mod deserializers;
 pub mod encoding_helpers;
+pub mod error_codes;
+
+pub use encoding_helpers::{check_roundtrip, check_roundtrip_with_root, decode_base64};
 
-pub use encoding_helpers::{check_roundtrip, check_roundtrip_with_root};
+pub static TESTING_VALIDATOR_PUBKEY: LazyLock<PublicKeyBytes> = LazyLock::new(|| {
```
diegomrsantos, line 14: Nit: a comment citing the upstream Go fixture that defines this key would help maintainers verify it stays in sync (e.g. a link to the relevant `ssv-spec` test file or constant).

anchor/spec_tests/src/types/signed_ssv_msg.rs
```diff
@@ -0,0 +1,164 @@
...
+                ));
+            }
+        }
+        Ok(())
+    }
+}
+
+impl SignedSSVMessageTest {
+    /// Validate via `SignedSSVMessage::new()` (which calls `validate()` internally).
+    fn validate_message(&self, msg: &TestSignedSSVMessage) -> Result<(), i64> {
```
diegomrsantos: **Spec-parity: validation order doesn't match Go's `SignedSSVMessage.Validate()`.**

This function checks:
1. SSVMessage nil (line 64-67) → `NIL_SSV_MESSAGE`
2. Signature normalization via `prepare_signatures` (line 69) → `EMPTY_SIGNATURE`
3. Signer validation inside `SignedSSVMessage::new()` (line 71) → `NO_SIGNERS`, `ZERO_SIGNER`, etc.

Go's `messages.go:156` checks in the opposite order — signers first, then signatures, then SSVMessage content. The Go test harness (`test.go:32`) calls `Validate()` directly, so error precedence matters.

For mixed-invalid inputs (e.g. both no signers AND nil SSVMessage), this will return `NIL_SSV_MESSAGE` where Go returns `NO_SIGNERS`. This can cause spec test failures on edge-case fixtures.

Fix: reorder to match Go — pass through `SignedSSVMessage::new()` first (signer checks), then nil-message / signature checks, or restructure so the harness mirrors Go's `Validate()` call order.

diegomrsantos: This seems to be a current issue, before this PR, validation seems duplicated in `new` and the `validate` called inside it.

.claude/agents/code-reviewer-subagent.md
diegomrsantos: Let's agree that the PR author does a first review pass to catch issues.

shane-moore: yes, accidentally deleted. resurrected in https://github.com/sigp/anchor/pull/837/changes/b8cdd862a7cd059539c391291a61710cdd5a1eb8. Totally agree on the first pass suggestion. I've been trying to move too fast to crank out these spec test PR's. Slowing down and catching things like this would respect reviewers time better, thanks!

diegomrsantos: Thanks!

anchor/spec_tests/src/types/signed_ssv_msg.rs
shane-moore: I see the inconsistency between the two tests. And yep, they follow the same flow as the Go code in:
https://github.com/ssvlabs/ssv-spec/blob/45153e4e4b8c61b929f701b7af52e3f725668421/types/spectest/tests/signedssvmsg/test.go

and
https://github.com/ssvlabs/ssv-spec/blob/45153e4e4b8c61b929f701b7af52e3f725668421/types/spectest/tests/partialsigmessage/test.go

Since they're different runners, seems fine for them to use different patterns, and the Rust impl mirrors the Go counterparts, so we're all good on this one

anchor/common/ssv_types/src/deserializers.rs
shane-moore: luv the idea of `DTO`'s! much cleaner than polluting the production structs with all this `serde` logic. implemented in https://github.com/sigp/anchor/pull/837/changes/281ff7dfaad8f960389c7815065f0160643ee5e9

anchor/message_validator/src/partial_signature.rs
shane-moore: certainly agree with this defense-in-depth suggestion, resolved in https://github.com/sigp/anchor/pull/837/changes/cb2c3bc34d878b41b96ecc5f05e3656ae60e1c45

anchor/common/ssv_types/src/partial_sig.rs
shane-moore: Great to check. The error ordering matches Go. In Go's https://github.com/ssvlabs/ssv-spec/blob/45153e4e4b8c61b929f701b7af52e3f725668421/types/partial_sig_message.go#L46-L62, the loop checks `signer != m.Signer` before calling https://github.com/ssvlabs/ssv-spec/blob/45153e4e4b8c61b929f701b7af52e3f725668421/types/partial_sig_message.go#L97-L102 (which is where the zero-signer check lives):
```
  for _, m := range msgs.Messages {
      if signer != m.Signer {           // consistency first
          return NewError(InconsistentSignersErrorCode, ...)
      }
      if err := m.Validate(); err != nil {  // then per-message validation (zero check)
          return errors.Wrap(err, ...)
      }
  }
```
And we do same in the rust version

anchor/spec_tests/src/utils/deserializers.rs
shane-moore: i was torn on this one, and I agree with you. easy to resurrect when we'll need it in subsequent PR. Killed the fn for now

anchor/spec_tests/src/utils/mod.rs
shane-moore: 💯 
https://github.com/sigp/anchor/pull/837/changes/18785204d05613ea178fd7d82cd5c9a53b6da0a5

anchor/spec_tests/src/types/signed_ssv_msg.rs
shane-moore: On the first point, looked a bit more deeply into the validation order mismatch. Seems fine actually since every fixture tests exactly one error condition in isolation. `Validate()` short circuits on the first error. The ordering divergence in our harness (nil check before signer checks) is a Rust type system constraint (SSVMessage is a value, not a pointer) and can't produce wrong results against the existing fixtures. A little extra context:
The test harness deserializes into `Option<SSVMessage>` to handle Go's null pointer case, but it must unwrap that `Option` before it can pass the value into `SignedSSVMessage::new()`. That forces the nil check to happen first, so we can't defer it like Go does with `*SSVMessage`.

shane-moore: On the second point about pre-existing issue of duplicate validation between `new` and `validate`, great point! made issue https://github.com/sigp/anchor/issues/842 to handle

shane-moore:
> ## Scope mismatch: `test:` PR modifies production code
> The PR is titled `test:` but makes substantive changes to two production crates:
> 
> * **`anchor/common/ssv_types/`**: adds `validate()` method, `PartialSignatureMessagesError` enum, `thiserror` dep, `serde` feature, `deserializers` module, and serde derives on `SSVMessage`, `PartialSignatureMessages`, `PartialSignatureMessage`, `OperatorId`, `MsgType`
> * **`anchor/message_validator/src/partial_signature.rs`**: refactors validation logic, introduces a new `ZeroSigner` error path, changes error ordering
> 
> A `test:` PR should only contain test changes. Production behavior changes belong in their own `refactor:` or `feat:` PR so reviewers can evaluate them independently.
> 
> Suggested split:
> 
> 1. `refactor: extract PartialSignatureMessages::validate() to ssv_types` — production changes
> 2. `test: add message type validation spec tests` — test-only, depends on (1)
> 
> **General note:** smaller, focused PRs (one concern per PR) are easier to review, less risky, and get merged faster. When production and test changes land together it's harder to reason about behavioral impact and rollback if needed.

makes sense! opened pr https://github.com/sigp/anchor/pull/843 to handle the production changes and will keep this as draft PR til that's merged

diegomrsantos:
> > ## Scope mismatch: `test:` PR modifies production code
> > The PR is titled `test:` but makes substantive changes to two production crates:
> > 
> > * **`anchor/common/ssv_types/`**: adds `validate()` method, `PartialSignatureMessagesError` enum, `thiserror` dep, `serde` feature, `deserializers` module, and serde derives on `SSVMessage`, `PartialSignatureMessages`, `PartialSignatureMessage`, `OperatorId`, `MsgType`
> > * **`anchor/message_validator/src/partial_signature.rs`**: refactors validation logic, introduces a new `ZeroSigner` error path, changes error ordering
> > 
> > A `test:` PR should only contain test changes. Production behavior changes belong in their own `refactor:` or `feat:` PR so reviewers can evaluate them independently.
> > Suggested split:
> > 
> > 1. `refactor: extract PartialSignatureMessages::validate() to ssv_types` — production changes
> > 2. `test: add message type validation spec tests` — test-only, depends on (1)
> > 
> > **General note:** smaller, focused PRs (one concern per PR) are easier to review, less risky, and get merged faster. When production and test changes land together it's harder to reason about behavioral impact and rollback if needed.
> 
> makes sense! opened pr #843 to handle the production changes and will keep this as draft PR til that's merged

Approved!

diegomrsantos:
I recommend hiding previous Claude reviews that aren't relevant anymore and choosing outdated, which decreases the noise.

shane-moore:
> I recommend hiding previous Claude reviews that aren't relevant anymore and choosing outdated, which decreases the noise.

ha didnt know this was possible. great suggestion, was taking forever to scroll down to see new messages lol

anchor/common/ssv_types/src/partial_sig.rs
diegomrsantos: Thanks a lot, should have checked better.

anchor/spec_tests/src/types/signed_ssv_msg.rs
diegomrsantos: Good point about fixtures testing one error at a time — that's why it works today.

Just to clarify though, `Option<SSVMessage>` is equivalent to Go's `*SSVMessage` so the type system isn't really the constraint here. What forces the order is routing validation through `SignedSSVMessage::new()`, which needs a non-optional `SSVMessage`. The harness could check `operator_ids`/`signatures` directly from the fixture struct before unwrapping the `Option` to match Go's order — worth considering so mixed-invalid fixtures don't surprise us later.

shane-moore: Agreed it would match Go's order by adding the signer/signature checks to the harness before `SignedSSVMessage::new()` is called. but would mean the error paths never reach the production code. the spec test would just be testing the harness itself.  Prefer keeping validation in `SignedSSVMessage::validate()` as to exercise the spec tests against real code

Looked more into mixed-invalid fixutres. The spec's test generator doesn't combine multiple invalid conditions in a single fixture. if they decide to do this in the future for some reason, probably makes sense to handle then, and it will likely be significant refactor across multiple tests

diegomrsantos: Makes sense, thank you

diegomrsantos (approved):
Great work!

PR #839, opened 2026-02-20
test: add encryption spec test

shane-moore:

- Anchor's spec test suite lacks coverage for the Go ssv-spec `EncryptionSpecTest` type.
- Maintains spec parity as the test matrix grows. Continuation of the spec test effort from PR #837.
- Proposer consensus data tests extracted to a separate PR per [review feedback](https://github.com/sigp/anchor/pull/839#issuecomment-4215024532). Tracked in #941.

## Change Overview

**`EncryptionSpecTest`**: Mirrors Go's `EncryptionSpecTest.Run()` step-by-step:
1. Parse RSA private key from fixture PEM
2. Derive public key and verify it matches fixture PEM
3. RSA-PKCS1v15 encrypt plaintext with public key
4. RSA-PKCS1v15 decrypt ciphertext with private key
5. Verify decrypted output matches original plaintext

This tests the same RSA-PKCS1v15 invariant that Go tests, and the same primitives that `validator_store` uses in production for keyshare decryption. The test calls `openssl` directly rather than going through the higher-level wrapper.

One minor divergence: Go compares raw PEM bytes for SK/PK consistency; we compare base64-encoded PEM (equivalent check, different encoding).

Changes:
- New `EncryptionSpecTest` type in `spec_tests/src/types/encryption.rs`
- Dispatcher wiring and module registration

- **Uses `openssl` directly instead of production wrapper**: The test validates the same RSA-PKCS1v15 primitives but doesn't exercise the `validator_store` wrapper itself. The wrapper adds BLS-specific encoding logic that doesn't apply to the spec test's arbitrary plaintext fixture.

## Validation

- `cargo test -p spec_tests`: all fixtures pass
- `cargo clippy` and `cargo fmt` clean
- Audited against [Go source](https://github.com/ssvlabs/ssv-spec/blob/main/types/spectest/tests/encryption/test.go) for invariant parity

## Blockers / Dependencies

- Depends on PR #837 for CI (branch chain). Local builds/tests pass independently.
- Proposer consensus data tests will follow in a separate PR (#941).

### DIFF
--- a/anchor/spec_tests/src/runner/types_dispatcher.rs
+++ b/anchor/spec_tests/src/runner/types_dispatcher.rs
@@ -55,6 +55,11 @@ fn dispatch_fixture_by_prefix(prefix: &str, path: &Path, contents: &str) -> Disp
             DispatchOutcome::Executed(run_test::<types::PartialSigMsgSpecTest>(path, contents))
         }
 
+        // Encryption tests
+        "encryption.EncryptionSpecTest" => {
+            DispatchOutcome::Executed(run_test::<types::EncryptionSpecTest>(path, contents))
+        }
+
         // TODO(spec-tests): Add more test types here as they are implemented.
         // This arm will be replaced with panic!() once all test types are added.
         _ => {
--- a/anchor/spec_tests/src/types/encryption.rs
+++ b/anchor/spec_tests/src/types/encryption.rs
@@ -1,15 +1,16 @@
 use base64::prelude::*;
-use operator_key::{encrypted::EncryptedKey, public, unencrypted};
+use openssl::{encrypt::Encrypter, pkey::PKey, rsa::Padding};
+use operator_key::{public, unencrypted};
 use serde::Deserialize;
 
 use crate::{SpecTest, utils::deserializers::deserialize_base64};
 
-/// Anchor-specific coverage using Go's `EncryptionSpecTest` fixtures.
+/// Mirrors Go's `EncryptionSpecTest.Run()`: parse RSA key pair from PEM,
+/// verify SK/PK consistency, then RSA-PKCS1v15 encrypt/decrypt roundtrip.
 ///
-/// Go tests RSA-PKCS1v15 encrypt/decrypt of arbitrary data. Anchor doesn't use
-/// RSA-PKCS1v15; operator keys are protected via EIP-2335 keystores instead.
-/// This test exercises that production path: parse the fixture's RSA key,
-/// verify the SK/PK pair, then roundtrip the key through EIP-2335 encrypt/decrypt.
+/// Uses the same RSA-PKCS1v15 primitives that `keysplit` and `validator_store`
+/// use in production for keyshare encryption/decryption, but calls `openssl`
+/// directly rather than going through those higher-level wrappers.
 #[derive(Debug, Deserialize)]
 #[serde(rename_all = "PascalCase")]
 pub struct EncryptionSpecTest {
@@ -23,16 +24,13 @@ pub struct EncryptionSpecTest {
 
 impl SpecTest for EncryptionSpecTest {
     fn run(&self) -> Result<(), String> {
-        if self.plain_text.is_empty() {
-            return Err("Fixture plain_text is empty — expected non-empty data".to_string());
-        }
-
-        // `from_base64` expects base64-encoded PEM, so re-encode the raw PEM bytes.
+        // Parse private key from PEM.
         let sk_b64 = BASE64_STANDARD.encode(&self.sk_pem);
         let private_key = unencrypted::from_base64(sk_b64.as_bytes())
             .map_err(|e| format!("Failed to parse private key: {e}"))?;
 
         // Verify derived public key matches fixture.
+        // Go compares raw PEM bytes; we compare base64-encoded PEM (equivalent).
         let derived_pk_b64 = public::to_base64(&private_key)
             .map_err(|e| format!("Failed to derive public key: {e}"))?;
         let expected_pk_b64 = BASE64_STANDARD.encode(&self.pk_pem);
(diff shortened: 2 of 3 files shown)


shane-moore:
@claude what are possible issues with this pr? even theoretical?

shane-moore:
@claude you alive?

diegomrsantos:
> @claude you alive?

Has the fix been aplied here?

shane-moore:
@claude, can you review in depth

shane-moore:
@claude, Following up on [this review comment](https://github.com/sigp/anchor/pull/839#issuecomment-4206990576), I went through each point in depth and here’s where I landed:

1. Issue 1: not resolving.
I traced the current `ProposerSpecTest_*` fixtures end-to-end and couldn’t reproduce the “both blinded and full decoders hit BLS errors” case on this branch (`both_bls = 0`). In the full-block validation fixture I get `full=BLST_BAD_ENCODING` and `blinded=OffsetOutOfBounds`; in the blinded validation fixture I get the inverse. So there isn’t a live ambiguity in blinded-state detection here.

2. Issues 2 and 3: resolved via documentation, not by changing the test shape.
I agree with the core point that `EncryptionSpecTest` is an adapted substitute rather than a faithful port of Go’s RSA encrypt/decrypt test, and that `plain_text` is only password material in Anchor’s path. I updated the PR description to state that explicitly, including that this does **not** prove RSA interoperability with Go SSV nodes. I’m not changing the test to mirror Go exactly because Anchor does not use raw RSA encrypt/decrypt in production.

3. Issue 4: not resolving.
`ForkName::from(self.version)` and `self.version.into()` are equivalent here because `DataVersion` derives both `From` and `Into`. This is just style, not behavior.

4. Issues 5-7: no action needed.
Those were either positive notes or documentation/naming observations that I don’t think need follow-up changes.

Separately from this review thread, the one follow-up I do think is worth tracking is CI coverage for the proposer spec test’s `fake_crypto` path. I opened [#938](https://github.com/sigp/anchor/issues/938) for adding an explicit Makefile/CI target there.

diegomrsantos:
Could you extract the encryption work into its own PR?

The two proposer additions share the new DTO/deserializer/block-decode plumbing and make sense together, but `EncryptionSpecTest` is independent and changes the review shape in a materially different way:

- it does not depend on the proposer DTOs, proposer error codes, or `can_decode_block()` helper
- it adapts the upstream RSA encrypt/decrypt spec into an `EncryptedKey` keystore roundtrip, which deserves separate review on its own terms
- pulling it out would make the proposer parity review smaller and easier to reason about

Suggested split:
- `test: proposer consensus-data spec coverage`
- `test: add adapted encryption spec coverage`

shane-moore:
Done! Since this branch was already named `test/encryption`, I kept the encryption work here and extracted the proposer consensus data tests into their own PR:

- #942 — proposer consensus data spec test coverage
- Tracked by #941

This PR is now just the `EncryptionSpecTest` addition.

shane-moore:
@claude, can you take a look again

shane-moore:
@claude, got you. updated the comment in latest commit

diegomrsantos:
Could you expand the rationale a bit more in the PR description?

I agree it is useful to say this test is intentionally different from the Go one. What still feels missing is why this is the right thing to test in Anchor instead.

That part matters for reviewers, because the main question is not just that Anchor diverges here, but why we are choosing not to test the same invariant as upstream. A short explanation of why this adaptation is the most relevant Anchor coverage would make the intent much clearer.

shane-moore:
@diegomrsantos, i feel you on that. added to description:

Why EIP-2335 instead of RSA-OAEP? Anchor stores operator keys using EIP-2335 keystores, not raw RSA-OAEP — the operator_key crate's EncryptedKey::encrypt/decrypt is the production code path for key handling. Testing RSA-OAEP would exercise code that doesn't exist in Anchor. By reusing the fixture's RSA key pair, this test validates the real production path: PEM parsing, public key derivation, EIP-2335 encrypt/decrypt roundtrip, and DER-

diegomrsantos:
Thanks, this helps. Just to make sure I’m following correctly: is the upstream Go spec test actually covering key storage / key handling in the same sense you describe here?

My reading was that it exercises RSA encrypt/decrypt of `PlainText`, so I’m trying to understand whether this `EncryptedKey` roundtrip is intended as a substitute for a different Anchor-specific invariant rather than coverage of the same upstream one.

shane-moore:
You're right, Go's test is a pure RSA-PKCS1v15 encrypt/decrypt roundtrip on arbitrary data (including BLS secret key bytes). It doesn't exercise key storage.

This is a substitute covering a different Anchor-specific invariant, not the same upstream one. Anchor doesn't have RSA-PKCS1v15 encrypt/decrypt; operator keys are protected via EIP-2335 keystores. So the test exercises that production path instead, reusing the fixture's RSA key pair as input.

Updated the PR description and code comments to make this explicit (92d487f):

> **This is Anchor-specific coverage, not a port of the Go test.** Go's `EncryptionSpecTest` tests RSA-PKCS1v15 encrypt/decrypt of arbitrary data (including BLS secret key bytes). Anchor doesn't use RSA-PKCS1v15; operator keys are protected via EIP-2335 keystores. This test exercises that real production path using the fixture's RSA key pair as input: PEM parsing, public key derivation, EIP-2335 encrypt/decrypt roundtrip, and DER-level key integrity. The invariant tested is different from upstream.

diegomrsantos:
Thanks, this is much clearer now.

I still have one remaining question though: if this is explicitly Anchor-specific coverage for a different invariant, should it really live in `spec_tests` as `EncryptionSpecTest` coverage?

Would it make more sense to do both instead: keep this as a normal Anchor test for the `operator_key` path, and cover the closer RSA encrypt/decrypt behavior in `spec_tests` so that layer stays aligned with the upstream invariant?

shane-moore:
The test confirms Anchor can deserialize ssv-spec's encryption fixtures and that the fixture's RSA key pair is valid (parses correctly, derived PK matches fixture PK), then exercises Anchor's production `operator_key` path with that key material.

The reason to keep it in `spec_tests` rather than as a unit test: if ssv-spec changes their encryption fixtures (different key format, new fields, etc.), the dispatcher catches the deserialization failure immediately. Moving it to a unit test with hardcoded data loses that signal.

Writing a closer RSA encrypt/decrypt spec test would mean adding RSA-PKCS1v15 code that Anchor doesn't use in production, purely to match Go's test shape. That seems like the wrong trade-off.

diegomrsantos:
I am still not sure the last part is right.

A closer RSA encrypt/decrypt test would not necessarily require adding PKCS#1 v1.5 code that Anchor does not use in production. Anchor already appears to have a production RSA keyshare flow, so this still looks less like the only production path and more like one production path that is easier to test here.

That is why I am still hesitant about keeping this as `EncryptionSpecTest` coverage in `spec_tests`. If the upstream invariant is closer to Anchor’s existing RSA keyshare flow than to `operator_key`, I am not yet convinced that `EncryptedKey` is the right mapping here.

shane-moore:
You were absolutely right to push on this. I should have dug deeper into the codebase before defending the EIP-2335 approach.

After looking more carefully, `keysplit/src/crypto.rs` and `validator_store/src/lib.rs` both use RSA-PKCS1v15 encrypt/decrypt in production for keyshare handling, which is exactly the invariant Go's test covers. I rewrote the test to mirror Go's `EncryptionSpecTest.Run()` step-by-step:

1. Parse RSA private key from fixture PEM
2. Derive public key, verify it matches fixture PK
3. RSA-PKCS1v15 encrypt plaintext with public key
4. RSA-PKCS1v15 decrypt ciphertext with private key
5. Verify decrypted output matches original plaintext

The test calls `openssl` directly (same primitives `keysplit`/`validator_store` use) rather than going through those wrappers, since the wrappers add BLS-specific encoding that doesn't apply to the spec fixture's arbitrary plaintext.

Updated the PR description and title to reflect this. Latest commit: f26b24e.

Thanks for pushing on this. The test is in a much better place now.

shane-moore:
@claude, please re-review, since the last commit fundamentally changed the test

shane-moore:
@claude, makes sense. updated the description in encryption.rs to handle issue 1

diegomrsantos (approved):
Thanks for iterating on this.

The current version addresses my earlier concern. It now covers the same RSA-PKCS1v15 encrypt/decrypt invariant as the upstream Go `EncryptionSpecTest`, so keeping it in `spec_tests` makes sense. Using `openssl` directly here also looks like the right trade-off for parity, rather than routing the fixture through higher-level Anchor-specific wrappers.

LGTM.

PR #835, opened 2026-02-17
test: add encoding spec  tests for SSV types

shane-moore:
## Issue Addressed

Continuation of spec test from #833 in support of https://github.com/sigp/anchor/issues/773#issuecomment-4000424894 and https://github.com/sigp/anchor/issues/867

## Proposed Changes

Add encoding spec tests for the remaining SSV types. Each test validates SSZ decode, encode roundtrip, and hash tree root against fixtures from the ssv-spec `boole` branch.

Encoding tests added:
  - `PartialSignatureMessages`
  - `SignedSSVMessage`
  - `SSVMessage`
  - `AggregatorCommitteeConsensusData`
  - `ProposerConsensusData`

## Additional Info
The remaining fixtures cover message validation, RSA encryption, duty mapping, and other test categories planned for future PRs.

PR #840, opened 2026-02-20
Test/simple spec test types

shane-moore:
## Issue Addressed
Continuation of spec test PR's

## Proposed Changes

Added `SszSpecTest` and `CommitteeMemberTest`

## Additional Info
Blocked by https://github.com/sigp/anchor/pull/839 since this PR is built on top of it's commits


diegomrsantos:
Can we avoid having many parallel draft PRs and instead focus only on one at a time? Also, please let's go back to small, focused, and easy-to-review PRs

shane-moore:
> Can we avoid having many parallel draft PRs and instead focus only on one at a time? Also, please let's go back to small, focused, and easy-to-review PRs

agree with having small PR's. each one of these spec test PR's builds commits on top of the previous PR, because the changes would clash if each PR's commits were each built off `unstable`. so this looks like 1800 line diff, but it's really just a few hundred lines. will be smaller once the upstream PR's are merged to `unstable`

I was thinking the strategy could be to leave the [PR](https://github.com/sigp/anchor/pull/835) that's ready for review as open. and put the rest as draft. then, once [PR](https://github.com/sigp/anchor/pull/835) is merged, I would take the next PR in the line out of draft.  so you'd only have to review a small PR at a time instead of making one big one with all the changes. kinda figured it's fine for folks to ignore the ones that are in draft. alternatively could keep the draft branches local and just submit them to the repo once the PR before it is merged.  I'm open to other suggestions to handle this

diegomrsantos:
> > Can we avoid having many parallel draft PRs and instead focus only on one at a time? Also, please let's go back to small, focused, and easy-to-review PRs
> 
> agree with having small PR's. each one of these spec test PR's builds commits on top of the previous PR, because the changes would clash if each PR's commits were each built off `unstable`. so this looks like 1800 line diff, but it's really just a few hundred lines. will be smaller once the upstream PR's are merged to `unstable`
> 
> I was thinking the strategy could be to leave the [PR](https://github.com/sigp/anchor/pull/835) that's ready for review as open. and put the rest as draft. then, once [PR](https://github.com/sigp/anchor/pull/835) is merged, I would take the next PR in the line out of draft. so you'd only have to review a small PR at a time instead of making one big one with all the changes. kinda figured it's fine for folks to ignore the ones that are in draft. alternatively could keep the draft branches local and just submit them to the repo once the PR before it is merged. I'm open to other suggestions to handle this

The right way of doing this is using stacked PRs, but I believe the current way we work doesn't allow it. @dknopik

diegomrsantos:
> > > Can we avoid having many parallel draft PRs and instead focus only on one at a time? Also, please let's go back to small, focused, and easy-to-review PRs
> > 
> > 
> > agree with having small PR's. each one of these spec test PR's builds commits on top of the previous PR, because the changes would clash if each PR's commits were each built off `unstable`. so this looks like 1800 line diff, but it's really just a few hundred lines. will be smaller once the upstream PR's are merged to `unstable`
> > I was thinking the strategy could be to leave the [PR](https://github.com/sigp/anchor/pull/835) that's ready for review as open. and put the rest as draft. then, once [PR](https://github.com/sigp/anchor/pull/835) is merged, I would take the next PR in the line out of draft. so you'd only have to review a small PR at a time instead of making one big one with all the changes. kinda figured it's fine for folks to ignore the ones that are in draft. alternatively could keep the draft branches local and just submit them to the repo once the PR before it is merged. I'm open to other suggestions to handle this
> 
> The right way of doing this is using stacked PRs, but I believe the current way we work doesn't allow it. @dknopik

In the meantime, it'd be helpful to add a more elaborate description explaining the context in which the PR was created, as you did in the previous comment.

shane-moore:
agreed that stacked PR's are the right approach for this type of thing. don't think it's easy or common to do for repo's where contributors work off of a fork.  i think a middle ground is that for the rest of the spec test PR's, i'll make them as stacked PR's in my forked repo. and then, upstream each one to `sigp/anchor` once the previous PR has been merged to `unstable`

shane-moore:
also updated the PR description to reference the PR that it's blocked by as you mentioned, great idea

shane-moore:
Closing this PR in favor of splitting into 4 independent PRs, one per test type. The tests have no dependencies on each other so they're better reviewed separately.

Tracking issue: #954

Individual issues:
- #955 - SSZSpecTest
- #956 - DutySpecTest
- #957 - BeaconDepositDataSpecTest
- #958 - CommitteeMemberTest

The rebased implementation (all 4 tests on top of current `unstable`) is preserved on branch `test/simple-spec-test-types` for reference.

PR #841, opened 2026-02-25
fix: offload index sync store to blocking task

dknopik:

- During historical sync, validator index sync runs on async code but still performs a blocking SQLite write inline at the end of each batch.
- That can stall the Tokio runtime long enough to surface beacon-node timeout / offline health-check noise during longer sync runs.
- This is worth doing now because the larger superseding attempt in `#885` was closed in favor of a smaller replacement stack from `unstable`, but the local runtime-blocking issue here is still valid and can be fixed with a narrower change.
- Relevant links: `#841`, `#735`, `#885`.

## Change Overview (Required)

- Move the validator index store step onto the blocking executor, but await it immediately.
- Keep the existing batch fetch flow, DB fallback lookup, and overall ordering / backpressure behavior.
- Remove the earlier fire-and-forget queueing / in-flight-write heuristic from this PR so the workaround stays incremental and easier to reason about.
- Add a focused regression test that verifies the blocking store path updates both durable DB state and in-memory state.
- Review guide: start with the sync flow change, then the blocking store helper, then the new test.
- Intentionally did not change: event-processing ownership, store architecture, or the broader persistence-boundary work discussed in `#735` / `#885`.

- Main trade-off: this gives up overlap between fetching the next beacon-node batch and writing the current batch to SQLite.
- That is intentional here. Preserving backpressure and avoiding a second queue is simpler and lower risk than the earlier fire-and-forget approach.
- Blast radius is limited to validator index sync.
- Risk is mitigated by keeping the existing DB fallback behavior and adding targeted test coverage.

## Validation (Required)

- `cargo fmt --all --check`
- `cargo test -p eth --lib index_sync::tests::test_store_validator_indices_blocking_updates_database_and_state -- --exact`
- `cargo clippy -p eth --all-targets -- -D warnings`

- Safe code rollback before merge: revert this PR.
- No schema, config, or external API changes.
- Operationally, rollback just restores the previous inline write behavior in validator index sync.

- N/A

- This is a tactical fix, not the long-term architecture.
- If broader store-boundary simplification continues on top of `unstable`, this PR should be evaluated as a narrow runtime-unblocking change rather than the end-state design.


anchor/eth/src/index_sync.rs
```diff
@@ -77,7 +86,9 @@ async fn validator_index_syncer(
         // next, fill up the rest of the batch with older validators that are unknown from the
         // database
         let space = MAX_BATCH_SIZE - batch.len();
-        if space > 0 {
+        // Only do this if we have any space remaining and there are no store tasks that might wait
+        // to write missing indices. If the count is 1, only we hold the Arc (no other tasks).
```
diegomrsantos: **Nit (non-blocking):** `Arc::strong_count` uses `Relaxed` ordering and is inherently subject to TOCTOU — the count can change between the read and the branch. In this case the consequences are benign (an extra or skipped DB sweep), so it's not a correctness issue.

Still, an `AtomicUsize` would express the intent more clearly and avoid future confusion:

```rust
let pending_stores = Arc::new(AtomicUsize::new(0));
// ...
if space > 0 && pending_stores.load(Ordering::Acquire) == 0 {
```

Up to you whether it's worth changing — the current code works fine in practice.

dknopik: I think the ordering makes only little difference here if I understand correctly.

I understand the argument to use a `AtomicUsize` for clarity though. What do you prefer?

diegomrsantos: I believe `AtomicUsize` would be better

anchor/eth/src/index_sync.rs
```diff
@@ -127,9 +138,25 @@ async fn validator_index_syncer(
                 .map(|v| (v.validator.pubkey, ValidatorIndex(v.index as usize)))
                 .collect::<HashMap<_, _>>();
             trace!(len = map.len(), "Got validators from BN");
-            if let Err(err) = db.set_validator_indices(map) {
-                error!(?err, "Failed to update validator indices");
-            }
+
+            // `set_validator_indices` may block as it start a database transaction and updates the
+            // in memory database. We do not want to do that on the async runtime, so we
+            // spawn a blocking task.
```
diegomrsantos: **Nit:** The comment explains *what* but not *why* — a reader unfamiliar with the PR context wouldn't know why blocking matters here. Consider something like:

```rust
// `set_validator_indices` performs synchronous SQLite I/O and takes a write lock
// on in-memory state, which would block the Tokio worker thread and starve other
// async tasks (e.g., beacon node health checks). Offload to the blocking thread pool.
```

Also minor typo: "it **start**" → "it **starts**".

diegomrsantos (commented):
Thanks for handling this, great catch. Mostly good, only two small nits.

dknopik:
@diegomrsantos 

Thanks for the review, I addressed it. I do not think the Ordering matters at all:

> Memory orderings specify the way atomic operations synchronize memory. In its weakest [Ordering::Relaxed](https://doc.rust-lang.org/std/sync/atomic/enum.Ordering.html#variant.Relaxed), only the memory directly touched by the operation is synchronized. On the other hand, a store-load pair of [Ordering::SeqCst](https://doc.rust-lang.org/std/sync/atomic/enum.Ordering.html#variant.SeqCst) operations synchronize other memory while additionally preserving a total order of such operations across all threads.

from https://doc.rust-lang.org/std/sync/atomic/enum.Ordering.html

We do not really touch any other shared memory beyond the `AtomicUsize`, so I just opted for `Relaxed`, the simplest option.

anchor/eth/src/index_sync.rs
```diff
@@ -77,7 +93,9 @@ async fn validator_index_syncer(
         // next, fill up the rest of the batch with older validators that are unknown from the
         // database
         let space = MAX_BATCH_SIZE - batch.len();
-        if space > 0 {
+        // Only do this if we have any space remaining and there are no store tasks that might wait
+        // to write missing indices. If the count is 1, only we hold the Arc (no other tasks).
+        if space > 0 && waiting_store_tasks.load(Ordering::Relaxed) == 1 {
```
diegomrsantos: `waiting_store_tasks` starts at 0, increments before spawning each blocking write, and decrements when it finishes.
So should this condition check `== 0` (not `== 1`) if we only want DB fallback when no store writes are currently in flight?

diegomrsantos:
Thanks for the follow-up — this is a good direction, and offloading `set_validator_indices` makes sense.

  If you’re open to a small readability pass, I think these changes would help:
  - rename intent-driven identifiers, e.g. `pending_store_writes`, `has_inflight_store_writes`, `missing_index_scan_cursor`
  - add a small helper/predicate (`fn has_inflight_store_writes(...) -> bool`) and gate with that name
  - keep comments on *why* fallback is gated (avoid redundant DB candidates while writes are active)
  - extract the DB fallback scan into a tiny helper with a brief doc comment
  - add a small regression test that verifies DB fallback scan is skipped while a store task is in flight

  The core idea is solid; this will make the intent much clearer and safer for future edits.

anchor/eth/src/index_sync.rs
dknopik: Ahh, damn, good catch, forgot to change this when editing earlier

dknopik:
Thanks! Addressed everything except the `has_inflight_store_writes` predicate — renamed the identifiers, added the gating comments, extracted the DB fallback scan helper, and added the regression test.

I skipped the predicate because the check is only used in one spot and reads clearly inline. Let me know if you'd still prefer it extracted.

diegomrsantos:
I think this PR is a good tactical fix for the immediate issue, but I also think it is important to frame it as a workaround, not the real fix.

What this PR fixes is clear: the index syncer is async, but it was still doing blocking SQLite work on the runtime thread. Moving that store step into `spawn_blocking` is the right local mitigation for that.

However, I think this is the same underlying design problem that came up in #735, just showing up in a different way.

In #735, the problem shows up as a correctness issue:
- shared in-memory state is updated before we have a clean commit/publication boundary
- side effects can escape before commit
- failure handling becomes hard to reason about

Here, the same problem shows up as an execution-model issue:
- async code is still directly responsible for database-backed writes
- so we need ad hoc `spawn_blocking` fixes at individual call sites

In both cases, the root issue is that there is no single store boundary that owns:
- blocking database execution
- durable commit
- publication of committed in-memory state

So I view this PR as a reasonable stopgap, but not the architectural fix.

The real fix, as discussed in #735, is to move toward a model where:
1. async tasks do not write to SQLite directly
2. they send commands / batch plans to a centralized store owner
3. that store owner runs off the async runtime thread
4. it commits durable state first
5. it publishes the new in-memory snapshot exactly once, after commit

That would solve this class of problem at the right layer instead of requiring local mitigations like this one.

So from my side: this looks like a sensible short-term fix, but I would not want us to treat it as the end-state design.

dknopik:
Sure. :)

Everything we work on has potential to be improved. Nothing I ever propose is implicitly supposed to be final. We improve the code incrementally. 

> So from my side: this looks like a sensible short-term fix, but I would not want us to treat it as the end-state design.

I agree.

Is there anything else you want to have addressed for this workaround PR?

diegomrsantos:
Thanks for working on this. After looking at the current shape of the workaround, I think we should close this PR and solve the problem properly instead.

The easiest way to see the issue is to compare the control flow before and after this PR.

Before this PR, `index_sync` effectively did this:

`fetch indices from BN -> write them to SQLite inline -> only then continue`

That was bad because the SQLite write could block the async runtime thread.

But it also had one useful property: natural backpressure. The syncer could not get ahead of the database.

After this PR, the flow becomes:

`fetch indices from BN -> spawn_blocking(write to SQLite) -> immediately continue fetching`

That fixes the local runtime-blocking symptom, but it removes the old backpressure.

The problem is that the database is still single-writer in practice. `NetworkDatabase` still uses a single SQLite connection / single writer model, so `spawn_blocking` does not make the writes meaningfully concurrent. It only makes them queue up off-thread behind one SQLite connection.

So during historical sync, we can now do something like this:

`fetch batch 1 -> enqueue writer 1`  
`fetch batch 2 -> enqueue writer 2`  
`fetch batch 3 -> enqueue writer 3`

But only one writer can actually run against SQLite. The rest just sit and wait. If we produce batches faster than SQLite can drain them, the queue grows. At that point we have traded:

- old problem: blocking the runtime thread
- new problem: building a backlog of writer tasks behind a single DB connection

That is not just a performance detail. It changes the failure mode. Late tasks can now wait a long time for the connection and potentially fail under load.

There is a second issue as well. The new `pending_store_writes` gating suppresses the DB fallback scan while writes are in flight. That avoids duplicate work, but it also means validators that are already in the DB with `index == None` now depend on that heuristic before they are retried. If queue-driven work keeps arriving, those older missing-index validators may be delayed for a long time. That makes liveness/progress harder to reason about than before.

So my concern is not that the PR identified the wrong symptom. The symptom is real: blocking SQLite work should not run on the async runtime.

My concern is that this workaround is now getting complex enough that it is changing write ordering, backpressure, and retry/progress behavior locally inside `index_sync`, while the real problem is architectural.

For that reason, I think we should close this PR rather than keep refining the workaround.

I'm already working on the proper fix: a dedicated store owner / centralized write boundary that preserves serialization and backpressure explicitly, instead of trying to recover them with local heuristics inside `index_sync`.

That direction addresses the runtime-blocking issue here and also moves us toward the broader persistence/cache-coherence model we discussed in #735.

Given that, I do not think it is worth landing this workaround in its current form.

dknopik:
We can also immediately await the offloaded task if that is what you prefer. That seems to be what you are doing in #885. We should regain backpressure that way. Let me know what you think.

The clear advantage of merging this is that we have this fixed for now. I feel larger refactors might be contentious and I want to avoid to be blocked on that as the Boole release approaches. This also aligns with our desire to have smaller PRs.

The effort to merge this "workaround" is miniscule, now that it exists. We are on track to spend a lot of time on discussing whether to merge **something that we both consider to be at least an improvement over the current situation**.

@shane-moore what is your opinion?

shane-moore:
> We can also immediately await the offloaded task if that is what you prefer. That seems to be what you are doing in #885. We should regain backpressure that way. Let me know what you think.
> 
> The clear advantage of merging this is that we have this fixed for now. I feel larger refactors might be contentious and I want to avoid to be blocked on that as the Boole release approaches. This also aligns with our desire to have smaller PRs.
> 
> The effort to merge this "workaround" is miniscule, now that it exists. We are on track to spend a lot of time on discussing whether to merge **something that we both consider to be at least an improvement over the current situation**.
> 
> @shane-moore what is your opinion?

when reading the beginning of the pr description about tokio runtime being blocked, my first thought was `spawn_blocking`. then, i read a few sentences later that is what you did ha

and I read through the threads, interesting points about losing backpressure, and yah, just doing an `await` for the spawn blocking work to complete seems a solid way to regain it.  think you could kill the `AtomicUsize` in that case as well.  tradeoff is that we lose the concurrency of fetching BN batches while waiting on db writes. but i think thats fine. I'm happy as long as tokio runtime gets unblocked tbh

then, I don't see why not merge this in the spirit of continuous improvement, which y'all have already discussed. and then iterating/rearchitecting in another pr which is already happening.  A phrase I like is, "progress over perfection"

diegomrsantos:
I agree with Shane’s point that, for the immediate problem here, dropping the fetch/write overlap and just awaiting the blocking DB work is probably fine. I think that concern about "losing the concurrency of fetching BN batches while waiting on DB writes" is also a good example of the broader issue, though: that overlap may improve throughput, but so far I have not seen evidence that the benefit is large enough to justify the complexity it introduces.

The work on `#885` gives us at least one concrete data point on that tradeoff. I reran historical sync on fresh `--data-dir`s, one run at a time, against the same Hoodi endpoints:

- `unstable`: `46.94s`
- event-by-event branch: `53.43s`

So the simpler design is somewhat slower on this setup, by about `6.5s` / `14%`. That is not irrelevant, but it is also not the kind of gap that, on its own, justifies the amount of complexity it introduced. We are not looking at an order-of-magnitude difference that would explain the current design.

More importantly, that complexity has not only contributed to the cache/publication issues discussed in `#735`, it also appears to have contributed to real persistence bugs. During this work I found that fresh `unstable` retains `8` orphan cluster rows in SQLite with no validators and no members. After digging into it, the failure mode is fairly concrete: `unstable` processes a whole fetched batch inside one outer transaction, `ValidatorAdded` inserts the cluster row before the validator row, and duplicate-validator failures (`UNIQUE constraint failed: validators.validator_pubkey`) are treated as malformed/skipped while the outer batch transaction is still committed later. So the partial cluster insert survives as a dead row. The event-by-event branch does not have that problem because it rejects duplicate `ValidatorAdded` earlier and no longer keeps a long-lived outer transaction alive across skipped events.

That is the broader concern I’ve been raising for a long time: we keep carrying complexity in the name of throughput/concurrency/caching without first establishing that it is actually needed. The result is usually weaker boundaries, less obvious ownership, and failure semantics that are much harder to reason about.

Given that, I don’t think `#841` should be merged separately. `#885` already addresses the main issue here in a broader and cleaner way: event processing and validator-index DB writes no longer run directly on async tasks, historical sync has a single bounded processor instead of overlapping DB work arbitrarily, and the database layer now owns the durable commit/publication boundary instead of spreading it across async callers. `#885` is also already in a reviewable state from my perspective: it has been tested, benchmarked, and the major regressions found during that work were fixed. If there is still a strong preference to merge `#841` as a narrow interim fix, that is fine, but I would see it as a stopgap rather than the direction we should keep building on.

Assuming `#885` is the direction we want, the more useful next step would be to apply that same scrutiny to the full in-memory `NetworkState` snapshot: test whether it is providing meaningful benefit for the paths that actually matter, and if it is not, remove it and simplify the system further.

shane-moore:
Really cool and thorough how you ran the #885 against hoodi and were able to make legit comparisons for historical sync.  Great finds with the current approach resulting in orphaned rows in the db.  It feels pretty risk free to merge this after adding the `spawn_blocking` + `.await` as a stop gap as you said. and in parallel, i can start gaining familiarity with what's happening in #885 so i can understand why it could perhaps be the more long-term fix. give me some time on that as it looks like a rather big PR

diegomrsantos:
Updated this PR to the narrower incremental version we discussed here.

What changed:
- offload `set_validator_indices` via `TaskExecutor::spawn_blocking_handle(...)`
- await that blocking task immediately, so we preserve the original backpressure / ordering
- remove the fire-and-forget queueing, `pending_store_writes` / `AtomicUsize`, and the extra DB-fallback gating
- keep the existing batch fill and DB fallback behavior otherwise unchanged
- add a focused regression test that verifies the helper updates both SQLite and in-memory state

Why:
- this still fixes the immediate issue from the PR description: the blocking SQLite write no longer runs on the Tokio runtime thread
- but it avoids changing retry / progress behavior locally inside `index_sync`
- that keeps this PR aligned with the incremental, less invasive direction we moved to after the larger superseding PR `#885` was closed in favor of the smaller replacement stack from `unstable`

shane-moore:
The PR description says “this conveniently also allows us to continue fetching indices while the store operation waits” and mentions an `Arc` tracking mechanism, but the current code `.await`s the blocking store synchronously. The loop is still fully sequential: batch → fetch → store → next batch.

shane-moore:
Might be worth a follow-up issue to add basic observability: batches processed, indices synced, store duration, and error counts. Maybe would help operators catch slow initial syncs or persistent db failures from dashboards instead of log scraping. wdyt? happy to open one if that sounds useful

diegomrsantos:
> The PR description says “this conveniently also allows us to continue fetching indices while the store operation waits” and mentions an `Arc` tracking mechanism, but the current code `.await`s the blocking store synchronously. The loop is still fully sequential: batch → fetch → store → next batch.

Good catch, updated it.

anchor/eth/src/index_sync.rs
```diff
@@ -125,30 +139,43 @@ async fn validator_index_syncer(
                 .flat_map(|v| v.data)
                 .map(|v| (v.validator.pubkey, ValidatorIndex(v.index as usize)))
                 .collect::<HashMap<_, _>>();
```
shane-moore, line 141: when all BNs fail, `unwrap_or_else` produces `None` → empty `HashMap` → no-op `store_validator_indices_blocking`. Harmless in practice, but a `continue` on empty map makes the intent clearer:
> ```rust
> if map.is_empty() {
>     continue;
> }
> ```

diegomrsantos:
@shane-moore I think we should be careful not to add implementation-shaped metrics just because we can. I'm not yet convinced that `index_sync` internals like batches processed, indices synced, store duration, or even "validators still missing indices" are actually useful operator-facing signals, especially since many of those validators may not even be theirs.

For operators, the more important questions are whether the client is healthy, sync is making forward progress, and validator readiness is being affected. If we can't tie a metric to a concrete operator action or alert, I'd rather keep the detail in logs than expand the default metrics surface.

So I probably would not open a follow-up issue for those specific metrics yet. If we revisit observability here later, I'd want it framed around a clear operator-visible symptom rather than internal `index_sync` mechanics.

anchor/eth/src/index_sync.rs
diegomrsantos: Thanks for the good catch, addressed in `517cab721`: `store_validator_indices_blocking` now returns early on empty maps, so we skip both the blocking task and the empty DB transaction.

shane-moore:
> @shane-moore I think we should be careful not to add implementation-shaped metrics just because we can. I'm not yet convinced that `index_sync` internals like batches processed, indices synced, store duration, or even "validators still missing indices" are actually useful operator-facing signals, especially since many of those validators may not even be theirs.
> 
> For operators, the more important questions are whether the client is healthy, sync is making forward progress, and validator readiness is being affected. If we can't tie a metric to a concrete operator action or alert, I'd rather keep the detail in logs than expand the default metrics surface.
> 
> So I probably would not open a follow-up issue for those specific metrics yet. If we revisit observability here later, I'd want it framed around a clear operator-visible symptom rather than internal `index_sync` mechanics.

ah ok, true true, i agree with you from the perspective of an operator

lh uses OpenTelemetry support (--telemetry-collector-url with a prefix-based sampler for lh_-prefixed spans). Exports to Jaeger/Tempo for visual trace timelines so is dev-facing, separate from the Prometheus pipeline operators watch. Could be useful for anchor too. instrumenting hot paths and having a trace explorer to find optimization opportunities. Way out of scope for this PR though. Maybe worth a separate discussion at some point.

shane-moore (approved):

PR #844, opened 2026-02-26
fix: claude mentions

shane-moore:
## Issue Addressed

`@claude` mentions on PRs stopped posting comments after PR #830

## Proposed Changes

  - Remove `prompt` input from `claude-mentions` workflow to restore tag mode (the prompt input switched the action to agent mode, which doesn't manage comment lifecycle)
  - Inject review criteria via `--append-system-prompt` in `claude_args` instead, preserving tag mode while keeping review guidelines
  - Extract shared review criteria to .github/prompts/review.md as single source of truth

## Additional Info
N/A

### DIFF
--- a/.github/prompts/review.md
+++ b/.github/prompts/review.md
@@ -1,0 +1,24 @@
+# Code Review Guidelines
+
+Only comment on issues you are CONFIDENT are real problems:
+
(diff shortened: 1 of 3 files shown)


dknopik (approved):
LGTM, lets see what happens

PR #846, opened 2026-02-26
refactor: group Node CLI fields into logical option sub-structs

diegomrsantos:
## Issue Addressed

N/A — structural improvement to CLI organization, prerequisite for CLI doc auto-generation (#526).

## Proposed Changes

The `Node` struct has grown to 43 flat fields covering unrelated concerns (networking, metrics, HTTP API, security, MEV). This makes the struct hard to scan, and forces every function that touches CLI args to receive the entire `Node` even when it only needs a few fields.

**Mental model:** each sub-struct maps to one independent concern a node operator configures. Six `#[clap(flatten)]` sub-structs replace the flat layout:

| Sub-struct | Concern | Fields |
|---|---|---|
| `SecurityOptions` | Operator key & password | 2 |
| `ExternalApis` | Beacon/execution endpoints, TLS certs, broadcast, sync tolerances | 7 |
| `HttpApiOptions` | REST API server | 5 |
| `MetricsOptions` | Prometheus metrics server | 5 |
| `NetworkOptions` | P2P listen/port/ENR/discovery/gossipsub | 24 |
| `PayloadBuildingOptions` | Gas limit, builder boost, MEV preferences | 4 |

Fields that don't fit a clear group (slashing protection, impostor mode, performance tuning, doppelganger protection, majority fork protection) stay directly on `Node`.

**Reading order:** start with `cli.rs` (struct definitions move out of `Node` into sub-structs), then `config.rs` (field accesses get prefixed, e.g. `cli_args.http` → `cli_args.http_api_options.http`).

**What did NOT change:** CLI argument names, defaults, `requires`/`conflicts_with` relationships, and all runtime behavior are preserved. `main.rs` is untouched — it only accesses `node.logging_flags`, which was already a flatten.

**Bonus cleanup:** `parse_listening_addresses` was narrowed from `&Node` to `&NetworkOptions` since it only reads network fields — makes the dependency explicit and the function easier to test in isolation.

## Additional Info

- Purely mechanical refactor — the compiler catches any missed field access.
- Verified: `cargo build --release`, `make cargo-fmt`, `make lint`, `make test` all pass.
- This lays groundwork for a future CLI docgen PR that can iterate sub-structs to produce grouped documentation sections.

### DIFF
--- a/anchor/client/src/cli.rs
+++ b/anchor/client/src/cli.rs
@@ -465,14 +465,6 @@ pub struct NetworkOptions {
 
     #[clap(long, help = "Disables gossipsub topic scoring.", hide = true)]
     pub disable_gossipsub_topic_scoring: bool,
-
-    #[clap(
-        long,
-        help = "Disable the latency measurement service.",
-        display_order = 0,
-        help_heading = FLAG_HEADER
-    )]
-    pub disable_latency_measurement_service: bool,
 }
 
 #[derive(Parser, Clone, Debug)]
@@ -540,6 +532,14 @@ pub struct Node {
 
     #[clap(flatten)]
     pub payload_building_options: PayloadBuildingOptions,
+
+    #[clap(
+        long,
+        help = "Disable the latency measurement service.",
+        display_order = 0,
+        help_heading = FLAG_HEADER
+    )]
+    pub disable_latency_measurement_service: bool,
 
     #[clap(
         long,
--- a/anchor/client/src/cli.rs
+++ b/anchor/client/src/cli.rs
@@ -15,8 +15,7 @@ use logging::FileLoggingFlags;
 pub const FLAG_HEADER: &str = "Flags";
 
 #[derive(Parser, Clone, Debug)]
-#[clap(name = "node", about = "Start Anchor node")]
-pub struct Node {
+pub struct SecurityOptions {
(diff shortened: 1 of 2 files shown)


anchor/client/src/cli.rs
```diff
@@ -450,38 +457,26 @@ pub struct Node {
...
-        value_delimiter = ',',
-        help = "Override size for a specific queue. Needs to be of the format \"queue_name=42\".",
-        hide = true,
-        display_order = 0
+        help = "Disable the latency measurement service.",
+        display_order = 0,
+        help_heading = FLAG_HEADER
     )]
-    pub work_queue_size: Vec<String>,
+    pub disable_latency_measurement_service: bool,
```
shane-moore: looks like `disable_latency_measurement_service` disables beacon node http latency metrics.  to me, it doesn't feel like it fits directly in `NetworkOptions` since that's all p2p. maybe keep it in top level `Node` since it doesnt really seem to fit in a specific category?

shane-moore:
Is more readable now, nice!

anchor/client/src/cli.rs
diegomrsantos: Good catch, done.

shane-moore (approved):

PR #851, opened 2026-02-27
chore: tweak minTargetPercent for testnet stability

dknopik:
## Issue Addressed

Intermittent CI testnet failures.

## Proposed Changes

The previous value of 98% was effectively 100%, as a single failing attestation of 40 was already 2.5%. 

Make this more lenient by reducing to 95%, allowing up to two failing attestations.


diegomrsantos (approved):
Thanks!

PR #853, opened 2026-02-27
fix: validate SignedSSVMessage on SSZ-decoded network messages

### LINKED ISSUE #842: refactor: remove unreachable size-bound checks from SignedSSVMessage::validate()
Follow-up to convo in https://github.com/sigp/anchor/pull/837#discussion_r2854348085

`SignedSSVMessage::validate` contains 4 size bound checks that are dead code:

  - https://github.com/sigp/anchor/blob/42cf6c6ebaaa58c52356169ba50a0cc2783bb734/anchor/common/ssv_types/src/message.rs#L612-L617 (signatures.len() > 13)
  - https://github.com/sigp/anchor/blob/42cf6c6ebaaa58c52356169ba50a0cc2783bb734/anchor/common/ssv_types/src/message.rs#L619-L627 (sig.len() != 256)
  - https://github.com/sigp/anchor/blob/42cf6c6ebaaa58c52356169ba50a0cc2783bb734/anchor/common/ssv_types/src/message.rs#L629-L634 (operator_ids.len() > 13)
  - https://github.com/sigp/anchor/blob/42cf6c6ebaaa58c52356169ba50a0cc2783bb734/anchor/common/ssv_types/src/message.rs#L636-L641 (full_data.len() > MAX_FULL_DATA_SIZE)

Why they're unreachable: The struct fields are `VariableList` types (e.g. `VariableList<_, U13>`) which enforce these bounds at the type level. There are only two construction paths -> `new()` (which converts via `try_to_variable_list`, rejecting oversized inputs before `validate()` runs) and SSZ decode (which enforces `VariableList` max length during deserialization). In both paths, by the time `validate()` executes, the type system has already guaranteed the bounds hold.

Let's remove these unnecessary checks

PoulavBhowmick03:
## Issue Addressed

Fixes #842

## Proposed Changes

- Remove 3 redundant size-bound checks from `SignedSSVMessage::validate()` (`TooManySignatures`, `TooManyOperatorIDs`, `FullDataTooLong`) — these are enforced by `VariableList` at the type level.
- Restore the RSA signature size check (`sig.len() != RSA_SIGNATURE_SIZE`) — `VariableList<u8, U256>` only guarantees `<= 256`, not `== 256`. SSZ-decoded messages can have undersized signatures. [Tracked upstream for the next fork](https://github.com/ssvlabs/SIPs/issues/58).
- Wire `validate()` into the network message path by calling it at the start of `validate_decoded_message()`, with a `From<SignedSSVMessageError> for ValidationFailure` mapping. This activates structural validation (RSA sig size, zero signer, duplicate signers, signer/signature count mismatch) on SSZ-decoded messages for the first time.
- Remove the `is_sorted()` check from `validate()` — Go's `Validate()` does not enforce sorting and Go's `Aggregate()` appends without sorting. Signature verification pairs `operator_ids[i]` with `signatures[i]` by index (consistent pairing, not sorted order). Enforcing sorted order would reject valid messages from Go nodes.
- Add 4 tests verifying `validate()` rejects structurally invalid messages that bypass `new()` (undersized RSA sig, zero signer, duplicate signers, length mismatch).

## Additional Info

- `MAX_FULL_DATA_SIZE` moved to test module (only used there now).
- Stale comment on `DuplicatedSigner` check removed — it claimed the check "assumes sorted" but the implementation uses `HashSet`.
- `ValidationFailure` variants `WrongRSASignatureSize`, `DuplicatedSigner`, `SignersAndSignaturesWithDifferentLength` were previously dead (defined but never returned). They are now active via the `From` impl.
- `SignersNotSorted` variant is now unreachable from `validate()` since the sort check was removed.

### DIFF
--- a/anchor/common/ssv_types/src/message.rs
+++ b/anchor/common/ssv_types/src/message.rs
@@ -607,8 +607,8 @@ impl SignedSSVMessage {
     }
 
     pub fn validate(&self) -> Result<(), SignedSSVMessageError> {
-        // Rule: Each RSA signature must be exactly RSA_SIGNATURE_SIZE bytes.
-        // VariableList<u8, U256> only guarantees <= 256, not == 256, so this
+        // Rule: Each RSA signature must be exactly `RSA_SIGNATURE_SIZE` bytes.
+        // `VariableList<u8, U256>` only guarantees <= 256, not == 256, so this
         // catches undersized signatures from SSZ-decoded network messages.
         for (i, sig) in self.signatures.iter().enumerate() {
             if sig.len() != RSA_SIGNATURE_SIZE {
@@ -629,11 +629,14 @@ impl SignedSSVMessage {
             return Err(SignedSSVMessageError::NoSignatures);
         }
 
-        if !self.operator_ids.is_sorted() {
-            return Err(SignedSSVMessageError::SignersNotSorted);
-        }
-
-        // Note: Len Signers & Operators will only be > 1 after commit aggregation
+        // Note: Len Signers & Operators will only be > 1 after commit aggregation.
+        //
+        // No `is_sorted()` check: the Go spec's `Validate()` does not enforce sorting,
+        // and Go's `Aggregate()` appends without sorting. Signature verification
+        // pairs `operator_ids[i]` with `signatures[i]` by index, requiring consistent
+        // pairing rather than sorted order. Both Go (append) and Anchor (sort pairs
+        // together) preserve this pairing. Enforcing sorted order here would reject
+        // valid messages from Go nodes.
 
         // Rule: Signer can't be zero
         if self.operator_ids.iter().any(|&id| *id == 0) {
@@ -985,20 +988,21 @@ mod tests {
         }
     }
 
-    /// Checks that unsorted operator IDs triggers `SignersNotSorted`.
+    /// Verifies that unsorted operator IDs are accepted. The Go spec's Validate()
+    /// does not enforce sorting, and Go's Aggregate() appends without sorting,
+    /// so network messages may have unsorted operator IDs.
     #[test]
-    fn test_signed_ssv_message_signers_not_sorted() {
+    fn test_signed_ssv_message_unsorted_signers_accepted() {
         let ssv_msg = valid_ssv_message();
         let sigs = vec![valid_signature(), valid_signature()];
-        // Not sorted
         let ops = vec![OperatorId(10), OperatorId(2)];
 
         let result = SignedSSVMessage::new(sigs, ops, ssv_msg, vec![]);
 
-        match result {
-            Err(SignedSSVMessageError::SignersNotSorted) => (),
-            other => panic!("Expected SignersNotSorted, got {other:?}"),
-        }
+        assert!(
+            result.is_ok(),
+            "Unsorted signers should be accepted: {result:?}"
+        );
     }
 
     /// Checks that operator ID = 0 triggers `ZeroSigner`.
--- a/anchor/common/ssv_types/src/message.rs
+++ b/anchor/common/ssv_types/src/message.rs
@@ -607,6 +607,19 @@ impl SignedSSVMessage {
     }
 
     pub fn validate(&self) -> Result<(), SignedSSVMessageError> {
+        // Rule: Each RSA signature must be exactly RSA_SIGNATURE_SIZE bytes.
+        // VariableList<u8, U256> only guarantees <= 256, not == 256, so this
+        // catches undersized signatures from SSZ-decoded network messages.
+        for (i, sig) in self.signatures.iter().enumerate() {
+            if sig.len() != RSA_SIGNATURE_SIZE {
+                return Err(SignedSSVMessageError::WrongRSASignatureSize {
+                    index: i,
+                    length: sig.len(),
+                    sig_length: RSA_SIGNATURE_SIZE,
+                });
+            }
+        }
+
         // Rule: Must have at least one signer
         if self.operator_ids.is_empty() {
             return Err(SignedSSVMessageError::NoSigners);
@@ -628,8 +641,6 @@ impl SignedSSVMessage {
         }
 
         // Rule: Signers must be unique
-        // This check assumes that signers is sorted, so this rule should be after the check for
-        // ErrSignersNotSorted.
         let mut seen_ids = HashSet::with_capacity(self.operator_ids.len());
         for &id in &self.operator_ids {
             if !seen_ids.insert(id) {
@@ -1220,6 +1231,123 @@ mod tests {
         assert_eq!(
             SSVMessageDataLen::to_usize(),
             std::cmp::max(MAX_PARTIAL_SIGNATURE_MSGS_SIZE, MAX_CONSENSUS_MSG_SIZE)
+        );
+    }
+
+    // ==================== SSZ-decoded SignedSSVMessage validation tests ====================
+    //
+    // These tests verify that `SignedSSVMessage::validate()` rejects structurally invalid
+    // messages that bypass the `new()` constructor. This matters because `from_ssz_bytes()`
+    // can produce structs with invalid state (e.g., undersized signatures, zero signers),
+    // and the `validate()` call in `validate_decoded_message()` must catch them.
+
+    /// Helper to build a `SignedSSVMessage` with direct field access, bypassing `new()`.
+    /// This simulates a struct produced by `from_ssz_bytes()` with arbitrary field values.
+    fn build_unvalidated_signed_ssv_message(
+        signatures: Vec<Vec<u8>>,
+        operator_ids: Vec<OperatorId>,
+    ) -> SignedSSVMessage {
+        let sig_variable_lists: Vec<VariableList<u8, U256>> = signatures
+            .into_iter()
+            .map(|sig| {
+                VariableList::new(sig).expect("signature should fit in VariableList<u8, U256>")
+            })
+            .collect();
+        SignedSSVMessage {
+            signatures: VariableList::new(sig_variable_lists)
+                .expect("signatures list should fit in VariableList<_, U13>"),
+            operator_ids: VariableList::new(operator_ids)
+                .expect("operator_ids should fit in VariableList<_, U13>"),
+            ssv_message: valid_ssv_message(),
+            full_data: VariableList::empty(),
+        }
+    }
+
+    /// Verify that `validate()` rejects signatures shorter than `RSA_SIGNATURE_SIZE` (256 bytes).
+    /// `from_ssz_bytes()` can produce a `VariableList<u8, U256>` with fewer than 256 bytes
+    /// because the type only enforces an upper bound, not an exact length.
+    #[test]
+    fn test_validate_rejects_undersized_rsa_signature() {
+        // Arrange: create a message with a 100-byte signature (well under the required 256)
+        let undersized_signature = vec![0u8; 100];
+        let signed_msg =
+            build_unvalidated_signed_ssv_message(vec![undersized_signature], vec![OperatorId(1)]);
+
+        // Act
+        let result = signed_msg.validate();
+
+        // Assert
+        assert_eq!(
+            result,
+            Err(SignedSSVMessageError::WrongRSASignatureSize {
+                index: 0,
+                length: 100,
+                sig_length: RSA_SIGNATURE_SIZE,
+            }),
+            "validate() must reject signatures shorter than RSA_SIGNATURE_SIZE"
+        );
+    }
+
+    /// Verify that `validate()` rejects a signer with operator ID = 0 on an SSZ-decoded message.
+    /// `from_ssz_bytes()` does not check operator ID values, so `validate()` must catch this.
+    #[test]
+    fn test_validate_rejects_zero_signer_on_ssz_decoded() {
+        // Arrange: valid 256-byte signature but operator_id = 0
+        let valid_sig = vec![0u8; RSA_SIGNATURE_SIZE];
+        let signed_msg = build_unvalidated_signed_ssv_message(vec![valid_sig], vec![OperatorId(0)]);
+
+        // Act
+        let result = signed_msg.validate();
+
+        // Assert
+        assert_eq!(
+            result,
+            Err(SignedSSVMessageError::ZeroSigner),
+            "validate() must reject operator_id = 0"
+        );
+    }
+
+    /// Verify that `validate()` rejects duplicate operator IDs on an SSZ-decoded message.
+    /// `from_ssz_bytes()` does not enforce uniqueness, so `validate()` must catch duplicates.
+    #[test]
+    fn test_validate_rejects_duplicate_signers_on_ssz_decoded() {
+        // Arrange: two valid signatures with the same operator_id (sorted, so passes sort check)
+        let sig1 = vec![0u8; RSA_SIGNATURE_SIZE];
+        let sig2 = vec![0u8; RSA_SIGNATURE_SIZE];
+        let signed_msg = build_unvalidated_signed_ssv_message(
+            vec![sig1, sig2],
+            vec![OperatorId(1), OperatorId(1)],
+        );
+
+        // Act
+        let result = signed_msg.validate();
+
+        // Assert
+        assert_eq!(
+            result,
+            Err(SignedSSVMessageError::DuplicatedSigner),
+            "validate() must reject duplicate operator IDs"
+        );
+    }
+
+    /// Verify that `validate()` rejects a mismatch between the number of signers and signatures
+    /// on an SSZ-decoded message. `from_ssz_bytes()` does not enforce length equality between
+    /// the `operator_ids` and `signatures` lists.
+    #[test]
+    fn test_validate_rejects_signers_signatures_length_mismatch() {
+        // Arrange: two signers but only one signature
+        let sig = vec![0u8; RSA_SIGNATURE_SIZE];
+        let signed_msg =
+            build_unvalidated_signed_ssv_message(vec![sig], vec![OperatorId(1), OperatorId(2)]);
+
+        // Act
+        let result = signed_msg.validate();
+
+        // Assert
+        assert_eq!(
+            result,
+            Err(SignedSSVMessageError::SignersAndSignaturesWithDifferentLength),
+            "validate() must reject mismatched signers/signatures lengths"
         );
     }
 }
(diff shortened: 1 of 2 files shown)


anchor/common/ssv_types/src/message.rs
```diff
@@ -609,36 +607,6 @@ impl SignedSSVMessage {
...
-                max: MAX_SIGNATURES,
-            });
-        }
-
-        if self.full_data.len() > MAX_FULL_DATA_SIZE {
-            return Err(SignedSSVMessageError::FullDataTooLong {
-                provided: self.full_data.len(),
-                max: MAX_FULL_DATA_SIZE,
-            });
-        }
```
shane-moore, line 622: lets remove this blank line

shane-moore:
lets update pr description to something like `refactor: remove unreachable size-bound checks from SignedSSVMessage::validate` to get CI to pass

PoulavBhowmick03:
> lets update pr description to something like `refactor: remove unreachable size-bound checks from SignedSSVMessage::validate` to get CI to pass

done!

anchor/common/ssv_types/src/message.rs
```diff
@@ -609,37 +607,6 @@ impl SignedSSVMessage {
...
 
     pub fn validate(&self) -> Result<(), SignedSSVMessageError> {
-        if self.signatures.len() > MAX_SIGNATURES {
-            return Err(SignedSSVMessageError::TooManySignatures {
-                provided: self.signatures.len(),
-                max: MAX_SIGNATURES,
-            });
-        }
-
-        for (i, sig) in self.signatures.iter().enumerate() {
```
diegomrsantos, line 619: Thanks for the cleanup. After investigation, we think this PR could go further and become a fix rather than just a refactor. We've pushed commits implementing the suggestions below.

**Agree with removing 3 checks:** `signatures.len() > MAX_SIGNATURES`, `operator_ids.len() > MAX_SIGNATURES`, and `full_data.len() > MAX_FULL_DATA_SIZE` are genuinely redundant with `VariableList` bounds.

**Suggest keeping the RSA sig size check:** `VariableList<u8, U256>` only guarantees `<= 256`, not `== 256`. RSA-2048 signatures are always exactly 256 bytes. This is a consequence of the Go spec using `ssz-max:"13,256"` (variable-length) rather than fixed-size encoding — [tracked upstream for the next fork](https://github.com/ssvlabs/SIPs/issues/58). Note: the Go spec's `Validate()` only checks `len(sig) == 0`, not exact size — so this is Anchor-specific defense-in-depth.

**Bigger issue: `validate()` never runs on network messages.** Currently `validate()` is only called from `new()`. SSZ-decoded messages from the network go through `from_ssz_bytes()` → `validate_decoded_message()`, which never calls `validate()`. Several `ValidationFailure` variants (`WrongRSASignatureSize`, `DuplicatedSigner`, `SignersNotSorted`, `SignersAndSignaturesWithDifferentLength`) are defined but dead — never returned by any code path. (`ZeroSigner` is active via `partial_signature.rs` but the others are not.)

**Changes pushed:**
1. Restored the RSA sig size check in `validate()`.
2. Added `signed_ssv_message.validate()` call at the start of `validate_decoded_message()` with a `From<SignedSSVMessageError> for ValidationFailure` mapping, activating structural validation on network messages.
3. Added 4 tests that build `SignedSSVMessage` structs bypassing `new()` (simulating `from_ssz_bytes()` output) with invalid state (undersized RSA sig, zero signer, duplicate signers, signers/signatures length mismatch) and assert `validate()` rejects them.
4. **Removed the `is_sorted()` check from `validate()`.** We verified that Go's `Validate()` does not enforce sorted operator IDs, and Go's `Aggregate()` appends without sorting. The signature verification code (`verify_message_signatures`) pairs `operator_ids[i]` with `signatures[i]` by index — it requires consistent pairing, not sorting. Both implementations maintain this pairing (Go appends both arrays together; Anchor sorts pairs together in `aggregate()`). Keeping `is_sorted()` would reject valid messages from Go nodes.
5. Removed a stale comment on the `DuplicatedSigner` check that said "assumes sorted" — the implementation uses `HashSet`, which doesn't depend on sort order.

diegomrsantos (changes requested):
let's investigate more what's going on with this function

anchor/common/ssv_types/src/message.rs
diegomrsantos: done

anchor/common/ssv_types/src/message.rs
diegomrsantos: done

diegomrsantos:
@shane-moore pls review again

anchor/common/ssv_types/src/message.rs
```diff
@@ -609,13 +607,9 @@ impl SignedSSVMessage {
...
 
     pub fn validate(&self) -> Result<(), SignedSSVMessageError> {
-        if self.signatures.len() > MAX_SIGNATURES {
-            return Err(SignedSSVMessageError::TooManySignatures {
-                provided: self.signatures.len(),
-                max: MAX_SIGNATURES,
-            });
-        }
-
+        // Rule: Each RSA signature must be exactly RSA_SIGNATURE_SIZE bytes.
```
shane-moore: great find btw. when modifying this kind of thing, important to always cross-reference against the [ssv-spec](https://github.com/ssvlabs/ssv-spec)

diegomrsantos:
> Some required checks have failed. Could you please take a look @PoulavBhowmick03? 🙏

@dknopik this seems a false positive

anchor/common/ssv_types/src/message.rs
diegomrsantos: I created https://github.com/ssvlabs/SIPs/issues/58#issuecomment-3999949883

anchor/common/ssv_types/src/message.rs
```diff
@@ -649,20 +629,21 @@ impl SignedSSVMessage {
...
         }
 
-        if !self.operator_ids.is_sorted() {
-            return Err(SignedSSVMessageError::SignersNotSorted);
-        }
-
-        // Note: Len Signers & Operators will only be > 1 after commit aggregation
+        // Note: Len Signers & Operators will only be > 1 after commit aggregation.
+        //
+        // No is_sorted() check: the Go spec's Validate() does not enforce sorting,
```
shane-moore: Good call — added backticks around the code references (`is_sorted()`, `Validate()`, `Aggregate()`, `operator_ids[i]`, etc.) in both `message.rs` and `lib.rs` comments. Pushed.

diegomrsantos:
@dknopik @shane-moore also claude doesn't review the PR if the author isn't in the org

anchor/common/ssv_types/src/message.rs
```diff
@@ -649,20 +629,21 @@ impl SignedSSVMessage {
...
+        // and Go's Aggregate() appends without sorting. Signature verification
+        // pairs operator_ids[i] with signatures[i] by index, requiring consistent
+        // pairing rather than sorted order. Both Go (append) and Anchor (sort pairs
+        // together) preserve this pairing. Enforcing sorted order here would reject
```
shane-moore, line 638: also great find!

anchor/common/ssv_types/src/message.rs
diegomrsantos: done

anchor/common/ssv_types/src/message.rs
```diff
@@ -1254,4 +1237,121 @@ mod tests {
...
+            ssv_message: valid_ssv_message(),
+            full_data: VariableList::empty(),
+        }
+    }
+
+    /// Verify that `validate()` rejects signatures shorter than `RSA_SIGNATURE_SIZE` (256 bytes).
+    /// `from_ssz_bytes()` can produce a `VariableList<u8, U256>` with fewer than 256 bytes
+    /// because the type only enforces an upper bound, not an exact length.
+    #[test]
+    fn test_validate_rejects_undersized_rsa_signature() {
```
shane-moore, line 1274: i think this is the only one of the 4 tests added here that is not already covered by our [signed_ssv_msg.rs](https://github.com/shane-moore/anchor/blob/5738e72f771b6c1bdf1b728b17b4151507bc2910/anchor/spec_tests/src/types/signed_ssv_msg.rs) spec tests. the rest are tested through spec test since it calls the `SignedSSVMessage::validate` just like the we're doing in the 3 subsequent tests

anchor/message_validator/src/lib.rs
```diff
@@ -391,6 +429,11 @@ impl<S: SlotClock + 'static, D: DutiesProvider> Validator<S, D> {
         signed_ssv_message: &SignedSSVMessage,
         topic_context: &TopicContext,
     ) -> Result<ValidatedMessage, ValidationFailure> {
+        // Structural validation: signer/signature invariants, RSA size, etc.
+        signed_ssv_message
+            .validate()
```
shane-moore, line 434: nice

anchor/common/ssv_types/src/message.rs
diegomrsantos: Good catch — removed the 3 redundant tests (zero signer, duplicate signers, length mismatch). They're already covered by the spec fixtures (`signedssvmsg_zero_signer.json`, `signedssvmsg_non_unique_signers.json`, `signedssvmsg_signers_and_signatures_with_different_length.json`).

Kept `test_validate_rejects_undersized_rsa_signature` since `prepare_signatures()` pads to 256 before `new()`, so the RSA size check never fires in the spec test path.

shane-moore:
> @dknopik @shane-moore also claude doesn't review the PR if the author isn't in the org

I'm guessing the original reason for this is cost. don't want to let external folks spam trigger claude pr reviews and claude mentions. If that is the reason, I think I agree with it. Feels like something that could be abused, but maybe you have some thoughts to the contrary? Open to them as well. I could ask Kingy how he feels about it in our next chat as well. 

wdyt of a middle ground kind of solution? we currently have the `claude-recheck` label that triggers the `claude-pr-review` workflow, so we could just apply that label for external contributors PR's when we start reviewing it.  We would just need to modify the `claude-pr-review` workflow's "Check if PR author is org member" to let the `claude-recheck` label event bypass the author being in the org check. seems safe since only folks in the org can add labels I'm pretty sure

shane-moore (approved):

diegomrsantos:
> > @dknopik @shane-moore also claude doesn't review the PR if the author isn't in the org
> 
> I'm guessing the original reason for this is cost. don't want to let external folks spam trigger claude pr reviews and claude mentions. If that is the reason, I think I agree with it. Feels like something that could be abused, but maybe you have some thoughts to the contrary? Open to them as well. I could ask Kingy how he feels about it in our next chat as well.
> 
> wdyt of a middle ground kind of solution? we currently have the `claude-recheck` label that triggers the `claude-pr-review` workflow, so we could just apply that label for external contributors PR's when we start reviewing it. We would just need to modify the `claude-pr-review` workflow's "Check if PR author is org member" to let the `claude-recheck` label event bypass the author being in the org check. seems safe since only folks in the org can add labels I'm pretty sure

I'm in favor of not fixing or avoiding an issue that doesn't exist yet

diegomrsantos:
> > > @dknopik @shane-moore also claude doesn't review the PR if the author isn't in the org
> > 
> > 
> > I'm guessing the original reason for this is cost. don't want to let external folks spam trigger claude pr reviews and claude mentions. If that is the reason, I think I agree with it. Feels like something that could be abused, but maybe you have some thoughts to the contrary? Open to them as well. I could ask Kingy how he feels about it in our next chat as well.
> > wdyt of a middle ground kind of solution? we currently have the `claude-recheck` label that triggers the `claude-pr-review` workflow, so we could just apply that label for external contributors PR's when we start reviewing it. We would just need to modify the `claude-pr-review` workflow's "Check if PR author is org member" to let the `claude-recheck` label event bypass the author being in the org check. seems safe since only folks in the org can add labels I'm pretty sure
> 
> I'm in favor of not fixing or avoiding an issue that doesn't exist yet

Btw, I'm not sure if this label ever worked

diegomrsantos (approved):

PoulavBhowmick03:
Whoa! A lot of catch up for me to do here haha

shane-moore:
> > > > @dknopik @shane-moore also claude doesn't review the PR if the author isn't in the org
> > > 
> > > 
> > > I'm guessing the original reason for this is cost. don't want to let external folks spam trigger claude pr reviews and claude mentions. If that is the reason, I think I agree with it. Feels like something that could be abused, but maybe you have some thoughts to the contrary? Open to them as well. I could ask Kingy how he feels about it in our next chat as well.
> > > wdyt of a middle ground kind of solution? we currently have the `claude-recheck` label that triggers the `claude-pr-review` workflow, so we could just apply that label for external contributors PR's when we start reviewing it. We would just need to modify the `claude-pr-review` workflow's "Check if PR author is org member" to let the `claude-recheck` label event bypass the author being in the org check. seems safe since only folks in the org can add labels I'm pretty sure
> > 
> > 
> > I'm in favor of not fixing or avoiding an issue that doesn't exist yet
> 
> Btw, I'm not sure if this label ever worked

I tried the label on this PR. it triggered the claude-pr-review workflow. but it quit early cuz the pr creator is not in the org, which is why i was suggesting the middle ground solution above. I think what you mean by "I'm in favor of not fixing or avoiding an issue that doesn't exist yet" is that we shouldn't change anything atm cuz not much external contribution?

diegomrsantos:
> > > > > @dknopik @shane-moore also claude doesn't review the PR if the author isn't in the org
> > > > 
> > > > 
> > > > I'm guessing the original reason for this is cost. don't want to let external folks spam trigger claude pr reviews and claude mentions. If that is the reason, I think I agree with it. Feels like something that could be abused, but maybe you have some thoughts to the contrary? Open to them as well. I could ask Kingy how he feels about it in our next chat as well.
> > > > wdyt of a middle ground kind of solution? we currently have the `claude-recheck` label that triggers the `claude-pr-review` workflow, so we could just apply that label for external contributors PR's when we start reviewing it. We would just need to modify the `claude-pr-review` workflow's "Check if PR author is org member" to let the `claude-recheck` label event bypass the author being in the org check. seems safe since only folks in the org can add labels I'm pretty sure
> > > 
> > > 
> > > I'm in favor of not fixing or avoiding an issue that doesn't exist yet
> > 
> > 
> > Btw, I'm not sure if this label ever worked
> 
> I tried the label on this PR. it triggered the claude-pr-review workflow. but it quit early cuz the pr creator is not in the org, which is why i was suggesting the middle ground solution above. I think what you mean by "I'm in favor of not fixing or avoiding an issue that doesn't exist yet" is that we shouldn't change anything atm cuz not much external contribution?

So seems the label is working now. But I'm not sure it'd be a robust solution, as the author can also add the label. Yes, that's what I meant. Contributors spamming the repo with no-value PRs isn't an issue atm.

shane-moore:
> > > > > > @dknopik @shane-moore also claude doesn't review the PR if the author isn't in the org
> > > > > 
> > > > > 
> > > > > I'm guessing the original reason for this is cost. don't want to let external folks spam trigger claude pr reviews and claude mentions. If that is the reason, I think I agree with it. Feels like something that could be abused, but maybe you have some thoughts to the contrary? Open to them as well. I could ask Kingy how he feels about it in our next chat as well.
> > > > > wdyt of a middle ground kind of solution? we currently have the `claude-recheck` label that triggers the `claude-pr-review` workflow, so we could just apply that label for external contributors PR's when we start reviewing it. We would just need to modify the `claude-pr-review` workflow's "Check if PR author is org member" to let the `claude-recheck` label event bypass the author being in the org check. seems safe since only folks in the org can add labels I'm pretty sure
> > > > 
> > > > 
> > > > I'm in favor of not fixing or avoiding an issue that doesn't exist yet
> > > 
> > > 
> > > Btw, I'm not sure if this label ever worked
> > 
> > 
> > I tried the label on this PR. it triggered the claude-pr-review workflow. but it quit early cuz the pr creator is not in the org, which is why i was suggesting the middle ground solution above. I think what you mean by "I'm in favor of not fixing or avoiding an issue that doesn't exist yet" is that we shouldn't change anything atm cuz not much external contribution?
> 
> So seems the label is working now. But I'm not sure it'd be a robust solution, as the author can also add the label. Yes, that's what I meant. Contributors spamming the repo with no-value PRs isn't an issue atm.

checked the gh permissions [docs](https://docs.github.com/en/organizations/managing-user-access-to-your-organizations-repositories/managing-repository-roles/repository-roles-for-an-organization). since external folks have `Read` access, they shouldnt be allowed to add labels. but I see your point that contributors spamming hte repo with no value prs isn't an issue atm, so would you say you're in favor of opening up claude pr review CI and `claude-mentions` to run for non-org members PR's? if so, I'll discuss with kingy/Daniel in my next chat with them

diegomrsantos:
> > > > > > > @dknopik @shane-moore also claude doesn't review the PR if the author isn't in the org
> > > > > > 
> > > > > > 
> > > > > > I'm guessing the original reason for this is cost. don't want to let external folks spam trigger claude pr reviews and claude mentions. If that is the reason, I think I agree with it. Feels like something that could be abused, but maybe you have some thoughts to the contrary? Open to them as well. I could ask Kingy how he feels about it in our next chat as well.
> > > > > > wdyt of a middle ground kind of solution? we currently have the `claude-recheck` label that triggers the `claude-pr-review` workflow, so we could just apply that label for external contributors PR's when we start reviewing it. We would just need to modify the `claude-pr-review` workflow's "Check if PR author is org member" to let the `claude-recheck` label event bypass the author being in the org check. seems safe since only folks in the org can add labels I'm pretty sure
> > > > > 
> > > > > 
> > > > > I'm in favor of not fixing or avoiding an issue that doesn't exist yet
> > > > 
> > > > 
> > > > Btw, I'm not sure if this label ever worked
> > > 
> > > 
> > > I tried the label on this PR. it triggered the claude-pr-review workflow. but it quit early cuz the pr creator is not in the org, which is why i was suggesting the middle ground solution above. I think what you mean by "I'm in favor of not fixing or avoiding an issue that doesn't exist yet" is that we shouldn't change anything atm cuz not much external contribution?
> > 
> > 
> > So seems the label is working now. But I'm not sure it'd be a robust solution, as the author can also add the label. Yes, that's what I meant. Contributors spamming the repo with no-value PRs isn't an issue atm.
> 
> checked the gh permissions [docs](https://docs.github.com/en/organizations/managing-user-access-to-your-organizations-repositories/managing-repository-roles/repository-roles-for-an-organization). since external folks have `Read` access, they shouldnt be allowed to add labels. but I see your point that contributors spamming hte repo with no value prs isn't an issue atm, so would you say you're in favor of opening up claude pr review CI and `claude-mentions` to run for non-org members PR's? if so, I'll discuss with kingy/Daniel in my next chat with them

Oh I didn't know this about the labels, thanks for sharing. In this case, I do agree it's a good middle-ground solution!

PR #855, opened 2026-02-27
refactor: validator store trait fns return streams

shane-moore:
## Issue Addressed
  Prepares `unstable` for PR #834 (attestation batching optimization) by migrating to the updated lighthouse `ValidatorStore` trait

## Proposed Changes

 - Update Lighthouse dependency to `shane-moore/lighthouse` fork [PR](https://github.com/sigp/lighthouse/pull/8880) with streaming `ValidatorStore` trait fn return types. 
   - **Note** that [PR](https://github.com/sigp/lighthouse/pull/8880) must be merged on the lh side, and anchor updated to use `sigp/lighthouse` for dependencies, before we merge this code to our `stable` branch
 - Implement four streaming trait methods to replace existing fn's:
    - `sign_aggregate_and_proofs`,  `sign_sync_committee_signatures`,
  `sign_sync_committee_contributions`, and `sign_attestations`

The impl's now have naive stream wrappers, and the functionality is equivalent to the previous functions in `unstable`. The idea is that we can have separate PR's built on top of this base that will refactor these trait methods to stream committee based batched signing

## Additional Info
All new methods use `stream::once` to wrap existing logic, no behavioral changes. PR #834 should rebase on top and only needs to optimize `sign_attestations`

PR #859, opened 2026-03-03
chore: update Lighthouse and rusqlite dependencies

shane-moore:
## Issue Addressed
 Splitting dependency updates out of #855 to keep that PR focused on validator store streaming changes.

## Proposed Changes
Update Lighthouse pin from `edba56b` to `f4b5b03` (latest unstable). This pulls in two upstream breaking changes:
  - Add `vendored` OpenSSL to `Cargo.toml` to fix testnet CI dynamic linking [failure](https://github.com/sigp/anchor/actions/runs/22505079954/job/65203886802?pr=855) after Lighthouse removed its transitive openssl-src dependency in https://github.com/sigp/lighthouse/pull/8768. This addition on our end has no changes to openssl functionality in anchor because `vendored` only controls how libssl is linked (static vs dynamic), not the Rust API surface.
  - **rusqlite 0.28 → 0.38** ([lighthouse#8866](https://github.com/sigp/lighthouse/pull/8866)): rusqlite removed
  `FromSql`/`ToSql` impls for `u64` and `usize` by default since SQLite's native integer is `i64`. We add `FromSql`/`ToSql` impls on `OperatorId` and `ValidatorIndex` newtypes so most call sites just work via type inference. A handful of raw `u64` columns (`block_number`, `schema_version`) use explicit `i64` casts.

  Also adapts to minor LH API changes (`get_builder_domain` → `get_builder_application_domain`, new `sign_execution_payload_envelope` trait method, `Timeouts` dropping its `events` field, `eth2` no longer re-exporting `reqwest`).

## Additional Info
 #855 will rebase on this once merged.

### DIFF
--- a/anchor/common/ssv_types/src/cluster.rs
+++ b/anchor/common/ssv_types/src/cluster.rs
@@ -3,6 +3,10 @@ use std::fmt::Debug;
 use bls::PublicKeyBytes;
 use derive_more::{Deref, Display, From};
 use indexmap::IndexSet;
+use rusqlite::{
+    ToSql,
+    types::{FromSql, FromSqlError, FromSqlResult, ToSqlOutput, Value, ValueRef},
+};
 use ssz_derive::{Decode, Encode};
 use types::{Address, Graffiti};
 
@@ -72,6 +76,20 @@ pub struct ClusterMember {
 )]
 #[ssz(struct_behaviour = "transparent")]
 pub struct ValidatorIndex(pub usize);
+
+impl FromSql for ValidatorIndex {
+    fn column_result(value: ValueRef<'_>) -> FromSqlResult<Self> {
+        let v = value.as_i64()?;
+        let v = usize::try_from(v).map_err(|_| FromSqlError::OutOfRange(v))?;
+        Ok(ValidatorIndex(v))
+    }
+}
+
+impl ToSql for ValidatorIndex {
+    fn to_sql(&self) -> rusqlite::Result<ToSqlOutput<'_>> {
+        Ok(ToSqlOutput::Owned(Value::Integer(self.0 as i64)))
+    }
+}
 
 impl From<ValidatorIndex> for u64 {
     fn from(value: ValidatorIndex) -> Self {
--- a/anchor/common/ssv_types/src/cluster.rs
+++ b/anchor/common/ssv_types/src/cluster.rs
@@ -87,7 +87,9 @@ impl FromSql for ValidatorIndex {
 
 impl ToSql for ValidatorIndex {
     fn to_sql(&self) -> rusqlite::Result<ToSqlOutput<'_>> {
-        Ok(ToSqlOutput::Owned(Value::Integer(self.0 as i64)))
+        let v = i64::try_from(self.0)
+            .map_err(|e| rusqlite::Error::ToSqlConversionFailure(Box::new(e)))?;
+        Ok(ToSqlOutput::Owned(Value::Integer(v)))
     }
 }
 
--- a/anchor/common/ssv_types/src/operator.rs
+++ b/anchor/common/ssv_types/src/operator.rs
@@ -41,7 +41,9 @@ impl FromSql for OperatorId {
 
 impl ToSql for OperatorId {
     fn to_sql(&self) -> rusqlite::Result<ToSqlOutput<'_>> {
-        Ok(ToSqlOutput::Owned(Value::Integer(self.0 as i64)))
+        let v = i64::try_from(self.0)
+            .map_err(|e| rusqlite::Error::ToSqlConversionFailure(Box::new(e)))?;
+        Ok(ToSqlOutput::Owned(Value::Integer(v)))
     }
 }
 
--- a/anchor/common/ssv_types/src/operator.rs
+++ b/anchor/common/ssv_types/src/operator.rs
@@ -2,6 +2,10 @@ use std::{cmp::Eq, fmt::Debug, hash::Hash};
 
 use derive_more::{Deref, Display, From};
 use openssl::{pkey::Public, rsa::Rsa};
+use rusqlite::{
+    ToSql,
+    types::{FromSql, FromSqlError, FromSqlResult, ToSqlOutput, Value, ValueRef},
+};
 use ssz_derive::{Decode, Encode};
 use tree_hash::{Hash256, PackedEncoding, TreeHash, TreeHashType};
 use types::Address;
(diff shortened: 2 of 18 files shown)


anchor/common/ssv_types/src/operator.rs
```diff
@@ -26,6 +30,21 @@ use types::Address;
...
+impl FromSql for OperatorId {
+    fn column_result(value: ValueRef<'_>) -> FromSqlResult<Self> {
+        let v = value.as_i64()?;
+        let v = u64::try_from(v).map_err(|_| FromSqlError::OutOfRange(v))?;
+        Ok(OperatorId(v))
+    }
+}
+
+impl ToSql for OperatorId {
+    fn to_sql(&self) -> rusqlite::Result<ToSqlOutput<'_>> {
```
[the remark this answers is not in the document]
shane-moore: Good catch, fixed. The newtype impls now use `i64::try_from` with proper error propagation

anchor/common/ssv_types/src/cluster.rs
```diff
@@ -73,6 +77,20 @@ pub struct ClusterMember {
...
+impl FromSql for ValidatorIndex {
+    fn column_result(value: ValueRef<'_>) -> FromSqlResult<Self> {
+        let v = value.as_i64()?;
+        let v = usize::try_from(v).map_err(|_| FromSqlError::OutOfRange(v))?;
+        Ok(ValidatorIndex(v))
+    }
+}
+
+impl ToSql for ValidatorIndex {
+    fn to_sql(&self) -> rusqlite::Result<ToSqlOutput<'_>> {
```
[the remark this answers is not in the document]
shane-moore: Good catch, fixed. The newtype impls now use `i64::try_from` with proper error propagation

anchor/database/src/lib.rs
```diff
@@ -208,7 +208,7 @@ impl NetworkDatabase {
         tx: &Transaction<'_>,
     ) -> Result<(), DatabaseError> {
         tx.prepare_cached(sql_operations::UPDATE_BLOCK_NUMBER)?
-            .execute(params![block_number])?;
+            .execute(params![block_number as i64])?;
```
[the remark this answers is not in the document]
shane-moore: These raw u64 values (block numbers, operator IDs) are domain-bounded and will
   never approach i64::MAX. Block numbers at 1 block/12s would take ~3.5 
  trillion years to overflow, and operator IDs are sequential in the low
  thousands. The as i64 cast is safe here since adding try_from with error
  construction adds verbosity for an impossible scenario

anchor/database/src/state.rs
```diff
@@ -136,13 +136,16 @@ impl NetworkState {
     // Get the last block that was processed and saved to db
     fn get_last_processed_block_from_db(conn: &PoolConn) -> Result<u64, DatabaseError> {
         conn.prepare_cached(sql_operations::GET_BLOCK_NUMBER)?
-            .query_row(params![], |row| row.get(0))
+            .query_row(params![], |row| row.get::<_, i64>(0).map(|v| v as u64))
```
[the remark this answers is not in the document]
shane-moore: Same reasoning as above — these values are written by us as non-negative integers (block numbers, operator IDs). Negative values in these columns would indicate a bug elsewhere, not something we'd recover from gracefully at the read site. The as u64 cast is fine here.

anchor/database/src/lib.rs
diegomrsantos: Reasoning makes sense — these values are clearly domain-bounded. Worth adding a brief comment near the casts (e.g. `// Safe: block numbers / operator IDs are domain-bounded well below i64::MAX`) so the assumption is explicit and future readers don't have to re-derive it.

diegomrsantos:
## Consider enabling `fallible_uint` feature

In rusqlite 0.28, `FromSql`/`ToSql` impls for `u64` and `usize` were available by default. In 0.38 they moved behind the `fallible_uint` feature flag — so this PR effectively lost a capability that was previously there.

Enabling it would let you remove the manual `as i64` / `as u64` casts in the database layer (`lib.rs`, `state.rs`, `keysplit_operations.rs`, tests) and get **checked conversions for free** — rusqlite's fallible impl does `i64::try_from` internally and returns a proper error, which is strictly safer than the current `as` casts.

One-line change in workspace `Cargo.toml`:

```toml
rusqlite = { version = "0.38.0", features = ["fallible_uint"] }
```

Then the raw `u64` call sites can go back to direct usage (`params![block_number]`, `row.get::<_, u64>(0)`) without manual casting. The `OperatorId` / `ValidatorIndex` newtype impls are still needed regardless since rusqlite doesn't know about newtypes.

Not blocking — the current approach works and the values are domain-bounded — but this would be less code and safer by default.

shane-moore:
> ## Consider enabling `fallible_uint` feature
> In rusqlite 0.28, `FromSql`/`ToSql` impls for `u64` and `usize` were available by default. In 0.38 they moved behind the `fallible_uint` feature flag — so this PR effectively lost a capability that was previously there.
> 
> Enabling it would let you remove the manual `as i64` / `as u64` casts in the database layer (`lib.rs`, `state.rs`, `keysplit_operations.rs`, tests) and get **checked conversions for free** — rusqlite's fallible impl does `i64::try_from` internally and returns a proper error, which is strictly safer than the current `as` casts.
> 
> One-line change in workspace `Cargo.toml`:
> 
> ```toml
> rusqlite = { version = "0.38.0", features = ["fallible_uint"] }
> ```
> 
> Then the raw `u64` call sites can go back to direct usage (`params![block_number]`, `row.get::<_, u64>(0)`) without manual casting. The `OperatorId` / `ValidatorIndex` newtype impls are still needed regardless since rusqlite doesn't know about newtypes.
> 
> Not blocking — the current approach works and the values are domain-bounded — but this would be less code and safer by default.

how cool! implemented in https://github.com/sigp/anchor/pull/859/changes/6cb7465b0d5971a716949e77bfce7fdbea71b9fd. will keep in mind to check if dependencies have some breaking changes, maybe they have just hid the functionality behind a feature flag ha

anchor/database/src/tests/metadata_tests.rs
```diff
@@ -181,7 +181,7 @@ mod tests {
         // Verify it's version 1
         {
             let conn = Connection::open(&db_path).expect("Failed to open database");
-            let version: u64 = conn
+            let version: i64 = conn
```
diegomrsantos: Nit: with `fallible_uint` now enabled, these three `i64` annotations (lines 184, 209, 232) can go back to `u64` — matching the rest of the codebase where the manual casts were reverted.

anchor/database/src/schema.rs
```diff
@@ -141,7 +141,7 @@ fn determine_database_type(conn: &Connection, network_name: &str) -> DatabaseTyp
                 .query_row(sql_operations::GET_LEGACY_BLOCK, [], |row| {
                     // Check if there is the expected column and no further columns.
                     Ok(
-                        row.get::<_, u64>("block_number").is_ok()
+                        row.get::<_, i64>("block_number").is_ok()
```
diegomrsantos: Same nit: `row.get::<_, i64>("block_number")` could stay `u64` now that `fallible_uint` is enabled. Only used for `.is_ok()` so no behavioral difference either way.

anchor/database/src/tests/metadata_tests.rs
shane-moore: no doubt! https://github.com/sigp/anchor/pull/859/changes/f6812edccced5e8576aca755cd96b21458995713

anchor/database/src/schema.rs
shane-moore: for sure, https://github.com/sigp/anchor/pull/859/changes/f6812edccced5e8576aca755cd96b21458995713

diegomrsantos (approved):
LGTM!

PR #861, opened 2026-03-05
chore: restructure CLAUDE.md into modular rules

diegomrsantos:
## Issue Addressed

N/A — improves AI-assisted development workflow by restructuring project instructions.

## Proposed Changes

The monolithic `CLAUDE.md` (474 lines) was well beyond the recommended size for effective context loading. This PR splits it into a concise core file (~90 lines) plus 10 modular rule files under `.claude/rules/`.

The core `CLAUDE.md` retains only essential project context: description, commands, architecture overview, code organization, specialized agents, and development tips. Detailed standards are extracted into dedicated rule files:

- **Rust-specific rules** (`rust-style.md`, `production-safety.md`, `testing.md`, `dependency-api.md`) use `paths` frontmatter so they load only when relevant files are touched
- **Workflow rules** (`verification.md`, `pr-workflow.md`, `dev-workflow.md`, `commands.md`, `session-learning.md`) and **architecture rules** (`architecture.md`) load always

**What did NOT change:** No code changes. All existing guidance content is preserved — it's reorganized, not removed. The rule files merge and deduplicate content that was previously split across `CLAUDE.md` and local rule files.

**Reading order:** Start with `CLAUDE.md` (the trimmed core), then scan any rule file of interest.

## Additional Info

- Rule files with `paths` frontmatter are conditionally loaded, reducing token overhead for unrelated tasks
- Verified: `make cargo-fmt-check`, `make lint`, `make sort` all pass (pre-commit hooks ran successfully)

### DIFF
--- a/.claude/rules/architecture.md
+++ b/.claude/rules/architecture.md
@@ -1,0 +1,26 @@
+## Architecture and design
+
+Anchor is a modular Rust workspace with service-oriented architecture. Preserve crate boundaries and typed interfaces.
+
+### Crate structure
+- Each major component is its own crate with a minimal, well-documented public API
+- Crates should only depend on what they need
+- Each crate has clear responsibility and ownership
+
+### Dependency flow
+- Common types and utilities live in `common/` subdirectories
+- Higher-level services depend on lower-level ones, not vice versa
+- Config flows down from client to components; events flow up to coordinators
+
+### Inter-component communication
+- Components communicate via typed message channels
+- System-wide events use the EventBus pattern
+- Components interact through trait interfaces, not concrete implementations
+- Errors are properly typed and propagated up the stack
+
+### Design principles
+1. **Question intermediaries**: If data flows A -> B with no transformation, question why an intermediate layer exists. Each layer must provide clear value (logging, transformation, validation, etc.)
+2. **Separation through interfaces, not layers**: Clean boundaries come from well-defined APIs, not intermediary components
+3. **Simplification is always valid**: Refactoring working code for simplicity is encouraged. Fewer lines and fewer components often indicates better design
+4. **Challenge complexity**: Every abstraction must justify its existence. "We might need it later" is not sufficient. Complexity must solve specific, current problems
+5. **Prefer direct communication**: Unless an intermediary adds clear value (validation, transformation, isolation, or observability), components should communicate directly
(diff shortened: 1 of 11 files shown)


diegomrsantos:
### Note on `paths` frontmatter in rule files

The `paths` field is used for conditional rule loading:

1. Rules **without** `paths` load at session start.
2. Rules **with** `paths` load only when Claude works with matching files.

Source: https://code.claude.com/docs/en/memory (section "Path-specific rules").

In this repo, keeping `paths` on Rust-specific rules still has value:

- Avoids loading Rust coding rules in docs/PR-only sessions
- Reduces instruction noise and potential conflicts
- Keeps intent explicit and future-proof if project scope expands

The current split is reasonable. Benefit is modest in a Rust-heavy repo, but real.

shane-moore (approved):
I think about the same amount of context will be injected when using an agent after this PR compared to before, but with some modest gains as mentioned since some rules will only be triggered when `.rs` files are hit. but what i like most is that the pr leans up the `Claude.md` and organizes rules per domain

i also think how we approach `.claude` should be updated pretty regularly, especially as new models come out. so this is definitely a step in the right direction, i.e. using `rules`

diegomrsantos:
> I think about the same amount of context will be injected when using an agent after this PR compared to before, but with some modest gains as mentioned since some rules will only be triggered when `.rs` files are hit. but what i like most is that the pr leans up the `Claude.md` and organizes rules per domain
> 
> i also think how we approach `.claude` should be updated pretty regularly, especially as new models come out. so this is definitely a step in the right direction, i.e. using `rules`

Totally agree. I was thinking if, in the end, this was adding any value, but I believe it's a good idea overall. IMO, splitting files in general into smaller ones when they get too big is usually the right call.

shane-moore:
> > I think about the same amount of context will be injected when using an agent after this PR compared to before, but with some modest gains as mentioned since some rules will only be triggered when `.rs` files are hit. but what i like most is that the pr leans up the `Claude.md` and organizes rules per domain
> > i also think how we approach `.claude` should be updated pretty regularly, especially as new models come out. so this is definitely a step in the right direction, i.e. using `rules`
> 
> Totally agree. I was thinking if, in the end, this was adding any value, but I believe it's a good idea overall. IMO, splitting files in general into smaller ones when they get too big is usually the right call.

🤝

PR #864, opened 2026-03-06
fix: anchor dappnode keys

shane-moore:
## Issue Addressed
https://github.com/dappnode/DAppNodePackage-anchor-generic/pull/1#issuecomment-3846021939

## Proposed Changes
DAppNode's `copyFileToDockerContainer` temporarily starts the container to `docker cp` a key file in. This races with Anchor's key generation -> `try_read` finds no file, `Rsa::generate` runs, then `docker cp` lands the file, and `File::create_new` fails with `EEXIST`.

Fix: re-check for key files after `Rsa::generate` before attempting to write. If one appeared, use it.

## Additional Info
The race window depends on how long `Rsa::generate(2048)` takes, which varies by hardware. Reproduced in Docker with an artificial sleep widening the window.


shane-moore:
@claude, what do you think?

shane-moore:
Closed because Dappnode team thinks the issue is on their end atm.  Will re-open this if proves useful later in the debugging process.

PR #870, opened 2026-03-07
ci: strengthen PR template with evidence, risk, and rollback prompts

diegomrsantos:

The previous PR template was too verbose in some places and too thin in others. It did not consistently force authors to state the real problem being solved, the evidence that it exists, the risks being taken, or how the change can be rolled back. That weakens review quality in a repo where many changes can carry runtime or protocol risk.

Relevant context: Shane's feedback on this PR pointed out that some earlier sections overlapped, so this version simplifies the structure while keeping the important discipline.

## Change Overview

This change replaces the earlier section split with a shorter template built around six core review questions:
- what real problem or value justifies the change
- what changed at a high level
- what risks and trade-offs exist
- what validation proves the change works
- how it can be rolled back
- what blockers or dependencies matter

The template stays concise and explicitly allows `N/A` where appropriate.

The main trade-off is adding a bit more structure to PR writing. The mitigation is keeping the prompts short and focused on the highest-value review context instead of expanding into a long checklist.

## Validation

- Reviewed the final template wording for overlap and unnecessary verbosity.
- Verified the branch changes only `.github/PULL_REQUEST_TEMPLATE.md`.

## Rollback

Revert this PR to restore the previous template.

## Blockers / Dependencies

N/A

If this lands and proves useful in practice, we can later refine prompt wording based on actual reviewer friction instead of adding more sections up front.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -2,9 +2,33 @@
 
 Which issue # does this PR address?
 
+## Motivation
+
+Why is this change needed now? What concrete gap or risk does it address?
+
 ## Proposed Changes
 
 Please list or describe the changes introduced by this PR.
+
+## Real Benefits
+
+What practical outcomes do we gain (e.g., risk reduction, reliability, performance, maintainability)?
+
+## Risk
+
+What could go wrong with this change? What is the expected blast radius if it fails?
+
+## Testing
+
+What verification was performed? Include relevant commands or evidence.
+
+## Rollback
+
+How can this change be reverted safely if needed?
+
+## Blockers / Dependencies
+
+What dependencies, prerequisite merges, external changes, or blockers are relevant?
 
 ## Additional Info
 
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -2,10 +2,34 @@
 
 Which issue # does this PR address?
 
+## Motivation
```
shane-moore: Motivation section feels pretty similar to Issue section

shane-moore:
all the new sections will force PR creators to think a bit more when writing PR descriptions, i.e. having to think about how a change can most easily be rolled back. but I can see that as being a good thing

diegomrsantos:
Agree with your comments. I pushed a new version that tries to balance a better context and overhead. Note that we can use Claude/Codex to fill out these sections, so the real overhead is more reviewing than writing them by hand. Let me know what you think.

shane-moore:
> Agree with your comments. I pushed a new version that tries to balance a better context and overhead. Note that we can use Claude/Codex to fill out these sections, so the real overhead is more reviewing than writing them by hand. Let me know what you think.

Makes sense, and I think the different categories are clearly distinct now.  I am sure there will be edge cases where a "required" field can be marked as N/A, but I'm fine with that

shane-moore (approved):

PR #872, opened 2026-03-09
chore: change gossipsub dependency to main repo

dknopik:
## Proposed Changes

As mainline gossipsub now has all changes of the explicit version we depended on, just use gossipsub via the "gossipsub" libp2p crate feature now.

## Additional Info

Also, bump dependencies.


jking-aus (dismissed):
lgtm

PR #875, opened 2026-03-09
feat: query multiple beacon nodes in parallel

petarjuki7:
# Issue Addressed
WAD - Weighted Attestation Data

## Proposed Changes
- Added parallel fetching from all configured beacon nodes instead of sequential fallback
- Implemented scoring algorithm based on checkpoint epochs and head slot proximity, matching Go-SSV's scoring logic
- Added dual timeout strategy: soft timeout at 1s (return best result if we have responses), hard timeout at 3s (return unconditionally)
- Feature is opt-in behind the `--with-weighted-attestation-data` CLI flag

## Additional Info
- Scoring algorithm intentionally matches Go-SSV for cross-client compatibility
- This PR replaces #751, rebased onto the latest `ValidatorStore` trait / `metadata_service.rs`


anchor/validator_store/src/metadata_service.rs
```diff
@@ -921,6 +985,172 @@ impl<E: EthSpec, T: SlotClock + 'static> MetadataService<E, T> {
...
+                Ok(scored.attestation_data)
+            }
+            None => Err(format!(
+                "No attestation data from {} beacon nodes (succeeded: {}, failed: {})",
+                num_clients, succeeded, failed
+            )),
+        }
+    }
+
+    async fn fetch_and_score(
```
shane-moore, line 1121: the non-WAD path records a `validator_metrics::ATTESTATION_SERVICE_TIMES` / `ATTESTATIONS_HTTP_GET` timer

lets add the same `start_timer_vec` call inside `fetch_and_score` and a WAD-specific histogram (i.e. `WAD_FETCH_TIMES`)

anchor/validator_store/src/metadata_service.rs
```diff
@@ -921,6 +985,172 @@ impl<E: EthSpec, T: SlotClock + 'static> MetadataService<E, T> {
...
+        })
+    }
+
+    async fn get_block_slot(client: &BeaconNodeHttpClient, block_root: Hash256) -> Option<Slot> {
+        tokio::time::timeout(BLOCK_SLOT_LOOKUP_TIMEOUT, async {
+            client
+                .get_beacon_headers_block_id(BlockId::Root(block_root))
+                .await
+                .ok()
+                .flatten()
```
shane-moore: pretty sure the `ok` and `flatten` silenty converts failure modes like http errors, timeouts, etc to `None`.  probably better to match on each failure case and log at each level

anchor/validator_store/src/metadata_service.rs
```diff
@@ -921,6 +985,172 @@ impl<E: EthSpec, T: SlotClock + 'static> MetadataService<E, T> {
...
+                    }
+
+                    if succeeded + failed == num_clients {
+                        break;
+                    }
+                }
+
+                () = &mut soft_timeout, if !soft_timeout_hit => {
+                    soft_timeout_hit = true;
+                    debug!(
```
shane-moore, line 1077: Could we add a metrics counter for the soft timeout path here? This is the point where WAD stops waiting for more BN responses and returns best-so-far after 1s, so tracking its frequency would make it much easier to tell how often we’re getting the full WAD benefit versus short-circuiting

anchor/validator_store/src/metadata_service.rs
```diff
@@ -46,6 +49,59 @@ type SyncByCommitteeMap = HashMap<CommitteeId, Vec<(SyncSubnetId, SyncAggregator
...
+
+#[derive(Debug)]
+struct AttestationScore {
+    score: f64,
+    base_score: f64,
+    distance: Option<u64>,
+    bonus: Option<f64>,
+}
+
+fn calculate_attestation_score(
```
shane-moore, line 83: Could we add a short doc comment here describing the scoring formula and why we’re using it? Since this is compatibility sensitive logic derived from ssv, a brief note on `base_score`, the proximity bonus, and the intent of the tie break would make future maintenance/review much easier.

anchor/validator_store/src/metadata_service.rs
```diff
@@ -46,6 +49,59 @@ type SyncByCommitteeMap = HashMap<CommitteeId, Vec<(SyncSubnetId, SyncAggregator
 /// because SSV has additional latency for QBFT consensus and P2P propagation.
 const BEACON_API_FETCH_TIMEOUT: Duration = Duration::from_secs(2);
 
+// Weighted Attestation Data (WAD) timeouts - inspired by SSV-Go but shorter
```
shane-moore: trying to understand this comment. is it saying that we're intentionally using shorter timeout durations compared to go-ssv? if so, what is the reasoning for the 1/3/500ms timings? could be useful to include in the comment as well

anchor/validator_store/src/metadata_service.rs
```diff
@@ -921,6 +985,172 @@ impl<E: EthSpec, T: SlotClock + 'static> MetadataService<E, T> {
...
+            client_addr,
+            attestation_data,
+            score: score.score,
+        })
+    }
+
+    async fn get_block_slot(client: &BeaconNodeHttpClient, block_root: Hash256) -> Option<Slot> {
+        tokio::time::timeout(BLOCK_SLOT_LOOKUP_TIMEOUT, async {
+            client
+                .get_beacon_headers_block_id(BlockId::Root(block_root))
```
shane-moore, line 1162: looks like the go-ssv impl does retries for these block root lookups:
https://github.com/ssvlabs/ssv/blob/34657cf895f054b23a08e8920810d3c84f05c1cf/beacon/goclient/attest.go#L337-L388

and a cache https://github.com/ssvlabs/ssv/blob/34657cf895f054b23a08e8920810d3c84f05c1cf/beacon/goclient/attest.go#L392-L426

In this flow, each BN returns `attestation_data.beacon_block_root`, and the scorer wants the slot for that root. If multiple BNs return the same root, then without a cache you do the same header lookup over and over.

could make a good follow-up gh issue to look into if we want to add similar, would say not a blocker for this PR

anchor/validator_store/src/metadata_service.rs
```diff
@@ -921,6 +985,172 @@ impl<E: EthSpec, T: SlotClock + 'static> MetadataService<E, T> {
...
+        let client_addr = client.to_string();
+
+        let attestation_data = client
+            .get_validator_attestation_data(slot, 0)
+            .await
+            .map_err(|e| format!("{client_addr}: {e:?}"))?
+            .data;
+
+        let head_slot = Self::get_block_slot(client, attestation_data.beacon_block_root).await;
+        let score = calculate_attestation_score(&attestation_data, slot, head_slot);
```
shane-moore: since the score is derived from attestation_data, it may be slightly cleaner to use `attestation_data.slot` instead of the passed in slot

anchor/validator_store/src/metadata_service.rs
petarjuki7: Good catch! We looked into this. SSV-Go's cache is long-lived on the client struct and their retry loop polls every 100ms to handle the race where a block was just proposed.

For us with typically 3-4 BNs, the benefit is pretty marginal. In the common case (all BNs return the same root), a per-round cache would save 2-3 lightweight header GETs. But since all futures start near-simultaneously they'd mostly miss the cache anyway and fire off their lookups in parallel. The natural stagger from different BN response times gives us redundancy, if one BN fails the lookup, another will likely succeed and score higher.

We had a similar discussion on #751 here: https://github.com/sigp/anchor/pull/751#discussion_r2613999464 and https://github.com/sigp/anchor/pull/751#discussion_r2663204820

Happy to revisit if the new `get_block_slot` failure logs or `WAD_SOFT_TIMEOUT_TOTAL` metric show issues in practice!

anchor/validator_store/src/metadata_service.rs
```diff
@@ -921,6 +1002,201 @@ impl<E: EthSpec, T: SlotClock + 'static> MetadataService<E, T> {
...
+        }
+    }
+
+    async fn fetch_and_score(
+        client: &BeaconNodeHttpClient,
+        slot: Slot,
+    ) -> Result<ScoredAttestationData, String> {
+        let client_addr = client.to_string();
+
+        let _timer = validator_metrics::start_timer_vec(
```
shane-moore: Nit: The `_timer` for `ATTESTATIONS_HTTP_GET` is scoped to the entire `fetch_and_score` function, so it measures `get_block_slot` (up to 500ms) and `calculate_attestation_score` in addition to the actual HTTP GET. In the non-WAD path (line 356), the same metric only covers `get_validator_attestation_data`.
>
> This means operators enabling WAD would see `ATTESTATIONS_HTTP_GET` latency increase and could misattribute it to slower BN responses rather than the added block-slot lookup.
>
> Scoping the timer to just the HTTP call would keep the metric semantics consistent:
> ```rust
> let attestation_data = {
>     let _timer = validator_metrics::start_timer_vec(
>         &validator_metrics::ATTESTATION_SERVICE_TIMES,
>         &[validator_metrics::ATTESTATIONS_HTTP_GET],
>     );
>     client.get_validator_attestation_data(slot, 0)
>         .await
>         .map_err(|e| format!("{client_addr}: {e:?}"))?
>         .data
> };
> ```

petarjuki7: Fixed in `312c07b`!

shane-moore:
The CLI help says "Only useful with multiple beacon nodes" but there's no runtime guard. A `warn!` at startup when WAD is enabled with 1 BN would prevent operators from silently degrading their setup

petarjuki7:
> The CLI help says "Only useful with multiple beacon nodes" but there's no runtime guard. A `warn!` at startup when WAD is enabled with 1 BN would prevent operators from silently degrading their setup

makes sense, added!

shane-moore (approved):

PR #881, opened 2026-03-12
fix: agg attestation batching

### LINKED ISSUE #882: Add ssv committee batching functionality to sign_aggregate_and_proofs
follow similar pattern as introduced in https://github.com/sigp/anchor/pull/834

shane-moore:
## Issue Addressed
Follows up on #834 and closes #882 -> extends committee-based batching from attestations to aggregate-and-proofs.

## Proposed Changes
  - Rewrote `sign_aggregate_and_proofs` to group aggregates by committee and process each concurrently via `FuturesUnordered`, matching the pattern established for attestations in #834
  - Added `sign_committee_aggregate_and_proofs` batch path for boole with single QBFT consensus per committee and batch signature collection
  - All existing pre-boole behavior preserved, just rearranged some functions for readability. The actual diff's are quite small
  - The `group_by_committee` helper now increments `other_error` metrics on pubkey/cluster lookup failures in `sign_attestations` and `sign_sync_committee_signatures`, where previously only warnings/errors were logged.

## Additional Info
- Mirrors the committee batching architecture from `sign_attestations`
- Follow-up PR will be to create integration tests as mentioned in https://github.com/sigp/anchor/issues/768?issue=sigp%7Canchor%7C886

### DIFF
--- a/anchor/validator_store/src/lib.rs
+++ b/anchor/validator_store/src/lib.rs
@@ -60,7 +60,7 @@ use tokio::{
     sync::{Barrier, RwLock, watch},
     time::{Instant, sleep},
 };
-use tracing::{debug, error, info, trace, warn};
+use tracing::{debug, error, info, info_span, trace, warn};
 use types::{
     AbstractExecPayload, Address, AggregateAndProof, AggregateAndProofBase,
     AggregateAndProofElectra, Attestation, AttestationBase, AttestationElectra, BeaconBlock,
@@ -108,18 +108,17 @@ struct SigningRequest<T> {
 impl<T> SigningRequest<T> {
     /// Look up this validator's collected signature, consuming the request.
     ///
-    /// Returns `Ok((duty_data, signature))` on success, or `Err(validator)` if
-    /// the validator index is missing or no signature was collected (so the
-    /// caller can log/metric with the validator metadata).
+    /// Returns `Ok((duty_data, signature))` on success, or `Err(pubkey)` if
+    /// the validator index is missing or no signature was collected.
     fn resolve(
         self,
         signatures: &HashMap<ValidatorIndex, Signature>,
-    ) -> Result<(T, Signature), ValidatorMetadata> {
+    ) -> Result<(T, Signature), PublicKeyBytes> {
         let sig = self
             .validator
             .index
             .and_then(|idx| signatures.get(&idx).cloned())
-            .ok_or(self.validator)?;
+            .ok_or(self.validator.public_key)?;
         Ok((self.duty_data, sig))
     }
 }
@@ -268,29 +267,24 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> AnchorValidator
         &self,
         items: Vec<I>,
         get_pubkey: impl Fn(&I) -> PublicKeyBytes,
-        log_context: &str,
-        counter: &validator_metrics::Result<IntCounterVec>,
-    ) -> HashMap<CommitteeId, Vec<I>> {
-        let mut mapping: HashMap<CommitteeId, Vec<I>> = HashMap::new();
+    ) -> HashMap<CommitteeId, (Cluster, Vec<(ValidatorMetadata, I)>)> {
+        let mut mapping: HashMap<CommitteeId, (Cluster, Vec<(ValidatorMetadata, I)>)> =
+            HashMap::new();
         for item in items {
             let pubkey = get_pubkey(&item);
             match self.get_validator_and_cluster(pubkey) {
-                Ok((_, cluster)) => {
-                    mapping
-                        .entry(cluster.committee_id())
-                        .or_default()
-                        .push(item);
+                Ok((validator, cluster)) => {
+                    let committee_id = cluster.committee_id();
+                    let (_, validators) = mapping
+                        .entry(committee_id)
+                        .or_insert_with(|| (cluster, Vec::new()));
+                    validators.push((validator, item));
                 }
                 Err(Error::UnknownPubkey(pk)) => {
-                    warn!(
-                        ?pk,
-                        "Unknown pubkey while grouping {}, skipping", log_context
-                    );
-                    validator_metrics::inc_counter_vec(counter, &[metrics::OTHER_ERROR]);
+                    warn!(?pk, "Unknown pubkey while grouping, skipping");
                 }
                 Err(e) => {
-                    error!(error = ?e, ?pubkey, "Failed to get cluster for {}, skipping", log_context);
-                    validator_metrics::inc_counter_vec(counter, &[metrics::OTHER_ERROR]);
+                    error!(error = ?e, ?pubkey, "Failed to get cluster, skipping");
                 }
             }
         }
@@ -301,7 +295,6 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> AnchorValidator
     ///
     /// Shared across all `sign_committee_*` methods. Builds `CollectionMode::Committee`,
     /// fans out `collect_signature` calls concurrently, and returns the collected signatures.
-    #[expect(clippy::too_many_arguments)]
     async fn collect_prepared_signatures<D>(
         &self,
         role: Role,
@@ -1302,17 +1295,15 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> AnchorValidator
     async fn sign_committee_sync_committee_signatures(
         &self,
         committee_id: CommitteeId,
-        messages: Vec<SyncMessageToSign>,
+        cluster: Cluster,
+        messages: Vec<(ValidatorMetadata, SyncMessageToSign)>,
     ) -> Result<Vec<SyncCommitteeMessage>, Error> {
-        let Some(first) = messages.first() else {
+        let Some((_, first)) = messages.first() else {
             warn!("sign_committee_sync_committee_signatures called with empty messages");
             return Ok(vec![]);
         };
         let slot = first.slot;
         let epoch = slot.epoch(E::slots_per_epoch());
-
-        // All validators in this committee share the same cluster
-        let (_, cluster) = self.get_validator_and_cluster(first.pubkey)?;
 
         let voting_context = self.get_voting_context(slot).await?;
         let validator_attestation_committees =
@@ -1346,32 +1337,18 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> AnchorValidator
             Completed::Success(data) => data,
         };
 
-        // Prepare all validators
+        // Prepare all validators (metadata already resolved by `group_by_committee`)
         let domain = self.get_domain(epoch, Domain::SyncCommittee);
         let signing_root = data.block_root.signing_root(domain);
 
-        let mut prepared: Vec<SigningRequest<u64>> = Vec::with_capacity(messages.len());
-        for msg in &messages {
-            let validator = match self.database.state().metadata().get_by(&msg.pubkey) {
-                Some(v) => v.clone(),
-                None => {
-                    warn!(
-                        pubkey = ?msg.pubkey,
-                        "Unknown pubkey while signing sync committee message, skipping"
-                    );
-                    validator_metrics::inc_counter_vec(
-                        &validator_metrics::SIGNED_SYNC_COMMITTEE_MESSAGES_TOTAL,
-                        &[metrics::OTHER_ERROR],
-                    );
-                    continue;
-                }
-            };
-            prepared.push(SigningRequest {
+        let prepared: Vec<SigningRequest<u64>> = messages
+            .into_iter()
+            .map(|(validator, msg)| SigningRequest {
                 validator,
                 signing_root,
                 duty_data: msg.validator_index,
-            });
-        }
+            })
+            .collect();
 
         // Collect signatures and assemble results
         let committee_validator_indices = self.get_committee_validator_indices(&committee_id);
@@ -1394,9 +1371,9 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> AnchorValidator
         for req in prepared {
             let (validator_index, signature) = match req.resolve(&signatures) {
                 Ok(resolved) => resolved,
-                Err(validator) => {
+                Err(pubkey) => {
                     warn!(
-                        pubkey = ?validator.public_key,
+                        ?pubkey,
                         "Missing validator index or signature, skipping sync committee message"
                     );
                     validator_metrics::inc_counter_vec(
@@ -1428,19 +1405,16 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> AnchorValidator
     async fn sign_committee_attestations(
         &self,
         committee_id: CommitteeId,
-        attestations: Vec<AttestationToSign<E>>,
+        cluster: Cluster,
+        attestations: Vec<(ValidatorMetadata, AttestationToSign<E>)>,
     ) -> Result<Vec<(u64, Attestation<E>, PublicKeyBytes)>, Error> {
         // Early return and log error for empty attestations
-        let Some(first_attestation) = attestations.first() else {
+        let Some((_, first_attestation)) = attestations.first() else {
             warn!("sign_committee_attestations called with empty attestations");
             return Ok(vec![]);
         };
         let slot = first_attestation.attestation.data().slot;
         let first_att_data = first_attestation.attestation.data();
-
-        // All validators in this committee share the same cluster (same set of operators).
-        // Look up once from the first attestation and reuse for QBFT + signing.
-        let (_, cluster) = self.get_validator_and_cluster(first_attestation.pubkey)?;
 
         let voting_context_tx = self.get_voting_context(slot).await?;
         let validator_attestation_committees =
@@ -1482,38 +1456,23 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> AnchorValidator
         let domain_hash = self.get_domain(data.target.epoch, Domain::BeaconAttester);
 
         // Prepare all validators and apply consensus results upfront
-        let mut prepared: Vec<SigningRequest<(u64, PublicKeyBytes, usize, Attestation<E>)>> =
-            Vec::with_capacity(attestations.len());
-        for att in attestations {
-            let (validator_index, pubkey, validator_committee_position, mut attestation) = (
-                att.validator_index,
-                att.pubkey,
-                att.validator_committee_index,
-                att.attestation,
-            );
-            let validator = match self.database.state().metadata().get_by(&pubkey) {
-                Some(v) => v.clone(),
-                None => {
-                    warn!(
-                        ?pubkey,
-                        "Unknown pubkey while signing attestation, skipping"
-                    );
-                    continue;
+        // (metadata already resolved by `group_by_committee`)
+        let prepared: Vec<SigningRequest<AttestationToSign<E>>> = attestations
+            .into_iter()
+            .map(|(validator, mut att)| {
+                // Apply consensus result to this attestation
+                att.attestation.data_mut().beacon_block_root = data.block_root;
+                att.attestation.data_mut().source = data.source;
+                att.attestation.data_mut().target = data.target;
+
+                let signing_root = att.attestation.data().signing_root(domain_hash);
+                SigningRequest {
+                    validator,
+                    signing_root,
+                    duty_data: att,
                 }
-            };
-
-            // Apply consensus result to this attestation
-            attestation.data_mut().beacon_block_root = data.block_root;
-            attestation.data_mut().source = data.source;
-            attestation.data_mut().target = data.target;
-
-            let signing_root = attestation.data().signing_root(domain_hash);
-            prepared.push(SigningRequest {
-                validator,
-                signing_root,
-                duty_data: (validator_index, pubkey, validator_committee_position, attestation),
-            });
-        }
+            })
+            .collect();
 
         // Collect signatures and assemble results
         let committee_validator_indices = self.get_committee_validator_indices(&committee_id);
@@ -1534,19 +1493,25 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> AnchorValidator
 
         let mut results = Vec::with_capacity(prepared.len());
         for req in prepared {
-            let ((validator_index, pubkey, validator_committee_position, mut attestation), signature) =
-                match req.resolve(&signatures) {
-                    Ok(resolved) => resolved,
-                    Err(validator) => {
-                        warn!(
-                            pubkey = ?validator.public_key,
-                            "Missing validator index or signature, skipping attestation"
-                        );
-                        continue;
-                    }
-                };
+            let (att, signature) = match req.resolve(&signatures) {
+                Ok(resolved) => resolved,
+                Err(pubkey) => {
+                    warn!(
+                        ?pubkey,
+                        "Missing validator index or signature, skipping attestation"
+                    );
+                    continue;
+                }
+            };
 
-            if let Err(e) = attestation.add_signature(&signature, validator_committee_position) {
+            let AttestationToSign {
+                validator_index,
+                pubkey,
+                validator_committee_index,
+                mut attestation,
+            } = att;
+
+            if let Err(e) = attestation.add_signature(&signature, validator_committee_index) {
                 error!(error = ?e, ?pubkey, "Failed to add signature to attestation, skipping");
                 continue;
             }
@@ -1562,17 +1527,15 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> AnchorValidator
     async fn sign_committee_aggregate_and_proofs(
         &self,
         committee_id: CommitteeId,
-        aggregates: Vec<AggregateToSign<E>>,
+        cluster: Cluster,
+        aggregates: Vec<(ValidatorMetadata, AggregateToSign<E>)>,
     ) -> Result<Vec<SignedAggregateAndProof<E>>, Error> {
-        let Some(first) = aggregates.first() else {
+        let Some((_, first)) = aggregates.first() else {
             warn!("sign_committee_aggregate_and_proofs called with empty aggregates");
             return Ok(vec![]);
         };
         let slot = first.aggregate.data().slot;
         let signing_epoch = first.aggregate.data().target.epoch;
-
-        // All validators in this committee share the same cluster
-        let (_, cluster) = self.get_validator_and_cluster(first.pubkey)?;
 
         // Get pre-built consensus data from AggregationAssignments (populated at 2/3 slot)
         let aggregation_assignments = self.get_aggregation_assignments(slot).await?;
@@ -1616,22 +1579,11 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> AnchorValidator
         let domain_hash = self.get_domain(signing_epoch, Domain::AggregateAndProof);
 
         // Prepare all validators: find each in decided data, decode aggregate, build message
+        // (metadata already resolved by `group_by_committee`)
         let mut prepared: Vec<SigningRequest<(u64, AggregateAndProof<E>)>> =
             Vec::with_capacity(aggregates.len());
 
-        for agg in aggregates {
-            let validator = match self.database.state().metadata().get_by(&agg.pubkey) {
-                Some(v) => v.clone(),
-                None => {
-                    warn!(?agg.pubkey, "Unknown pubkey while signing aggregate, skipping");
-                    validator_metrics::inc_counter_vec(
-                        &validator_metrics::SIGNED_AGGREGATES_TOTAL,
-                        &[metrics::OTHER_ERROR],
-                    );
-                    continue;
-                }
-            };
-
+        for (validator, agg) in aggregates {
             let validator_index = match validator.index {
                 Some(idx) => idx,
                 None => {
@@ -1765,9 +1717,9 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> AnchorValidator
         for req in prepared {
             let ((aggregator_index, message), signature) = match req.resolve(&signatures) {
                 Ok(resolved) => resolved,
-                Err(validator) => {
+                Err(pubkey) => {
                     warn!(
-                        pubkey = ?validator.public_key,
+                        ?pubkey,
                         "Missing validator index or signature for aggregator, skipping"
                     );
                     validator_metrics::inc_counter_vec(
@@ -2515,21 +2467,17 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> ValidatorStore
             >= Fork::Boole
         {
             // Boole+: group by committee, stream per committee via FuturesUnordered
-            let committee_mapping = self.group_by_committee(
-                aggregates,
-                |a| a.pubkey,
-                "aggregates",
-                &validator_metrics::SIGNED_AGGREGATES_TOTAL,
-            );
+            let _span = info_span!("sign_aggregate_and_proofs").entered();
+            let committee_mapping = self.group_by_committee(aggregates, |a| a.pubkey);
 
             let committee_futures: FuturesUnordered<_> = committee_mapping
                 .into_iter()
-                .map(|(committee_id, aggregates)| {
+                .map(|(committee_id, (cluster, aggregates))| {
                     let this = Arc::clone(self);
                     async move {
                         let num_aggregates = aggregates.len();
                         match this
-                            .sign_committee_aggregate_and_proofs(committee_id, aggregates)
+                            .sign_committee_aggregate_and_proofs(committee_id, cluster, aggregates)
                             .await
                         {
                             Ok(signed) => Ok(signed),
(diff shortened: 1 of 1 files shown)


petarjuki7 (dismissed):
Great, thanks, everything looks good! The only thing I would mention (non blocking for this, could be in a separate PR) is that all of these new improved "sign_*" functions all follow an identical pattern:

1. Extract slot/epoch from first item, get cluster
2. Run QBFT consensus once
3. Prepare validators loop (lookup metadata, skip with warn + metrics on failure)
4. Compute num_signatures_to_collect + data_hash
5. Call collect_committee_signatures
6. Assemble results loop (lookup signature, skip with warn + metrics on failure)

The consensus step (1-2) differs per duty type, but steps 3, 5, and 6 (validator preparation, signature collection, and result assembly) are nearly identical across all three.
It would maybe make sense to extract these into separate functions, so the functions become less crowded and every step is more clear - consensus, prepare, sign, assemble...
What do you think?

anchor/validator_store/src/lib.rs
```diff
@@ -2402,19 +2457,75 @@ impl<T: SlotClock, E: EthSpec> ValidatorStore for AnchorValidatorStore<T, E> {
...
+                            .await
+                        {
+                            Ok(signed) => Ok(signed),
+                            Err(e) => {
+                                error!(
+                                    ?committee_id,
+                                    error = ?e,
+                                    "Failed to sign committee aggregate and proofs"
+                                );
+                                Ok(Vec::new())
```
diegomrsantos: **[Concern] Metrics gap + log-level inconsistency for committee-wide errors**

When `sign_committee_aggregate_and_proofs` returns an `Err` (QBFT timeout, missing consensus data, etc.), this catch-all swallows the error and returns `Ok(Vec::new())`. No `SIGNED_AGGREGATES_TOTAL` metric is updated for the affected validators.

In the old code every aggregate went through `sign_single_aggregate_and_proof` → `run_and_update_metrics`, which incremented the per-validator metric with the appropriate label (`timeout`, `other_error`, etc.). Now committee-wide failures produce **zero** metric updates. The attestation path has a similar gap at the committee level, but attestation metrics are tracked downstream in `slashing_protection_attestations` — aggregates have no equivalent downstream tracking point.

Separately, this uses `error!` for all failures including timeouts. The existing convention in `run_and_update_metrics` uses `warn!` for timeouts ("other operators might be offline"). Using `error!` for timeouts could produce false alerts in monitoring.

Suggestion: consider incrementing `SIGNED_AGGREGATES_TOTAL` for each aggregate in the failed committee (matching the old per-validator behavior), and using `warn!` for `SpecificError::Timeout`.

anchor/validator_store/src/lib.rs
```diff
@@ -2402,19 +2457,75 @@ impl<T: SlotClock, E: EthSpec> ValidatorStore for AnchorValidatorStore<T, E> {
...
+        let Some(first) = aggregates.first() else {
+            return Either::Left(Either::Left(stream::once(futures::future::ready(Ok(
+                Vec::new(),
+            )))));
+        };
 
-            let results = join_all(futures).await;
+        if self
+            .fork_schedule
+            .active_fork(first.aggregate.data().target.epoch)
```
diegomrsantos, line 2455: **[Question] Fork determined from the first aggregate only**

Is it guaranteed that all aggregates in a single `sign_aggregate_and_proofs` call share the same target epoch? If a batch could straddle a fork boundary, some aggregates would be processed under the wrong fork path.

Likely safe since aggregates come from a single slot's duties, but worth confirming.

anchor/validator_store/src/lib.rs
```diff
@@ -2402,19 +2457,75 @@ impl<T: SlotClock, E: EthSpec> ValidatorStore for AnchorValidatorStore<T, E> {
...
     ) -> impl Stream<Item = Result<Vec<SignedAggregateAndProof<E>>, Error>> + Send {
-        let this = Arc::clone(self);
-        stream::once(async move {
-            let futures = aggregates.into_iter().map(|aggregate| {
-                let this = Arc::clone(&this);
-                async move { this.sign_single_aggregate_and_proof(aggregate).await }
-            });
+        // Early return for empty input —> no fork to determine, nothing to sign
+        let Some(first) = aggregates.first() else {
+            return Either::Left(Either::Left(stream::once(futures::future::ready(Ok(
```
diegomrsantos: **[Nit] Nested `Either` readability**

`Either::Left(Either::Left(...))` / `Either::Left(Either::Right(...))` / `Either::Right(...)` is functional but harder to parse than the two-way `Either` in `sign_attestations`.

Minor: an early return with `.boxed()` for the empty case, or a helper enum, could simplify this. Not blocking.

diegomrsantos (commented):
Architecture mirrors the attestation committee-batching pattern cleanly. Main concern is a metrics regression in the Boole+ error path — see inline comments.

diegomrsantos:
This PR copies the same structural pattern from `sign_attestations` / `sign_committee_attestations` almost line-for-line instead of extracting shared logic. The reviewer's approval already identifies steps 3, 5, and 6 as near-identical — that's exactly the kind of finding that should block merge, not be deferred to a follow-up that may never happen.

I've previously asked that PRs be self-reviewed for this kind of issue before requesting review. Copy-paste duplication across methods in the same file is something the author should catch before it reaches reviewers.

I'd like us to agree on a rule: if a reviewer identifies structural duplication during review, the refactor happens in the same PR, not "later."

diegomrsantos:
I don’t think we should keep extending validator-store batching changes on a “tests later” basis.
We already merged the attestation batching path without the targeted coverage requested in review, and the follow-up test issue is still open. This PR extends the same pattern into another production path while again deferring the direct tests.

If this gap is considered non-blocking, then we are effectively setting the standard that behavior-changing batching work in this area can merge without the targeted tests that prove it. I don’t think that is the right bar here.

I think this should either add the aggregate-path coverage now, or wait until the earlier batching-path tests land.

anchor/validator_store/src/lib.rs
shane-moore: Good question! And yep, lh call paths guarantee that all aggregates in a single `sign_aggregate_and_proofs` share same target epoch since lh will groups duties by  `committee_index` and then fetches an aggregated attestation per `slot, committee_index` where `slot` is for current slot

anchor/validator_store/src/lib.rs
shane-moore: yahh, I really wanted to do the `.boxed` for the early return to avoid the two way `Either` but doesn't work cuz the `impl Stream` resolves to a single concrete type, and `.boxed` returns `Pin<Box<dyn Stream>>`. Really don't like the two way `Either` though, so I played with it a bit more and will now return `Either::Right(FuturesUnordered::new())` for the early return case. then, we should be able to kill the either right/left after we don't need the alan logic anymore 🙌

shane-moore:
> I don’t think we should keep extending validator-store batching changes on a “tests later” basis. We already merged the attestation batching path without the targeted coverage requested in review, and the follow-up test issue is still open. This PR extends the same pattern into another production path while again deferring the direct tests.
> 
> If this gap is considered non-blocking, then we are effectively setting the standard that behavior-changing batching work in this area can merge without the targeted tests that prove it. I don’t think that is the right bar here.
> 
> I think this should either add the aggregate-path coverage now, or wait until the earlier batching-path tests land.

hey, yah, my thoughts were that we could make these `sign_*` pr's to `unstable` while @petarjuki7 is working on the integration test suite, which we would want to use for all these `sign_*` fn's. Just to keep development unblocked.  I think what you're saying is, "it was ok to merge one pr without tests that prove the new pattern to `unstable`, but let's not merge additional ones til the tests are done." I think this makes sense in the normal case. My idea was more like, "let's get these merged to `unstable`, so if ssv guys ask for another round of devnet testing, we can test all this logic there. and be working on the integration tests on the side in seperate PR. But I can tell you feel strongly about this, so I'm ok with holding off on merging this til we get the integration test foundations merged to `unstable` -> make commit to this pr to extend the tests for `sign_aggregate_and_proofs`

anchor/validator_store/src/lib.rs
shane-moore: makes sense, updated in https://github.com/sigp/anchor/pull/881/changes/570d0f589a81d836662ae2fb134f86eba3a19f18

shane-moore:
> This PR copies the same structural pattern from `sign_attestations` / `sign_committee_attestations` almost line-for-line instead of extracting shared logic. The reviewer's approval already identifies steps 3, 5, and 6 as near-identical — that's exactly the kind of finding that should block merge, not be deferred to a follow-up that may never happen.
> 
> I've previously asked that PRs be self-reviewed for this kind of issue before requesting review. Copy-paste duplication across methods in the same file is something the author should catch before it reaches reviewers.
> 
> I'd like us to agree on a rule: if a reviewer identifies structural duplication during review, the refactor happens in the same PR, not "later."

I agree with the rule. Duplication identified in review should be addressed in the same PR.

To clarify, I did consider doing this while building. I chose to defer extraction until all 4 `sign_*` functions were converted, so the helpers could be designed against the full set rather than prematurely abstracting based off just `sign_attestations`, which is all I had at the time. That said, I should have created a tracking issue, that's on me.

But now that PR https://github.com/sigp/anchor/pull/887 is merged, we have 3 of the 4 `sign_*` functions converted to the committee pattern, so I feel better about extracting shared helpers in this PR.

I did a thorough analysis of Petar's 6-step pipeline across all three inner functions. Here's what I found:

 Step 5 (collect signatures): Already shared -> `collect_committee_signatures` is the helper all `sign_committee_*` functions call. The ~10 lines of setup around it (`building validators_for_signing`, `computing num_signatures_to_collect`) vary by tuple shape and counting source, but each is only 3-4 lines and the accessor logic differs per function.

Step 6 (assemble results): The shared core is the index → signature resolution (two nested matches repeated in each
  function). I inlined this as a single let-else with `and_then` -> each call site is now one line
  (`validator.index.and_then(|idx| signatures.get(&idx).cloned())`) instead of two nested matches. The remaining
  assembly logic stays inline since it's duty-specific: different output types, different metrics, and different construction (3-8 lines per function).
  
Step 3 (prepare validators): This is where the functions genuinely diverge. Sync sigs: ~15 lines (shared signing root, simple lookup). Attestations: ~40 lines (mutate attestation data with consensus result, per-attestation signing
  root). Aggregates: ~130 lines (multi-step decided-data lookup + fork-dependent SSZ decode). The only shared piece is a 5-line metadata lookup, but the warn messages and metrics on failure differ per function.

I also looked at extracting the QBFT consensus call (step 2) since attestations and sync share nearly identical parameters, but it only applies to 2 of the 3 functions (aggregates use completely different instance IDs, consensus types, validators, and timeouts), and the consensus call is the most domain-important part of each function where hiding the parameters behind a helper reduces auditability.

Results:
- created `group_by_committee` in https://github.com/sigp/anchor/pull/881/changes/0ba963d6fcc085bad88c931e16140ed98b399b57 ->  eliminates ~15-20 lines of identical grouping logic from each outer function (×3 now, ×4 once contributions PR lands)

anchor/validator_store/src/lib.rs
```diff
@@ -1725,28 +1537,282 @@ impl<T: SlotClock, E: EthSpec> AnchorValidatorStore<T, E> {
...
+
+            results.push((validator_index, attestation, pubkey));
+        }
+
+        Ok(results)
+    }
+
+    /// Boole+ committee-based aggregate signing: single QBFT consensus per committee,
+    /// batch signature collection for all aggregators.
+    async fn sign_committee_aggregate_and_proofs(
```
diegomrsantos, line 1617: **[Concern] Structural duplication — prep and assembly loops are now tripled**

This method, `sign_committee_attestations`, and `sign_committee_sync_committee_signatures` all follow the same 5-phase pattern:

```
1. Get cluster from first item
2. Run QBFT consensus
3. Prepare validators (loop → DB lookup → skip if missing → build signing data)
4. Call collect_committee_signatures()
5. Assemble results (loop → get signature → skip if missing → push)
```

`group_by_committee` already shares phase 0 (grouping). `collect_committee_signatures` already shares phase 4. But phases 3 and 5 are fully duplicated across all three methods — the prep scaffolding (DB lookup per validator, skip-with-warn+metric) and the assembly scaffolding (index lookup, signature lookup, SUCCESS increment) are nearly identical.

The aggregate-specific logic (decided-data extraction, SSZ decoding by fork) is domain-specific and can't easily be shared — but the surrounding prep/assembly boilerplate could be.

Would you be open to creating an issue to track extracting shared `prepare_validators_from_db` and `assemble_results_from_signatures` helpers? Whether that belongs in this PR or a follow-up is a judgment call, but this is now three copies of the same pattern.

anchor/validator_store/src/lib.rs
```diff
@@ -228,6 +228,42 @@ impl<T: SlotClock, E: EthSpec> AnchorValidatorStore<T, E> {
...
+                        .entry(cluster.committee_id())
+                        .or_default()
+                        .push(item);
+                }
+                Err(Error::UnknownPubkey(pk)) => {
+                    warn!(
+                        ?pk,
+                        "Unknown pubkey while grouping {}, skipping", log_context
+                    );
+                    validator_metrics::inc_counter_vec(counter, &[metrics::OTHER_ERROR]);
```
diegomrsantos: **[Note] Behavioral change: new metrics on grouping failure for existing paths**

The old inline grouping for attestations and sync committee messages only logged warnings/errors on lookup failure — no metric increment. This new helper also increments `counter[other_error]` on every failure.

This is arguably an improvement (better observability), but it's an undocumented behavior change for two existing code paths (`sign_attestations` and `sign_sync_committee_signatures`). Could surprise anyone monitoring these metrics across releases.

Worth mentioning in the PR description.

anchor/validator_store/src/lib.rs
```diff
@@ -1725,28 +1537,282 @@ impl<T: SlotClock, E: EthSpec> AnchorValidatorStore<T, E> {
...
+        aggregates: Vec<AggregateToSign<E>>,
+    ) -> Result<Vec<SignedAggregateAndProof<E>>, Error> {
+        let Some(first) = aggregates.first() else {
+            return Ok(vec![]);
+        };
+        let slot = first.aggregate.data().slot;
+        let signing_epoch = first.aggregate.data().target.epoch;
+
+        // All validators in this committee share the same cluster
+        let (_, cluster) = self.get_validator_and_cluster(first.pubkey)?;
```
diegomrsantos: **[Note] Redundant database lookups per item (pre-existing pattern)**

Each aggregate goes through 3 DB lookups:
1. `group_by_committee` → `get_validator_and_cluster(pubkey)` — gets committee ID; **validator metadata discarded**
2. Here → `get_validator_and_cluster(first.pubkey)` — gets cluster
3. Prep loop below → `database.state().metadata().get_by(&agg.pubkey)` — gets validator metadata **again**

Same pattern exists for attestations and sync, so not a regression. But if `group_by_committee` returned the `Cluster` from the first item alongside the grouped items, lookups #2 and #3 could be eliminated. Could be part of the duplication cleanup above.

anchor/validator_store/src/lib.rs
```diff
@@ -1725,28 +1537,282 @@ impl<T: SlotClock, E: EthSpec> AnchorValidatorStore<T, E> {
...
+    }
+
+    /// Boole+ committee-based aggregate signing: single QBFT consensus per committee,
+    /// batch signature collection for all aggregators.
+    async fn sign_committee_aggregate_and_proofs(
+        &self,
+        committee_id: CommitteeId,
+        aggregates: Vec<AggregateToSign<E>>,
+    ) -> Result<Vec<SignedAggregateAndProof<E>>, Error> {
+        let Some(first) = aggregates.first() else {
```
diegomrsantos: **[Nit]** The other two committee methods (`sign_committee_attestations`, `sign_committee_sync_committee_signatures`) log a `warn!` when called with empty input. This one silently returns `Ok(vec![])`. Minor inconsistency — though in practice `group_by_committee` shouldn't produce empty groups.

anchor/validator_store/src/lib.rs
```diff
@@ -1541,31 +1368,16 @@ impl<T: SlotClock, E: EthSpec> AnchorValidatorStore<T, E> {
...
-                    );
-                    validator_metrics::inc_counter_vec(
-                        &validator_metrics::SIGNED_SYNC_COMMITTEE_MESSAGES_TOTAL,
-                        &[metrics::OTHER_ERROR],
-                    );
-                    continue;
-                }
+            let Some(signature) = validator
+                .index
+                .and_then(|idx| signatures.get(&idx).cloned())
```
diegomrsantos: **[Nit]** This `and_then` simplification merges two previously distinct failure cases — "validator has no index" vs "signature not found" — into a single log message. The old code had separate warnings for each case, which was useful for debugging. Same applies to the attestation assembly at line 1542.

Minor observability loss — both cases are equally rare and the action is the same (skip), so not blocking.

anchor/validator_store/src/lib.rs
shane-moore: Good call on the structural duplication. Pushed a commit addressing phases 4+5: https://github.com/shane-moore/anchor/commit/a449b604f30fdfaa81f67da5ae8e941b7854e514

  What changed:
  - Introduced `SigningRequest<T>` — a generic struct that standardizes the prepared data (validator, signing_root, duty_data) across all three methods.
  - Added `SigningRequest::resolve()` which handles the shared assembly-loop kernel: validator index lookup → signature lookup → return (duty_data, signature) or Err(validator) for logging. All three assembly loops now use this instead of duplicating the index/signature lookup logic.
  - `collect_committee_signatures` now takes `&[SigningRequest<D>]` directly (renamed to `collect_prepared_signatures`), so phases 4+5 flow naturally from the same prepared data.

  What I didn't extract (and why):

  - `prepare_validators_from_db` (phase 3): The shared kernel is ~5 lines (DB lookup + skip-on-miss). The domain-specific logic between DB lookup and SigningRequest construction varies too much — sync committee is ~10 lines, attestations ~20, and aggregates ~100+ lines (aggregator lookup, committee index positioning, SSZ decoding by fork, AggregateAndProof construction). Extracting a helper would require closures or trait impls that add more code than they save.
  - Full `assemble_results loop` (phase 5): `resolve()` shares the signature lookup pattern, but the post-signature assembly is genuinely different per method — attestations call `add_signature()`, aggregates wrap in `SignedAggregateAndProof`, sync committee builds `SyncCommitteeMessage`. A generic loop helper would need a closure or trait impl per method, which felt like it would trade duplication for indirection without a net win.

anchor/validator_store/src/lib.rs
shane-moore: certainly! added to description

anchor/validator_store/src/lib.rs
shane-moore: added the `warn` in https://github.com/sigp/anchor/pull/881/changes/a449b604f30fdfaa81f67da5ae8e941b7854e514

anchor/validator_store/src/lib.rs
```diff
@@ -228,6 +228,42 @@ impl<T: SlotClock, E: EthSpec> AnchorValidatorStore<T, E> {
...
     }
 
+    /// Group items by SSV committee, looking up each item's cluster.
+    ///
+    /// Items with unknown pubkeys or cluster lookup failures are logged and skipped.
+    fn group_by_committee<I>(
+        &self,
+        items: Vec<I>,
+        get_pubkey: impl Fn(&I) -> PublicKeyBytes,
+        log_context: &str,
```
diegomrsantos: **[Suggestion] `log_context` could be replaced by a tracing span on the caller**

This parameter exists so the generic helper can produce duty-specific log messages like `"Unknown pubkey while grouping attestations, skipping"` vs `"...aggregates, skipping"`. But in practice, the **pubkey** is the actionable diagnostic — knowing which duty type failed doesn't change debugging steps (you look up the pubkey, check if the validator was removed, check DB state).

If caller context in logs is valuable, the idiomatic `tracing` approach is a span on the caller:

```rust
// In sign_attestations:
let _span = tracing::info_span!("sign_attestations").entered();
let mapping = self.group_by_committee(attestations, |a| a.pubkey);

// In sign_aggregate_and_proofs:
let _span = tracing::info_span!("sign_aggregate_and_proofs").entered();
let mapping = self.group_by_committee(aggregates, |a| a.pubkey);
```

Then any `warn!` inside `group_by_committee` automatically inherits the caller's span in the log output — no parameter needed:
```
WARN sign_attestations: Failed to resolve committee, skipping pubkey=0xabc...
```

The codebase doesn't use spans widely today (only ~14 across the project), so this isn't urgent. But it's the right direction — and it would let `group_by_committee` drop this parameter, keeping the signature focused on its actual job (grouping).

Not blocking.

anchor/validator_store/src/lib.rs
diegomrsantos: Revisiting this after the `group_by_committee` API discussion above — I'd actually recommend removing the metrics from this helper and dropping the `counter` parameter.

My original note said "arguably an improvement," but thinking about it more: the old inline grouping had no metrics on failure, just logging. Adding `OTHER_ERROR` increments here is a behavioral change to two existing paths (attestations and sync messages) that's not the intent of this PR. It could show new bumps on dashboards after deploy for something that was previously silent.

The metrics also don't add much at this level — a coarse `OTHER_ERROR` that doesn't distinguish between failure reasons. The prep loops inside each `sign_committee_*` method already track per-validator errors with specific cases downstream.

This also ties into the `log_context` suggestion — removing both parameters gives a clean signature:
```rust
fn group_by_committee<I>(
    &self,
    items: Vec<I>,
    get_pubkey: impl Fn(&I) -> PublicKeyBytes,
) -> HashMap<CommitteeId, Vec<I>>
```

anchor/validator_store/src/lib.rs
shane-moore: luv this! cleaned up now

anchor/validator_store/src/lib.rs
shane-moore: really like the idea of using spans here, updated in https://github.com/sigp/anchor/pull/881/changes/bca0e5a60019d9837a4cd09a72fcc2312357fc97

anchor/validator_store/src/lib.rs
shane-moore: for sure, intentionally traded a small amount of diagnostic specificity for code conciseness

Both failure cases ("no index" and "signature not found") are equally rare, the action is the same (skip + warn + increment error metric), and the combined message is still sufficient to investigate — you'd check the pubkey and quickly see which case it was. Splitting them back out adds ~10 lines per call site for marginal debuggability

anchor/validator_store/src/lib.rs
```diff
@@ -2458,19 +2450,73 @@ impl<T: SlotClock, E: EthSpec> ValidatorStore for AnchorValidatorStore<T, E> {
...
+                                );
+                                for _ in 0..num_aggregates {
+                                    validator_metrics::inc_counter_vec(
+                                        &validator_metrics::SIGNED_AGGREGATES_TOTAL,
+                                        &[metrics::OTHER_ERROR],
+                                    );
+                                }
+                                Ok(Vec::new())
+                            }
+                        }
```
diegomrsantos: **[Concern] This 30-line error-handling closure is copy-pasted across committee methods**

The pattern — clone Arc, capture count, call method, match 3 error arms, increment metrics N times — is repeated nearly identically for aggregates (here), sync committee (L2753-2791), and attestations (similar). It buries the actual intent ("call the committee signer, handle failure gracefully") under boilerplate.

A small helper would collapse each closure from ~30 lines to ~5:

```rust
async fn run_committee_signing<T>(
    committee_id: CommitteeId,
    count: usize,
    counter: &LazyLock<validator_metrics::Result<IntCounterVec>>,
    fut: impl Future<Output = Result<Vec<T>, Error>>,
) -> Result<Vec<T>, Error> {
    match fut.await {
        Ok(signed) => Ok(signed),
        Err(Error::SpecificError(SpecificError::Timeout)) => {
            warn!(?committee_id, "Committee signing timed out");
            for _ in 0..count {
                validator_metrics::inc_counter_vec(counter, &[metrics::TIMEOUT]);
            }
            Ok(Vec::new())
        }
        Err(e) => {
            error!(?committee_id, error = ?e, "Committee signing failed");
            for _ in 0..count {
                validator_metrics::inc_counter_vec(counter, &[metrics::OTHER_ERROR]);
            }
            Ok(Vec::new())
        }
    }
}
```

Then each closure becomes:
```rust
.map(|(committee_id, (cluster, aggregates))| {
    let this = Arc::clone(self);
    async move {
        run_committee_signing(
            committee_id,
            aggregates.len(),
            &validator_metrics::SIGNED_AGGREGATES_TOTAL,
            this.sign_committee_aggregate_and_proofs(committee_id, cluster, aggregates),
        ).await
    }
})
```

The error policy is defined once, and each call site reads as intent rather than plumbing.

anchor/validator_store/src/lib.rs
```diff
@@ -1644,109 +1452,293 @@ impl<T: SlotClock, E: EthSpec> AnchorValidatorStore<T, E> {
...
+        };
+
+        let domain_hash = self.get_domain(signing_epoch, Domain::AggregateAndProof);
+
+        // Prepare all validators: find each in decided data, decode aggregate, build message
+        // (metadata already resolved by `group_by_committee`)
+        let mut prepared: Vec<SigningRequest<(u64, AggregateAndProof<E>)>> =
+            Vec::with_capacity(aggregates.len());
+
+        for (validator, agg) in aggregates {
```
diegomrsantos, line 1676: **[Concern] This prep loop is ~112 lines — extract the per-validator decided-data resolution**

`sign_committee_aggregate_and_proofs` is ~222 lines total, well over the ~50-line guideline. The main contributor is this loop (L1582-L1694) which does 7 sequential steps per validator: check index → find aggregator in decided data → get committee index → find position → get bytes → decode SSZ by fork → build message.

All of that is decided-data resolution logic that's independent of the surrounding orchestration. Extracting it into a helper would make the main function read as a clear sequence of phases:

```rust
fn resolve_decided_aggregate(
    &self,
    decided_data: &AggregatorCommitteeConsensusData,
    validator: &ValidatorMetadata,
    agg: &AggregateToSign<E>,
    domain_hash: Hash256,
) -> Result<SigningRequest<(u64, AggregateAndProof<E>)>, Error> {
    // Find aggregator, resolve committee index, decode SSZ, build message
    // ~60 lines of focused logic with proper error returns
}
```

Then the prep loop collapses to:
```rust
for (validator, agg) in aggregates {
    match self.resolve_decided_aggregate(&decided_data, &validator, &agg, domain_hash) {
        Ok(request) => prepared.push(request),
        Err(e) => {
            debug!(pubkey = ?agg.pubkey, error = ?e, "Skipping aggregate");
            validator_metrics::inc_counter_vec(
                &validator_metrics::SIGNED_AGGREGATES_TOTAL,
                &[metrics::OTHER_ERROR],
            );
        }
    }
}
```

The main function then reads at one abstraction level: setup → consensus → prepare → collect signatures → assemble. The low-level protocol logic (SSZ decoding, aggregator lookup, committee index resolution) lives in a focused helper that's easy to review independently.

anchor/validator_store/src/lib.rs
```diff
@@ -1345,40 +1116,118 @@ impl<T: SlotClock, E: EthSpec> AnchorValidatorStore<T, E> {
         .map(|signature| SignedContributionAndProof { message, signature })
     }
 
-    /// Sign a single aggregate and proof, handling fork-aware production and metrics.
+    /// Sign a single aggregate and proof (pre-Boole per-validator path).
     async fn sign_single_aggregate_and_proof(
```
diegomrsantos, line 1153: **[Concern] Four PR-touched functions exceed the ~50-line guideline — broader pattern**

Sizing the functions this PR introduced or rewrote:

| Function | ~Lines | Status |
|---|---|---|
| `sign_committee_aggregate_and_proofs` | ~228 | New |
| `sign_committee_attestations` | ~122 | Modified |
| `sign_single_aggregate_and_proof` | ~121 | Rewritten |
| `sign_committee_sync_committee_signatures` | ~110 | Modified |

The two inline comments above (`run_committee_signing` helper and `resolve_decided_aggregate` extraction) would fix the biggest offender. But the same readability issue applies to the other committee methods — they all mix orchestration (setup → consensus → prepare → collect → assemble) with low-level protocol logic in one long function body.

Concrete asks:
1. **`run_committee_signing`** — collapses the error-handling closure in `sign_aggregate_and_proofs` and `sign_sync_committee_signatures` (saves ~25 lines each)
2. **`resolve_decided_aggregate`** — extracts the 112-line prep loop from `sign_committee_aggregate_and_proofs`
3. **`sign_single_aggregate_and_proof`** — acceptable at 121 lines since it is the inlined alan logic that will be deleted when the alan fork is removed

After (1) and (2), the function sizes would be roughly:
- `sign_committee_aggregate_and_proofs`: ~60 lines (down from ~228)
- `sign_aggregate_and_proofs`: ~48 lines (down from ~73)
- `sign_sync_committee_signatures`: ~26 lines (down from ~51)
- `sign_committee_attestations` and `sign_committee_sync`: unchanged, but these were already closer to the boundary and are pre-existing

anchor/validator_store/src/lib.rs
shane-moore: removed, nice!

anchor/validator_store/src/lib.rs
shane-moore: makes sense! https://github.com/sigp/anchor/pull/881/changes/317f4ae360ddd3267c1bf2bfc566853d822a10e5

anchor/validator_store/src/lib.rs
shane-moore: && same here! https://github.com/sigp/anchor/pull/881/changes/317f4ae360ddd3267c1bf2bfc566853d822a10e5

anchor/validator_store/src/lib.rs
shane-moore: should be all good on these, hurray for smaller functions that are easier to reason about! 🥳

shane-moore:
@diegomrsantos   i feel like a lot of your findings could have been caught by `claude` while it was building if it had actually been taking the `rules` into account. but claude seems to pick and choose when it follows rules and otherwise. not sure if you've run into similar issues. i just asked my claude to create an agent hook that will spawn a subagent to actually review the main agent's edits against the rules. More expensive (LLM call per edit), but actually catches violations. Not sure if you do similar approach or different.  But let's see if that helps.

diegomrsantos:
Thanks for addressing all the feedback Shane — the `SigningRequest<T>`, `collect_prepared_signatures`, `resolve_decided_aggregate`, `run_committee_signing`, and the `group_by_committee` cleanup are all solid improvements.

That said, I'm still struggling to understand what the committee methods (`sign_committee_aggregate_and_proofs`, `sign_committee_attestations`, `sign_committee_sync_committee_signatures`) are doing when I read them. They're each 100-120 lines where everything sits at the same abstraction level — consensus setup, timer management, timeout construction, `decide_instance` arguments, `Completed` matching, validator prep, signature collection, result assembly, metric increments — all interleaved in one function body. I can't tell what the high-level flow is without reading every line.

Ideally this would be a separate PR, but as-is I can't really review the logic because I can't see the forest for the trees.

The ask isn't to make everything generic or reusable across duty types. It's just readability — I want to look at a function and immediately see:

```
1. Run consensus for this committee
2. Prepare signing requests from the decided data
3. Collect signatures
4. Assemble final results
```

Then I can drill into each step for details. Right now all four phases are flattened into one wall of code.

One approach: extract private methods that just **name the phases** — e.g. `run_consensus(...)`, `assemble_results(...)` — even if they're only called from one place and aren't reusable. The goal is that each committee method reads as a short sequence of named steps, not a sequence of low-level operations. Is there another way you'd prefer to tackle this?

diegomrsantos:
Thinking about this more — the reason the committee methods are hard to follow might not be something fixable with helpers alone. `AnchorValidatorStore` is a 3,000+ line struct with 80+ methods that owns everything: QBFT, signature collection, slashing protection, key management, fork scheduling, voting contexts, metrics. Every committee method has to reach into all of these concerns because they all live on `self`, so the functions end up as a flat sequence of low-level operations across unrelated subsystems.

The real fix is probably a deeper `AnchorValidatorStore` refactor — extracting a concept like a committee signing pipeline that encapsulates the shared orchestration. But that's clearly out of scope for this PR.

I can approve if you'd like to move forward — the functional changes look correct and the improvements you've made (`SigningRequest<T>`, `run_committee_signing`, `resolve_decided_aggregate`, etc.) are genuinely good. I just want to register that as a reviewer I struggle to understand what these committee methods are doing at a high level, and I think that's a sign we need to revisit the structure of this file at some point.

shane-moore:
> Thinking about this more — the reason the committee methods are hard to follow might not be something fixable with helpers alone. `AnchorValidatorStore` is a 3,000+ line struct with 80+ methods that owns everything: QBFT, signature collection, slashing protection, key management, fork scheduling, voting contexts, metrics. Every committee method has to reach into all of these concerns because they all live on `self`, so the functions end up as a flat sequence of low-level operations across unrelated subsystems.
> 
> The real fix is probably a deeper `AnchorValidatorStore` refactor — extracting a concept like a committee signing pipeline that encapsulates the shared orchestration. But that's clearly out of scope for this PR.
> 
> I can approve if you'd like to move forward — the functional changes look correct and the improvements you've made (`SigningRequest<T>`, `run_committee_signing`, `resolve_decided_aggregate`, etc.) are genuinely good. I just want to register that as a reviewer I struggle to understand what these committee methods are doing at a high level, and I think that's a sign we need to revisit the structure of this file at some point.

I'm with you. Difficult for me to reason about a specific piece without reading a few hundred lines.  do you think the initiative you're describing should be rolled into https://github.com/sigp/anchor/issues/896 or another clean up after that?

And sure sure, feel free to approve, thank you so much for spending the time helping with this

diegomrsantos:
I think it rolls into #896 — the readability issue here is a symptom of the same monolith problem that issue targets. Extracting a committee signing pipeline would naturally fall out of decomposing `AnchorValidatorStore` into focused components. No need for a separate issue.

diegomrsantos (dismissed):
Approving — thanks for all the iterations Shane. The functional changes look correct and the refactoring improvements are solid.

Readability concern with the committee methods stands and is tracked in #896.

shane-moore:
hey @petarjuki7 @diegomrsantos, please give another green checkmark on this then throw on the `ready-to-merge` label whenever you're online tomorrow. had to rebase `unstable` and lost the previous approval ha

then, @petarjuki7 can rebase his sync contributions pr off `unstable` and have claude analyze what changes were made to `sign_aggregate_and_proofs` and do same for sync contributions

diegomrsantos (approved):

Issue #880, opened 2026-03-12
Rework event processing around a single store boundary


diegomrsantos:
## Problem

The discussions in #735 and #841 are two symptoms of the same architectural problem: we do not have a single boundary that owns blocking SQLite execution, durable commit, in-memory `NetworkState` publication, and post-commit side effects.

Today `EventProcessor::process_logs` opens a transaction, processes logs one by one, calls database helpers during the batch, updates `last_processed_block`, and commits only at the end. During that window:

- database helpers mutate the shared watched state before commit
- notification behavior depends on whether a path uses silent `modify_state` or `send_modify`
- side effects can be emitted before commit
- async services still perform SQLite-backed writes directly, which is why #841 needed a local `spawn_blocking` workaround

That makes the system hard to reason about because there is no simple answer to these questions:

- when is a change durable?
- when is it visible through `watch<NetworkState>`?
- when is it valid to notify subscribers?
- when is it valid to run side effects?
- where should blocking DB work actually live?

## Why #735 and #841 are the same issue

In #735, the problem shows up as a correctness/coherence issue:
- shared in-memory state is updated during batch processing
- state publication is not cleanly tied to commit
- side effects can escape before commit

In #841, the same problem shows up as an execution-model issue:
- async code is still directly responsible for blocking SQLite writes
- so we need ad hoc `spawn_blocking` fixes at individual call sites

The `send_modify` / `modify_state` split and the extra `spawn_blocking` calls are not the root cause. They are consequences of the same missing store boundary.

## What a correct design needs

A single store boundary should own all of the following:

- where blocking SQLite work runs
- the atomic unit of persistence
- publication of committed in-memory state
- post-commit side effects

The invariant should be simple:

- before commit, changes are local only
- after commit, durable state and published in-memory state advance together
- side effects happen only from committed state

### 1. Keep batch-level atomicity

In this model:

- `EventProcessor` parses logs into batch-local domain changes
- later logs validate against a batch-local staged `NetworkState`
- `NetworkDatabase` owns the transaction internally
- committed state is published exactly once, after commit

This preserves the current batch/block-style processing model, but it is inherently more complex because later events in a batch need to see earlier uncommitted changes.

### 2. Simplify to event-by-event commits

In this model:

- each event validates against committed state
- each event commits in its own transaction
- `NetworkState` is updated only after commit
- side effects run only after commit

[...]

PR #885, opened 2026-03-12
fix: restore a post-commit boundary in event processing

diegomrsantos:

- Addresses #880.
- Related discussions: #735, #841.
- Event processing currently mixes durable DB writes, progress tracking, in-memory `NetworkState` publication, and downstream side effects across different layers.
- That makes it possible to reason about tentative state as if it were committed state, and it leaves sync progress too coarse to safely resume in the middle of a block.
- This is worth doing now because the current ownership model has already produced multiple correctness and maintainability issues, and local mitigations were diverging instead of fixing the boundary.

## Change Overview

- Keep RPC log fetching batched, but process and commit logs one event at a time.
- Add an exact processed-event cursor to metadata so sync can safely resume within a block.
- Move the durable-write boundary into the database layer so persisted state, processed-event progress, and in-memory publication advance in the right order.
- Update sync to resume from the finer-grained cursor and skip already committed logs from the same block.
- Keep slashing protection registration as the one pre-commit safety precondition for validator activation.
- Intentionally did not change event decoding, contract-facing semantics, or the existing test-only transaction helpers outside the production event/sync path.

- Main trade-off: more SQLite transactions during sync, in exchange for a much simpler and safer commit/publication model.
- Main data-plane risk: schema version increases to v4.
- Main runtime risk: sync resume semantics change from block-level to event-level.
- Mitigations:
  - log fetching remains batched for RPC efficiency
  - progress is now persisted with exact `(block_number, transaction_index, log_index)` information
  - same-block resume behavior is covered by targeted tests
  - the production event/sync path now publishes in-memory state only after commit

## Validation

- `cargo fmt --all`
- `cargo fmt --all --check`
- `cargo clippy -p database -p eth --all-targets -- -D warnings`
- `cargo test -p database -p eth --features database/test-utils`
- Added coverage for cursor persistence across restart and resuming within the same block without replaying already committed logs.

## Rollback

- Safe code rollback before merge: revert this PR.
- Operational caveat: this migrates the DB schema to v4. Rolling back to an older binary that does not understand schema v4 is not safe on an already-upgraded DB.
- If a rollback is needed after deploying a build with this change, use a compatible binary or restore the DB from a pre-migration backup.

## Blockers / Dependencies

- N/A

- The production path now uses the new boundary, but some legacy transaction-taking database helpers still exist for tests and non-production utility paths. They can be cleaned up separately once this behavior change lands.


diegomrsantos:
The earlier review guide is outdated after the history rewrite. The easiest way to review this PR now is to follow the 4-commit stack in order and focus on a small set of functions.

The core model is:

1. `EventProcessor` decides what one log means.
2. `NetworkDatabase` makes that event durable and persists the exact progress cursor in the same transaction.
3. Only after commit do we update `NetworkState` and notify watchers.
4. Once a whole fetched range succeeds, we collapse partial per-event progress back to a fully processed block.

If any production path violates that ordering, it is suspicious.

## 1. Database-owned commit boundary
Commit:
- `d4a6e60a9` `refactor(database): add post-commit event progress boundary`

Read these first:

- `anchor/database/src/lib.rs`
  - `mark_event_processed`: persists only the exact `(block, tx, log)` cursor for a committed or intentionally skipped event
  - `advance_processed_block`: persists the coarse “whole block completed” boundary and clears the finer cursor
  - `commit_db_update`: the core boundary of the PR; applies the SQL change, persists matching progress in the same transaction, commits, and only then publishes the corresponding in-memory state update
  - `apply_progress_to_tx` / `apply_progress_to_state`: mirror the same progress model in SQLite and `NetworkState`

- `anchor/database/src/state.rs`
  - `get_last_processed_event_from_db`: treats the three cursor columns as one logical value and rejects mixed NULL/non-NULL rows as corruption
  - `next_block_to_fetch`: defines resume semantics by preferring the partial cursor block when present so the same block can be re-fetched and already-committed logs skipped deterministically

- `anchor/database/src/operator_operations.rs`
  - `commit_operator_added`: commits operator insert + `max_operator_id_seen` + exact cursor together
  - `commit_seen_operator_id`: commits only “we have seen operator id N” plus the exact cursor for malformed/skipped operator events
  - `commit_operator_removed`: commits operator removal plus exact cursor

- `anchor/database/src/cluster_operations.rs`
  - `commit_validator_added`: commits nonce bump + validator/cluster/share insert + exact cursor together
  - `commit_owner_nonce`: commits nonce bump + exact cursor for malformed/skipped `ValidatorAdded`
  - `commit_cluster_status`: commits liquidation/reactivation status plus exact cursor
  - `commit_validator_removed`: commits validator removal plus exact cursor

- `anchor/database/src/validator_operations.rs`
  - `commit_fee_recipient_updated`: commits fee-recipient update plus exact cursor
  - `set_validator_indices`: commits index updates and then mirrors them into state
  - `update_graffiti`: same post-commit boundary, but without sync-progress changes

What to check:
- no shared state mutation before `tx.commit()`
- durable event change and durable progress always commit together
- `NetworkState` is only updated after commit
- processed block progress is monotonic and clears any partial cursor
- production writes go through the DB-owned commit boundary

## 2. Per-event processing model
Commit:
- `ba5fffa2d` `refactor(eth): process logs with per-event commits`

Read these next:

- `anchor/eth/src/event_processor.rs`
  - `process_logs`: skips already-committed logs before processing the fetched range
  - `process_logs_inner`: runs the range sequentially and only advances the full block boundary if the whole range succeeded
  - `process_single_log`: derives the exact cursor for one log and routes it
  - `dispatch_log`: only chooses the handler
  - `finish_processed_log`: the key function where cursor ownership becomes explicit:
    - handler already committed cursor
    - handler succeeded and caller must advance cursor
    - handler was skippable and caller advances cursor
    - handler was fatal
  - `mark_event_processed`: persists cursor-only progress when the caller owns advancement
  - `advance_processed_block_if_needed`: collapses partial in-block progress back to a full block boundary only when appropriate
  - `skip_processed_logs` / `log_at_or_before_cursor` / `cursor_for_log`: the replay/resume mechanics that make “re-fetch the same block, skip already-committed logs, continue” actually work

- `anchor/eth/src/util.rs`
  - `validate_operators`: validates operator-set structure and treats missing operators as invalid event data rather than internal DB failure

Then read the handlers with the most important semantics:

- `process_operator_added`: duplicate operator ids are skippable, malformed/duplicate operator payloads can still persist `max_operator_id_seen`, and later valid operator ids should not be blocked by a bad earlier one
- `process_validator_added`: malformed validator-add events still consume the owner nonce, duplicate validator-add events are skipped without letting the SQL unique constraint abort sync, slashing registration happens before the main DB commit and explicitly relies on idempotence, and fee recipient is no longer read before the write
- `process_validator_removed`: missing or mismatched validator state is treated as invalid event data rather than a local DB failure
- `process_validator_exited`: intentionally different from the others, because this is mostly a side-effect event, so it returns `NeedsCursorAdvance` and lets the caller mark progress afterwards

What to check:
- there is no long-lived outer DB transaction anymore
- one log is now the meaningful unit of progress
- cursor ownership is explicit for every handler outcome
- block progress is only advanced after the whole fetched range succeeds

## 3. Sync semantics and hardening
Commit:
- `ffad62c65` `fix(eth): harden per-event sync semantics`

Read:

- `anchor/eth/src/sync.rs`
  - `historical_sync`: fetching is still overlapped, but processing is serialized behind one `running_processor`; each batch is processed in `spawn_blocking(move || process_logs(...))`, and after a round it re-reads the committed resume point from DB instead of assuming `end_block + 1`
  - `live_sync`: computes `start_block` from the same `next_block_to_fetch` model, skips already-synced blocks correctly on reconnect/reorg, and immediately awaits the blocking event-processing work so there is backpressure

- `anchor/eth/src/event_processor.rs`
  - `advance_processed_block_if_needed`: refuses to regress the processed block boundary and only collapses to block progress when the whole range actually succeeded

This commit also contains the benchmark-driven correctness fixes that made the branch actually match the intended semantics:
- duplicate `ValidatorAdded`
- `OperatorAdded` malformed/duplicate handling preserving `max_operator_id_seen`
- re-reading committed sync progress instead of assuming the outer loop state

## 4. Tests and cleanup
Commit:
- `e5066390f` `refactor(database): align tests with commit boundary`

Read this last.

Highest-signal tests:
- `anchor/database/src/tests/state_tests.rs`
  - `test_processed_event_cursor_after_restart`: proves exact event cursor persists across restart and is cleared once the full block boundary is committed
- `anchor/eth/tests/integration.rs`
  - duplicate validator-added test: proves duplicate `ValidatorAdded` is skipped without re-queueing index sync and while still advancing nonce/progress
  - malformed/duplicate operator-added tests: prove bad operator history no longer blocks later valid operator ids and still preserves `max_operator_id_seen`
  - multi-event processing test: proves multiple events in one fetched range still produce the expected final block progress

What to check:
- restart preserves the exact event cursor
- replay in the same block skips already-committed logs
- malformed duplicate history no longer blocks later valid events
- tests now use the same committed API shape as production instead of older tx-only helpers

## Shortest useful path

If you only want the highest-signal path through the diff, I would read these in order:
1. `anchor/database/src/lib.rs`
2. `anchor/database/src/state.rs`
3. `anchor/database/src/operator_operations.rs`
4. `anchor/database/src/cluster_operations.rs`
5. `anchor/database/src/validator_operations.rs`
6. `anchor/eth/src/event_processor.rs`
7. `anchor/eth/src/sync.rs`
8. `anchor/database/src/tests/state_tests.rs`
9. `anchor/eth/tests/integration.rs`

## What to ignore on first pass

On a first review, I would ignore:
- most helper/test cleanup details
- small parsing helper refactors
- doc-only changes

The main review question for this PR is still:

does every production event now cross one DB-owned boundary where the durable event change and exact sync progress commit together, with all in-memory publication strictly after commit?

anchor/eth/src/event_processor.rs
```diff
@@ -306,45 +517,78 @@ impl EventProcessor {
...
+        };
 
         // Get the fee recipient if one has been stored, otherwise default to the owner address
-        let fee_recipient = match self.db.fee_recipient_for_owner(&owner, tx) {
-            Ok(Some(address)) => address,
-            _ => owner,
-        };
+        let fee_recipient = self
+            .db
+            .with_state(|state| state.fee_recipient_for_owner(&owner).unwrap_or(owner));
```
diegomrsantos: `ValidatorAdded` does not need `fee_recipient` to persist the validator. The write path only needs `cluster_id`, `owner`, validator metadata, and shares; `fee_recipient` lives in the `owners` table and is joined later when materializing a full `Cluster` on reads. Reading it here was forcing a read-model concern into the write path and is what made the earlier design more confusing than necessary.

diegomrsantos:
@claude review this PR

diegomrsantos:
@claude review this PR again

anchor/eth/src/event_processor.rs
```diff
@@ -306,70 +534,99 @@ impl EventProcessor {
...
         let cluster_id = compute_cluster_id(owner, &operatorIds);
         let operator_ids: Vec<_> = operatorIds.into_iter().map(OperatorId).collect();
 
         // Perform verification on the operator set and make sure they are all registered in the
         // network
-        validate_operators(&operator_ids, &cluster_id, &self.db.state())?;
+        let operators_valid = self
+            .db
+            .with_state(|state| validate_operators(&operator_ids, &cluster_id, state));
+        if let Err(err) = operators_valid {
```
diegomrsantos, line 573: @dknopik  @shane-moore I reverted `validate_operators` back to `ExecutionError::Database`, but this call site is still where the ambiguity matters. If required operators are missing from committed state, we know the validator event is not safely applicable, but we do not know whether that comes from malformed history or local state inconsistency. This PR is not meant to redefine that boundary, so I kept the old classification and wanted to call out the ambiguity explicitly.

anchor/eth/src/event_processor.rs
```diff
@@ -386,53 +643,59 @@ impl EventProcessor {
...
+            .with_state(|state| state.metadata().get_by(&validator_pubkey).cloned())
+        {
             Some(data) => data,
             None => {
                 debug!(
                     cluster_id = ?cluster_id,
                     "Failed to fetch validator metadata from database"
                 );
-                return Err(ExecutionError::Database(
+                return Err(EventActionError::Fatal(ExecutionError::Database(
```
diegomrsantos: @dknopik @shane-moore I reverted these missing-metadata and missing-cluster branches back to DB-style failures. From this code we cannot prove the `ValidatorRemoved` event is invalid; all we know is that committed local state is missing the validator state the event expects. That could still be malformed history, but it could also be local inconsistency, so treating it as skippable `InvalidEvent` felt out of scope for this PR.

diegomrsantos:
Closing in favor of the smaller replacement stack discussed on #880:
https://github.com/sigp/anchor/issues/880#issuecomment-4081328884

The problem is valid, but this PR is too large for the boundary fix and it introduces event-level resume machinery that we do not want to take forward. The replacement path is a smaller series from `unstable`: database prep first, then the block-scoped transaction fix, then cleanup.

PR #887, opened 2026-03-13
fix: sync committee message batch by committee

### LINKED ISSUE #883: Add ssv committee batching functionality to sign_sync_committee_signatures
Follow patterns as introduced in https://github.com/sigp/anchor/pull/834

petarjuki7:
## Issue Addressed
Follows up on #834 and closes #883 -> extends committee-based batching from attestations to sync committee messages.

## Proposed Changes

- Rewrote `sign_sync_committee_signatures` to group messages by committee and process each concurrently via `FuturesUnordered`, matching the pattern established for attestations in #834
- Added `sign_committee_sync_committee_signatures` batch path with single QBFT consensus per committee and batch signature collection
- All existing behavior preserved, just rearranged for committee-level batching

## Additional Info

- Mirrors the committee batching architecture from sign_attestations
- Companion PR for sync committee contributions #888

### DIFF
--- a/anchor/validator_store/src/lib.rs
+++ b/anchor/validator_store/src/lib.rs
@@ -93,7 +93,6 @@ const VALIDATOR_REGISTRATION_LOG_NAME: &str = "validator registration";
 const AGGREGATE_LOG_NAME: &str = "aggregate";
 const SELECTION_PROOF_LOG_NAME: &str = "selection proof";
 const SYNC_SELECTION_PROOF_LOG_NAME: &str = "sync selection proof";
-const SYNC_COMMITTEE_SIGNATURE_LOG_NAME: &str = "sync committee signature";
 const SYNC_COMMITTEE_CONTRIBUTION_LOG_NAME: &str = "sync committee contribution";
 
 pub struct AnchorValidatorStore<T: SlotClock + 'static, E: EthSpec> {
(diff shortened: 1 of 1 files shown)


shane-moore:
@claude

shane-moore:
> **Claude finished @shane-moore's task in 5m 19s** —— [View job](https://github.com/sigp/anchor/actions/runs/23061113622)
> 
> ### PR Review: Sync Committee Message Batching
> * [x]  Gather context and read PR diff
> * [x]  Read surrounding context and comparison code (attestation batching pattern)
> * [x]  Analyze changes for correctness, performance, and design
> * [x]  Post review feedback
> 
> Overall this is a clean refactor that correctly mirrors the committee-batching pattern established for attestations in #834. The structure is consistent, the single-QBFT-per-committee optimization is correctly applied, and error handling follows established patterns. I found one correctness issue and a couple of minor items:
> 
> ### Issues
> **1. `sign_sync_committee_signatures` swallows committee-level errors differently than `sign_attestations`** (`lib.rs:2731-2745`)
> 
> The new `sign_sync_committee_signatures` converts all errors from `sign_committee_sync_committee_signatures` into `Ok(Vec::new())`:
> 
> ```rust
> Err(e) => {
>     error!(?committee_id, error = ?e, "Failed to sign committee sync committee messages");
>     Ok(Vec::new())
> }
> ```
> 
> This means the stream **never yields an `Err`** — every item is `Ok(...)`. This is arguably fine since the old code also silently dropped per-validator errors via `filter_map(|r| r.ok())`, and the attestation path does the same (`lib.rs:2898-2901`). But it means downstream consumers cannot distinguish "committee timed out" from "committee had zero validators." If that's intentional for consistency with attestations, it's acceptable — just calling it out.
> 
> **2. Metrics regression: `run_and_update_metrics` no longer used for sync committee messages** (`lib.rs:1439-1585`)
> 
> The old `sign_single_sync_committee_signature` wrapped its entire operation in `run_and_update_metrics(SYNC_COMMITTEE_SIGNATURE_LOG_NAME, ...)`, which tracked fine-grained error categories (timeout, slashable, same-data, unregistered, etc.) via the centralized metric handler.
> 
> The new code does inline metric tracking only for `SUCCESS` and `OTHER_ERROR` paths. Specific error types like `Timeout`, `SameData`, and `Slashable` that previously got their own metric labels are now:
> 
> * **Timeout**: Propagated as `Err` from `sign_committee_sync_committee_signatures` and converted to `Ok(Vec::new())` in the caller — logged as a generic error, no dedicated `TIMEOUT` metric increment.
> * **SameData / Slashable / Unregistered**: Not possible in this path since there's no slashing protection for sync committee messages, so this is actually fine.
> 
> The `SYNC_COMMITTEE_SIGNATURE_LOG_NAME` constant was also removed. Consider whether the timeout case should still increment the `TIMEOUT` metric label for observability parity. The attestation path also doesn't use `run_and_update_metrics` at the committee level, so this may be intentional — but it's a behavior change worth noting.
> 
> [Add timeout metric tracking →](https://claude.ai/code?q=In%20anchor%2Fvalidator_store%2Fsrc%2Flib.rs%20in%20the%20sign_sync_committee_signatures%20method%20around%20line%202731%2C%20add%20a%20check%20for%20SpecificError%3A%3ATimeout%20in%20the%20error%20handler%20to%20increment%20the%20TIMEOUT%20metric%20label%20on%20SIGNED_SYNC_COMMITTEE_MESSAGES_TOTAL%20before%20returning%20Ok(Vec%3A%3Anew())&repo=sigp/anchor)
> 
> ### Observations (non-blocking)
> * The pattern is structurally identical to `sign_committee_attestations` — shared cluster lookup from first message, single QBFT, batch signature collection via `collect_committee_signatures`, and assembly loop. This is good for maintainability.
> * The `signing_root` is the same for all validators in the committee (since sync committee messages sign the beacon block root with the same domain), so passing it per-validator in `prepared` is slightly redundant but consistent with the attestation path where signing roots can differ. Not worth changing.
> * The `validator.index` `None` check at line 1544 is technically reachable even though `collect_committee_signatures` filters `None`-indexed validators internally — the `prepared` vec still contains those entries. The check is correct.
> 
> |

@claude
for 1, I think it's fine new since the new code logs at `error!` with `committee_id` context and has per validator metrics in the assembly loop
for 2, yah, I mentioned in the sister pr https://github.com/sigp/anchor/pull/888#issuecomment-4056592793 that we should make a gh issue to evaluate whether or not we need better committee level observability across all these `sign_*` methods

shane-moore (approved):

PR #888, opened 2026-03-13
fix: committee contributions batch by committee

### LINKED ISSUE #884: Add ssv committee batching functionality to sign_sync_committee_contributions
Follow similar patterns as introduced in https://github.com/sigp/anchor/pull/834

petarjuki7:
## Issue Addressed
Follows up on #834 and #881 and closes #884 -> extends committee-based batching from attestations/aggregates to sync committee contributions.

## Proposed Changes

- Rewrote `sign_sync_committee_contributions` to group contributions by committee (Boole+) and process each concurrently via `FuturesUnordered`
- Added `sign_committee_sync_committee_contributions` Boole+ batch path with single QBFT consensus per committee and batch signature collection
- All existing pre-Boole behavior preserved via `sign_single_sync_committee_contribution` which now only handles the Alan path
- Uses `collect_signature` directly (instead of `collect_committee_signatures`) because a validator can appear in multiple subcommittees with different signing roots

## Additional Info

- Mirrors the committee batching architecture from `sign_aggregate_and_proofs`
- Companion to sync committee message batching PR #887

### DIFF
--- a/anchor/validator_store/src/lib.rs
+++ b/anchor/validator_store/src/lib.rs
@@ -376,6 +376,52 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> AnchorValidator
         Ok(signatures)
     }
 
+    /// Run `AggregatorCommittee` QBFT consensus for a committee at 2/3 slot.
+    ///
+    /// Shared by `sign_committee_aggregate_and_proofs` and
+    /// `sign_committee_sync_committee_contributions`.
+    async fn run_aggregator_committee_consensus(
+        &self,
+        committee_id: CommitteeId,
+        slot: Slot,
+        cluster: &Cluster,
+        metric_label: &str,
+    ) -> Result<AggregatorCommitteeConsensusData<E>, Error> {
+        let aggregation_assignments = self.get_aggregation_assignments(slot).await?;
+        let our_consensus_data = aggregation_assignments
+            .get_consensus_data(&committee_id)
+            .ok_or(Error::SpecificError(SpecificError::ConsensusDataNotFound))?;
+
+        let timer = metrics::start_timer_vec(&metrics::CONSENSUS_TIMES, &[metric_label]);
+        let timeout_mode = TimeoutMode::SlotTime {
+            instance_start_time: self.get_instant_in_slot(
+                slot,
+                Duration::from_secs(self.spec.seconds_per_slot) * 2 / 3,
+            )?,
+        };
+
+        let completed = self
+            .consensus
+            .decide_instance(
+                AggregatorCommitteeInstanceId {
+                    committee: committee_id,
+                    instance_height: slot.as_usize().into(),
+                },
+                (*our_consensus_data).clone(),
+                Box::new(AggregatorCommitteeDataValidator::new()),
+                timeout_mode,
+                &cluster.cluster_members,
+            )
+            .await
+            .map_err(SpecificError::from)?;
+        drop(timer);
+
+        match completed {
+            Completed::TimedOut => Err(Error::SpecificError(SpecificError::Timeout)),
+            Completed::Success(data) => Ok(data),
+        }
+    }
+
     /// Compute the signing root for a sync committee selection proof.
     ///
     /// Each subnet has a different signing root based on `SyncAggregatorSelectionData{Slot,
@@ -1143,46 +1189,14 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> AnchorValidator
         let slot = first.contribution.slot;
         let epoch = slot.epoch(E::slots_per_epoch());
 
-        // Get pre-built consensus data from AggregationAssignments (populated at 2/3 slot)
-        let aggregation_assignments = self.get_aggregation_assignments(slot).await?;
-        let our_consensus_data = aggregation_assignments
-            .get_consensus_data(&committee_id)
-            .ok_or(Error::SpecificError(SpecificError::ConsensusDataNotFound))?;
-
-        // Run QBFT consensus once for the entire committee
-        let timer = metrics::start_timer_vec(
-            &metrics::CONSENSUS_TIMES,
-            &[metrics::SYNC_CONTRIBUTION_AND_PROOF],
-        );
-        let timeout_mode = TimeoutMode::SlotTime {
-            instance_start_time: self.get_instant_in_slot(
+        let decided_data = self
+            .run_aggregator_committee_consensus(
+                committee_id,
                 slot,
-                Duration::from_secs(self.spec.seconds_per_slot) * 2 / 3,
-            )?,
-        };
-
-        let completed = self
-            .consensus
-            .decide_instance(
-                AggregatorCommitteeInstanceId {
-                    committee: committee_id,
-                    instance_height: slot.as_usize().into(),
-                },
-                (*our_consensus_data).clone(),
-                Box::new(AggregatorCommitteeDataValidator::new()),
-                timeout_mode,
-                &cluster.cluster_members,
+                &cluster,
+                metrics::SYNC_CONTRIBUTION_AND_PROOF,
             )
-            .await
-            .map_err(SpecificError::from)?;
-        drop(timer);
-
-        let decided_data = match completed {
-            Completed::TimedOut => {
-                return Err(Error::SpecificError(SpecificError::Timeout));
-            }
-            Completed::Success(data) => data,
-        };
+            .await?;
 
         let domain_hash = self.get_domain(epoch, Domain::ContributionAndProof);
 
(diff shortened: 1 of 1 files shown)


shane-moore:
@claude

shane-moore:
> **Claude finished @shane-moore's task in 3m 38s** —— [View job](https://github.com/sigp/anchor/actions/runs/23061106121)
> 
> ### Code Review: `fix: committee contributions batch by committee`
> * [x]  Gather context and read PR diff
> * [x]  Review code changes in detail
> * [x]  Post review findings
> 
> **CI Status:** All 4 checks passing (Pull Request, test-suite, local testnet, Claude PR Review)
> 
> ### Summary
> The refactoring cleanly mirrors the committee-batching pattern from `sign_attestations`/`sign_committee_attestations`. The old per-validator Boole path (`produce_signed_contribution_and_proof_boole`) is correctly replaced with a committee-level function (`sign_committee_sync_committee_contributions`) that runs QBFT consensus once per committee and then collects signatures for all contributors.
> 
> Overall this is well-structured and correct. A few items worth noting:
> 
> ### Issues
> **1. Early failure on first validator lookup aborts entire committee** — `lib.rs:1379`
> 
> ```rust
> let (_, cluster) = self.get_validator_and_cluster(first.aggregator_pubkey)?;
> ```
> 
> If the first contribution in the `contributions` vec has an unknown pubkey, the `?` propagates the error and aborts the entire committee's contributions. This is inconsistent with the outer `sign_sync_committee_contributions` (line 2730-2748) which already filters unknown pubkeys during grouping. Since grouping already handles this, the first element should always be valid — but if a caller were to invoke `sign_committee_sync_committee_contributions` directly with a bad first element, it would fail the whole batch. This is a minor concern since the only call site pre-filters, but worth noting.
> 
> **2. Validator metadata re-fetched inside the batch loop** — `lib.rs:1427-1445`
> 
> The validator metadata was already fetched during grouping in `sign_sync_committee_contributions` via `get_validator_and_cluster`, but `sign_committee_sync_committee_contributions` re-fetches each validator from `self.database.state().metadata()` in the preparation loop. While this is functionally correct (and consistent with the per-validator old code), it means two database lookups per validator. Not a correctness issue, but a potential optimization — `ContributionToSign` could carry the `ValidatorMetadata` to avoid this. This mirrors how the existing pattern works, so acceptable for this PR.
> 
> **3. No metrics emitted for contributions skipped during grouping** — `lib.rs:2739-2748`
> 
> When a contribution is skipped during the grouping phase (unknown pubkey or cluster error), no `SIGNED_SYNC_COMMITTEE_CONTRIBUTIONS_TOTAL` metric counter is incremented. The `sign_committee_sync_committee_contributions` function correctly tracks metrics for validators skipped during preparation, but contributions dropped at grouping time are silently lost from metrics. Compare with the inner function which increments `OTHER_ERROR` for every skip case.
> 
> [Fix this →](https://claude.ai/code?q=In%20anchor%2Fvalidator_store%2Fsrc%2Flib.rs%20around%20line%202739-2748%2C%20add%20validator_metrics%3A%3Ainc_counter_vec%20calls%20for%20SIGNED_SYNC_COMMITTEE_CONTRIBUTIONS_TOTAL%20with%20OTHER_ERROR%20label%20in%20the%20Err%20branches%20of%20the%20grouping%20match%20statement%20in%20sign_sync_committee_contributions&repo=sigp/anchor)
> 
> ### Observations (non-blocking)
> * The comment at line 1518-1520 explaining why `collect_signature` is used instead of `collect_committee_signatures` is helpful and addresses a non-obvious design choice.
> * The `Either::Left(Either::Left(...))` / `Either::Left(Either::Right(...))` / `Either::Right(...)` pattern for the three return paths is the standard approach for `impl Stream` with multiple branches. It's a bit noisy but correct.
> * The `sign_single_sync_committee_contribution` method is now exclusively the Alan path — the doc comment update and simplification are clean.
> 
> |
@claude
1. No action needed. The upstream code pre-filters during grouping, so this is safe.

2. I'm ok with this since it's a lightweight fetch and prevents having to pass around more fn params

3. This PR can be updated to have same metrics handling as handled in https://github.com/sigp/anchor/pull/881

shane-moore:
To be clear, I'm not fully convinced we need more metrics here as I haven't thought about it enough, but I think the gh issue should be to evaluate if it makes sense to add soem and then to add them if so

anchor/validator_store/src/lib.rs
```diff
@@ -1490,35 +1338,20 @@ impl<T: SlotClock, E: EthSpec> AnchorValidatorStore<T, E> {
...
-                    contribution.selection_proof,
-                    slot,
-                    epoch,
-                    subcommittee_index,
-                    validator,
-                    cluster,
-                )
-                .await
-            }
+            self.produce_signed_contribution_and_proof_alan(
```
shane-moore: I would just inline the logic from  `produce_signed_contribution_and_proof_alan` here. too thin of a wrapper otherwise

shane-moore:
@petarjuki7, I'm converting this to draft til #881 is merged. Then, we can rebase this pr on it and use the new helper function `group_by_committee` (supposing it survives review) and add the validator specific failure metrics like in `sign_aggregate_and_proofs`

shane-moore:
looks like this was rebased from `unstable`. feel free to mark as ready for review when I should take another look 🤙

anchor/validator_store/src/lib.rs
```diff
@@ -2959,9 +2995,6 @@ impl<E: EthSpec> SignableBlock<E> for BeaconBlock<E, BlindedPayload<E>> {
     }
 }
 
-#[cfg(test)]
-mod testing;
```
shane-moore, line 2963: this perhaps got killed during rebase or something. think we need it so that the `testing/` integration tests will run?

anchor/validator_store/src/lib.rs
```diff
@@ -1318,6 +1124,207 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> AnchorValidator
...
+        contributions: Vec<(ValidatorMetadata, ContributionToSign<E>)>,
+    ) -> Result<Vec<SignedContributionAndProof<E>>, Error> {
+        let Some((_, first)) = contributions.first() else {
+            warn!("sign_committee_sync_committee_contributions called with empty contributions");
+            return Ok(vec![]);
+        };
+        let slot = first.contribution.slot;
+        let epoch = slot.epoch(E::slots_per_epoch());
+
+        // Get pre-built consensus data from AggregationAssignments (populated at 2/3 slot)
```
shane-moore: [Suggestion] Extract shared QBFT consensus setup with sign_committee_aggregate_and_proofs

  These ~35 lines (data fetch → decide_instance → Completed match) are identical to sign_committee_aggregate_and_proofs
   — same instance ID type, consensus data type, data validator, and 2/3-slot timeout. The only difference is the
  metrics label string.

```
 I tried the extraction locally — compiles clean, -15 net lines. Each call site drops from ~35 lines to 7:

  async fn run_aggregator_committee_consensus(
      &self,
      committee_id: CommitteeId,
      slot: Slot,
      cluster: &Cluster,
      metric_label: &str,
  ) -> Result<AggregatorCommitteeConsensusData<E>, Error> {
      let aggregation_assignments = self.get_aggregation_assignments(slot).await?;
      let our_consensus_data = aggregation_assignments
          .get_consensus_data(&committee_id)
          .ok_or(Error::SpecificError(SpecificError::ConsensusDataNotFound))?;

      let timer = metrics::start_timer_vec(&metrics::CONSENSUS_TIMES, &[metric_label]);
      let timeout_mode = TimeoutMode::SlotTime {
          instance_start_time: self.get_instant_in_slot(
              slot,
              Duration::from_secs(self.spec.seconds_per_slot) * 2 / 3,
          )?,
      };

      let completed = self
          .consensus
          .decide_instance(
              AggregatorCommitteeInstanceId {
                  committee: committee_id,
              },
              (*our_consensus_data).clone(),
              Box::new(AggregatorCommitteeDataValidator::new()),
              timeout_mode,
              &cluster.cluster_members,
          )
          .await
      drop(timer);

      match completed {
          Completed::TimedOut => Err(Error::SpecificError(SpecificError::Timeout)),
          Completed::Success(data) => Ok(data),
      }
  }

  Then both callers become:
  let decided_data = self
      .run_aggregator_committee_consensus(committee_id, slot, &cluster, metrics::SYNC_CONTRIBUTION_AND_PROOF)
      .await?;
```

anchor/validator_store/src/lib.rs
petarjuki7: Yeah, my fault

shane-moore (approved):

shane-moore:
lgtm, nice work breh!

Issue #856, opened 2026-03-02
feat: auto-generate CLI reference docs from clap definitions


diegomrsantos:
## Summary

We need a mechanism to auto-generate CLI reference documentation from clap struct definitions, so docs stay in sync with code changes. This replaces the manual approach and the previous `scripts/cli.sh` which uses `--help` output.

Closes the intent of #526 with a refined approach based on team discussion.

### 1. Use `clap_markdown` + slim post-processor

Instead of a custom renderer (~300 lines with edge cases), use [`clap_markdown`](https://crates.io/crates/clap_markdown) or a similar crate to handle the rendering, with a small post-processor to adjust the output for our MDX docs layout (table format, section grouping, etc.).

### 2. CI integration (the critical piece)

Add a CI check that:
- Runs the doc generator
- Compares output against the committed docs
- Fails if they differ

This ensures CLI option changes always come with a docs update. Similar to how [Lighthouse handles this](https://github.com/sigp/lighthouse/blob/unstable/scripts/cli.sh).

Update `scripts/cli.sh` and the [test-suite workflow](https://github.com/sigp/anchor/blob/stable/.github/workflows/test-suite.yml#L188-L201) accordingly.

### 3. Architecture

- The generator should **not** live in the `client` crate (avoids adding `keygen`/`keysplit` as runtime deps).
- Preferred location: a standalone binary crate, or alongside `main.rs` in `anchor/src/` where `keygen` and `keysplit` are already dependencies.

### 4. Scope of generation

- **Auto-generate**: options tables (flag names, descriptions, defaults) for all subcommands (`node`, `keygen`, `keysplit`), including global options (`--network`, `--data-dir`, etc.) and all `Node`-level flags.
- **Hand-written**: examples, prose descriptions, and supplementary sections should remain in the mdx files and not be overwritten by the generator.

## Prior Art

- PR #526 by @shubhexists — initial implementation with custom renderer
- Lighthouse's [CLI doc generation](https://github.com/sigp/lighthouse/blob/unstable/scripts/cli.sh)

## Acceptance Criteria

- [ ] `anchor docgen` (or equivalent) generates CLI reference docs from clap definitions
- [ ] CI check fails when generated docs don't match committed docs
- [ ] All subcommands and global flags are covered
- [ ] No incorrect flag names or types in generated output
- [ ] `keygen`/`keysplit` not added as dependencies of the `client` crate

PR #891, opened 2026-03-17
feat: add cli documentation generator and CI check for anchor docs

jnhsigmap:
**TODO: Implement `make cli-reference` and `make cli-reference-check` into CI/CD.**

## Issue Addressed
Issue #856.

The objective was to introduce an automated CLI reference doc generator that produces markdown tables from clap struct definitions

In addition a CI check that fails when CLI documentation is out of sync with the source code was desired.

## Proposed Changes

The following changes were included in this PR:
-   A standalone binary and library crate (`anchor-docgen`) that introspects the Anchor CLI tree at compile time and produces structured markdown documentation.
-   In `client` the FLAG_HEADER = "Flags" constant was replaced with semantic detail that allowed categorisation when `docgen` created `.mdx` files. A `help_heading` attribute was added to every visible field in the `Node` struct.
- Each crate that owned a CLI subcommand exports a `pub fn cli() -> clap::Command` function to allow introspection without booting the `anchor` application.

## Additional Info
I have added an integration test to ensure that every subcommand in `docgen::construct_anchor_cli_tree()` appears in `anchor --help`. This is a common sense check to try to catch instances where a CLI flag was added to the `anchor` binary and not to `docgen`.

Convenience `Makefile` targets
- `make cli-reference` runs `anchor-docgen` update to regenerate `cli-*.mdx` files given the current state of `anchor-docgen`. It overwrites the relevant sections in the doc files.
- `make cli-reference-check` is a check for consistency between the `cli-*.mdx` files and the CLI subcommands.


shane-moore:
directionally this looks good, but `docgen` is currently rebuilding a parallel top level cli instead of reusing the real one from `main.rs`. That creates a second source of truth: the binary can change while `docgen` stays “close enough” but wrong

I’d recommend moving top-level cli construction into one shared module/crate and having both the binary and docgen use that same `clap::Command`

Concretely, a small shared crate/module like `anchor/cli` would work well:

- define `Cli` and `AnchorSubcommands` there
- expose `pub fn command() -> clap::Command or use Cli::command()`
- use that from both the binary and docgen

just this change could be the first PR. basically just move the top-level cli construction into one shared module/crate and make the binary use it, with no doc generation yet

shane-moore:
I get that doing this all in one PR was a way to validate that the overall approach is valid. From a reviewer's perspective, I think we should consider whether the implementation can be split across multiple PR's that can individually be merged to `unstable`. Something like:
1. shared CLI extraction
2. docgen consuming the shared CLI
3. MDX generation/update flow
4. CI enforcement

But it is only safe to do them seperately if:
PR 1 preserves current binary behavior exactly
PR 2 does not yet become the required source for docs enforcement unless it is complete enough
PR 3 updates committed docs in the same PR
PR 4 is added only after generation is stable

In particular, I’d only split it if the first PR is a pure CLI refactor with no docgen behavior change, and later PRs layer docgen, MDX writing, and CI on top one at a time.  Perhaps this is not possible, but it's worth trying

jnhsigmap:
Thanks heaps for your feedback @shane-moore. I'm closing this PR as I will raise smaller ones with phased changes as we discussed :slightly_smiling_face:

Issue #878, opened 2026-03-10
Add targeted tests for committee-batching path introduced in #834


petarjuki7:
## Context

PR #834 introduced a new committee-batching path with significant behavioral changes concentrated in `anchor/validator_store/src/lib.rs` (specifically `sign_attestations`, `sign_committee_attestations`, and `collect_committee_signatures`). 
However, the diff does not include targeted test coverage for the new streaming / per-committee semantics. The existing tests don't lock in this new behavior, which leaves the door open for silent regressions during future cleanup or refactoring of `validator_store`.

## Proposed test cases

- Multiple SSV committees in one `sign_attestations()` call.
Assert that the stream yields separate batches per committee and that the test does not rely on completion order.
- Mixed success/failure across committees.
Assert that one failing committee does not prevent another committee from producing a successful batch.
- Committee attestation batching with existing sync-committee post-consensus collection contract.
Assert that the attestation path still uses the same` CollectionMode::Committee` / `num_signatures_to_collect` / base_hash semantics expected by the shared collector path.

diegomrsantos:
PR #897 covers all three proposed test cases and adds additional coverage beyond the issue scope:

**Additional tests not in the original proposal:**
- `sign_attestations_not_synced`: verifies the sync-gate early return at the top of `sign_attestations` produces a `NotSynced` error without touching consensus or the database.
- Full signing pipeline exercise: the mock consensus returns `Completed::Success` (not just timeout), so tests 1 and 2 now exercise the complete post-consensus path — applying consensus results to attestation data, collecting committee signatures, running slashing protection on a blocking thread, and assembling the final signed attestations. This covers the bulk of `sign_committee_attestations` which was previously untested.
- Test 1 also asserts that all `sign_and_collect` calls use `SignatureRequester::Committee` with correct parameters, verifying the integration between `sign_committee_attestations` and `collect_committee_signatures` end-to-end (not just when calling `collect_committee_signatures` directly).

Depends on #899 (`ConsensusDecider` trait) for the mock.

diegomrsantos:
Current unstable state: this looks implemented for the attestation path.

- The proposed multi-committee streaming coverage is present.
- The proposed mixed success/failure isolation coverage is present.
- The committee collection semantics are asserted via the committee requester checks.
- There is also additional not-synced coverage beyond the original issue scope.

So I think #878 is no longer tracking missing attestation-path test coverage on unstable; it is effectively pending promotion to stable and/or any explicitly remaining follow-up work.

PR #893, opened 2026-03-17
test: add ValidatorStore test harness and committee attestation test

petarjuki7:

- The `ValidatorStore` committee attestation path lacks integration test coverage, making it hard to verify correctness of committee grouping, QBFT consensus, and stream behavior during development
- Addresses #878
- This is the first of two PRs — introduces the test harness and one foundational test to keep review scope manageable
- Follow-up PR adds 3 additional tests (failure isolation, collection mode, not-synced guard)

## Change Overview

- Added `ValidatorStoreTestHarness` in `testing/common.rs` — wires up a real `AnchorValidatorStore` with in-memory database, QBFT consensus managers, signature collection, and manual slot clock
- Added one integration test verifying that `sign_attestations()` correctly groups attestations by `CommitteeId` and produces one stream item per committee
- Infrastructure supports multi-operator QBFT consensus routing for future tests

- Test-only code, no production changes beyond what was already merged
- The harness includes multi-operator consensus routing infrastructure that isn't exercised by this PR's single test — this is intentional to avoid duplicating setup code across PRs

## Validation

- `cargo test -p anchor_validator_store` — all 22 tests pass including the new one
- Test verifies stream produces exactly 2 items (one per committee) with single-operator QBFT timeout

## Rollback

N/A — test-only changes

## Blockers / Dependencies

N/A

- Follow-up PR: `feat/committee-attestation-tests-remaining` adds tests for failure isolation, committee collection mode, and not-synced early exit


anchor/validator_store/src/testing/common.rs
```diff
@@ -0,0 +1,820 @@
...
+/// Genesis duration in seconds for the manual slot clock.
+const GENESIS_DURATION_SECS: u64 = 0;
+
+/// Slot duration in seconds.
+const SLOT_DURATION_SECS: u64 = 12;
+
+// ==================== Key generation helpers ====================
+
+/// A set of BLS key shares for a single validator, split across operators.
+pub struct SplitKeySet {
```
shane-moore, line 71: looks like `SplitKeySet` is only used in this file so perhaps don't need `pub`?

anchor/validator_store/src/testing/common.rs
```diff
@@ -0,0 +1,820 @@
+#![allow(dead_code)]
```
shane-moore, line 1: I get that you're removing this `#![allow(dead_code)]` in #894.  Seems the idea here is to include all the infra in this PR so that the follow-up PR's are mostly just building out the test cases.  The fn's `start_consensus_router`, `RouterContext`, `message_router`, `route_consensus_message`, and `route_partial_signature` are the dead code it looks like. it's a little hard to review them in this pr without flipping over to #894 to see their call sites.  In the future, let's try to prevent dead code in PR's, even if they're about to immediately be used in a subsequent PR.  But in this situation, i don't think it's worth the effort to rip them out of this PR and moving them over to #894. Just thoughts for next time

diegomrsantos (changes requested):
## Summary

The intent is right — we need integration tests for the committee-batching path. But the approach introduces disproportionate complexity for what it verifies.

## Complexity vs value

The harness (`common.rs`) is ~820 lines. The single test is ~136 lines. The test asserts `results.len() == 2` — confirming that a `HashMap` grouping + `FuturesUnordered` fan-out yields one stream item per committee, both `Ok(empty)` because QBFT times out.

Of the 820 lines of harness:
- **~370 lines** are the multi-operator consensus router (`start_consensus_router`, `RouterContext`, `message_router`, `route_consensus_message`, `route_partial_signature`), which is **not used by this PR's test at all**. It simulates QBFT message routing between operators and other operators' partial signatures — infrastructure for future tests.
- The `#[allow(dead_code)]` at the top of `common.rs` suppresses warnings for this unused code.

Shipping ~370 lines of untested infrastructure that simulates an SSV network — ahead of the tests that would exercise it — means we're reviewing complex code with no way to verify its correctness.

## Root cause: tight coupling

The reason the harness is this large is that `AnchorValidatorStore::new()` takes 16 parameters, most of which are concrete types. Constructing one in tests requires wiring up the entire dependency chain: `SignatureCollectorManager` (which needs `Senders`, `MockMessageSender`, `ForkSchedule`), `QbftManager`, `NetworkDatabase` with operators/validators/shares, RSA encryption, BLS key splitting, etc.

The most impactful coupling is `SignatureCollectorManager` — it's the reason the consensus router needs to simulate partial signatures from other operators. PR #895 extracts a `SignatureCollecting` trait (35 lines, no behavioral change) that decouples this boundary. With a mock implementing that trait, tests can:

1. **Capture the `SignatureRequester`** passed to `sign_and_collect`, directly verifying `CollectionMode::Committee` receives the correct `num_signatures_to_collect` — no partial signature routing needed
2. **Return canned signatures or errors**, enabling failure isolation tests without simulating the full SSV network
3. **Remove the entire partial signature routing** (~200 lines of the consensus router)

The QBFT message routing part of the router would still be needed for tests that require multi-operator consensus, but those tests belong in the follow-up PR that actually exercises them.

## Specific suggestions

1. **Defer the consensus router to #894** where it's actually used. This PR should only contain the harness needed for its own test.
2. **Rebase on #895** (`SignatureCollecting` trait) and use a mock instead of `SignatureCollectorManager::new()`. This removes the need for processor queues and message senders for the signature collector.
3. **Remove `#[allow(dead_code)]`** — if code is unused, it shouldn't be in the PR.
4. The visibility changes (`pub(crate)` on `VotingContext` and `update_voting_context`) are fine.

anchor/validator_store/src/testing/common.rs
```diff
@@ -0,0 +1,820 @@
...
+/// Slots per epoch (mainnet default).
+pub const SLOTS_PER_EPOCH: u64 = 32;
+
+/// RSA key size in bits for test key generation.
+const RSA_KEY_SIZE: u32 = 2048;
+
+/// Genesis duration in seconds for the manual slot clock.
+const GENESIS_DURATION_SECS: u64 = 0;
+
+/// Slot duration in seconds.
```
shane-moore, line 65: can probably kill this comment

anchor/validator_store/src/testing/common.rs
```diff
@@ -0,0 +1,820 @@
...
+        );
+
+        // Populate the database with operators and validators from all committee setups
+        {
+            let mut conn = database
+                .connection()
+                .expect("database connection should succeed");
+            let tx = conn.transaction().expect("transaction should start");
+
+            let mut inserted_operators = std::collections::HashSet::new();
```
shane-moore, line 349: could just do `HashSet::new()` since it's already imported at the top of file

anchor/validator_store/src/testing/committee_attestation.rs
```diff
@@ -0,0 +1,136 @@
...
+async fn test_sign_attestations_produces_one_stream_item_per_committee() {
+    // Arrange
+    let rsa_private_key = generate_rsa_keypair();
+    let rsa_pubkey = rsa_public_from_private(&rsa_private_key);
+
+    let our_operator_id = OperatorId(1);
+
+    // Create 2 committees with distinct operator sets to guarantee different `CommitteeId`s.
+    // Committee A uses operators [1, 2, 3, 4], Committee B uses operators [1, 5, 6, 7].
+    let operator_ids_a: Vec<OperatorId> =
```
shane-moore, line 44: notice each test doing the generate private key -> create_committee_setup -> validatorstoretestharness::new, so lets use a bulder pattern instead:
```
Before (current):
  let rsa_private_key = generate_rsa_keypair();
  let rsa_pubkey =
  rsa_public_from_private(&rsa_private_key);

  let our_operator_id = OperatorId(1);

  let operator_ids_a = vec![OperatorId(1),
  OperatorId(2), OperatorId(3), OperatorId(4)];
  let operator_ids_b = vec![OperatorId(1),
  OperatorId(5), OperatorId(6), OperatorId(7)];

  let committee_a = create_committee_setup(
      &operator_ids_a,
      2,  // num_validators
      our_operator_id,
      &rsa_pubkey,
      0,  // starting_validator_index
  );
  let committee_b = create_committee_setup(
      &operator_ids_b,
      1,
      our_operator_id,
      &rsa_pubkey,
      100, // manually offset to avoid
  collision
  );

  let (executor, _signal) =
  create_test_executor();

  let harness = ValidatorStoreTestHarness::new(
      vec![committee_a, committee_b],
      our_operator_id,
      rsa_private_key,
      executor,
  );

  After (builder):
  let (executor, _signal) =
  create_test_executor();

  let harness = ValidatorStoreTestHarness::buil
  der(OperatorId(1), executor)
      .committee(&[1, 2, 3, 4], 2)  //
  operators, num_validators
      .committee(&[1, 5, 6, 7], 1)
      .build();

  The builder itself would look roughly like:

  pub struct HarnessBuilder {
      our_operator_id: OperatorId,
      executor: TaskExecutor,
      rsa_private_key:
  Rsa<openssl::pkey::Private>,
      rsa_pubkey: Rsa<openssl::pkey::Public>,
      committees: Vec<CommitteeSetup>,
      next_validator_index: usize,
  }

  impl HarnessBuilder {
      fn new(our_operator_id: OperatorId,
  executor: TaskExecutor) -> Self {
          let rsa_private_key =
  generate_rsa_keypair();
          let rsa_pubkey =
  rsa_public_from_private(&rsa_private_key);
          Self {
              our_operator_id,
              executor,
              rsa_private_key,
              rsa_pubkey,
              committees: Vec::new(),
              next_validator_index: 0,
          }
      }

      pub fn committee(mut self, operator_ids:
  &[u64], num_validators: usize) -> Self {
          let operator_ids: Vec<OperatorId> =
              operator_ids.iter().map(|&id|
  OperatorId(id)).collect();

          let setup = create_committee_setup(
              &operator_ids,
              num_validators,
              self.our_operator_id,
              &self.rsa_pubkey,
              self.next_validator_index,
          );
          self.next_validator_index +=
  num_validators;
          self.committees.push(setup);
          self
      }

      pub fn build(self) ->
  ValidatorStoreTestHarness {
          ValidatorStoreTestHarness::new(
              self.committees,
              self.our_operator_id,
              self.rsa_private_key,
              self.executor,
          )
      }
  }

  impl ValidatorStoreTestHarness {
      pub fn builder(our_operator_id:
  OperatorId, executor: TaskExecutor) ->
  HarnessBuilder {
          HarnessBuilder::new(our_operator_id,
  executor)
      }
  }
```

anchor/validator_store/src/testing/common.rs
shane-moore: closing in favor of Diego's comments below

diegomrsantos:
Superseded by #897, which has already been merged.

PR #894, opened 2026-03-17
test: add failure isolation, collection mode, and not-synced tests

petarjuki7:

- Continues test coverage for the committee attestation path started in #893
- Addresses #878
- Adds 3 remaining integration tests that exercise multi-operator consensus, signature batching, and error handling

## Change Overview

- **Test 2 (failure isolation):** A stuck committee doesn't block other committees from completing QBFT consensus and producing signed attestations
- **Test 3 (collection mode):** Partial signature messages use `DutyExecutor::Committee` and batch all validators' signatures into a single outgoing message
- **Test 4 (not-synced guard):** `sign_attestations` returns immediate `NotSynced` error without entering consensus

- Test-only code, no production changes
- Tests 2 and 3 exercise real multi-operator QBFT consensus with message routing between simulated operators

## Validation

- `cargo test -p anchor_validator_store` — all 25 tests pass (22 existing + 1 from PR 1 + 3 new)
- Tests 2 and 3 verify actual signed attestations are produced (not just timeout paths)

## Rollback

N/A — test-only changes

## Blockers / Dependencies

- **Depends on #893** — should be merged first. This PR's branch is based on #893's branch.


diegomrsantos (changes requested):
## Summary

This PR adds the 3 remaining integration tests (failure isolation, collection mode, not-synced guard) for the committee attestation path. The tests themselves verify important behavioral contracts, but the approach carries forward the complexity concerns from #893.

## What works well

- The test cases are the right ones — failure isolation, collection mode verification, and the not-synced guard are exactly the behavioral contracts that need automated coverage.
- Test 2 (failure isolation) genuinely needs multi-operator QBFT for one committee to succeed while another fails — that complexity is justified.
- Test 4 (not-synced) is simple and focused.

## What should change

### Test 3 (collection mode) can be dramatically simplified

Test 3 verifies that partial signature messages use `DutyExecutor::Committee` and batch all validators' signatures into a single outgoing message. This currently requires:
- Multi-operator QBFT consensus to complete (so signature collection is reached)
- The full consensus router to simulate partial signatures from other operators
- Capturing outgoing `SignedSSVMessage` instances and decoding them to verify the `DutyExecutor`

With the `SignatureCollecting` trait from PR #895, a mock can **directly capture the `SignatureRequester`** passed to `sign_and_collect`. The core assertion — that `CollectionMode::Committee` receives the correct `num_signatures_to_collect` — becomes:

```rust
let reqs = captured.lock().unwrap();
assert!(matches!(
    &reqs[0],
    SignatureRequester::Committee { num_signatures_to_collect: expected, .. }
));
```

No consensus router, no partial signature simulation, no message decoding. The mock replaces ~370 lines of routing infrastructure for this specific test.

### The consensus router is complex enough to have its own bugs

The `route_partial_signature` function (lines 986-1062 of common.rs across both PRs) simulates other operators signing the same roots and feeding results into the signature collector. This is ~80 lines of logic involving SSZ decoding, committee member lookup, BLS signing, and `PartialSignatureMessages` construction. It's untested infrastructure — if it has a bug, the integration tests it enables may pass or fail for the wrong reasons.

### Recommended approach

1. **Rebase on #895** (`SignatureCollecting` trait extraction)
2. **Simplify test 3** to use a mock instead of the full consensus router
3. **Keep the consensus router only for test 2** (failure isolation), where multi-operator QBFT is genuinely needed
4. This should reduce the combined diff of #893+#894 by ~400-500 lines while preserving the same behavioral coverage

## Dependency

This PR should wait for #895 (35-line trait extraction) to merge first, then rebase. The simplification is worth the short delay.

diegomrsantos:
Superseded by #897, which has already been merged.

PR #898, opened 2026-03-18
refactor(database): prepare tx-scoped block processing

diegomrsantos:

- PR 1 of the smaller replacement stack for the work previously attempted in #885.
- The immediate problem is that the database layer does not yet support a future block-scoped transaction boundary cleanly; DB writes and in-memory publication are still coupled at the method level.
- This is worth landing now because it lets the follow-up block-tx PR stay focused on the runtime boundary change instead of mixing that with additive database helper work.
- Evidence: the replacement direction is documented on issue #880: https://github.com/sigp/anchor/issues/880#issuecomment-4081328884
- Relevant links: #880, #885.

## Change Overview (Required)

- Add an explicit `PendingStateUpdates` staging object in `database` for deferred in-memory state replay.
- Add tx-scoped helpers for operator, cluster, validator, nonce, and block metadata updates/reads that the later block-tx PR will use.
- Keep existing wrapper methods behavior-preserving by continuing to apply or publish state the same way they did before.
- Add focused tests for the two key prep-only guarantees: tx-visible block number updates stay unpublished until explicit publish, and tx-visible own-operator resolution works before publication.
- Review commit-by-commit: commit 1 is the database-layer prep; commit 2 is the focused tests.
- Intentionally did not change: schema, `NetworkState` removal, `event_processor`, or runtime block processing behavior.

- Main risk: `publish_pending_state_updates()` is an explicit commit-boundary API and relies on caller discipline.
- Mitigation: existing wrapper methods preserve current behavior, and the new API is only additive in this PR.
- Trade-off: `PendingStateUpdates` keeps ordered replay instead of collapsing to a flat field struct.
- Rationale: later block-scoped processing can interleave different update kinds inside one transaction, so preserving order avoids hidden state divergence.

## Validation (Required)

- `cargo check -p database --lib`
- `cargo check -p database --tests --features test-utils`
- `cargo test -p database --lib --features test-utils tests::state_tests::state_database_tests::test_processed_block_tx_defers_publication_until_explicit_publish -- --exact`
- `cargo test -p database --lib --features test-utils tests::operator_tests::operator_database_tests::test_insert_operator_tx_reads_own_id_before_publish -- --exact`

- Safe to revert as a normal code rollback.
- No schema, config, or operational rollback steps.

- The next stacked PR should switch block processing to the new tx-scoped helpers and publish once at the explicit block boundary.

- This keeps PR 1 prep-only on purpose so the runtime behavior change can be reviewed separately.


shane-moore:
- `get_cluster_tx`
- `get_cluster_by_validator_tx` (only caller: itself calls `get_cluster_tx`)
- `has_own_share_tx`
- `get_validator_index_tx`
- `get_max_operator_id_seen_tx`

these seem unused. consider moving to follow-up PR?

anchor/database/src/operator_operations.rs
```diff
@@ -46,37 +47,35 @@ impl NetworkDatabase {
...
-                }
-            }
-            // Store the operator in memory
-            state
-                .single_state
-                .operators
-                .insert(operator.id, operator.to_owned());
-        });
+        let is_own_operator = match &self.operator {
+            PubkeyOrId::Pubkey(pubkey) => pem_key == pubkey.public_key_to_pem().unwrap_or_default(),
```
shane-moore: The original unwrap_or_default() was justified — this ran inside a `send_modify` closure that couldn't return errors. But insert_operator_tx returns `Result<(),DatabaseError>`, so the constraint no longer applies. This should now propagate the error properly, e.g. `.map_err(|e| DatabaseError::SQLError(format!("Failed to encode own RSA key: {e}")))?`

anchor/database/src/lib.rs
```diff
@@ -200,30 +437,89 @@ impl NetworkDatabase {
...
+            return Ok(Some(operator_id));
+        }
+
+        match &self.operator {
+            PubkeyOrId::Id(id) => Ok(Some(*id)),
+            PubkeyOrId::Pubkey(pubkey) => {
+                let encoded = BASE64_STANDARD.encode(
+                    pubkey
+                        .public_key_to_pem()
+                        .expect("Failed to encode RsaPublicKey"),
```
shane-moore: the reasoning for keeping `expect` is that:
`public_key_to_pem()` on a valid `Rsa<Public>` key is effectively infallible in practice — it's just serializing an already-parsed key back to PEM. If it fails, something is catastrophically wrong with OpenSSL. So `expect()` here is saying "this is a programmer error / impossible state, not a runtime condition."

could be nice if we had comments when we use `expect` in production code as to explain why we made that decision. i know this was pre-existing. just my thoughts so future devs understand why we accept `expect` in production paths situationally

anchor/database/src/lib.rs
```diff
@@ -133,6 +134,242 @@ enum PubkeyOrId {
...
+                    cluster.liquidated = status;
+                }
+            }
+            Self::DeleteValidator { validator_pubkey } => {
+                state.multi_state.shares.remove(&validator_pubkey);
+                let metadata = state
+                    .multi_state
+                    .validator_metadata
+                    .remove(&validator_pubkey)
+                    .expect("Data should have existed");
```
shane-moore: This `.expect()` is safe today — `process_validator_removed` pre-checks memory before calling `delete_validator`. Worth noting that future `direct _tx `callers won't have that guard, and `SQL DELETE` silently succeeds on missing rows. Might be a comment documenting the precondition that callers must ensure the validator exists in memory before staging a delete.

shane-moore:
you probably have already considered this but just flagging for the next pr just in case:
  `validate_operators` reads operatorsfrom in-memory state via `self.db.state()`. Once the follow-up switches to _tx variants with deferred publication, same-block. `OperatorAdded` + `ValidatorAdded` will fail here — the operator won't be in memory yet, so this check returns "One or more operators do not exist" and the valid event gets silently skipped. The same issue applies to `self.db.state().operator_exists()` (line 185) and `self.db.state().get_max_operator_id_seen()` (line 191). These reads will need to go through the transaction too.

diegomrsantos:
Addressed in `f1f713d90` by moving the PR 2-only tx reads out of PR 1. Thanks for the good catch on keeping the prep surface tighter here.

anchor/database/src/lib.rs
```diff
@@ -133,6 +134,242 @@ enum PubkeyOrId {
     Id(OperatorId),
 }
 
+#[derive(Debug)]
+struct InsertValidatorStateUpdate {
```
shane-moore: `lib.rs` nearly doubled (~350 → ~672 lines) StateUpdate, PendingStateUpdates, and InsertValidatorStateUpdate are a distinct responsibility from NetworkDatabase connection/pool management. Consider extracting to pending_updates.rs in the follow-up — keeps apply logic reviewable in isolation as more update types land.

anchor/database/src/operator_operations.rs
diegomrsantos: Thanks, good catch. Addressed in `bbc5a04f6` by propagating PEM serialization failures through `DatabaseError` in the new tx path instead of keeping `unwrap_or_default()`.

anchor/database/src/lib.rs
diegomrsantos: Thanks, good catch. I kept PR 1 behavior unchanged and added a short comment in `c76716425` documenting that this staged delete still relies on the existing in-memory precheck before we hit the `expect`.

anchor/database/src/lib.rs
diegomrsantos: Agreed. Since this `expect` was newly introduced in PR 1, I removed it rather than documenting it. `get_own_operator_id_tx` now uses the same `DatabaseError` path as `insert_operator_tx` in `bbc5a04f6`.

anchor/database/src/lib.rs
diegomrsantos: Thanks, good catch. Addressed in `3027197aa` by extracting `PendingStateUpdates` and `StateUpdate` into `pending_updates.rs` so the apply logic is easier to review in isolation.

diegomrsantos:
@shane-moore on your point here: https://github.com/sigp/anchor/pull/898#issuecomment-4091694534, I agree with the follow-up concern. PR 1 is intended to stay prep-only and preserve existing behavior; it does not switch `event_processor` to deferred publication yet. The actual behavior change and fix will land in the next PR, where those same-block reads move to transaction-aware reads in the same change.

If you are happy with PR 1 staying at that boundary, could you approve it and resolve the comments you are satisfied with?

shane-moore (approved):

diegomrsantos:
@claude do a final review

Issue #896, opened 2026-03-18
refactor: decompose AnchorValidatorStore into focused components


diegomrsantos:
## Context

`AnchorValidatorStore` is a 3200-line monolithic struct with 16 constructor parameters serving 5 distinct responsibilities. This makes it difficult to test in isolation — constructing one in tests requires wiring up the entire dependency chain (QBFT manager, signature collector, database with operators/validators/shares, RSA key encryption, slashing DB, processor queues, message senders, etc.).

PRs #893 and #894 demonstrate this cost: ~1400 lines of test harness to simulate an SSV network, primarily to satisfy the constructor's dependency requirements.

## Goal

Incrementally decompose `AnchorValidatorStore` into focused components, each independently testable. Each step should be a small, independently reviewable PR.

## Current responsibilities (5 concerns in one struct)

1. **Duty context management** — watch channels for `VotingAssignments`, `VotingContext`, `AggregationAssignments`
2. **Key management** — RSA decryption of BLS shares, LRU caching of decrypted keys
3. **Consensus + signature collection pipeline** — QBFT consensus → partial signature collection → reconstruction (duplicated per duty type)
4. **Slashing protection** — attestation/block slashing checks, DB pruning
5. **Configuration/plumbing** — database, slot clock, spec, fork schedule, MEV config, sync state

### Step 1: Trait extraction at dependency boundaries
Extract traits for concrete dependencies so tests can use mocks instead of real infrastructure.

- [x] `SignatureCollecting` trait — #895 (merged or in review)
- [ ] `ConsensusDecider` trait over `QbftManager` — extract `decide_instance`. Removes the need for processor/message sender in tests that don't need real consensus.
- [ ] Database read trait — the store only calls `state().shares().get_by()` and `state().validators()`. A small read trait decouples from `NetworkDatabase`.

Each is a ~30-line PR. Together they make `AnchorValidatorStore` constructible in tests with just mocks.

### Step 2: Extract `KeyStore`
The `decrypted_keys` LRU cache + `private_key` + `decrypt_key_share` is a cohesive unit used only by `collect_signature`. Extract into a `KeyStore` struct:

```rust
pub struct KeyStore {
    decrypted_keys: Mutex<LruCache<[u8; ENCRYPTED_KEY_LENGTH], SecretKey>>,
    private_key: Option<Rsa<Private>>,
}

impl KeyStore {
[...]

PR #899, opened 2026-03-18
refactor: extract ConsensusDecider trait for testability

diegomrsantos:

- Testing the signing pipeline in `AnchorValidatorStore` currently requires a real `QbftManager`, which needs a full networking stack and slot clock. This makes integration tests slow and complex.
- Extracting a `ConsensusDecider` trait allows future tests to swap in a mock that returns pre-determined consensus results, enabling focused testing of the signing logic without real QBFT.
- This is step 1 of #896.

## Change Overview

- **`qbft_manager`**: Added a `ConsensusDecider<E>` trait with a single method `decide_instance`, plus a blanket implementation for `QbftManager<E, S>` that delegates to the existing method.
- **`validator_store`**: Added a third type parameter `C: ConsensusDecider<E>` with default `= QbftManager<E, T>`, so all existing code using `AnchorValidatorStore<T, E>` compiles unchanged. Renamed the internal field from `qbft_manager` to `consensus` and updated all 7 call sites.
- **`client`**: Updated one turbofish from `<_, E>` to `<_, E, _>` to account for the new type parameter.

The trait uses static dispatch (`impl Future` return) rather than `dyn` because `decide_instance` is generic over `D: QbftDecidable<E>`, which prevents object safety. This differs from `SignatureCollecting` (which uses `dyn`) for that technical reason.

- No behavioral changes. All runtime code paths are identical.

- **Low risk**: The default type parameter means all downstream code is unchanged. The only non-test caller (`client/src/lib.rs`) just needs a `_` in the turbofish.
- **Trade-off**: Static dispatch means test mocks must also be generic-compatible, but this is unavoidable given the generic `D` parameter on `decide_instance`.

## Validation

- `cargo clippy -p anchor_validator_store --tests -- -D warnings` passes
- `cargo clippy -p qbft_manager --tests -- -D warnings` passes
- `cargo test -p anchor_validator_store` passes
- `cargo test -p qbft_manager` passes
- Full `make lint` passes

## Rollback

- Pure refactor with no runtime behavior change. Safe to revert at any time.

## Next Steps

- Step 2 (#896): Add a `MockConsensusDecider` and use it in `validator_store` integration tests to test signing pipelines without real QBFT.

### DIFF
--- a/anchor/client/src/lib.rs
+++ b/anchor/client/src/lib.rs
@@ -551,7 +551,7 @@ impl Client {
         // Spawn the network listening task
         executor.spawn(network.run::<E>(), "network");
 
-        let validator_store = AnchorValidatorStore::<_, E>::new(
+        let validator_store = AnchorValidatorStore::<_, E, _>::new(
             database.clone(),
             Box::new(signature_collector),
             qbft_manager,
--- a/anchor/qbft_manager/src/lib.rs
+++ b/anchor/qbft_manager/src/lib.rs
@@ -1,4 +1,4 @@
-use std::{fmt::Debug, hash::Hash, num::NonZeroU64, sync::Arc};
+use std::{fmt::Debug, future::Future, hash::Hash, num::NonZeroU64, sync::Arc};
 
 use bls::PublicKeyBytes;
 use dashmap::DashMap;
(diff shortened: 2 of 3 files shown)


shane-moore (approved):
i'm liking the trait abstractions to make the codebase more testable

PR #901, opened 2026-03-19
feat: introduces cli crate and migrates cli definitions to commonly accessible module

jnhsigmap:
This is the first of several phased changes to address issue #856. The original PR is #891, however team discussions lead to a preference for smaller phased changes that are easier to review and reason with.

- What real problem does this solve, or what real value does it add now?
The objective is to use the anchor CLI tree in other crates without having to redefine it in the caller module. This change allows usage of a single definition by other crates by moving it from `src/main.rs` to a new module `cli`.

- Why is this worth doing now?
It's required to eventually create a standalone CLI documentation generator that depends on all `clap` CLI components that are defined in the anchor project.

- What evidence supports that?
n/a
- Relevant links: issue, discussion, spec, related PRs.
n/a

## Change Overview (Required)

**New crate: `anchor/cli`**:
   - Extracts `Cli`, `AnchorSubcommands`, version statics, and `get_color_style()` from `anchor/src/main.rs`. Functions previously just used by `src/main.rs`.
   - Moved the `client/src/cli.rs` module to `cli` crate and instead exported what other crates currently depend on.
   - `build.rs` ensures `OUT_DIR` is available for `build_profile_name()`. From the stack overflow post that gave us the idea :slightly_smiling_face: 

This touches the anchor CLI interface, so it could break the whole app. It is also easy to spot where problems arise as the change is relatively simple.
Adding another crate that depends on keysplit, keygen, and some minor components is paid in return for a commonly available CLI struct that can be used everywhere without large unnecessary installs.
By incrementally building changes toward the desired end state.

## Validation (Required)

Current tests pass and tested this end-to-end with all phases included and the documentation and doc generator output is as expected. 
- Examples: targeted tests, manual repro, benchmark/profiling result, before/after behaviour, spec-parity check.

## Rollback (Required for behaviour or runtime changes; optional otherwise)


anchor/cli/build.rs
```diff
@@ -0,0 +1,5 @@
+fn main() {
```
shane-moore, line 1: we should be able to kill `anchor/build.rs` now since the `build_profile_name` was moved to this crate

shane-moore:
The CLI crate extraction looks good and is what we discussed in [#891](https://github.com/sigp/anchor/issues/891). One thing I'd like to change: can we drop the semantic help headings from this PR and keep `--help` output identical to what it is today?

  The heading reorganization (replacing `FLAG_HEADER` with `SECURITY_OPTIONS, EXTERNAL_APIS`, etc. and adding `next_help_heading` to each struct) changes user facing `--help` output. For Phase 1, I'd prefer this to be a pure refactor with no visible behavior change.

  For the doc generator, grouping can be handled in the post-processor -> which is what [#856](https://github.com/sigp/anchor/issues/856) describes: "a small
  post-processor to adjust the output for our MDX docs layout (table format, section grouping, etc.)". The structs
  already have natural grouping via flatten boundaries (SecurityOptions, ExternalApis, etc.) that the generator can use without changing the clap source definitions.

If we do want to improve `--help` with semantic headings, that's perhaps a worthwhile change on its own that i haven't thought about. maybe @diegomrsantos has an opinion. but regardless would be seperate PR so it can be evaluated independently

anchor/cli/src/lib.rs
```diff
@@ -0,0 +1,69 @@
+use std::sync::LazyLock;
+
+use clap::{
+    Parser,
+    builder::{Styles, styling::AnsiColor},
+};
+use client::Node;
```
shane-moore: The dependency direction here is backwards -> the CLI crate depends on the application crate (client) just for Node. This means cli inherits all 45 of client's dependencies (network, database, qbft, processor, etc.), none of which Node actually uses.

Node and its sub-structs (the entire `client/src/cli.rs` — `SecurityOptions`, `ExternalApis`, `HttpApiOptions`, `MetricsOptions`, `NetworkOptions`, `PayloadBuildingOptions`, heading constants) should move into this crate. Then client depends on cli for Node, not the reverse.

The change is mechanical:
  - `cli/Cargo.toml`: drop client, add axum, beacon_node_fallback, logging
  - `client/Cargo.toml`: add cli
  - `client/src/config.rs`: use cli::{NetworkOptions, Node} instead of use
  crate::cli::

jnhsigmap: cool thanks :slightly_smiling_face: to recap:
- `build.rs` no longer required at top level.
- `cli/src/lib.rs` exports `Node` and `NetworkOptions` used by `client`, and `Cli` and `AnchorSubcommands` used by `src/main.rs`.
- `cli` no longer depends on `client`.
- `cli` depends on `logging`, `beacon_node_fallback` and `axum`. Things not required removed from `client`.
- `client/src/cli.rs` moved to `cli` crate and `client` crate now depends on `cli`.

Hope that's not a handful :sweat_smile:

anchor/logging/Cargo.toml
```diff
@@ -6,8 +6,8 @@ authors = ["Sigma Prime <contact@sigmaprime.io>"]
 
 [dependencies]
 chrono = { version = "0.4", default-features = false, features = [
-  "clock",
-  "std",
+    "clock",
```
diegomrsantos: Nit: This indentation change (2-space → 4-space) is unrelated to the CLI crate extraction. Consider removing it from this PR to keep the diff focused on the stated goal.

Cargo.toml
```diff
@@ -128,7 +130,7 @@ libp2p = { version = "0.56", default-features = false, features = [
     "dns",
     "metrics",
     "gossipsub",
-    "upnp"
+    "upnp",
```
diegomrsantos: Nit: This trailing-comma addition is unrelated to the CLI crate extraction. Consider removing it to keep the diff scoped to the PR goal.

anchor/cli/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+name = "cli"
+version = "0.1.0"
+edition = { workspace = true }
+authors = ["Sigma Prime <contact@sigmaprime.io>"]
+
+[dependencies]
+axum = { workspace = true }
+beacon_node_fallback = { workspace = true }
+clap = { workspace = true }
+ethereum_hashing = "0.7.0"
```
diegomrsantos: Minor: This carries over the bare version spec from the old `anchor/Cargo.toml`. Since `ethereum_hashing` isn't in `[workspace.dependencies]` yet, this would be a good time to add it there and switch to `ethereum_hashing = { workspace = true }` for consistency with the other deps in this file.

diegomrsantos:
**Title mismatch**: The PR title says "semantic help headings" but the final diff contains no help heading changes — `cli.rs` is moved with `similarity index 100%`. The first commit mentioned headings, but a later commit ("client depends on cli and no --help changes") appears to have removed them. Consider updating the title to just `feat: introduce cli crate`.

diegomrsantos:
**Description accuracy**: The PR description states "client depends on cli with no large transient dependencies ending up in users of cli." However, the `cli` crate depends on `keygen`, `keysplit`, `beacon_node_fallback`, `axum`, and `logging` — which are not lightweight. This is inherent to the design (the CLI types reference `Keysplit`, `Keygen`, `HeaderValue`, `FileLoggingFlags`, `ApiTopic`), but the description could be corrected to avoid misleading reviewers.

diegomrsantos:
Thanks for the clean extraction, Jason! Only minor comments from my side.

I agree with Shane's earlier feedback — keeping Phase 1 as a pure structural refactor with no `--help` output changes is the right approach.

anchor/cli/Cargo.toml
jnhsigmap: cool thanks @diegomrsantos ! I've pushed changes for these

shane-moore:
pr looks overall in good shape! just merged https://github.com/sigp/anchor/pull/907 to handle the cargo deps vulnerability fix since its out of scope for this pr. hopefully if you rebase on `unstable`, it'll remove the vulnerability fix diffs from this pr, else killing your last commit will probably work

diegomrsantos:
On the `help_heading` / `next_help_heading` attrs for the Clap option structs: if this is a supported Clap feature and it gives us better grouping, I don’t see a reason not to use it. That said, I’d treat it as a small CLI UX improvement rather than something required for the docgen work. So I’d keep it out of scope for this issue and, if we want it, do it in a follow-up that can be evaluated on its own.

diegomrsantos:
I do think #856 is a bit ambiguous here since it explicitly mentions section grouping in the MDX post-processor. My view on scope is still that the core of that issue is generating accurate docs from the Clap definitions and checking them in CI. Grouping itself, especially if it also changes runtime `--help`, feels like follow-up scope to me.

jnhsigmap:
happy to defer the clap headers. I think the `Node` docs (`Keygen` and `Keysplit` are flat) were organised by semantic grouping (by the struct names), which made the `help_heading` an easy win to arrange them. The alternative is a function with hard-coded mapping logic to produce the same. The next PR would contain the `docgen` crate with some additional `Makefile` targets. This would not be behaviour altering, but the debate would be how to post-process in order to achieve the desired documentation structure.

jnhsigmap:
> pr looks overall in good shape! just merged #907 to handle the cargo deps vulnerability fix since its out of scope for this pr. hopefully if you rebase on `unstable`, it'll remove the vulnerability fix diffs from this pr, else killing your last commit will probably work

thanks @shane-moore I just manually removed it. my original upgrade was an automated fix

shane-moore (approved):

PR #904, opened 2026-03-19
fix(eth): process contract logs one block at a time

diegomrsantos:

- This fixes the behavior bug that remains after PR 1: later contract events in the same block need to see earlier writes through the active SQLite transaction before `NetworkState` is published.
- This is worth doing now because PR `#885` is being replaced by a smaller reviewed stack, and PR `#898` already prepared the database layer for this boundary change.
- Evidence: `event_processor` previously mixed tx writes with in-memory reads, which breaks same-block flows once publication is deferred. This PR adds focused same-block integration coverage for `OperatorAdded` + `ValidatorAdded` and `ValidatorAdded` + `ValidatorRemoved`.
- Relevant links: `#880`, `#898`, replacement for the behavior-fix portion of `#885`.

## Change Overview (Required)

- Process fetched contract logs one block at a time inside `event_processor`, with one SQLite transaction per block.
- Route same-block reads through tx-scoped database helpers instead of `self.db.state()`.
- Publish `NetworkState` and trigger index-sync / exit side effects only after the block transaction commits.
- Add position-aware test helpers so integration tests can model real same-block ordering.
- Intentionally did not change schema, sync resume machinery, or `NetworkState` consumers outside this boundary.

- Main risk is changing the event-processing boundary in historical and live sync.
- The diff stays scoped to `event_processor`, the small set of tx-scoped reads it needs, and targeted integration coverage.
- Trade-off: fetched multi-block batches are now committed block-by-block instead of as one larger transaction. That is intentional so durable progress, state publication, and post-commit side effects align with block processing.

## Validation (Required)

- `cargo check -p eth`
- `cargo fmt --all --check`
- `cargo clippy -p database -p eth --all-targets -- -D warnings`
- `cargo test -p eth --tests`

- Safe revert after PR `#898` if needed.
- No schema or migration impact.
- Reverts to the pre-PR2 batch-wide event-processing boundary.

- Depends on merged PR `#898`.

- A follow-up cleanup PR can remove any dead helper or cursor-adjacent code left over from the replaced direction.

### DIFF
--- a/anchor/database/src/cluster_operations.rs
+++ b/anchor/database/src/cluster_operations.rs
@@ -1,8 +1,6 @@
-use std::str::FromStr;
-
 use bls::PublicKeyBytes;
 use rusqlite::{OptionalExtension, Transaction, params};
-use ssv_types::{Cluster, ClusterId, OperatorId, Share, ValidatorMetadata};
+use ssv_types::{Cluster, ClusterId, ClusterMember, OperatorId, Share, ValidatorMetadata};
 use types::Address;
 
 use super::{DatabaseError, NetworkDatabase, PendingStateUpdates, sql_operations};
@@ -144,43 +142,32 @@ impl NetworkDatabase {
         validator_pubkey: &PublicKeyBytes,
         tx: &Transaction<'_>,
     ) -> Result<Option<Cluster>, DatabaseError> {
-        let cluster_row = tx
+        let Some(cluster_id) = tx
             .prepare_cached(sql_operations::GET_CLUSTER_BY_VALIDATOR)?
             .query_row(params![validator_pubkey.to_string()], |row| {
-                Ok((
-                    ClusterId(row.get("cluster_id")?),
-                    row.get::<_, String>("owner")?,
-                    row.get::<_, Option<String>>("fee_recipient")?,
-                    row.get::<_, bool>("liquidated")?,
-                ))
+                Ok(ClusterId(row.get("cluster_id")?))
             })
-            .optional()?;
-
-        let Some((cluster_id, owner, fee_recipient, liquidated)) = cluster_row else {
+            .optional()?
+        else {
             return Ok(None);
         };
 
         let cluster_members = tx
             .prepare_cached(sql_operations::GET_CLUSTER_MEMBERS)?
-            .query_map([cluster_id.0], |row| row.get(0))?
+            .query_map([cluster_id.0], |row| {
+                Ok(ClusterMember {
+                    cluster_id,
+                    operator_id: row.get(0)?,
+                })
+            })?
             .collect::<Result<Vec<_>, _>>()?;
-        let owner = Address::from_str(&owner).map_err(|e| {
-            DatabaseError::SQLError(format!("Failed to parse cluster owner from database: {e}"))
-        })?;
-        let fee_recipient = match fee_recipient {
-            Some(fee_recipient) => Address::from_str(&fee_recipient).map_err(|e| {
-                DatabaseError::SQLError(format!("Failed to parse fee recipient from database: {e}"))
-            })?,
-            None => owner,
-        };
 
-        Ok(Some(Cluster {
-            cluster_id,
-            owner,
-            fee_recipient,
-            liquidated,
-            cluster_members: cluster_members.into_iter().collect(),
-        }))
+        tx.prepare_cached(sql_operations::GET_CLUSTER_BY_VALIDATOR)?
+            .query_row(params![validator_pubkey.to_string()], |row| {
+                Cluster::try_from((row, cluster_members))
+            })
+            .optional()
+            .map_err(DatabaseError::from)
     }
 
     /// Bump the nonce of the owner in the active transaction and queue the matching state update.
--- a/anchor/database/src/cluster_operations.rs
+++ b/anchor/database/src/cluster_operations.rs
@@ -1,3 +1,5 @@
+use std::str::FromStr;
+
 use bls::PublicKeyBytes;
 use rusqlite::{OptionalExtension, Transaction, params};
 use ssv_types::{Cluster, ClusterId, ClusterMember, OperatorId, Share, ValidatorMetadata};
@@ -142,10 +144,15 @@ impl NetworkDatabase {
         validator_pubkey: &PublicKeyBytes,
         tx: &Transaction<'_>,
     ) -> Result<Option<Cluster>, DatabaseError> {
-        let Some(cluster_id) = tx
+        let Some((cluster_id, owner, fee_recipient, liquidated)) = tx
             .prepare_cached(sql_operations::GET_CLUSTER_BY_VALIDATOR)?
             .query_row(params![validator_pubkey.to_string()], |row| {
-                Ok(ClusterId(row.get("cluster_id")?))
+                Ok((
+                    ClusterId(row.get("cluster_id")?),
+                    row.get::<_, String>("owner")?,
+                    row.get::<_, Option<String>>("fee_recipient")?,
+                    row.get("liquidated")?,
+                ))
             })
             .optional()?
         else {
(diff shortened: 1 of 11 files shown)


anchor/eth/src/util.rs
```diff
@@ -165,13 +167,15 @@ pub fn validate_operators(
...
-    {
-        return Err(ExecutionError::Database(
-            "One or more operators do not exist".to_string(),
-        ));
+    for operator_id in operator_ids {
+        let exists = db
+            .operator_exists_tx(*operator_id, tx)
+            .map_err(|e| ExecutionError::Database(e.to_string()))?;
+        if !exists {
+            return Err(ExecutionError::Database(format!(
```
shane-moore, line 175: This returns `ExecutionError::Database` but the query succeeded — the operator just isn't registered. Same pattern at `event_processor.rs:573` (missing validator metadata) and `event_processor.rs:591` (missing cluster). All three should be `InvalidEvent`.
>
> Pre-existing and harmless today since all errors skip-and-continue, but #351 makes `Database` → `AbortBatch` — these would become sync-stalling errors on legitimate "data not yet present" conditions. Worth reclassifying in either this PR or #351 before #351 merges.

anchor/database/src/cluster_operations.rs
shane-moore: nit: Comment says `liquidated = false, active = true` but the callers pass `true` for liquidation and the SQL is `SET liquidated = ?1` — so the mapping is inverted. Consider `// liquidated flag (true = liquidated, false = active)` or just removing the comment since the SQL is self-documenting.

anchor/eth/src/event_processor.rs
shane-moore, line 497: nit (pre-existing): The `_` arm conflates `Ok(None)` (no custom fee recipient — legitimate) with `Err(...)` (DB failure). Splitting into explicit arms with a `warn!` on the error path would make DB failures observable without blocking validator registration. Low probability but free insurance on a financial-correctness path.

anchor/database/src/cluster_operations.rs
```diff
@@ -136,6 +138,57 @@ impl NetworkDatabase {
         Ok(())
     }
 
+    /// Load a cluster by validator public key through the transaction's view of the database.
+    pub fn get_cluster_by_validator_tx(
```
shane-moore, line 140: nit: This method inlines the same owner/fee_recipient parsing, `None → owner` fallback, and `Cluster` construction that `Cluster::try_from((&Row, Vec<ClusterMember>))` already handles in `sql_conversions.rs`. Not blocking, but a shared `cluster_from_row`-style helper would keep the two paths in sync and shave ~30 lines. Fine as a follow-up.

anchor/eth/tests/integration.rs
```diff
@@ -109,6 +111,157 @@ async fn test_multiple_events_processing() {
...
+        "cluster should be removed after its only validator is removed"
+    );
+    assert_eq!(
+        test.processor.db.state().get_last_processed_block(),
+        same_block
+    );
+}
+
 #[tokio::test]
 async fn test_database_transaction_rollback_on_error() {
```
shane-moore: The test does NOT test rollback — it tests that a valid event commits while a malformed event is skipped. With per-block transactions, the old "entire batch rolls back on error" behavior no longer exists. The test name is actively misleading.

**Recommendation:** Rename to `test_malformed_events_skipped_without_affecting_valid_events`.

i know this was untouched as part of this pr, but easy fix either here or new gh issue, whichever you think

anchor/eth/src/event_processor.rs
```diff
@@ -55,120 +60,252 @@ impl EventProcessor {
...
+
+        // Buffer the current block so each transaction/process step owns exactly one block.
+        let mut current_block = None;
+        let mut block_logs = Vec::new();
+
+        for log in logs {
+            let block_number = log.block_number.unwrap_or(end_block);
+
+            // We hit a block boundary, so flush the previous block before buffering this one.
+            if current_block
```
shane-moore, line 97: The block-boundary flush (lines 97-108) is the core new logic in this PR but doesn't have direct test coverage yet. Every existing test either puts all logs in the same block or uses separate `process_logs` calls, so this path is never exercised.

A test that passes logs spanning 2+ blocks in a single `process_logs` call and verifies each block committed independently (e.g., operator added in block N is visible to a validator added in block N+1 within the same batch) would close the gap.

anchor/database/src/validator_operations.rs
```diff
@@ -78,6 +78,37 @@ impl NetworkDatabase {
...
+    }
+
+    /// Check whether the current operator has a share for a validator in the transaction view.
+    pub fn has_own_share_tx(
+        &self,
+        validator_pubkey: &PublicKeyBytes,
+        tx: &Transaction<'_>,
+    ) -> Result<bool, DatabaseError> {
+        let Some(operator_id) = self.get_own_operator_id_tx(tx)? else {
+            return Ok(false);
```
shane-moore, line 97: Could we add a unit test for the `operator_id = None` path (line 99-101)? Currently all integration tests populate the operator, so this branch is never exercised. Something like:
> ```rust
> #[test]
> fn test_has_own_share_tx_no_operator_returns_false() {
>     let fixture = InMemoryTestFixture::new_empty();
>     let conn = fixture.db.pool.get().unwrap();
>     let tx = conn.unchecked_transaction().unwrap();
>     let result = fixture.db.has_own_share_tx(&PublicKeyBytes::default(), &tx);
>     assert_eq!(result.unwrap(), false);
> }
> ```

anchor/database/src/cluster_operations.rs
diegomrsantos: Fixed in a21ce6b15 by renaming the parameter/field to `liquidated`. This naming/comment inversion was pre-existing on `unstable`; this PR just touched the area.

anchor/database/src/validator_operations.rs
diegomrsantos: Added in 9229bdd75. There is now a unit test covering the `operator_id = None` branch: `test_has_own_share_tx_no_operator_returns_false()`.

anchor/database/src/cluster_operations.rs
diegomrsantos: Fixed in 694db5b69. `get_cluster_by_validator_tx()` now reuses `Cluster::try_from((row, cluster_members))` instead of duplicating the owner/fee-recipient parsing and fallback logic.

anchor/eth/src/util.rs
diegomrsantos: I agree the ambiguity is real, but I don’t think `#904` should change that classification. In these paths we cannot actually prove whether the event is invalid or local committed state is unexpectedly missing data, so reclassifying to `InvalidEvent` here would be a behavior change, not just cleanup. That risk existed before this PR; `#904` is only changing the block-scoped processing boundary, so I’d rather keep the existing semantics here and handle the classification question separately.

anchor/eth/src/event_processor.rs
diegomrsantos: Addressed in `30345dcb2` and clarified in `5c09ad916`.

I added direct block-batching coverage for the PR2 path in `anchor/eth/tests/integration.rs`:
- `test_cross_block_operator_and_validator_processing_succeeds`: one `process_logs(...)` call spanning multiple blocks on the happy path
- `test_cross_block_failure_preserves_previous_block_commit`: stronger regression test that panics in block `N+1` and asserts block `N` stayed committed, which is the actual proof of the block-boundary flush
- `test_cross_block_empty_tail_advances_processed_block`: covers the `end_block`/empty-tail path

So the comment was correct: the original suite did not exercise that path directly. It is covered now.

anchor/eth/src/event_processor.rs
diegomrsantos: I agree this would be a cleaner shape, but I’m leaving it out of `#904`.

This `_` arm is pre-existing from `unstable`, and changing it here would be a small behavior/logging change on the validator-add path rather than part of the block-boundary work in this PR. If we touch it separately, I agree the better version is to split `Ok(None)` from `Err(...)` and log the DB-failure path explicitly.

anchor/eth/tests/integration.rs
diegomrsantos: Fixed in `f1d4f592f`.

I renamed it to `test_malformed_events_skipped_without_affecting_valid_events` and added a short doc comment making the preserved semantics explicit. This was a pre-existing test name on `unstable`; `#904` just made the old name more misleading.

anchor/eth/src/event_processor.rs
shane-moore: I feel you, made small ticket via https://github.com/sigp/anchor/issues/908

shane-moore (dismissed):

shane-moore (approved):

PR #906, opened 2026-03-20
test: add seeded message delivery shuffling to QBFT tests

diegomrsantos:

- QBFT integration tests deliver messages to operators in a fixed order (1, 2, 3, ..., N), meaning they never explore what happens when messages arrive at different operators in different orders.
- Real networks don't deliver messages simultaneously or in a fixed order — message timing and arrival order vary per topology and load.
- Inspired by commonware-runtime's deterministic simulation approach: same seed = same interleaving = reproducible failures.

## Change Overview (Required)

- `QbftTester` gains an optional seeded RNG that shuffles operator delivery order in `process_network_message`. When a seed is provided, each message broadcast delivers to operators in a different (but reproducible) shuffled order.
- `TestContext::new_seeded()` threads the seed through for ergonomic test construction.
- Three new multi-seed tests exercise basic consensus, f-faulty scenarios across committee sizes, and concurrent instances — each across multiple seeds.
- Existing tests are unchanged (pass `None` for seed, preserving fixed-order delivery).

- Test-only change, zero production code modified.
- Seeded shuffling is deterministic: if a seed fails in CI, re-running with that seed reproduces the exact same behavior locally.
- Existing tests preserve their original behavior.

## Validation (Required)

- All 19 tests pass (16 existing + 3 new seeded tests).
- `make cargo-fmt-check`, `make lint` clean.
- New seeded tests cover: 10 seeds × basic consensus, 5 seeds × 4 committee sizes with faults, 10 seeds × concurrent instances.

N/A — test-only change, safe to revert.

- This is the first step toward commonware-style deterministic testing. Future work could add clock abstraction to the QBFT orchestration layer, enabling full deterministic simulation of timing and scheduling interleavings.

### DIFF
--- a/anchor/qbft_manager/src/tests.rs
+++ b/anchor/qbft_manager/src/tests.rs
@@ -1,7 +1,7 @@
 use std::{
     collections::HashMap,
     num::NonZeroU64,
-    sync::{Arc, LazyLock, RwLock, RwLockWriteGuard},
+    sync::{Arc, LazyLock, Mutex, RwLock, RwLockWriteGuard},
     time::{Duration, SystemTime, UNIX_EPOCH},
 };
 
@@ -9,6 +9,7 @@ use fork::{Fork, ForkSchedule};
 use message_sender::testing::MockMessageSender;
 use processor::Senders;
 use qbft::InstanceHeight;
+use rand::prelude::*;
 use slot_clock::{ManualSlotClock, SlotClock};
 use ssv_types::{
     Cluster, ClusterId, CommitteeId, IndexSet, OperatorId,
@@ -64,14 +65,25 @@ where
     D: QbftDecidable<E>,
     D::Id: Send + Sync + Clone,
 {
-    // Create a new test context with default setup
+    // Create a new test context with default setup (no delivery shuffling)
     pub async fn new(
         clock: ManualSlotClock,
         executor: TaskExecutor,
         size: CommitteeSize,
         test_data: Vec<(D, D::Id)>,
     ) -> Self {
-        Self::new_with_delays(clock, executor, size, test_data, HashMap::new()).await
+        Self::new_inner(clock, executor, size, test_data, HashMap::new(), None).await
+    }
+
+    // Create a new test context with seeded delivery order shuffling
+    pub async fn new_seeded(
+        clock: ManualSlotClock,
+        executor: TaskExecutor,
+        size: CommitteeSize,
+        test_data: Vec<(D, D::Id)>,
+        seed: u64,
+    ) -> Self {
+        Self::new_inner(clock, executor, size, test_data, HashMap::new(), Some(seed)).await
     }
 
     pub async fn new_with_delays(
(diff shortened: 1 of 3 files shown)


anchor/qbft_manager/src/tests.rs
```diff
@@ -913,6 +949,72 @@ mod manager_tests {
...
+            let setup = setup_test(1);
+            let mut context = TestContext::<types::MainnetEthSpec, BeaconVote>::new_seeded(
+                setup.clock,
+                setup.executor,
+                CommitteeSize::Four,
+                setup.all_data,
+                seed,
+            )
+            .await;
+
```
petarjuki7, line 967: nit: Could we maybe print the seed here so if the QBFT failed we know what ordering of operators caused it?

petarjuki7 (approved):
Nice, lgtm, added one small comment

anchor/qbft_manager/src/tests.rs
```diff
@@ -913,6 +949,72 @@ mod manager_tests {
...
+            context.verify_consensus().await;
+        }
+    }
+
+    // Run f-faulty consensus across multiple seeds and committee sizes.
+    #[tokio::test]
+    async fn test_seeded_f_faulty() {
```
shane-moore, line 974: `test_seeded_f_faulty` doesn’t seem to exercise seeded round-change behavior as written. `setup_test(1)` gives this case instance_height = 1, which makes operator 2 the round-1 leader with the current leader function, and the faulty sets here never take operator 2 offline. So this looks more like additional seeded happy-path coverage than seeded round-change coverage. Not blocking for me, but if we want shuffle coverage for round changes too, I think that would need a separate leader-offline or delayed/split-round seeded case

diegomrsantos: Good catch — though worth noting this is a pre-existing gap: the non-seeded `test_consensus_f_faulty` uses the same faulty sets with the same `setup_test(1)`, so it also never takes the round-1 leader (operator 2) offline. The seeded version just mirrors it.

A seeded leader-offline test (similar to `test_fault_leader` but with shuffled delivery) would be a good addition — could be done in a follow-up PR.

diegomrsantos:
@claude review this PR

diegomrsantos:
Closing this for now.

The extra shuffle coverage is useful, but the PR is framed as deterministic/replayable testing and it does not fully provide that yet: the seed only controls recipient order, while the harness still relies on fresh random test data and async scheduling outside that seed.

I’d rather revisit this later with a cleaner deterministic harness so the replay guarantees are actually true.

anchor/qbft_manager/src/tests.rs
shane-moore: for sure, opened issue for a follow-up https://github.com/sigp/anchor/issues/927

shane-moore: ah, noticed this was closed for regions described in https://github.com/sigp/anchor/pull/906#issuecomment-4200218743. closed issue 927 for now as well

PR #910, opened 2026-03-25
chore: trivial refactor and docs adjustments for automated doc generator

jnhsigmap:

A minor PR for #856 that does not alter the default behaviour of the Anchor project, however does change how default configuration is assigned in `client`. Default values previously only provided in `Config.new` are now assigned as default values in `clap` derived CLI struct `ExternalApis` . Defaults are now not assigned via `Config.new` unless explicitly invoked.

A few small changes were required to better facilitate a new executable `anchor-docgen` that automatically generates project documentation.

## Change Overview (Required)

- [Moved configuration constants](https://github.com/sigp/anchor/pull/910/changes#diff-5879bfb8ba38183fba56481d4795057b71d0ffcb91ee259d65c435d60533289aR13) (`DEFAULT_BEACON_NODE`, `DEFAULT_EXECUTION_NODE`, `DEFAULT_EXECUTION_NODE_WS`)  from `client` to `global_config` to ensure that CLI flag detail is consistent with application behaviour. This did not create a behavioural change or touch any files beyond where these were accessed.
- Created `defaults` module in `common/global_config` to store all other hard-coded options in the project that were liked to CLI flags and instead use them as constants. This extended to values in `http_api` and `http_metrics`.
- I deliberately chose duplication of `port_str` and `port` to avoid the possibility of runtime panics if invalid values were included as defaults. The application would otherwise compile and panic at runtime.
- Minor edits to `clap` struct definitions to either address rendering errors, enforce better readability or consistency, or correct existing errors.

There should be no risk here as this is standalone and does not impact Anchor's behaviour.

## Validation (Required)

Tests passing as normal.

n/a

- PR to follow adds the `docgen` crate to create automated documentation generator.


anchor/cli/src/cli.rs
```diff
@@ -224,7 +225,7 @@ pub struct MetricsOptions {
     #[clap(
         long,
         help = "Enable per validator metrics for > 64 validators. \
-                Note: This flag is automatically enabled for <= 64 validators. \
+                Note: This flag is automatically enabled for < 65 validators. \
```
jnhsigmap, line 231: the `=` character broke my browser render 😕

jnhsigmap:
~Spellcheck appears to be playing up spuriously. The README is using those terms correctly.~

anchor/cli/src/cli.rs
```diff
@@ -172,7 +173,7 @@ pub struct HttpApiOptions {
         long,
         value_name = "PORT",
         requires = "http",
-        help = "Set the listen TCP port for the RESTful HTTP API server.",
+        help = "Set the listen TCP port for the RESTful HTTP API server. Defaults to 5062 if --http is set.",
```
diegomrsantos: Could we use constants for these hardcoded ports?

Cargo.toml
```diff
@@ -155,6 +155,7 @@ serde_yaml = "0.9"
 sha2 = "0.10.8"
 ssz_types = "0.14.0"
 strum = { version = "0.27.0", features = ["derive"] }
+tempfile = "3.27.0"
```
diegomrsantos, line 158: Is it necessary?

diegomrsantos: Ah cause you use the workspace version now

anchor/cli/src/cli.rs
jnhsigmap: They are :) I just mentioned it there because it was previously manually inputted into the book defaults column

https://github.com/sigp/anchor/pull/910/changes/BASE..cb2a536264b0c9cbf73eaa7b50303eb74be17429#diff-5879bfb8ba38183fba56481d4795057b71d0ffcb91ee259d65c435d60533289aR13

diegomrsantos: but here it's hardcoded and could drift

jnhsigmap: @diegomrsantos I have moved hard-coded values into a separate module in `global_config`

anchor/common/global_config/src/defaults.rs
```diff
@@ -0,0 +1,61 @@
...
+pub const DEFAULT_NETWORK: NetworkDefaults = NetworkDefaults { address: "0.0.0.0" };
+
+pub struct ExternalApiDefaults {
+    pub sync_tolerances: &'static str,
+}
+
+/// Default external API configuration.
+pub const DEFAULT_EXTERNAL_API: ExternalApiDefaults = ExternalApiDefaults {
+    sync_tolerances: "8,8,48",
+};
```
shane-moore, line 59: nit: Since `port` and `port_str` are maintained separately (necessary because clap's `default_value_if` requires `&str`), could we add a quick test at the bottom of this file to guard against them drifting?
>
> ```rust
> #[cfg(test)]
> mod tests {
>     use super::*;
>
>     #[test]
>     fn port_and_port_str_are_consistent() {
>         assert_eq!(DEFAULT_METRICS.port.to_string(), DEFAULT_METRICS.port_str);
>         assert_eq!(DEFAULT_HTTP_API.port.to_string(), DEFAULT_HTTP_API.port_str);
>     }
> }
> ```
>
> `port` feeds `Config::default()` impls and `port_str` feeds clap — if they ever diverge the CLI default and programmatic default silently disagree.

anchor/common/global_config/src/defaults.rs
```diff
@@ -0,0 +1,61 @@
...
+pub struct HttpApiDefaults {
+    pub host: &'static str,
+    pub port: u16,
+    pub port_str: &'static str, /* Used as an alternative to .parse().expect() risking runtime
+                                 * panics. */
+}
+
+/// Default HTTP API configuration.
+pub const DEFAULT_HTTP_API: HttpApiDefaults = HttpApiDefaults {
+    host: "127.0.0.1",
```
shane-moore: nit: `DEFAULT_HTTP_API.host` appears to have no consumer — `http_api::Config::default()` uses `Ipv4Addr::LOCALHOST` directly, and `http_address` in the CLI has no `default_value_if` (by design, since it requires `--unencrypted-http-transport`). `DEFAULT_METRICS.host` has a consumer (`cli.rs:212` `default_value_if`), so the asymmetry is intentional.
>
> Consider removing `host` from `HttpApiDefaults` until the docgen crate (or another consumer) needs it -> avoids a dead constant that suggests a single source of truth but isn't actually wired in.

jnhsigmap: good pickup, done :slightly_smiling_face:

anchor/common/global_config/src/defaults.rs
jnhsigmap: good point too, I added this test

diegomrsantos:
@claude review this PR

jnhsigmap:
> **Claude finished @diegomrsantos's task in 2m 21s** —— [View job](https://github.com/sigp/anchor/actions/runs/24078373467)
> 
> ### Review of PR #910
> * [x]  Gather context and read changed files
> * [x]  Analyze the diff against `origin/unstable`
> * [x]  Review for correctness, security, and design issues
> * [x]  Post review feedback
> 
> ### Behavioral change: `default_value` on `Option<Vec<String>>` fields
> **`anchor/cli/src/cli.rs:48-57`, `cli.rs:59-67`, `cli.rs:69-76`**
> 
> Adding `default_value` to `beacon_nodes`, `execution_rpc`, and `execution_ws` changes these from `None` (when the user doesn't pass the flag) to `Some(...)` with the default value. This means the `if let Some(...)` guards in `anchor/client/src/config.rs:143-153` will now _always_ match, causing `parse_urls` / `SensitiveUrl::parse` to run on every startup even when the user didn't pass these flags.
> 
> Functionally the output is the same because the default strings match what `Config::new()` was already constructing. However, this is a subtle semantic change worth being aware of:
> 
> 1. The `Config::new()` defaults at lines 89-99 of `config.rs` are now dead code for the normal CLI path — they're always overwritten by `from_cli`. They only matter if someone calls `Config::new()` directly from non-CLI code (e.g., tests or the upcoming docgen).
> 2. If `SensitiveUrl::parse` ever returned an error for one of these defaults, the old path would panic in `Config::new` (`expect`), while the new path returns `Err` from `from_cli`. This is arguably better error handling, so not necessarily a problem.
> 
> This is not a bug, but it's worth confirming this is the intended behavior since the PR description says "does not alter the behaviour."
> 
> ### Remaining findings (minor)
> **Stray blank line** — `cli.rs:54`: There's an extra blank line between `help = ...` and `display_order = 0` in the `beacon_nodes` attribute. Trivial, but inconsistent with the other fields.
> 
> **`tempfile` workspace consolidation is incomplete** — The PR adds `tempfile = "3.27.0"` to the workspace `Cargo.toml` and updates `anchor/Cargo.toml` to `{ workspace = true }`, but several other crates still pin older non-workspace versions (`keysplit`, `validator_store`, `database`, `ssv_network_config`, `operator_key`, `subnet_service`). This isn't a problem, but if the intent was to consolidate, it's incomplete. Fine to defer to a follow-up.
> 
> ### What looks good
> * The `defaults.rs` module is well-structured with clear ownership of each constant group.
> * The `port` / `port_str` duplication rationale (avoiding runtime panics from `parse().expect()` in clap attributes) is sound, and the consistency test added per Shane's suggestion guards against drift.
> * Removing `host` from `HttpApiDefaults` per Shane's feedback was the right call — avoids a dead constant.
> * The `Ipv4Addr::new(127, 0, 0, 1)` → `Ipv4Addr::LOCALHOST` change in `http_api/src/config.rs` is a nice cleanup.
> 
> ### Summary
> No blocking issues. The `default_value` addition on `Option<Vec<_>>` fields is a semantic shift worth acknowledging explicitly, but functionally equivalent for the current code paths.
> 
> |

It makes a good point about `Config::new`. This was being used previously because there were no defaults. This change moves them from being handled here internally to being more visible to users at CLI invocation.

jnhsigmap:
Updated the description slightly

shane-moore (approved):

PR #912, opened 2026-03-26
chore: bump deps in stable

shane-moore:
As the last `stable` release has been a while, bump all dependencies for a bugfix release.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -144,7 +144,7 @@ install-audit:
 	cargo install --force cargo-audit
 
 audit-CI:
-	cargo audit --ignore RUSTSEC-2026-0049
+	cargo audit
 
 # Runs `cargo vendor` to make sure dependencies can be vendored for packaging, reproducibility and archival purpose.
 vendor:
--- a/Makefile
+++ b/Makefile
@@ -144,7 +144,7 @@ install-audit:
 	cargo install --force cargo-audit
 
 audit-CI:
-	cargo audit
+	cargo audit --ignore RUSTSEC-2026-0049
 
 # Runs `cargo vendor` to make sure dependencies can be vendored for packaging, reproducibility and archival purpose.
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -144,7 +144,7 @@ install-audit:
 	cargo install --force cargo-audit
 
 audit-CI:
-	cargo audit
```
shane-moore, line 147: did the same as lh [here](https://github.com/sigp/lighthouse/blob/176cce585c1ba979a6210ed79b6b6528596cdb8c/Makefile#L327) in their pr https://github.com/sigp/lighthouse/pull/9034/changes#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542

when i back merge to `unstable`, this will likely make a merge conflict. in that case, i should take `unstable` branche's `cargo audit` without the `--ignore RUSTSEC-2026-0049` since `rustls-webpki` is at a fixed version of `0.103.10` in `unstable` whereas `stable` is using the vulnerable `0.103.9` version

anchor/network/Cargo.toml
```diff
@@ -5,7 +5,6 @@ edition = { workspace = true }
 authors = ["Sigma Prime <contact@sigmaprime.io>"]
 
 [dependencies]
-async-trait = "0.1.85"
```
shane-moore, line 8: We removed `async-trait` because the new libp2p rev changed the `request_response::Codec `trait from using `#[async_trait]` to native impl Future return types

Makefile
dknopik: Can't we run a `cargo update` to avoid using `0.103.9`?

shane-moore: that approach works for me! just updated

dknopik (approved):
LGTM

PR #926, opened 2026-04-07
test: add aggregate committee coverage

### LINKED ISSUE #886: test: integration tests for aggregation committee signature count and collection
## Context

  The committee-batching semantics introduced in #834 apply equally to
  `sign_aggregate_and_proofs`, but #878 only covers the attestation path (`sign_attestations`).

  The aggregation path has the same per-committee isolation and collection semantics and should have equivalent test coverage.

 Ideally reuses test infrastructure built for #878.

  ## Proposed test cases
  - Multiple committees in one `sign_committee_aggregate_and_proofs` call -> assert separate batches per committee, no ordering dependency
  - Mixed success/failure across committees -> one failing committee does not block others from producing aggregates
  - `CollectionMode::Committee` receives correct `num_signatures_to_collect` derived from local committee membership

diegomrsantos:
Closes #886

- Addresses #886.
- `sign_aggregate_and_proofs` already has committee-path logic on `unstable`, but it was still missing focused integration coverage for the aggregate-specific cases:
  - multiple committees in one call
  - one `sign_and_collect` call per validator in the committee path
  - isolation when one committee is blocked waiting for aggregate assignments
- This rebased branch drops the stale attestation-side churn and keeps the PR focused on aggregate coverage only.

## Change Overview (Required)

- Adds a dedicated aggregate committee integration test module.
- Extends the shared `validator_store` test harness only where the aggregate test needs it: seeding `AggregationAssignments` and building aggregate fixtures for specific committees and slots.
- Keeps the scenario-specific committee setup and the batch-size assertions in `committee_aggregate.rs`, so reviewers do not need to decode shared constants in `common.rs`.
- Makes the committee-path assertion explicit: we expect one `sign_and_collect` call per validator, and the requester's batch size should match that validator's committee size.
- Intentionally unchanged: production aggregate signing logic and runtime behavior.

- Risk is low because this is test-only coverage.
- The shared harness grows slightly, but only with aggregate-specific helpers that are reused directly by the new test.
- No production code paths, configuration, or runtime behavior changed.

## Validation (Required)

- `cargo fmt --all --check`
- `cargo test -p anchor_validator_store committee_ -- --nocapture`
- `cargo clippy -p anchor_validator_store --all-targets -- -D warnings`
- The commit hook also ran `cargo fmt --all`, `cargo clippy --all`, and `cargo sort workspace`.

- Revert this PR.

- N/A

- Reviewer reading order: start with `committee_aggregate.rs`, then open the new aggregate helpers in `common.rs` if needed.

### DIFF
--- a/anchor/validator_store/src/testing/committee_aggregate.rs
+++ b/anchor/validator_store/src/testing/committee_aggregate.rs
@@ -10,7 +10,7 @@ use crate::Error;
 type SignAggregatesResult = Vec<Result<Vec<SignedAggregateAndProof<MainnetEthSpec>>, Error>>;
 
 const PRIMARY_COMMITTEE_VALIDATOR_COUNT: usize = 2;
-const SECONDARY_COMMITTEE_VALIDATOR_COUNT: usize = 1;
+const SINGLE_VALIDATOR_COMMITTEE_COUNT: usize = 1;
 const EXPECTED_TOTAL_SIGNED_AGGREGATES: usize = 3;
 const EXPECTED_SIGN_AND_COLLECT_CALLS: usize = 3;
 const EXPECTED_REQUESTED_COUNTS: [usize; 3] = [1, 2, 2];
@@ -22,7 +22,7 @@ const EXPECTED_REQUESTED_COUNTS: [usize; 3] = [1, 2, 2];
 async fn sign_aggregate_and_proofs_produces_one_stream_item_per_committee() {
     // Arrange
     let committee_a = create_primary_committee_setup(PRIMARY_COMMITTEE_VALIDATOR_COUNT);
-    let committee_b = create_secondary_committee_setup(SECONDARY_COMMITTEE_VALIDATOR_COUNT);
+    let committee_b = create_secondary_committee_setup(SINGLE_VALIDATOR_COMMITTEE_COUNT);
     assert_ne!(
         committee_a.cluster.committee_id(),
         committee_b.cluster.committee_id(),
@@ -85,8 +85,8 @@ async fn sign_aggregate_and_proofs_produces_one_stream_item_per_committee() {
 #[tokio::test(flavor = "multi_thread")]
 async fn sign_aggregate_and_proofs_failure_isolation() {
     // Arrange
-    let committee_a = create_primary_committee_setup(SECONDARY_COMMITTEE_VALIDATOR_COUNT);
-    let committee_b = create_secondary_committee_setup(SECONDARY_COMMITTEE_VALIDATOR_COUNT);
+    let committee_a = create_primary_committee_setup(SINGLE_VALIDATOR_COMMITTEE_COUNT);
+    let committee_b = create_secondary_committee_setup(SINGLE_VALIDATOR_COMMITTEE_COUNT);
     let harness = ValidatorStoreTestHarness::new(vec![committee_a, committee_b], OUR_OPERATOR_ID);
     harness.seed_aggregation_assignments_for_slot(TEST_SLOT, &[PRIMARY_COMMITTEE_INDEX]);
     let aggregates = vec![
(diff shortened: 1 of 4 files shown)


diegomrsantos:
@claude review this PR

anchor/validator_store/src/testing/committee_aggregate.rs
```diff
@@ -0,0 +1,126 @@
...
+        .sign_aggregate_and_proofs(aggregates);
+    tokio::pin!(stream);
+    let first = tokio::time::timeout(STREAM_ITEM_TIMEOUT, stream.next()).await;
+    let second = tokio::time::timeout(BLOCKED_STREAM_TIMEOUT, stream.next()).await;
+
+    // Assert
+    let first_item = first
+        .expect("first committee should complete within timeout")
+        .expect("stream should yield an item");
+    assert!(first_item.is_ok());
```
shane-moore: **nit (isolation tests):** `assert!(first_item.is_ok())` is a bit too weak here. It only proves the first stream item was `Ok(_)`, not that the successful committee actually produced signed data. Could we assert the batch length as well?
>
> ```rust
> let signed = first_item.expect("first committee should succeed");
> assert_eq!(signed.len(), 1, "successful committee should produce one signed item");
> ```
>
> I’d make the same change in both the aggregate and attestation isolation tests.

anchor/validator_store/src/testing/committee_aggregate.rs
```diff
@@ -0,0 +1,126 @@
...
+
+type SignAggregatesResult = Vec<Result<Vec<SignedAggregateAndProof<MainnetEthSpec>>, Error>>;
+
+const PRIMARY_COMMITTEE_VALIDATOR_COUNT: usize = 2;
+const SINGLE_VALIDATOR_COMMITTEE_COUNT: usize = 1;
+const EXPECTED_TOTAL_SIGNED_AGGREGATES: usize = 3;
+const EXPECTED_SIGN_AND_COLLECT_CALLS: usize = 3;
+const EXPECTED_REQUESTED_COUNTS: [usize; 3] = [1, 2, 2];
+
+/// `sign_aggregate_and_proofs` groups aggregates by `CommitteeId`, runs consensus once per
```
shane-moore, line 53: nit: could we add a sibling test here for `sign_aggregate_and_proofs(vec![])`? The Boole path has an explicit empty-input early return that should yield zero stream items, and this file doesn’t exercise that contract yet. A `collect().await` asserting zero stream items would pin it down cheaply.

anchor/validator_store/src/testing/committee_aggregate.rs
```diff
@@ -0,0 +1,126 @@
...
+            SignatureRequester::Committee {
+                num_signatures_to_collect,
+                ..
+            } => *num_signatures_to_collect,
+            other => panic!("expected SignatureRequester::Committee, got: {other:?}"),
+        })
+        .collect();
+    requested_counts.sort_unstable();
+    // Committee A has two aggregators and committee B has one, so the per-validator collection
+    // requests should be [2, 2, 1] irrespective of stream ordering.
```
shane-moore: **nit:** after `requested_counts.sort_unstable()`, the nearby comment still says `[2, 2, 1]`. Could we change it to `[1, 2, 2]` so it matches the asserted sorted order?

anchor/validator_store/src/testing/committee_aggregate.rs
diegomrsantos: Addressed in the latest push. I tightened both isolation tests to unwrap the successful batch and assert that it contains one signed item.

anchor/validator_store/src/testing/committee_aggregate.rs
diegomrsantos: Added in the latest push. sign_aggregate_and_proofs_empty_input now collects the stream and asserts that it stays empty.

anchor/validator_store/src/testing/committee_aggregate.rs
diegomrsantos: Addressed in the latest push. That sorted-count comment is gone now; the test spells out the per-validator calls and expected batch sizes directly.

shane-moore (approved):

Issue #934, opened 2026-04-07
feat: duplicate signature resolution (PR #890 split 4/4)


shane-moore:
## Context

Part 3 of the [PR #890 split](https://github.com/sigp/anchor/issues/931), per [review comment](https://github.com/sigp/anchor/pull/890#issuecomment-4118041676). Depends on #933.

## Scope

Spec function `resolveDuplicateSignature` (~60 LOC prod + ~265 LOC tests, 1 crate).

### Files

| File | Change |
|------|--------|
| `signature_collector/src/lib.rs` | `resolve_duplicate_signature`, `fetch_share_pubkeys_by_validator_index`, duplicate detection in message loop |
| `signature_collector/src/tests.rs` | Unit tests for `resolve_duplicate_*`; integration tests for duplicate resolution (including pre-registration path) |

### Details
- Triggered when the same operator sends conflicting signatures (independent from the fallback path in PR 2)
- Verifies both old and new sig against share pubkey from DB, keeps valid one, removes both if neither valid
- Uses `fetch_share_pubkeys_by_validator_index` (DB query from PR 1)

## Parent issue
- #931

Issue #933, opened 2026-04-07
feat: post-reconstruction signature verification with fallback (PR #890 split 3/4)


shane-moore:
## Context

Part 2 of the [PR #890 split](https://github.com/sigp/anchor/issues/931), per [review comment](https://github.com/sigp/anchor/pull/890#issuecomment-4118041676). Depends on #932.

## Scope

Core spec compliance (~250 LOC prod + ~400 LOC tests, 3 crates). Implements `ReconstructSignature` + `FallBackAndVerifyEachSignature`.

### Files

| File | Change |
|------|--------|
| `client/src/lib.rs` | Pass `database.clone()` to signature collector (1 line) |
| `validator_store/src/lib.rs` | Add `validator_pubkey` to `ValidatorSigningData` (1 line) |
| `signature_collector/Cargo.toml` | Add `metrics` + `database` deps |
| `signature_collector/src/lib.rs` | `try_combine_and_verify`, `verify_reconstructed_signature`, `find_invalid_shares`, `fetch_share_pubkeys`; `CombineOutcome` enum; `while` loop retry; DB wiring through manager; `SignatureCollecting` trait |
| `signature_collector/src/metrics.rs` | `SIGNATURE_VERIFICATION_FAILURES_TOTAL` counter |
| `signature_collector/src/tests.rs` | Unit tests for `find_invalid_shares`, `try_combine_and_verify`; integration tests for happy path + fallback eviction/recovery |

### Key design decisions
- `CombineOutcome` enum for explicit control flow (Success / CombineFailed / VerificationFailed)
- `while` loop replaces `if` to retry after evicting invalid shares
- `SignatureCollecting` trait for testability
- `spawn_blocking` for DB access to avoid blocking async runtime

### Known issue to address
- The `while` loop + eviction interaction when remaining shares still >= threshold needs review (flagged in original PR comment)

## Parent issue
- #931

shane-moore:
[...]

Issue #932, opened 2026-04-07
feat: add share pubkey DB queries (PR #890 split 1/4)


shane-moore:
## Context

Part 1 of the [PR #890 split](https://github.com/sigp/anchor/issues/931), per [review comment](https://github.com/sigp/anchor/pull/890#issuecomment-4118041676).

## Scope

Pure data-layer addition (~75 LOC, 1 crate). Zero behavior change.

### Files

| File | Change |
|------|--------|
| `database/src/sql_operations.rs` | `GET_SHARE_PUBKEYS_FOR_VALIDATOR`, `GET_SHARE_PUBKEYS_FOR_VALIDATOR_INDEX` SQL constants |
| `database/src/share_operations.rs` | `get_share_pubkeys_for_validator()`, `get_share_pubkeys_for_validator_index()` methods |

### Details
- Two queries that fetch operator share public keys by validator pubkey or validator index
- Used downstream by PR 2 (fallback verification) and PR 3 (duplicate resolution)
- Can merge immediately with no dependencies

## Parent issue
- #931

shane-moore:
closed via https://github.com/sigp/anchor/pull/936

Issue #931, opened 2026-04-07
Track PR #890 split: partial sig with fallback (4-PR stack)


shane-moore:
## Context

PR #890 implements three spec functions, DB queries, plumbing, metrics, and tests across 4 crates (+995 LOC). Per [this review comment](https://github.com/sigp/anchor/pull/890#issuecomment-4118041676), it should be split into a 4-PR stack.

### PR 1: `refactor: add share pubkey DB queries` (~75 LOC, 1 crate) — #932
- `database/src/sql_operations.rs` — two SQL constants (`GET_SHARE_PUBKEYS_FOR_VALIDATOR`, `GET_SHARE_PUBKEYS_FOR_VALIDATOR_INDEX`)
- `database/src/share_operations.rs` — `get_share_pubkeys_for_validator`, `get_share_pubkeys_for_validator_index`
- Pure data layer, zero behavior change, can merge immediately

### PR 2: `refactor: add validator_index DB index` (~10 LOC, 1 crate) — #940 — depends on PR 1
- Schema migration v3 → v4: `CREATE INDEX` on `validators.validator_index`
- Backs the `GET_SHARE_PUBKEYS_FOR_VALIDATOR_INDEX` query from PR 1
- Eliminates full table scan; matters under concurrent access with `POOL_SIZE = 1` ([context](https://github.com/sigp/anchor/pull/936#discussion_r3057785811))

### PR 3: `feat: post-reconstruction signature verification with fallback` (~250 LOC prod + ~400 LOC tests, 3 crates) — #933 — depends on PR 2
- Core spec compliance: `ReconstructSignature` + `FallBackAndVerifyEachSignature`
- `try_combine_and_verify`, `verify_reconstructed_signature`, `find_invalid_shares`, `fetch_share_pubkeys`
- `CombineOutcome` enum, `while` loop retry, `SignatureCollecting` trait
- DB wiring through `SignatureCollectorManager`, metrics
- 1-line mechanical changes in `client/src/lib.rs` and `validator_store/src/lib.rs`

### PR 4: `feat: duplicate signature resolution` (~60 LOC prod + ~265 LOC tests, 1 crate) — #934 — depends on PR 3
- Spec function: `resolveDuplicateSignature`
- `resolve_duplicate_signature`, `fetch_share_pubkeys_by_validator_index`, duplicate detection in message loop
- Logically independent code path (triggered by conflicting sigs from same operator, not by reconstruction failure)

## Why this split
[...]

PR #936, opened 2026-04-08
feat: add share pubkey queries

shane-moore:

- PR #890 was flagged for splitting into a 3-PR stack per [review comment](https://github.com/sigp/anchor/pull/890#issuecomment-4118041676)
- This is PR 1/3: adds DB queries needed by the signature verification and duplicate resolution features in PRs 2 and 3
- Tracking issue: #931, sub-issue: #932

## Change Overview

- Two new SQL constants in `sql_operations.rs`: `GET_SHARE_PUBKEYS_FOR_VALIDATOR` and `GET_SHARE_PUBKEYS_FOR_VALIDATOR_INDEX`
- Two new query methods in `share_operations.rs`: `get_share_pubkeys_for_validator()` and `get_share_pubkeys_for_validator_index()`
- Pure data layer additions — no behavior change, no new callers yet (consumers arrive in PRs 2 and 3)

- Minimal risk — new queries only, no existing code modified
- Queries have no callers until PR 2 lands; shipping them first keeps each PR small and independently reviewable

## Validation

- Existing test suite passes (no behavior change)
- Queries follow established patterns in `share_operations.rs`

## Rollback

N/A — pure additive, no runtime impact.

## Blockers / Dependencies

- None — can merge independently

- PR 2 (#933): post-reconstruction signature verification with fallback
- PR 3 (#934): duplicate signature resolution

### DIFF
--- a/anchor/database/src/share_operations.rs
+++ b/anchor/database/src/share_operations.rs
@@ -36,19 +36,7 @@ impl NetworkDatabase {
         let conn = self.connection()?;
         let mut stmt = conn.prepare(sql_operations::GET_SHARE_PUBKEYS_FOR_VALIDATOR)?;
         let mut rows = stmt.query(params![validator_pubkey.to_string()])?;
-
-        let mut result = HashMap::new();
-        while let Some(row) = rows.next()? {
-            let operator_id: u64 = row.get(0)?;
-            let share_pubkey_str: String = row.get(1)?;
-            let share_pubkey = PublicKeyBytes::from_str(&share_pubkey_str).map_err(|e| {
-                DatabaseError::SQLError(format!(
-                    "Invalid share pubkey for operator {operator_id}: {e}"
-                ))
-            })?;
-            result.insert(OperatorId(operator_id), share_pubkey);
-        }
-        Ok(result)
+        Self::collect_share_pubkeys(&mut rows)
     }
 
     /// Fetch all operator share public keys for a given validator index.
@@ -59,7 +47,13 @@ impl NetworkDatabase {
         let conn = self.connection()?;
         let mut stmt = conn.prepare(sql_operations::GET_SHARE_PUBKEYS_FOR_VALIDATOR_INDEX)?;
         let mut rows = stmt.query(params![validator_index])?;
+        Self::collect_share_pubkeys(&mut rows)
+    }
 
+    /// Decode rows of `(operator_id, share_pubkey)` into a map.
+    fn collect_share_pubkeys(
+        rows: &mut rusqlite::Rows<'_>,
+    ) -> Result<HashMap<OperatorId, PublicKeyBytes>, DatabaseError> {
         let mut result = HashMap::new();
         while let Some(row) = rows.next()? {
             let operator_id: u64 = row.get(0)?;
--- a/anchor/database/src/share_operations.rs
+++ b/anchor/database/src/share_operations.rs
@@ -58,7 +58,6 @@ impl NetworkDatabase {
     ) -> Result<HashMap<OperatorId, PublicKeyBytes>, DatabaseError> {
         let conn = self.connection()?;
         let mut stmt = conn.prepare(sql_operations::GET_SHARE_PUBKEYS_FOR_VALIDATOR_INDEX)?;
-        let validator_index: u64 = validator_index.into();
         let mut rows = stmt.query(params![validator_index])?;
 
         let mut result = HashMap::new();
--- a/anchor/database/src/share_operations.rs
+++ b/anchor/database/src/share_operations.rs
@@ -1,6 +1,8 @@
+use std::{collections::HashMap, str::FromStr};
+
 use bls::PublicKeyBytes;
 use rusqlite::{Transaction, params};
-use ssv_types::Share;
+use ssv_types::{OperatorId, Share};
 
 use super::{DatabaseError, NetworkDatabase, sql_operations};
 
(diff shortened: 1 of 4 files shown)


shane-moore:
@claude please review in depth

shane-moore:
@claude, Addressed point 1 — removed the unnecessary `ValidatorIndex` → `u64` conversion (b689d57).

For point 2 (extracting a shared row-parsing helper): skipping this for now. The duplication is only ~8 lines across 2 methods in the same file, and a helper adds indirection for minimal gain. Prefer to keep it inline unless a third caller appears.

jnhsigmap:
Would it be worth giving [`HashMap<OperatorId, PublicKeyBytes>`](https://github.com/sigp/anchor/pull/936/changes#diff-1b3485cef519696c3b80418773b011d596db55bfc90de119e5c9e95aa1b1a748R35) a concrete type alias for cleanliness?

shane-moore:
> Would it be worth giving [`HashMap<OperatorId, PublicKeyBytes>`](https://github.com/sigp/anchor/pull/936/changes#diff-1b3485cef519696c3b80418773b011d596db55bfc90de119e5c9e95aa1b1a748R35) a concrete type alias for cleanliness?

Good suggestion! I considered it but leaning towards leaving it as is. `HashMap<OperatorId, PublicKeyBytes>` is already pretty readable as is imo, and an alias would add indirection for readers. If this type were to be embedded in struct fields for example, that'd perhaps be a good signal to introduce a type alias. but even then, not sure i'd do it, since `HashMap<OperatorId, PublicKeyBytes>` is already pretty simple

anchor/database/src/sql_operations.rs
```diff
@@ -88,6 +88,16 @@ pub const GET_SHARES: &str = r#"
...
 "#;
+pub const GET_SHARE_PUBKEYS_FOR_VALIDATOR: &str = r#"
+    SELECT operator_id, share_pubkey
+    FROM shares WHERE validator_pubkey = ?1
+"#;
+pub const GET_SHARE_PUBKEYS_FOR_VALIDATOR_INDEX: &str = r#"
+    SELECT s.operator_id, s.share_pubkey
+    FROM shares s
+    JOIN validators v ON v.validator_pubkey = s.validator_pubkey
+    WHERE v.validator_index = ?1
```
diegomrsantos, line 99: Could we benchmark this lookup against a production-like DB before merging the first caller PR? `validator_index` is not indexed in the current schema, so this join may end up scanning `validators`. I’m fine with adding the query here, but I’d like at least `EXPLAIN QUERY PLAN` plus timings on a realistically sized DB to decide whether we should add an index or a small cache before this lands on the signature-collector path.

diegomrsantos: One concrete data point from checking the live SSV API on April 8, 2026: mainnet currently reports `126,647` active validators and `1,972` operators (`/api/v4/mainnet/validators/countActiveValidators`, `/api/v4/mainnet/operators/count`). I also sampled 100 validators from `/api/v4/mainnet/validators?page=1&perPage=100`, and all 100 had 4 operators, so a full local DB is likely on the order of ~126k `validators` rows and ~500k `shares` rows, not something tiny.

That said, Anchor's in-memory state only keeps our own shares; the network-wide validator/share view lives on disk. So I no longer think this is obviously too big for SQLite, but I do think it's worth validating on a production-like DB with `EXPLAIN QUERY PLAN` + timings before the first caller PR lands on the signature-collector path.

diegomrsantos:
@shane-moore On the helper-extraction point: I think the benefit here isn't really "wait until we have 3+ callers." We already have one shared row-decoding contract in 2 places, and that's enough for the methods to drift if the selected columns or parse/error handling change later.

Since the helper would stay private and local to this file, the added indirection seems pretty small compared to keeping the decode logic in sync manually.

Low severity from my side, though — the duplicated block is short and adjacent, so this is a maintainability preference rather than something I'd block on.

anchor/database/src/sql_operations.rs
shane-moore: Per your suggestion, I benchmarked the first caller path on a synthetic SQLite DB sized to roughly current SSV mainnet scale: `140k` validator rows and `13` shares per validator (`1.82M` share rows total). This is synthetic rather than a copied real node DB, but it should be in the right order of magnitude for the query shape here.

Without an index on `validators.validator_index`, `EXPLAIN QUERY PLAN` shows:

```sql
SCAN v
SEARCH s USING INDEX sqlite_autoindex_shares_1 (validator_pubkey=?)
```

After adding `CREATE INDEX idx_validators_validator_index ON validators(validator_index)`, the plan becomes:

```sql
SEARCH v USING INDEX idx_validators_validator_index (validator_index=?)
SEARCH s USING INDEX sqlite_autoindex_shares_1 (validator_pubkey=?)
```

For timings, I used 500 repeated lookups to make the difference measurable:

- without index: `~4.10s / 500` lookups, about `8.2ms` per lookup
- with index: `~0.07s / 500` lookups, about `0.14ms` per lookup

So this does confirm the planner concern is real and that an index materially changes the lookup shape. That said, in the first caller PR this query is only used on fallback / duplicate-resolution paths in `signature_collector`, not on the happy path, and the synthetic one-off cost still looks like single-digit ms. Because of that, I don’t think a small cache is warranted here, and I’m not yet convinced this benchmark alone shows we need a schema migration before the first caller lands.

^^ wdyt of this analysis? also curious if you would have taken a similar approach to get the data or if you would have copied one of our mainnet production node's db's as to run the query? or otherwise?

shane-moore:
> @shane-moore On the helper-extraction point: I think the benefit here isn't really "wait until we have 3+ callers." We already have one shared row-decoding contract in 2 places, and that's enough for the methods to drift if the selected columns or parse/error handling change later.
> 
> Since the helper would stay private and local to this file, the added indirection seems pretty small compared to keeping the decode logic in sync manually.
> 
> Low severity from my side, though — the duplicated block is short and adjacent, so this is a maintainability preference rather than something I'd block on.

sure sure, updated in https://github.com/sigp/anchor/pull/936/changes/ff735b2bb7e80df96a3c1328f4956774e366a42a

anchor/database/src/sql_operations.rs
diegomrsantos: This looks good to me. I wouldn’t insist on copying a prod DB for a first pass. For this query shape, the main thing that mattered was confirming whether SQLite really falls back to `SCAN v`, and the synthetic dataset is already the right order of magnitude to answer that.

Looking at `#890`, these lookups are done via `spawn_blocking`, so I don’t think one `8ms` lookup is obviously bad on its own, and I also don’t think this is about blocking a Tokio worker directly. The reason I’m still leaning toward the index is that `NetworkDatabase` currently has `POOL_SIZE = 1`, so if multiple collectors hit duplicate-resolution / fallback around the same time, those lookups still serialize on the single SQLite connection.

That’s why I think the tail case matters more than the single-query number. `~8.2ms` once is probably fine. But once a few of these stack up behind the same connection, the difference gets a lot more noticeable. Using your numbers, `20` lookups is already roughly `~160ms` serialized without the index vs `~3ms` with it.

So I agree this doesn’t justify a cache, and I wouldn’t block `#936` itself on it. But I do think the benchmark is enough to justify adding the `validator_index` index by the first caller PR, or in an immediate follow-up, rather than knowingly shipping the table scan.

shane-moore: valid points! made gh issue to add the `validator_index` index https://github.com/sigp/anchor/issues/940 and will tackle that before the first caller PR

diegomrsantos (dismissed):
Thanks!

anchor/database/src/share_operations.rs
```diff
@@ -22,4 +24,47 @@ impl NetworkDatabase {
             ])?;
         Ok(())
     }
+
+    /// Fetch all operator share public keys for a given validator.
+    ///
+    /// Returns a map from `OperatorId` to the operator's BLS share public key.
+    /// Used during signature reconstruction to verify individual shares.
+    pub fn get_share_pubkeys_for_validator(
```
diegomrsantos, line 32: Non-blocking: can we add a focused DB test for these new query helpers? Even though the callers land in follow-up PRs, a direct test here would lock in the selected column order / row-decoding contract and cover the empty-result path instead of relying only on higher-level integration once the feature stack is complete.

shane-moore:
@claude, can you take a look?

shane-moore:
@claude Re: the `generators::pubkey::random()` fixed-seed observation — agreed it's pre-existing and not introduced here.

We're not planning a follow-up issue for it either. The scenario it would guard against (column-order swap going undetected) can't actually happen silently here: `operator_id` is a `u64` and `share_pubkey` is a BLS hex string, so any swap would fail at `PublicKeyBytes::from_str` before assertions even run. The type mismatch between the two columns is already a sufficient guard, so making the fixture pubkeys distinct would add complexity for a case that's already caught.

anchor/database/src/share_operations.rs
shane-moore: Added `test_get_share_pubkeys` covers the column order / row-decoding contract, and `test_get_share_pubkeys_empty` covers the empty-result path.

diegomrsantos (approved):

PR #937, opened 2026-04-08
refactor: rename committee batch-size field

diegomrsantos:

- The committee signing path used names such as `num_signatures_to_collect`, `collected_signatures`, and `CommitteeSignatures`.
- Those names sounded like the network-side reconstruction path, even though this code is only batching the partial signatures produced locally by one operator before it sends a committee message.
- `SignatureMetadata::threshold` already describes a different concept: how many operator shares must arrive before a full validator signature can be reconstructed.
- That collision in naming made the committee path harder to read than it should be.
- Evidence: the committee attestation test previously had to explain the behavior with an opaque sorted list, `[1, 2, 2]`.

## Change Overview (Required)

- Rename `num_signatures_to_collect` to `validator_partial_signature_batch_size` in `signature_collector` and `validator_store`.
- Rename the local storage from `collected_signatures` to `batched_validator_partial_signatures`, and rename the corresponding local variable to `validator_partial_signature_batch`.
- Rename the private helper type `CommitteeSignatures` to `CommitteePartialSignatureBatch`, and rename its map from `committee_signatures` to `committee_partial_signature_batches`.
- Rewrite the surrounding comments and documentation in plain English so the local batching path is clearly separated from the network reconstruction path.
- Keep the committee attestation test explicit about two facts: we make one `sign_and_collect` call per validator, and each call carries the batch size for that validator's committee.
- What did not change: runtime behavior, batching rules, and reconstruction thresholds.
- Suggested review order:
  - `anchor/signature_collector/src/lib.rs`
  - `anchor/validator_store/src/lib.rs`
  - `anchor/validator_store/src/testing/committee_attestation.rs`

- Main risk: a mechanical rename could miss a call site or leave old terminology behind.
- Trade-off: the new names are longer, but they say what the code is actually doing and no longer collide with `threshold`.
- Mitigation: the affected call sites, comments, and tests were updated together, and the focused validation set was rerun.

## Validation (Required)

- `cargo fmt --all`
- `cargo fmt --all --check`
- `cargo test -p signature_collector`
- `cargo test -p anchor_validator_store committee_`
- `cargo clippy -p signature_collector -p anchor_validator_store --all-targets -- -D warnings`
- The commit hook also reran `cargo fmt --all`, `cargo clippy --all`, and `cargo sort workspace` during the amend.

- Safe rollback: revert commit `6630d23a1`.

- N/A.

- N/A.

### DIFF
--- a/anchor/signature_collector/src/lib.rs
+++ b/anchor/signature_collector/src/lib.rs
@@ -63,10 +63,11 @@ struct SignatureCollector {
     for_slot: Slot,
 }
 
-/// Outgoing partial signature messages that collected for a committee
-/// As soon as the partial signature for every validator in the committee is ready, it is sent.
-struct CommitteeSignatures {
-    collected_signatures: Vec<PartialSignatureMessage>,
+/// Locally accumulated validator partial signatures for one outgoing committee message.
+/// As soon as this operator has produced the full validator batch for the committee round, the
+/// message is sent.
+struct CommitteePartialSignatureBatch {
+    batched_validator_partial_signatures: Vec<PartialSignatureMessage>,
     for_slot: Slot,
 }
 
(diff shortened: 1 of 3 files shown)


shane-moore (approved):
especially like the updated comments!

PR #939, opened 2026-04-09
feat: Implements docgen crate with makefile invocation for checking and updating anchor documentation

jnhsigmap:

- PR for #856 implementing a crate (`docgen`) that generates CLI help documentation from Anchor's `clap` CLI interface.
- Follows #901 and #910 by utilising the CLI definitions in `cli` for generating help documentation.
- #960 prepared Anchor for separation of concern of documentation ownership by creating the `generated` directory in `docs` as a target for machine generated output based upon Anchor's source code.
- This logically follows as it does not alter any behaviour of the Anchor project and does not require booting the `anchor` application to run. It instead is implemented as a standalone binary executable `anchor-docgen`.

## Change Overview (Required)

-  A new crate `docgen` created to programmatically render CLI help docs with outputs stored in the `docs/generated` directory.
  - Includes a post-processor that ingests, groups (by subcommand), and formats canonical outputs from `docgen` into a structure that differs to what is currently used in the `cli-*.mdx` files in `docs/docs/generated`, however is easily rendered in a wrapped code block on the Anchor CLI Reference pages. 
    - `update` optionally writes to the `.mdx` files.
    - `check` checks for consistency.
- `Makefile` targets created for convenience to perform update and check functionality.
  - `cli-reference-update`: The `update` path, modifies `.mdx` files in `docs/docs/pages`.
  - `cli-reference-check`: The `check` path, tests `.mdx` files for consistency with output from `anchor-docgen`.
  - `cli-reference`: A convenience pipeline with both the above commands.
- `next_help_heading` fields added to CLI command line option groups (structs) in `anchor/cli`. Used to define visual grouping of options when `anchor-docgen` produces its output.
- Existing `help_heading = FLAG_HEADER` attributes on boolean options was removed as it silently preceded `next_help_heading`.
- While no changes have occurred to alter the runtime behaviour of `anchor`, the `--help` output text was affected by `next_help_heading` being added and `help_heading` removed. `anchor --help` is now aligned with the outputs of `docgen`.
- Trivial - added `docgen`, `subcommand`, and `stdout` to spellcheck `wordlist`.
 
### Discussion Points

- Actual option descriptions originally defined in `generated` inevitably changes (following #910 which crystallised changes to the existing descriptions and attributes) when `anchor-docgen` generates them going forward due to static descriptions being updated with information defined in the `clap` definitions.
- The `anchor --help` command outputs will no longer contain the `Flags` section for boolean options.
- `anchor-docgen` outputted documentation structure differs materially from existing Anchor CLI Reference on the website. This is a UI/UX question and can be addressed in more detail in a follow up PR.

None, as the main anchor binary does not depend on `docgen`. No runtime behaviours affected, however structure of `--help` has changed.

## Validation (Required)

To render docs:
- `make cli-reference`
- `make cli-reference-update`
- `make cli-reference-check`
- `npm run dev` in `docs/` renders the documentation.

Tests should succeed as normal.
- `cargo test --workspace`
- `cargo clippy`
- `make cargo-fmt-check`

`docgen` crate can be removed without impact on Anchor.

- Regenerate documentation in `docs/docs/generated` to contain machine-owned snippets.
- Implement a CI check workflow to test for documentation consistency.


jnhsigmap:
spellcheck is going to fail again :laughing:

jnhsigmap:
There are 3 big files that warrant review
- `checks.rs`
- `format.rs`
- `render.rs`

everything else should be safe. I went through a few iterations with claude and fixed the primary issues that it raised. Notably there are ~300 lines of tests, simply to ensure that 'all cases are canvassed' in some areas (i.e. the short and long option combinations) so that issues don't arise later. for me it's more boilerplate than anything that's going to cause too much debt in the future, but open to ideas :)

diegomrsantos:
I found a correctness bug in the generator that also highlights why this PR is hard to review as a single unit. Required clap args are currently rendered as `Default: None`, which changes their meaning in the generated docs. In `keysplit`, required args like `rpc`, `owner`, `output_path`, `operators`, and `public_keys` now read as if they were optional. That is not just formatting drift; it is a mismatch between the CLI contract and the reference docs.

This kind of issue is much easier to miss when one PR combines the new `docgen` crate, the marker/scaffolding changes in the mdx files, and the large mechanical regeneration of the docs. We are moving toward smaller, easier-to-review PRs, and I think this one should be split before going further.

Suggested split:
1. A small PR that only adds the `CLI_REFERENCE_START` / `CLI_REFERENCE_END` markers to the mdx pages.
2. A PR for the `docgen` implementation itself, covering `format.rs`, `render.rs`, `checks.rs`, tests, and Makefile wiring.
3. A final mechanical PR that regenerates and commits the mdx output.

If needed, the implementation PR could be split one more time into rendering logic versus update/check plumbing. That would make correctness issues like the required-arg bug much easier to spot and discuss.

jnhsigmap:
> I found a correctness bug in the generator that also highlights why this PR is hard to review as a single unit. Required clap args are currently rendered as `Default: None`, which changes their meaning in the generated docs. In `keysplit`, required args like `rpc`, `owner`, `output_path`, `operators`, and `public_keys` now read as if they were optional. That is not just formatting drift; it is a mismatch between the CLI contract and the reference docs.
> 
> This kind of issue is much easier to miss when one PR combines the new `docgen` crate, the marker/scaffolding changes in the mdx files, and the large mechanical regeneration of the docs. We are moving toward smaller, easier-to-review PRs, and I think this one should be split before going further.
> 
> Suggested split:
> 
> 1. A small PR that only adds the `CLI_REFERENCE_START` / `CLI_REFERENCE_END` markers to the mdx pages.
> 2. A PR for the `docgen` implementation itself, covering `format.rs`, `render.rs`, `checks.rs`, tests, and Makefile wiring.
> 3. A final mechanical PR that regenerates and commits the mdx output.
> 
> If needed, the implementation PR could be split one more time into rendering logic versus update/check plumbing. That would make correctness issues like the required-arg bug much easier to spot and discuss.

fair enough, I originally created that function to say "There are no defaults for this flag" but following what it currently looks like, an empty column is better as `None` is a type.

I'm most interested your thoughts on the _content of the description fields_ as they're the most change-heavy and there may have been some legacy semantic or behavioural drift between what is currently in the documentation and what is stated in the struct definitions. I decided to use the `help` fields to generate them as there needs to be a link to the source code somewhere.

Unfortunately I designed this with a lot of coupling between rendering the documentation and post-processing as I saw more importance in ensuring that there was **no opportunity** to render them differently in another execution path when writing/checking to rendering/ingesting than interface separation. Any drift would jeopardise the whole process as the CI job would see a different structure to the post-processing pipeline. This is irreconcilable on surface level inspection.

You can see this where the default `generate` flag, `run_check` (`--check`), and `run_update` (`--update`) all depend on `generate_cli_page_content` and `generate_subcommand_page_content` to ensure that the docs are rendered the same way regardless of the code path. This follows to the internals where `render` depends on `format` to ensure the same level of consistency.

The `update` and `check` flags use logic implemented in `checks` that doesn't have dependants. Excluding these would leave `docgen` as a command that renders the documentation from `cli` to `stdout`. 

My suggested split -
- PR 1. Combine `CLI_REFERENCE_` and automated regeneration of the content in the docs based on what's currently in `cli`. This will allow scrutiny of all changes to docs that occur. The headers are an implementation detail as a requirement from `docgen`. They wrap the docs content that should and will change when the automated tool is run. **The change that needs analysis is the documentation content that has changed.**

- PR 2. `docgen` crate. As mentioned, the place where this can be split is only removing `checks.rs`, and probably `README.md` as a cosmetic component as it's there to explain what the `check` and `update` flags do. I'd push back a bit on this split as `checks.rs` is only 85 lines of code (excluding tests) and is more mechanistic than an important implementation detail as risk can be removed by test coverage. In `checks.rs` there are also only 2 functions accessed by callers from other modules.

What are your thoughts @diegomrsantos ? :slightly_smiling_face:

jnhsigmap:
I pushed a fix to address the Defaults issue too!

Makefile
```diff
@@ -105,6 +105,14 @@ check-benches:
 # test vectors.
 test: test-release
 
+# Update CLI reference docs from current clap definitions
+cli-reference:
+	cargo run --bin anchor-docgen -- update
```
chong-he: I wonder should we build with the release feature `--release`? I tested building the binary and it comes to 250MB for default build (debug). With `--release`, it is just ~8MB

The same goes to `cli-reference-check`. (but if you or the Anchor team decides to combine check and update, then that's this to change here, which is simpler)

anchor/docgen/src/lib.rs
```diff
@@ -0,0 +1,106 @@
...
+    Ok(())
+}
+
+/// Renders documentation and updates/checks existing documentation based on the provided
+/// `DocGenCommand`.
+pub fn render_docs(
+    docgen_command: Option<DocGenCommand>,
+    anchor_command: &Command,
+) -> Result<(), DocGenError> {
+    match docgen_command.unwrap_or(DocGenCommand::Generate) {
```
chong-he: With the binary, if we just run: `anchor-docgen`, it will also generate the doc. This is not very user friendly. For example, if we run `anchor`, it display the available commands. It would be good if the binary can do something similar like the `anchor` binary

anchor/docgen/src/lib.rs
```diff
@@ -0,0 +1,106 @@
...
+        let content = generate_subcommand_page_content(cmd, name)?;
+        update_file(&docs_dir.join(file), &content)?;
+    }
+
+    eprintln!("CLI reference documentation updated successfully.");
+    Ok(())
+}
+
+/// Checks if the .mdx files are up to date with the current CLI definitions.
+fn run_check(cmd: &Command, docs_dir: &Path) -> Result<(), DocGenError> {
```
chong-he: In practice, the developers and users would probably prefer the option of a 1 step check+update, instead of separating the check and update. So we can combine the check and update together, which I think is better. 

Something like: when we run `anchor-docgen update` or `make cli-reference`, it will include both check and update, there are 2 possible outcomes:
- if all good, it displays: `CLI reference documentation is up to date.`
- if needs update, it will also updates it and display what files are updated

This can help to simplify things also

anchor/docgen/src/format.rs
```diff
@@ -0,0 +1,404 @@
...
+        "FileLoggingFlags" => "Logging Options".to_string(),
+        _ => split_pascal_case(group_id),
+    }
+}
+
+/// Split a PascalCase identifier into space-separated words.
+/// e.g. "SecurityOptions" → "Security Options"
+/// In addition, handles consecutive uppercase letters as a single word, e.g. "HTTPApi" → "HTTP
+/// Api".
+fn split_pascal_case(s: &str) -> String {
```
chong-he: This feels a little bit over complicated to me, that having this whole function just to format a single line of heading. This is nothing wrong, but I just feel it's a bit of unnecessary to have to handle like this for a single line of heading. 

Maybe we we could just match the other headings using the function `group_display_name`?  Though a bit of manual, but so long as there is no new headings in the future, then there should be very low or no maintenance required on this part in the future

anchor/docgen/src/lib.rs
```diff
@@ -0,0 +1,106 @@
...
+fn run_update(cmd: &Command, docs_dir: &Path) -> Result<(), DocGenError> {
+    let cli_content = generate_cli_page_content(cmd)?;
+    update_file(&docs_dir.join("cli.mdx"), &cli_content)?;
+
+    for (name, file) in SUBCOMMAND_PAGES {
+        let content = generate_subcommand_page_content(cmd, name)?;
+        update_file(&docs_dir.join(file), &content)?;
+    }
+
+    eprintln!("CLI reference documentation updated successfully.");
```
chong-he: This is printed regardless of there is an update or not, which could be confusing to the users (when there is no update).

chong-he (commented):
I left some comments below, feel free to take it or leave it as the Anchor team decides.

diegomrsantos:
One way to make this easier to review without giving up a single render path for `generate`, `update`, and `check` would be to split it into 3 PRs:

1. Integration / ownership
Keep `docs/docs/pages/cli*.mdx` as the canonical user-facing pages, move the generated option reference into machine-owned MDX snippets under `docs/docs/generated/*.mdx`, and import those snippets from the pages. Seed the snippets from the current docs so this PR does not also change wording, defaults, headings, or examples. The review question here is only: is this the right ownership boundary and integration model?

2. Generator semantics
Change `docgen` itself: formatting, grouping, required/default rendering, description sourcing, and any other behavior that affects what the generated reference means. The review question here is: what should the generated reference actually say?

3. Mechanical regeneration
Regenerate the snippet files and commit only the generated artifacts. The review question here is simply whether the checked-in generated output matches the approved generator behavior.

That keeps the good part of the current design, namely one canonical render path, while separating three different review concerns that are currently mixed together: how the docs are wired, what the generator means semantically, and the large mechanical output diff.

anchor/docgen/src/lib.rs
shane-moore: i think check and update are separate so that when we can add CI which will run check and fail if we didn't manually run update before pushing the commit

chong-he: Ah I see, sure, feel free to do the way the Anchor team prefers. 

In Lighthouse what we do is: check and update at the same time, if there is update then `exit 1` so CI will fail: https://github.com/sigp/lighthouse/blob/176cce585c1ba979a6210ed79b6b6528596cdb8c/scripts/cli.sh#L92-L94

shane-moore: ah, nice, makes sense. i'm cool with that approach as well

jnhsigmap: I see a lot of utility in having the ability to check or update as separate commands too as it would answer the question for the developer of "have my changes affected the CLI interface?" and "what differences to the documentation will this create and do I agree?" which would be a consideration if I was working on `cli`.

I agree that the standard `anchor-docgen` call (maybe `run`? or something along those lines?) should do both though as before I commit and push I'll run that to verify what I have done is reflected there correctly :slightly_smiling_face: so good suggestion.

I will still have CI just run `check` for differences and report/fail if there are any. It's ultimately up to the developer to ensure that they understand all they have changed when the PR is being reviewed though.

chong-he: All good, it's your call, I am only giving suggestion here

Makefile
jnhsigmap: thanks mate, added this :+1:

anchor/docgen/src/lib.rs
jnhsigmap: and this :+1:

anchor/docgen/src/format.rs
jnhsigmap: This was a deliberate choice for generality sake as my objective was to not alter the core structure but still allow things to grow naturally. @diegomrsantos or @shane-moore might be able to better predict the likelihood of this being an issue. How often are things added?

anchor/docgen/src/lib.rs
jnhsigmap: thanks :+1:

jnhsigmap:
> One way to make this easier to review without giving up a single render path for `generate`, `update`, and `check` would be to split it into 3 PRs:
> 
> 1. Integration / ownership
>    Keep `docs/docs/pages/cli*.mdx` as the canonical user-facing pages, move the generated option reference into machine-owned MDX snippets under `docs/docs/generated/*.mdx`, and import those snippets from the pages. Seed the snippets from the current docs so this PR does not also change wording, defaults, headings, or examples. The review question here is only: is this the right ownership boundary and integration model?
> 2. Generator semantics
>    Change `docgen` itself: formatting, grouping, required/default rendering, description sourcing, and any other behavior that affects what the generated reference means. The review question here is: what should the generated reference actually say?
> 3. Mechanical regeneration
>    Regenerate the snippet files and commit only the generated artifacts. The review question here is simply whether the checked-in generated output matches the approved generator behavior.
> 
> That keeps the good part of the current design, namely one canonical render path, while separating three different review concerns that are currently mixed together: how the docs are wired, what the generator means semantically, and the large mechanical output diff.

I've addressed @chong-he 's feedback and made a few changes, will get into this tomorrow, looks well thought out :+1:

jnhsigmap:
what's the deal with this?
https://github.com/sigp/anchor/actions/runs/24451620292/job/71441673300?pr=939
no dependencies changed.

chong-he:
> what's the deal with this? https://github.com/sigp/anchor/actions/runs/24451620292/job/71441673300?pr=939 no dependencies changed.

A new audit failure, Lighthouse also faces the same issue: https://github.com/sigp/lighthouse/pull/9130

jnhsigmap:
#960 will address point 1. that will then be merged into this which will be the PR for point 2.

> One way to make this easier to review without giving up a single render path for `generate`, `update`, and `check` would be to split it into 3 PRs:
> 
> 1. Integration / ownership
>    Keep `docs/docs/pages/cli*.mdx` as the canonical user-facing pages, move the generated option reference into machine-owned MDX snippets under `docs/docs/generated/*.mdx`, and import those snippets from the pages. Seed the snippets from the current docs so this PR does not also change wording, defaults, headings, or examples. The review question here is only: is this the right ownership boundary and integration model?
> 2. Generator semantics
>    Change `docgen` itself: formatting, grouping, required/default rendering, description sourcing, and any other behavior that affects what the generated reference means. The review question here is: what should the generated reference actually say?
> 3. Mechanical regeneration
>    Regenerate the snippet files and commit only the generated artifacts. The review question here is simply whether the checked-in generated output matches the approved generator behavior.
> 
> That keeps the good part of the current design, namely one canonical render path, while separating three different review concerns that are currently mixed together: how the docs are wired, what the generator means semantically, and the large mechanical output diff.

anchor/docgen/src/render.rs
```diff
@@ -0,0 +1,242 @@
...
+            write_arg_table_row(&mut output, arg)?;
+        }
+        writeln!(output).map_err(|e| DocGenError::RenderOptionGroup {
+            group: group_name,
+            source: e,
+        })?;
+    }
+    Ok(output)
+}
+
```
diegomrsantos: All the `writeln!(output, ...).map_err(|e| DocGenError::RenderOptionGroup { ... })?` calls in this file (and the matching ones in `format.rs`) are handling an error that cannot occur. `impl fmt::Write for String` is infallible — it just calls `push_str` and returns `Ok(())` — so the `Err` branch is statically unreachable, but we still pay for it with a public `DocGenError::RenderOptionGroup` variant, per-call closures, and `group_name.clone()` on the happy path.

The idiomatic fix is to drop the `Result` from the string-building helpers and use `push_str` / `write!` + `.unwrap()` (or `writeln!(...).expect("infallible")`). For example:

```rust
fn generate_formatted_option_table_doc(
    groups: &[(Option<String>, Vec<&Arg>)],
    heading_prefix: &str,
) -> String {
    let mut output = String::new();
    for (heading, group_args) in groups {
        if let Some(heading_text) = heading {
            writeln!(output, "{heading_prefix} {heading_text}\n").unwrap();
        }
        output.push_str("| Option | Description | Default |\n");
        output.push_str("| --- | --- | --- |\n");
        for arg in group_args {
            write_arg_table_row(&mut output, arg);
        }
        output.push('\n');
    }
    output
}
```

Functions that have a real failure mode (`SubcommandNotFound`, `format_description` if you want to keep the possible-values path fallible) keep `Result<_, DocGenError>`; the pure string-writers lose it. Net effect: `DocGenError::RenderOptionGroup` variant can be deleted, ~30 LOC of `map_err` plumbing disappears, and the `group_name.clone()`s go away.

If `docgen` ever switches its sink from `String` to `io::Write`, the errors become real — but that's a different error type (I/O, not `fmt::Error`) and should be introduced then, with accurate context.

anchor/docgen/src/format.rs
diegomrsantos: I agree with @chong-he that this is overcomplicated, but I'd go a step further than a manual match table — both `split_pascal_case` and `group_display_name` can be deleted entirely.

`clap` already has a first-class mechanism for this: `next_help_heading`. The `cli` crate already uses `help_heading` on individual args (13 occurrences today). Adding `next_help_heading` at the struct level gives docgen the display name directly:

```rust
// in cli/src/cli.rs
#[derive(Parser, Clone, Debug)]
#[clap(next_help_heading = "HTTP API")]
pub struct HttpApiOptions { ... }
```

Then in `docgen`, grouping becomes: read `arg.get_help_heading()` — it returns the literal string. No PascalCase splitting, no override table, no drift when someone adds a new struct.

Benefits:
- **Single source of truth** — the heading appears in `--help` output *and* generated docs from one declaration.
- **Reviewable where it matters** — a PR adding a new options struct in `cli/` carries its heading with it, impossible to forget.
- **~45 LOC deleted** — `split_pascal_case`, its tests, and `group_display_name` all go away.

The only trade-off is ~1 line per struct in `cli/`, which is where you'd want to review the heading anyway.

jnhsigmap: This makes sense, best we follow clap's recommendation to use `help_heading` for visual aid structuring and not Arg groups, which are apparently meant for internal reference and validation.

@diegomrsantos, some side effects were encountered from doing this. The existing use of `help_heading` collides with our intended use as `anchor node --help` requires them to organise its headings using a field called `Flags`, however in our case that results in a separate section containing the 13 existing fields grouped into one section incorrectly. I'll try find a way to keep both, but it looks like these are competing objectives

jnhsigmap: I'll revert my last commit, it's insufficient alone

anchor/docgen/src/render.rs
jnhsigmap: yep this was a good catch. I can confirm that is true of those `RenderOptionGroup` checks. Happy path can safely call `.expect` while the bytes are just basic text/strings and not panic in all reasonable cases. The sink should be able to remain as `String`, I can't think of simple cases where `io::Write` would be needed here.

anchor/docgen/src/format.rs
jnhsigmap: Regarding this point, we have opted for option A in @shane-moore 's [note](https://hackmd.io/@blockshane/SybGzpP6Zx). 
Side effects of this change were a change to the `anchor --help` call where it is now aligned to the output of `docgen` which allows use of `get_help_heading` as @diegomrsantos mentioned above. This provides benefits of simplicity and a single place where help descriptions are defined and arranged.

diegomrsantos:
Thanks for the updates here. After looking at the latest version again, I am less convinced that the custom table generator is worth the amount of code and maintenance.

The tricky part is that once `docgen` formats options itself, it has to reproduce a lot of Clap behavior correctly. We already have an example with repeated values: Clap renders `--listen-addresses [<ADDRESS>...]`, `--keystore-paths <PATH>...`, and `--public-keys <KEYS>...`, while the generated table currently flattens those to plain `<VALUE>` forms. That is fixable, but it points to the broader maintenance cost.

I wonder if a simpler approach would be better: keep the generated snippets, but have `docgen` ask Clap to render the help text directly with `render_long_help()`, then write that into fenced `text` blocks in the MDX files. The hand written docs pages can still provide examples and explanation around those snippets, but the reference block would be exactly what users see from the binary.

The tradeoff is that terminal help blocks are less polished than tables. The upside is that Clap remains the source of truth for usage lines, required args, defaults, possible values, headings, and repeated values. It would also delete most of the formatter code here.

Curious what you think. Would you be open to simplifying this PR in that direction, or do you feel the table format is important enough to justify maintaining the custom formatter?

jnhsigmap:
> Thanks for the updates here. After looking at the latest version again, I am less convinced that the custom table generator is worth the amount of code and maintenance.
> 
> The tricky part is that once `docgen` formats options itself, it has to reproduce a lot of Clap behavior correctly. We already have an example with repeated values: Clap renders `--listen-addresses [<ADDRESS>...]`, `--keystore-paths <PATH>...`, and `--public-keys <KEYS>...`, while the generated table currently flattens those to plain `<VALUE>` forms. That is fixable, but it points to the broader maintenance cost.
> 
> I wonder if a simpler approach would be better: keep the generated snippets, but have `docgen` ask Clap to render the help text directly with `render_long_help()`, then write that into fenced `text` blocks in the MDX files. The hand written docs pages can still provide examples and explanation around those snippets, but the reference block would be exactly what users see from the binary.
> 
> The tradeoff is that terminal help blocks are less polished than tables. The upside is that Clap remains the source of truth for usage lines, required args, defaults, possible values, headings, and repeated values. It would also delete most of the formatter code here.
> 
> Curious what you think. Would you be open to simplifying this PR in that direction, or do you feel the table format is important enough to justify maintaining the custom formatter?

yeah, absolutely valid. It would be frustrating to have some internal changes to clap between minor or major versions to completely derail this feature. Especially given that future time invested in this won't yield the same return for us as core production code.

Should we run this by @jking-aus ? Otherwise give me your thoughts on the below?

I have created a PoC of this suggestion and the trade-off is purely cosmetic (user experience) vs the additional maintenance burden. The formatting requirement collapsed to a simple `format!("```text\n{help}\n```\n")` call where a `clap::builder::StyledStr` was passed.

Noting that we are currently coupled to several calls such as `get_arguments()`, `get_help_heading()`, `get_subcommands()`, and some internal components of a clap `Command` such as `is_positional()` and `is_hide_set()`. This is a trade-off between visual appearance and some complexity in these `clap` internals.

Example:

**`render_long_help` generated:**
image

**`stable` branch:**
image

diegomrsantos:
Thanks for trying the PoC. That mostly settles the tradeoff for me.

I think we should move forward with Clap-rendered snippets, using the shorter `render_help()` output if `render_long_help()` feels too wordy in the docs. The important part for me is that Clap owns the formatting and semantics, not that we specifically use the long help variant.

I also do not think this needs to block on @jking-aus unless he wants to weigh in. This feels like a small docs tooling decision that we can make as the Anchor team. The table format looks nicer, but it comes with custom Clap formatting logic that we would have to maintain. I do not think the visual improvement is worth that extra code.

jnhsigmap:
> Thanks for trying the PoC. That mostly settles the tradeoff for me.
> 
> I think we should move forward with Clap-rendered snippets, using the shorter `render_help()` output if `render_long_help()` feels too wordy in the docs. The important part for me is that Clap owns the formatting and semantics, not that we specifically use the long help variant.
> 
> I also do not think this needs to block on @jking-aus unless he wants to weigh in. This feels like a small docs tooling decision that we can make as the Anchor team. The table format looks nicer, but it comes with custom Clap formatting logic that we would have to maintain. I do not think the visual improvement is worth that extra code.

cool, I think we should keep to using `render_long_help` as `render_help` truncates the output to a single line, while `render_long_help` formats with multi-line layout. That is more suitable for a documentation page.

I have identified a couple of subsequent issues that should be addressed as a result of going through this exercise that I'll create separate issues for -
- We mostly just use `help` descriptions in the CLI struct definitions, which means typically `render_long_help` and `render_help` will yield the same descriptions. It may be a good idea to generate more concise summaries in `help` that render shorter snippets when `anchor -h` is run.
- Some of the descriptions are stale and no longer reflect the behaviour of their respective paths execution. This is likely due to the legacy way of maintaining docs where this would have been easily missed.

Given we've made some changes to the architecture of the CLI it makes sense to address these issues once this PR has been merged to avoid having to resolve conflicts and have a clean slate for discussion.

jnhsigmap:
@diegomrsantos I'm still struggling with the need to scroll slightly left and right (as shown in the image above) to view completely the full descriptions in the code snippet on the CLI reference pages.

I've tried working through some overrides at the `.css` level but none seem to result in changes to that page element. 

I'm not very experienced with front-end so I don't have an intuitive direction to go in but the change to the page seems like quite a big downgrade in UI and as a user I'd be a little frustrated with that layout. My experience would now be better in the terminal. 

do you guys have a final view on this? I've pushed the changes that you can run locally to see. my process to start the docs server was
```
# Required for the version of vocs.
$ nvm use 22;

$ cd docs;

$ npm run dev;
```

shane-moore:
fixed the mobile/scroll UX in febccb7.

clap render now emits descriptions as single long lines (`term_width(MAX)`
+ `next_line_help(false)` for docs only); a small CSS+JS shim in
`vocs.config.ts` hangs wrapped lines at the original indent column so
clap's hierarchy holds across viewports.

tradeoff: the shim hooks Shiki's `.line` spans, so a future deliberate
vocs upgrade could break it (lockfile pins us to `1.0.13` today, so
nothing drive-by). nothing more native available that I found. open
to other ideas

jnhsigmap:
> fixed the mobile/scroll UX in [febccb7](https://github.com/sigp/anchor/commit/febccb729b6eb00fdf1819f1cf0c7af0662cd5ee).
> 
> clap render now emits descriptions as single long lines (`term_width(MAX)`
> 
> * `next_line_help(false)` for docs only); a small CSS+JS shim in
>   `vocs.config.ts` hangs wrapped lines at the original indent column so
>   clap's hierarchy holds across viewports.
> 
> tradeoff: the shim hooks Shiki's `.line` spans, so a future deliberate vocs upgrade could break it (lockfile pins us to `1.0.13` today, so nothing drive-by). nothing more native available that I found. open to other ideas

Nice :slightly_smiling_face: unfortunately it comes with fragility in that it depends directly on client-side DOM elements, and subsequently changes to Shiki or Vocs will break the injected JS functions.. To me that drifts from our core competency and introduces a bit of uncomfortable React maintenance that could be avoided.

It gave me another line of interrogation to follow though.

**Summary**
Aim: Replace brittle client-side JS/CSS dependency with Rust code that achieves the same visual result - readable CLI reference, clear indentation hierarchy, properly wrapped descriptions with no coupling to Shiki or Vocs internals.                                                             

Instead of injecting a `.js` script and re-applying whitespace as indentation, I instead opted for a Rust function to handle boosting indentation by setting a slightly lower starting indentation and than adding a slightly larger indent boost within the fenced code block.

Your solution was good, but hopefully this way we can keep our volatile dependencies to our known Rust crates that can be intuitively managed. This doesn't require as much boilerplate as the `md` tables approach from the original PR.

Let me know what you guys think.

anchor/docgen/src/lib.rs
```diff
@@ -0,0 +1,143 @@
...
+use interface::DocGenCommand;
+use render::{render_help_string, render_subcommand_help_snippet, to_title_case};
+
+const CLI_REFERENCE_FILE: &str = "cli-global-options.mdx";
+
+/// Subcommand-to-file mapping for generated CLI reference snippets.
+const SUBCOMMAND_REFERENCE_PAGES: [(&str, &str); 3] = [
+    ("node", "cli-node-options.mdx"),
+    ("keygen", "cli-keygen-options.mdx"),
+    ("keysplit", "cli-keysplit-options.mdx"),
```
diegomrsantos, line 21: I think `keysplit` needs one more level of rendering before this can merge. This mapping renders only the parent `anchor keysplit` help, so `cli-keysplit-options.mdx` lists `manual` and `onchain` as commands but does not document either child command.

The actual required flags are under the child commands: `anchor keysplit manual --help` includes `--nonce`, `--public-keys`, and the shared keysplit options like `--owner`, `--output-path`, and `--operators`; `anchor keysplit onchain --help` includes `--rpc` plus the shared options. None of those appear in the generated keysplit snippet at the moment.

Since `anchor-docgen check` passes in this state, the check is currently proving that the checked-in file matches the incomplete generator output, not that the keysplit reference is complete. Could we render nested help for `keysplit manual` and `keysplit onchain` here, or otherwise special-case nested commands so the generated reference includes those flags?

docs/docs/pages/installation.mdx
```diff
@@ -35,8 +35,8 @@ wget https://github.com/sigp/anchor/releases/download/<version>/anchor-<version>
 tar -xvf anchor-<version>-<platform>.tar.gz
 
 # Specific version example
-wget https://github.com/sigp/anchor/releases/download/v1.1.0/anchor-v1.1.0-unknown-linux-gnu.tar.gz
-tar -xvf anchor-v1.1.0-unknown-linux-gnu.tar.gz
+wget https://github.com/sigp/anchor/releases/download/v1.2.3/anchor-v1.2.3-gnu.tar.gz
```
diegomrsantos: This example now points at a release asset that does not exist. The `v1.2.3` release uses full target triples, for example `anchor-v1.2.3-x86_64-unknown-linux-gnu.tar.gz` and `anchor-v1.2.3-aarch64-unknown-linux-gnu.tar.gz`; there is no `anchor-v1.2.3-gnu.tar.gz` asset.

Can we keep the concrete example on one of the real asset names, or leave this as the generic `<platform>` form?

docs/README.md
```diff
@@ -77,10 +77,10 @@ npm run preview
 
 All documentation content is in `docs/pages/` as `.mdx` files.
 
-Specific CLI option detail is stored as separate imported components:
+CLI option details are stored as separate components:
 
-- `docs/docs/generated/*.mdx` contain specific CLI option implementation details generated using the structure of Anchor CLI source code.
-- The `docs/docs/pages/cli*.mdx` pages remain hand-written and import those generated snippets.
+- `docs/docs/generated/*.mdx` are machine-owned CLI option information snippets generated using the `achor-docgen` binary.
```
shane-moore: `anchor-docgen` instead of `achor-docgen`

anchor/docgen/README.md
```diff
@@ -0,0 +1,50 @@
...
+| Subcommand | Description |
+| --- | --- |
+| `generate` | Print generated reference snippets to stdout (default when no subcommand is given) |
+| `update` | Write generated reference snippets into `docs/docs/generated/*.mdx` files |
+| `check` | Verify generated reference snippets match current CLI definitions; exits non-zero if stale |
+
+## How it works
+
+1. Builds the full `clap::Command` tree from `cli::Cli` via `CommandFactory`. This leverages struct definitions in the `cli` crate.
+2. Groups arguments by their `clap` `ArgGroup` (derived from `#[derive(Parser)]` structs).
```
shane-moore: think this is a bit stale. Lines 39-41 describe the rejected table-formatter:

▎ 2. Groups arguments by their clap ArgGroup (derived from #[derive(Parser)] structs).
▎ 3. Renders each group as a markdown table with Option, Description, and Default columns.

Current impl is clap::Command::render_long_help() → fenced text

Makefile
```diff
@@ -113,6 +113,17 @@ check-benches:
 # test vectors.
 test: test-release
 
+# Update generated CLI reference snippets from current clap definitions
+cli-reference-update:
+	cargo run --release --bin anchor-docgen update
+
+# Check generated CLI reference snippets are up to date+
```
shane-moore: can we remove the + at the end of the comment?

anchor/docgen/src/render.rs
```diff
@@ -0,0 +1,168 @@
...
+use crate::errors::DocGenError;
+
+/// Target rendered line width minus some arbitrary amount.
+/// Clap wraps at this narrower width to leave room for a small indent boost.
+const WRAP_WIDTH: usize = 74;
+/// Extra spaces added per indent level so Clap's 2-space hierarchy reads clearly in the docs.
+const INDENT_BOOST: usize = 2;
+
+pub(crate) fn render_help_string(command: &mut Command) -> String {
```
shane-moore: render_help_string takes &mut Command but only does command.clone(). Could be &Command. (Call sites that thread through find_subcommand_mut would need to switch to find_subcommand.)

anchor/docgen/src/lib.rs
```diff
@@ -0,0 +1,143 @@
...
+        let content = render_subcommand_help_snippet(cmd, name, &cli_name)?;
+        append_to_out_of_date_files(&docs_dir.join(file), file, &content, &mut out_of_date)?;
+    }
+
+    check_out_of_date_files(&out_of_date)
+}
+
+/// Renders generated CLI reference snippets from the `clap` struct definitions to stdout.
+fn display_help_docs(anchor_command: &mut Command) -> Result<(), DocGenError> {
+    let cli_content = &anchor_command.render_long_help();
```
shane-moore: `display_help_docs` bypasses `render_help_string` for the global section only — the subcommand calls below go through the pipeline (`render_subcommand_help_snippet` → `render_help_string`), but this line uses raw `render_long_help()`.

Effect: `anchor-docgen` generate previews the global section as terminal help (no fence, no indent boost) while update writes the fenced/boosted form to `cli-global-options.mdx`. Preview doesn't match disk for the global page; matches for subcommand pages.

Looks like a missed rename during a682eed — the unified call existed before that commit.

 -    let cli_content = &anchor_command.render_long_help();
 +    let cli_content = render_help_string(anchor_command);

but maybe i'm missing something

diegomrsantos:
After looking at the full diff again, I want to clarify my comment on the bad `v1.2.3` release asset.

The immediate issue is that `anchor-v1.2.3-gnu.tar.gz` does not exist. But I do not think the best fix is just to patch that filename in this PR. The broader issue is that this PR is about CLI reference generation, while the changes to the homepage version, stars, contributor count, install example, and Vocs nav version look like `sync-version` output rather than part of `anchor-docgen`.

Can we remove that unrelated docs version/stat churn from this PR unless it is required for the docs build? If it is required, then the PR should say so explicitly and fix the sync path so it preserves real release asset names.

More broadly, I do not think the amount of review time going into this PR is proportional to the value of the change. The useful part is the generated CLI reference path. We should keep the remaining scope to correctness issues that affect that path, avoid further cosmetic iteration here, and move anything else to follow-up issues.

diegomrsantos:
One more scope clarification: I think we had agreed that this PR would not regenerate the checked-in docs content.

The split we discussed was: #960 handles the docs ownership/integration, this PR handles the `docgen` implementation and generator behavior, and a later mechanical PR regenerates the MDX output once the generator behavior is approved.

Current head still changes `docs/docs/generated/*.mdx` substantially. That makes this PR both the generator PR and the mechanical regeneration PR again, which is exactly the review shape we were trying to avoid. It also makes bugs like the missing nested `keysplit` flags harder to separate from the generated-output diff.

Can we drop the generated MDX changes from this PR and keep this focused on the generator code, Makefile wiring, and tests? Once the generator behavior is settled, a follow-up PR can run `anchor-docgen update` and commit only the generated output.

anchor/docgen/src/lib.rs
jnhsigmap: yep I think that was originally to not render the nested help documentation. Within addressing @diegomrsantos' request regarding the `keysplit` nested commands, I was able to address this too. In summary, call site changes to `let cli_content = render_help_flat(anchor_command);`

anchor/docgen/src/render.rs
jnhsigmap: Nice I have made this change where `clap` doesn't require access internally to a `&mut`.

anchor/docgen/README.md
jnhsigmap: yeah agreed, I've updated `README` with new detail.

docs/docs/pages/installation.mdx
jnhsigmap: I will remove this from HEAD and restore to `unstable`'s state. I'm not sure if it's required for the docs build..

anchor/docgen/src/lib.rs
jnhsigmap: addressed by changing `render_help_string` to two specific functions
- `render_help_with_subcommands`
- `render_help_flat`
Self explanatory, one recursively renders everything to exhaustion, the other one only renders top level

jnhsigmap:
> One more scope clarification: I think we had agreed that this PR would not regenerate the checked-in docs content.
> 
> The split we discussed was: #960 handles the docs ownership/integration, this PR handles the `docgen` implementation and generator behavior, and a later mechanical PR regenerates the MDX output once the generator behavior is approved.
> 
> Current head still changes `docs/docs/generated/*.mdx` substantially. That makes this PR both the generator PR and the mechanical regeneration PR again, which is exactly the review shape we were trying to avoid. It also makes bugs like the missing nested `keysplit` flags harder to separate from the generated-output diff.
> 
> Can we drop the generated MDX changes from this PR and keep this focused on the generator code, Makefile wiring, and tests? Once the generator behavior is settled, a follow-up PR can run `anchor-docgen update` and commit only the generated output.

I've reset changes to `docs` to what the current state of `unstable` was. The only file remaining is `docs/docs/styles.css` which was updated as part of the code-block wrapping change. They are coupled, so probably necessary.

shane-moore (approved):
LGTM. Nice work resolving the nested `keysplit` rendering. `render_help_with_subcommands` is a clean way to do it, and the split into `render_help_flat` for the global page also resolves the `display_help_docs` divergence as a bonus. Generator-only scope reads well now.

jnhsigmap:
I'll wait for @diegomrsantos' blessing before we merge :slightly_smiling_face:

jnhsigmap:
I'm going to draw the line here to merge this as I've got some follow up work to consider. If there's anything pressing please just ping me or reply for a follow up PR. It's reached the stage where any further discussion is on cosmetic detail and can be addressed in the future :slightly_smiling_face:

Issue #938, opened 2026-04-08
Add a Makefile target for spec tests with fake_crypto coverage


shane-moore:
## Summary

PR [#839](https://github.com/sigp/anchor/pull/839) adds `ConsensusDataProposerTest`, but the fuller success-path coverage for that test currently requires `--features fake_crypto` and there is no dedicated Makefile target for it.

Today the default test entry points still run with empty `TEST_FEATURES`:
- [Makefile](https://github.com/sigp/anchor/blob/stable/Makefile)
- [test-suite.yml](https://github.com/sigp/anchor/blob/stable/.github/workflows/test-suite.yml)

That means the branch can stay green without ever exercising the full block-verification path added in PR #839.

## Why this matters

In the new proposer spec test, successful fixtures can hit a BLS-validation fallback when `fake_crypto` is not enabled:
- [PR #839 discussion](https://github.com/sigp/anchor/pull/839#issuecomment-4206990576)
- [`consensus_data_proposer.rs` in PR #839](https://github.com/sigp/anchor/pull/839/files)

Without `fake_crypto`, `ConsensusDataProposerTest` can still pass while skipping the assertions that verify:
- block blinded state
- block root
- block SSZ roundtrip

Only the consensus-data root / SSZ roundtrip remains covered in that fallback path.

This is easy to miss because both of these pass locally:
```bash
cargo test -p spec_tests types_spec_tests
cargo test -p spec_tests --features fake_crypto types_spec_tests
```

## Suggested change

Add an explicit Makefile target for the full proposer/spec coverage path, for example:

```make
# Example name only
# test-spec-full:
# 	cargo test -p spec_tests --features fake_crypto types_spec_tests
```

A matching `nextest` target would also be reasonable if preferred.

## Follow-up

Once there is a dedicated Makefile target, CI can opt into it directly instead of relying on someone remembering the feature flag.

I am not suggesting changing the production/test logic here; just making the stronger spec-test path explicit and easy to run.

PR #942, opened 2026-04-09
test: proposer consensus-data spec coverage

### LINKED ISSUE #941: test: proposer consensus-data spec coverage
## Context

Extracted from PR #839 per [review feedback](https://github.com/sigp/anchor/pull/839#issuecomment-4215024532). The proposer consensus data tests are independent of the encryption test and deserve separate review.

## Proposed Changes

Two new spec test types:

1. **`ProposerConsensusDataTest`** — Mirrors Go's `ConsensusData.Validate()`: checks duty type is proposer, then verifies the block data is decodable (blinded-first, then full).

2. **`ConsensusDataProposerTest`** — Mirrors Go's `ProposerSpecTest.Run()`: SSZ-decodes `ProposerConsensusData`, extracts block data, verifies blinded state, block root, block SSZ, consensus data root, and consensus data SSZ roundtrip. Includes a BLS fallback path because Lighthouse validates BLS points during SSZ deserialization while Go's `fastssz` does not.

Shared infrastructure: `RawConsensusData`/`RawDuty` DTOs, `can_decode_block` helper, `is_bls_validation_error` helper, `deserialize_hex_hash256`/`deserialize_base64_or_null`/`deserialize_base64_or_empty` serde helpers, proposer-related error codes, and `fake_crypto` feature flag.

## Additional Info

- With `fake_crypto` enabled, full parity with Go is achieved.
- CI should run with `fake_crypto` for full coverage (see #938).
- Depends on PR #837 for CI (branch chain).

shane-moore:

- Anchor's spec test suite lacks coverage for the Go ssv-spec proposer consensus data test types.
- Needed to maintain spec parity as the test matrix grows. Continuation of the spec test effort from PR #837.
- Extracted from PR #839 per [review feedback](https://github.com/sigp/anchor/pull/839#issuecomment-4215024532). Closes #941.

## Change Overview

Two new spec test types:

1. **`ProposerConsensusDataTest`** — Mirrors Go's `ConsensusData.Validate()`: checks duty type is proposer, then verifies the block data is decodable (blinded-first, then full). Good parity with Go.

2. **`ConsensusDataProposerTest`** — Mirrors Go's `ProposerSpecTest.Run()`: SSZ-decodes `ProposerConsensusData`, extracts block data, verifies blinded state, block root, block SSZ, consensus data root, and consensus data SSZ roundtrip. Includes a BLS fallback path because Lighthouse validates BLS points during SSZ deserialization while Go's `fastssz` does not. With `fake_crypto` enabled, full parity with Go is achieved.

Shared infrastructure added: `RawConsensusData`/`RawDuty` DTOs, `can_decode_block` helper, `is_bls_validation_error` helper, `deserialize_hex_hash256`/`deserialize_base64_or_null`/`deserialize_base64_or_empty` serde helpers, and proposer-related error codes.

- **BLS validation gap without `fake_crypto`**: Without `--features fake_crypto`, `ConsensusDataProposerTest` skips block root, block SSZ, and blinded state checks for success fixtures (falls through to BLS fallback path which only verifies consensus data). Core assertions of the most complex test are bypassed in default mode. CI should run with `fake_crypto` for full coverage.
- **`is_bls_validation_error` string matching**: Uses `msg.contains("BLST")` — fragile against Lighthouse error message changes but stable in practice and test-only.
- **Full block SSZ comparison**: Uses input bytes instead of re-encoding `FullBlockContents` (which doesn't implement `Encode`). Functionally correct but weaker than Go's marshal-and-compare.

## Validation

- `cargo test -p spec_tests --features fake_crypto` — all fixtures pass
- `cargo clippy` and `cargo fmt` clean

## Rollback

N/A — test-only changes, no runtime impact.

## Blockers / Dependencies
- `fake_crypto` CI target tracked in #938.

### DIFF
--- a/anchor/spec_tests/src/types/consensus_data_proposer.rs
+++ b/anchor/spec_tests/src/types/consensus_data_proposer.rs
@@ -1,0 +1,192 @@
+use serde::Deserialize;
+use ssv_types::consensus::ProposerConsensusData;
+use ssz::{Decode, DecodeError, Encode};
+use tree_hash::TreeHash;
+use types::{Hash256, MainnetEthSpec};
+
+use crate::{
+    SpecTest,
+    utils::{
+        deserializers::{deserialize_base64, deserialize_base64_or_null, deserialize_hex_hash256},
+        error_codes, is_bls_validation_error,
+    },
+};
+
+/// Mirrors Go's `ProposerSpecTest.Run()`: SSZ-decode consensus data, extract block,
+/// then verify blinded state, roots, and SSZ roundtrips.
+#[derive(Debug, Deserialize)]
+#[serde(rename_all = "PascalCase")]
+pub struct ConsensusDataProposerTest {
+    #[serde(rename = "DataCd", deserialize_with = "deserialize_base64")]
+    consensus_data_ssz: Vec<u8>,
+
+    #[serde(
+        rename = "DataBlk",
+        deserialize_with = "deserialize_base64_or_null",
+        default
+    )]
+    expected_block_ssz: Option<Vec<u8>>,
+
+    blinded: bool,
+
+    #[serde(
+        rename = "ExpectedBlkRoot",
+        deserialize_with = "deserialize_hex_hash256"
+    )]
+    expected_block_root: Hash256,
+
+    #[serde(
+        rename = "ExpectedCdRoot",
+        deserialize_with = "deserialize_hex_hash256"
+    )]
+    expected_consensus_data_root: Hash256,
+
+    expected_error_code: i64,
+}
+
+impl SpecTest for ConsensusDataProposerTest {
+    fn run(&self) -> Result<(), String> {
+        // Decode ProposerConsensusData from SSZ
+        let proposer_data = match ProposerConsensusData::from_ssz_bytes(&self.consensus_data_ssz) {
+            Ok(data) => data,
+            Err(DecodeError::NoMatchingVariant) => {
+                return self.expect_error(error_codes::UNKNOWN_BLOCK_VERSION);
+            }
+            Err(_) => return self.expect_error(error_codes::UNMARSHAL_SSZ),
+        };
+
+        // Decode block — try blinded first, then full (mirrors Go's GetBlockData)
+        let blinded_err = match proposer_data.decode_blinded_block::<MainnetEthSpec>() {
+            Ok(blinded_block) => {
+                self.expect_success()?;
+                self.verify_block(
+                    true,
+                    blinded_block.tree_hash_root(),
+                    &blinded_block.as_ssz_bytes(),
+                )?;
+                return self.verify_consensus_data(&proposer_data);
+            }
+            Err(e) => e,
+        };
+
+        let full_err = match proposer_data.decode_block_contents::<MainnetEthSpec>() {
+            Ok(full_contents) => {
+                self.expect_success()?;
+                // Go's `vBlk.Root()` returns the inner `BeaconBlock` root, but
+                // `vBlk.Deneb.MarshalSSZ()` encodes the full block contents (block + blobs).
+                // `FullBlockContents` doesn't implement `Encode`, so we compare against the
+                // raw `data_ssz` bytes from `ProposerConsensusData` which is what was decoded.
+                let inner_block = full_contents.block();
+                self.verify_block(false, inner_block.tree_hash_root(), &proposer_data.data_ssz)?;
+                return self.verify_consensus_data(&proposer_data);
+            }
+            Err(e) => e,
+        };
+
+        // Both decoders failed. Go fixtures use synthetic BLS points that Go's fastssz
+        // accepts but Lighthouse rejects (`BLST_BAD_ENCODING`). Without `fake_crypto`,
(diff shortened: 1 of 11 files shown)


shane-moore:
@claude, can you take a look?

shane-moore:
@claude, thanks for the thorough review. Investigated both findings:

### Finding 1: `NoMatchingVariant` → `UNKNOWN_BLOCK_VERSION` mapping

This mapping is structurally guaranteed by the types, not fragile:

- **`BeaconRole`** uses `#[ssz(struct_behaviour = "transparent")]` — it's a transparent `u64` wrapper where every 8-byte value is accepted. No variant matching occurs.
- **All other `ValidatorDuty` fields** (`PublicKeyBytes`, `Slot`, `ValidatorIndex`, `u64`s, `VariableList<u64, _>`) are primitives that fail with `InvalidByteLength`, never `NoMatchingVariant`.
- **`data_ssz`** is `VariableList<u8, _>` — raw bytes, no variant matching.
- **`DataVersion`** has a manual `Decode` impl with explicit `_ => return Err(DecodeError::NoMatchingVariant)` for unknown fork versions. This is the only source.

If `BeaconRole` ever gained variant validation, it would change from a transparent wrapper to an enum — a major refactor that would naturally trigger review of all decode error handling. No comment needed.

### Finding 2: BLS fallback blinded-state check

The edge case analysis (both errors being BLS → blinded wins) is correct, but this is the intended behavior — it mirrors Go's `GetBlockData()` which also tries blinded first. The existing comment on line 96 already explains the logic. In practice, a block can't be structurally valid as both blinded and full simultaneously (different SSZ structures), so the "both are BLS errors" scenario doesn't arise.

petarjuki7 (approved):
Nice, looks good to me!

anchor/spec_tests/src/types/consensus_data_proposer.rs
```diff
@@ -0,0 +1,192 @@
...
+                // raw `data_ssz` bytes from `ProposerConsensusData` which is what was decoded.
+                let inner_block = full_contents.block();
+                self.verify_block(false, inner_block.tree_hash_root(), &proposer_data.data_ssz)?;
+                return self.verify_consensus_data(&proposer_data);
+            }
+            Err(e) => e,
+        };
+
+        // Both decoders failed. Go fixtures use synthetic BLS points that Go's fastssz
+        // accepts but Lighthouse rejects (`BLST_BAD_ENCODING`). Without `fake_crypto`,
```
diegomrsantos, line 87: why?

diegomrsantos:
I don't think `fake_crypto` is the right fix here.

The underlying issue looks to be the shared proposer fixtures, not Lighthouse. In `ssv-spec/types/testingutils/beacon_node_consts.go`, the Capella/Deneb fixtures are copied from upstream go-eth2-client JSON codec tests and contain placeholder-looking BLS values like sequential `0x000102...` pubkeys and `0x606162...` signatures. Those are the correct byte lengths, but they are not guaranteed to be valid compressed BLS points.

Go accepts them because go-eth2-client's `BLSPubKey` / `BLSSignature` JSON and SSZ decoding only checks hex + length and then copies bytes. Lighthouse's default BLS path actually validates the compressed point, so `BLST_BAD_ENCODING` is expected here.

Adding `spec_tests/fake_crypto` papers over that fixture-quality mismatch by disabling the stricter decoder on the Rust side. I don't think we should normalize invalid cross-client fixtures this way. If these tests are meant to assert proposer compatibility, the better fix is to update the `ssv-spec` fixtures to use BLS-valid encodings and then bump the submodule, rather than adding a crypto-bypass feature to make the current data pass.

As written, this weakens what the new test is proving: we pass by skipping real BLS decoding, instead of making the shared fixtures valid.

diegomrsantos:
Opened upstream issue for the fixture problem here: https://github.com/ssvlabs/ssv-spec/issues/622

That issue narrows the scope to the Capella/Deneb proposer fixtures copied from codec tests. Given that, I still think `fake_crypto` is working around a fixture-quality mismatch in `ssv-spec`, rather than fixing the underlying compatibility target.

shane-moore:
Agree with Diego. The Capella/Deneb proposer fixtures contain invalid BLS points (correct length, not valid curve points). Go accepts them because go-eth2-client only checks hex+length, while Lighthouse validates compressed points during SSZ deserialization.

Without `fake_crypto`, the `ConsensusDataProposerTest` can't decode blocks at all, so the core assertions (block root, block SSZ, blinded state) are skipped via the BLS fallback path. The test doesn't pass *with* real crypto — it passes by *skipping checks*. Using `fake_crypto` restores those checks, but the right fix is valid fixtures upstream so the tests work end-to-end without a crypto bypass.

Confirmed via Lighthouse's codebase that `fake_crypto` replaces all BLS operations with byte-copying stubs — any correct-length bytes are accepted as valid points during SSZ decode. Lighthouse uses this for their own SSZ static round-trip tests, but that's because those tests are auto-generated across every consensus type. Here we control the fixture pipeline via ssv-spec, so fixing the source data is the better path.

+1 on ssv-spec #622. Opened #953 to track removing `fake_crypto` once the upstream fixtures are fixed.

Issue #940, opened 2026-04-09
refactor: add validator_index DB index (PR #890 split 2/4)


shane-moore:
## Context

Sub-issue of #931. Adds a `CREATE INDEX` on `validators.validator_index` to back the `GET_SHARE_PUBKEYS_FOR_VALIDATOR_INDEX` query added in PR 1.

## Motivation

Per [this review comment](https://github.com/sigp/anchor/pull/936#discussion_r3057785811), without an index SQLite falls back to a full table scan (`SCAN v`). With `NetworkDatabase` `POOL_SIZE = 1`, concurrent lookups serialize on the single connection — 20 stacked lookups cost ~160ms without the index vs ~3ms with it.

## Scope

- Add schema migration v3 → v4 with `CREATE INDEX idx_validators_validator_index ON validators(validator_index)`
- Update `SCHEMA_VERSION` constant
- Minimal, isolated change — pure DB migration

## Depends on
- #932 (PR 1: share pubkey queries)

PR #943, opened 2026-04-09
chore: add validator index db index

shane-moore:

- Without an index on `validators.validator_index`, the `GET_SHARE_PUBKEYS_FOR_VALIDATOR_INDEX` query (added in #936) falls back to a full table scan (`SCAN v`).
- With `POOL_SIZE = 1`, concurrent lookups serialize on the single SQLite connection. Benchmarks on a mainnet-scale synthetic DB (~140k validators, ~1.82M shares) showed ~8.2ms per lookup without the index vs ~0.14ms with it. At 20 stacked lookups, that's ~160ms serialized vs ~3ms.
- This was identified during [review of #936](https://github.com/sigp/anchor/pull/936#discussion_r3057785811) and tracked as #940.

## Change Overview (Required)

- Schema migration v3 → v4: adds `CREATE INDEX idx_validators_validator_index ON validators(validator_index)`.
- Updates `table_schema.sql` so new databases get the index from the start.
- Adds migration test (`test_migration_v3_to_v4`) and updates existing migration tests (`v1_to_v4`, `v2_to_v4`) to pass through the new migration step.

**Design choice — `CREATE INDEX` vs `UNIQUE` constraint:** `validator_index` is unique on the beacon chain, so a `UNIQUE` constraint would be semantically correct. However, a plain index was chosen because: (1) it's the minimal change, (2) `validator_index` can be `NULL` before activation, and (3) adding a uniqueness constraint to existing databases risks breaking the migration if any edge-case duplicates or NULLs exist. The performance is identical — SQLite uses the same B-tree structure for both.

- Low risk — pure schema migration, no behavior change. The index is additive.
- Migration tested against v1, v2, and v3 starting schemas.
- Verified with `EXPLAIN QUERY PLAN` that SQLite uses the index after migration.

## Validation (Required)

- `test_migration_v3_to_v4`: creates a v3 database, runs migration, asserts version is 4 and the index exists.
- `test_migration_v1_to_v4` / `test_migration_v2_to_v4`: verify the full migration chain from older schemas still works.
- Manual `sqlite3` validation confirming query plan changes from `SCAN` to `SEARCH ... USING INDEX`.

N/A — additive index only. Removing it would require a v4 → v5 migration dropping the index, but the index is harmless if unused.

- Depends on #936 (adds the query that benefits from this index).

- Part 2/4 of the #931 split. Next: #941 (signature collector integration).

### DIFF
--- a/anchor/database/src/schema.rs
+++ b/anchor/database/src/schema.rs
@@ -28,6 +28,15 @@ const MIGRATION_V2_TO_V3: &str = r#"
     UPDATE metadata SET schema_version = 3;
 "#;
 
+/// Migration from schema version 3 to 4: Add index on `validators.validator_index`
+///
+/// The `GET_SHARE_PUBKEYS_FOR_VALIDATOR_INDEX` query joins validators and shares
+/// filtering by `validator_index`. Without an index this causes a full table scan.
+const MIGRATION_V3_TO_V4: &str = r#"
+    CREATE INDEX idx_validators_validator_index ON validators(validator_index);
+    UPDATE metadata SET schema_version = 4;
+"#;
+
 enum UpgradeAction {
     UpToDate,
     DoUpdate {
@@ -183,7 +192,11 @@ fn get_upgrade_action(version: Option<SchemaVersion>) -> UpgradeAction {
             script: MIGRATION_V2_TO_V3,
             new_version: 3,
         },
-        Some(3) => UpgradeAction::UpToDate,
-        Some(4..) => UpgradeAction::Future,
+        Some(3) => UpgradeAction::DoUpdate {
+            script: MIGRATION_V3_TO_V4,
+            new_version: 4,
+        },
+        Some(4) => UpgradeAction::UpToDate,
+        Some(5..) => UpgradeAction::Future,
     }
 }
--- a/anchor/database/src/table_schema.sql
+++ b/anchor/database/src/table_schema.sql
@@ -1,8 +1,8 @@
--- SCHEMA VERSION 3
+-- SCHEMA VERSION 4
 
 -- we should avoid removing columns from this to keep compatibility between anchor Versions
 CREATE TABLE metadata (
-    schema_version INTEGER NOT NULL DEFAULT 3,
+    schema_version INTEGER NOT NULL DEFAULT 4,
     -- DEPRECATED: domain_type is no longer used for network isolation as of schema v3.
     -- Previously, we used domain_type (a 4-byte value derived from the fork) to prevent
     -- cross-network database usage. However, domain_type changes at each fork activation,
@@ -57,6 +57,7 @@ CREATE TABLE validators (
     graffiti BLOB DEFAULT X'0000000000000000000000000000000000000000000000000000000000000000',
     FOREIGN KEY (cluster_id) REFERENCES clusters(cluster_id)
 );
+CREATE INDEX idx_validators_validator_index ON validators(validator_index);
 
 CREATE TABLE shares (
     validator_pubkey TEXT NOT NULL,
--- a/anchor/database/src/tests/metadata_tests.rs
+++ b/anchor/database/src/tests/metadata_tests.rs
@@ -172,7 +172,7 @@ mod tests {
     }
 
     #[test]
-    fn test_migration_v1_to_v3() {
+    fn test_migration_v1_to_v4() {
         let temp_dir = TempDir::new().expect("Failed to create temp dir");
         let db_path = temp_dir.path().join("test.db");
 
@@ -220,7 +220,7 @@ mod tests {
     }
 
     #[test]
-    fn test_migration_v2_to_v3() {
+    fn test_migration_v2_to_v4() {
         let temp_dir = TempDir::new().expect("Failed to create temp dir");
         let db_path = temp_dir.path().join("test.db");
 
--- a/anchor/database/src/tests/metadata_tests.rs
+++ b/anchor/database/src/tests/metadata_tests.rs
@@ -33,8 +33,8 @@ mod tests {
         let metadata = queries::get_metadata(&conn).expect("Failed to get metadata");
 
         assert_eq!(
-            metadata.schema_version, 3,
-            "Initial schema version should be 3"
+            metadata.schema_version, 4,
+            "Initial schema version should be 4"
         );
         assert_eq!(
             metadata.network_name, TEST_NETWORK_1,
@@ -196,8 +196,8 @@ mod tests {
             let conn = Connection::open(&db_path).expect("Failed to open database");
             let metadata = queries::get_metadata(&conn).expect("Failed to get metadata");
             assert_eq!(
-                metadata.schema_version, 3,
-                "Should be upgraded to version 3"
+                metadata.schema_version, 4,
+                "Should be upgraded to version 4"
             );
 
             // Verify the network_name column was set
@@ -244,8 +244,8 @@ mod tests {
             let conn = Connection::open(&db_path).expect("Failed to open database");
             let metadata = queries::get_metadata(&conn).expect("Failed to get metadata");
             assert_eq!(
-                metadata.schema_version, 3,
-                "Should be upgraded to version 3"
+                metadata.schema_version, 4,
+                "Should be upgraded to version 4"
             );
 
             // Verify the network_name column was set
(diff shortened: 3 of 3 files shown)


shane-moore:
@claude, can you take a look?

shane-moore:
@claude Re: the `CREATE INDEX` idempotency observation — the migration statements run within a single SQLite transaction, so a partial application (index created but version not bumped) would require a crash between two statements in the same atomic commit, which SQLite prevents. Non-issue in practice.

shane-moore:
@claude, Good catch. You're right that `execute_batch` doesn't implicitly wrap in a transaction. I was wrong about the atomicity guarantee. Agreed it's still a non-issue in practice given the vanishingly small crash window, but appreciate the correction.

diegomrsantos:
> **Claude finished @shane-moore's task in 16s** —— [View job](https://github.com/sigp/anchor/actions/runs/24202019483)
> 
> * [x]  Understand the context of the conversation
> * [x]  Respond to acknowledgment
> 
> Thanks for confirming. To summarize the thread for anyone following along: `rusqlite::Connection::execute_batch` does **not** wrap multiple statements in an implicit transaction — each statement executes under autocommit individually. This means a crash between statements could theoretically leave the DB in a partially-migrated state. In practice the window is vanishingly small, and this is a pre-existing pattern across all migrations in the codebase, so not something to address in this PR.
> 
> No further action needed here. The PR itself remains LGTM — clean migration, correct tests, minimal scope.
> 
> [View job](https://github.com/sigp/anchor/actions/runs/24202019483)

@claude should the current pattern be fixed?

shane-moore:
@claude, makes sense! made ticket https://github.com/sigp/anchor/issues/945 to handle

diegomrsantos:
One broader concern on the migration tests. I do not think this should block this PR, but I do think the current pattern is weak and worth cleaning up in a follow-up.

The issue is not that we have `v1 -> latest` / `v2 -> latest` coverage. That part is correct, since `ensure_up_to_date()` still claims to support those versions. The issue is *how* we test it today:

- We hand-recreate old schemas in Rust test helpers instead of using authoritative historical schema snapshots.
- That duplicates schema history in test code, which is brittle and drift-prone.
- In this PR, a small `v3 -> v4` index migration required updating the synthetic `v1` / `v2` builders to create `validators`, just so the later migration step can run. That is a smell.
- The assertions are also fairly shallow: mostly schema version bumps and a spot-check that the new index exists, rather than proving that a migrated DB matches a freshly created latest-schema DB.

So I think this PR is fine within the current system, but the system itself is not very strong. A better approach would be:

- keep versioned SQL snapshots for each supported historical schema
- create old DBs from those snapshots rather than handwritten Rust builders
- migrate each snapshot to `latest`
- compare the migrated schema against a fresh latest DB via `sqlite_master` / `PRAGMA` introspection
- keep migration-specific tests only for cases where data semantics change, not for every bit of structural boilerplate

Not a blocker here, but this PR is a good example of why the current migration-test setup is costing more than it should while still providing fairly weak assurance.

shane-moore:
Agree with the diagnosis, that the `create_validators_table` retrofit into the v1/v2 builders shows the current pattern is getting brittle. Creating a GH issue for this.

I'm less convinced we need SQL snapshots right now though. At 4 schema versions the hand-built helpers are manageable, and snapshots don't actually fix the weaker problem: shallow assertions. What I think we need is a "compare migrated vs fresh" assertion pattern:

```rust
// 1. Create old-schema DB with existing Rust helpers
create_v1_database(&db_path, TEST_NETWORK_1);

// 2. Migrate it
schema::ensure_up_to_date(&db_path, TEST_NETWORK_1).unwrap();

// 3. Create a fresh DB (always latest schema via table_schema.sql)
schema::create_initial_schema(&fresh_path, TEST_NETWORK_1).unwrap();

// 4. Compare sqlite_master between the two
let migrated_schema = get_schema_dump(&db_path);  // SELECT * FROM sqlite_master ORDER BY type, name
let fresh_schema = get_schema_dump(&fresh_path);
assert_eq!(migrated_schema, fresh_schema, "migrated DB schema doesn't match fresh DB");
```

This catches every structural difference automatically: missing indexes, wrong column defaults, missing columns -> without needing per-migration spot-checks. One generic assertion covers all future migrations too.

If maintaining the Rust helpers becomes painful, we can layer SQL snapshots on top of that. But the comparison assertion is where the real assurance gap is. wdyt?

diegomrsantos:
Replying to https://github.com/sigp/anchor/pull/943#issuecomment-4237942777

I agree that comparing a migrated DB against a fresh latest DB is the highest-value incremental improvement. That closes the biggest immediate assurance gap.

But I do not think that is the whole issue here. There are several separate problems in the current setup.

1. The assertions are weak.

Right now the migration tests mostly check version bumps and a few local details such as whether the new index exists. A structural comparison against a fresh latest DB would improve that a lot.

2. The historical inputs are synthetic.

Functions like `create_v1_database()` and `create_v2_database()` are hand-written Rust reconstructions of supported historical schemas. That is a bigger problem than it looks, because the test input itself can drift from what those old schemas actually were.

This PR shows that clearly. A small `v3 -> v4` index migration required retrofitting `create_validators_table()` into the synthetic `v1` and `v2` builders just so the later migration step could run. Once that happens, the tests are no longer exercising authoritative old states. They are exercising old states that we reconstructed and then patched forward by hand.

3. The current test helpers are doing too much.

I do not think we should keep fat helpers whose job is to manually define historical schema in Rust. That is the wrong source of truth for supported historical versions.

What is fine is a thin helper that loads `schema_v1.sql` or a real historical DB fixture and then seeds a row. That is very different from a helper that re-encodes the schema itself.

4. Schema truth is duplicated.

In the current system we have schema definition spread across migration SQL in `schema.rs`, fresh-schema SQL in `table_schema.sql`, and synthetic historical schemas in the test helpers. Even if we add a migrated-vs-fresh comparison on top, that still leaves multiple places that must stay aligned by hand.

5. The maintainability argument is already failing now.

I do not think the right bar is whether 4 schema versions are still manageable. The brittleness is already visible. `create_validators_table()` being bolted into older builders is the proof. The problem is not hypothetical future pain. It is current drift and current maintenance churn.

So my view is:

1. Yes, add the structural comparison between migrated and fresh latest DBs.
2. No, do not keep the handwritten historical-schema builders as the foundation for supported versions.
3. Replace those with SQL snapshots or real historical DB fixtures.
4. Keep synthetic builders only for negative cases such as legacy, unknown, or future-schema DBs, because those are not pretending to be authoritative historical versions.

I also put together a draft spike in #948, tracked by #947, that tries to address the full shape of the problem rather than only the assertion layer.

That spike does four things together:

1. moves schema evolution into migration files
2. uses the same source of truth for fresh DB creation and upgrades
3. replaces hand-built historical schemas with SQL fixtures
4. compares migrated DBs against fresh latest DBs structurally

One useful thing that spike exposed is that once you add real structural comparison, migrated DBs and fresh DBs did not actually converge cleanly without an extra canonicalization migration. That is exactly the kind of issue the current tests hide.

So I think your proposed assertion pattern is a good first step, but I would not stop there. The `create_v1_database()` / `create_v2_database()` style helpers are still the wrong foundation for supported historical migration tests.

shane-moore:
Your spike is compelling -> especially the V4 canonicalize migration. The fact that adding structural comparison immediately exposed that migrated and fresh DBs diverge (nullable vs NOT NULL on network_name, different defaults, DDL text mismatch in sqlite_master) is a strong argument that the current tests are hiding real issues. The approach of deriving old schemas from the real migration chain via `Target::Version(N)` is also cleaner than either hand-built helpers or static SQL snapshots -> it keeps a single source of truth.

That said, I think #943 should still merge in the interim:

1. It's the actual performance fix -> the index is what matters for the query added in #936
2. #948 is a draft spike with a dependency blocker (refinery doesn't support our rusqlite 0.38 in any published release)
3. When #948 lands, this migration just becomes a `V5__add_validator_index.sql` file -- straightforward rebase

wdyt?

diegomrsantos (approved):

diegomrsantos:
> Your spike is compelling -> especially the V4 canonicalize migration. The fact that adding structural comparison immediately exposed that migrated and fresh DBs diverge (nullable vs NOT NULL on network_name, different defaults, DDL text mismatch in sqlite_master) is a strong argument that the current tests are hiding real issues. The approach of deriving old schemas from the real migration chain via `Target::Version(N)` is also cleaner than either hand-built helpers or static SQL snapshots -> it keeps a single source of truth.
> 
> That said, I think #943 should still merge in the interim:
> 
> 1. It's the actual performance fix -> the index is what matters for the query added in [feat: add share pubkey queries #936](https://github.com/sigp/anchor/pull/936)
> 2. [refactor(database): spike refinery-backed migrations #948](https://github.com/sigp/anchor/pull/948) is a draft spike with a dependency blocker (refinery doesn't support our rusqlite 0.38 in any published release)
> 3. When [refactor(database): spike refinery-backed migrations #948](https://github.com/sigp/anchor/pull/948) lands, this migration just becomes a `V5__add_validator_index.sql` file -- straightforward rebase
> 
> wdyt?

Agreed. Approved ;)

PR #944, opened 2026-04-10
refactor: localize qbft_manager scheduling

diegomrsantos:

- `qbft_manager` reaches directly into raw `processor::Senders` at each scheduling site.
- That makes runtime plumbing leak across cleaner startup, instance spawning, and urgent message dispatch.
- This is worth doing now because it creates a single local seam before follow-up cleanup around the test harness and executor lifetime handling.
- Relevant links: #917

## Change Overview (Required)

- Added a private `QbftScheduler` wrapper in `anchor/qbft_manager/src/lib.rs`.
- Moved cleaner startup, instance task spawning, urgent message dispatch, and processor-closed checks behind named helper methods on that wrapper.
- Review order: start with `QbftScheduler`, then the call-site replacements in `QbftManager::new`, `decide_instance`, `pass_to_instance`, `cleaner`, and `QbftDecidable::get_or_spawn_instance`.
- Intentionally unchanged: public APIs, queue selection, task names, permit handling, and QBFT behavior.

- Risk: this touches the scheduling path used to start cleaner and instance tasks and to enqueue urgent QBFT work.
- Trade-off: one extra internal wrapper type in exchange for fewer direct queue operations spread across the manager.
- Mitigation: the wrapper is thin, private to the crate, and the change is limited to one file with no behavior change intended.

## Validation (Required)

- `cargo fmt --all`
- `cargo check -p qbft_manager`
- `cargo test -p qbft_manager`
- `cargo clippy -p qbft_manager --tests -- -D warnings`
- Commit hook also ran `cargo clippy --all` and `cargo sort workspace`

- Safe to revert as a single commit.

- None for merge.
- Follow-up cleanup around test/runtime plumbing can build on this seam.

- This is intentionally the first prep refactor before deeper work to simplify `qbft_manager` runtime plumbing.

### DIFF
--- a/anchor/qbft_manager/src/lib.rs
+++ b/anchor/qbft_manager/src/lib.rs
@@ -127,7 +127,6 @@ pub struct QbftInitialization<D: QbftData> {
 // Map from an identifier to a sender for the instance
 type Map<I, D> = DashMap<I, UnboundedSender<QbftMessage<D>>>;
 
-#[derive(Clone)]
 struct QbftScheduler {
     senders: Senders,
 }
--- a/anchor/qbft_manager/src/lib.rs
+++ b/anchor/qbft_manager/src/lib.rs
@@ -127,10 +127,52 @@ pub struct QbftInitialization<D: QbftData> {
 // Map from an identifier to a sender for the instance
 type Map<I, D> = DashMap<I, UnboundedSender<QbftMessage<D>>>;
 
+#[derive(Clone)]
+struct QbftScheduler {
+    senders: Senders,
+}
+
+impl QbftScheduler {
+    fn new(senders: Senders) -> Self {
+        Self { senders }
+    }
+
+    fn spawn_background(
+        &self,
+        task: impl Future<Output = ()> + Send + 'static,
+        name: &'static str,
+    ) -> Result<(), QbftError> {
+        self.senders.permitless.send_async(task, name)?;
+        Ok(())
+    }
+
+    fn dispatch<D: QbftData + Send + 'static>(
+        &self,
+        sender: UnboundedSender<QbftMessage<D>>,
+        kind: QbftMessageKind<D>,
+        name: &'static str,
+    ) -> Result<(), QbftError> {
+        self.senders.urgent_consensus.send_immediate(
+            move |drop_on_finish| {
+                let _ = sender.send(QbftMessage {
+                    kind,
+                    drop_on_finish: Some(drop_on_finish),
+                });
+            },
+            name,
+        )?;
+        Ok(())
+    }
+
+    fn is_closed(&self) -> bool {
+        self.senders.permitless.is_closed()
+    }
+}
+
 // Top level QBFTManager structure
 pub struct QbftManager<E: EthSpec, S: SlotClock> {
-    // Senders to send work off to the central processor
-    processor: Senders,
+    // Internal scheduler used to queue QBFT work onto the shared processor.
+    scheduler: QbftScheduler,
     // OperatorID
     operator_id: OwnOperatorId,
     // All of the QBFT instances that are voting on proposer consensus data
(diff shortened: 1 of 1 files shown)


anchor/qbft_manager/src/lib.rs
```diff
@@ -127,10 +127,52 @@ pub struct QbftInitialization<D: QbftData> {
 // Map from an identifier to a sender for the instance
 type Map<I, D> = DashMap<I, UnboundedSender<QbftMessage<D>>>;
 
+#[derive(Clone)]
```
shane-moore: Nit: `QbftScheduler` doesn’t appear to be cloned anywhere right now, so `#[derive(Clone)]` may be removable unless a follow-up needs it.

anchor/qbft_manager/src/lib.rs
```diff
@@ -127,10 +127,52 @@ pub struct QbftInitialization<D: QbftData> {
...
+struct QbftScheduler {
+    senders: Senders,
+}
+
+impl QbftScheduler {
```
shane-moore, line 134: The stated goal is to create a single local seam for follow-up cleanup, but the wrapper is not a trait, not a mockable boundary, and not substitutable

Can you say a bit more about what a dedicated `QbftScheduler` type is buying us here over private `QbftManager` helper methods, maybe in pr description?
>
> As written this looks like a very thin wrapper over `Senders`, so I’m trying to understand whether the follow-up value is the type itself or just the locality/readability win from having named helpers in one place.

anchor/qbft_manager/src/lib.rs
diegomrsantos: Removed in `affa767`. Good catch.

anchor/qbft_manager/src/lib.rs
```diff
@@ -127,10 +127,52 @@ pub struct QbftInitialization<D: QbftData> {
...
+    fn new(senders: Senders) -> Self {
+        Self { senders }
+    }
+
+    fn spawn_background(
```
shane-moore, line 140: The helper method names erase queue-selection information at the call sites. Previously the cleaner loop read `while !self.processor.permitless.is_closed()` — the reader could see *exactly* which queue governs termination. After the refactor it reads `while !self.scheduler.is_closed()`, and you have to open `QbftScheduler::is_closed` to learn it only polls `permitless` (and not `urgent_consensus`, which also carries QBFT work). The same holds for `spawn_background` and `dispatch` — the `permitless` vs `urgent_consensus` choice is now implicit in the helper name.

 For a scheduling-adjacent type, method names should make queue selection auditable at the call site.

**Suggested fix :**
- Rename for honesty: `is_closed → permitless_closed`, `spawn_background → spawn_permitless`, `dispatch → send_urgent` (or similar).

**Adversarial take:** Partly fair, but overstated. Hiding raw processor queue names is arguably the point of the seam. `dispatch()` and `spawn_background()` are reasonable domain-level helpers. `is_closed()` is the only method here that meaningfully hides a policy choice, because cleaner liveness is still keyed to `permitless` only.

In the end, i leave it to you to decide whether to rename or not cuz i'm a bit torn. it's potentially nice to see in the fn name whether the task is being added to permitless or permit required queues, but perhaps you feel differently which is also fine

shane-moore: or maybe we keep the naming and add comments like this:

```rust
/// Spawn a long-running async task on the `permitless` queue. Bypasses
/// `max_workers` throttling — only appropriate for tasks that are themselves
/// well-behaved (yield regularly, handle their own backpressure).
fn spawn_background(...) -> Result<(), QbftError> { ... }

/// Enqueue an immediate dispatch on the `urgent_consensus` queue. Requires a
/// worker permit; the permit is held for the full lifetime of the resulting
/// `QbftMessage` via `DropOnFinish`.
fn dispatch<D: QbftData + Send + 'static>(...) -> Result<(), QbftError> { ... }
```

or rename and add comments. again, whatever you prefer

diegomrsantos:
Thanks @shane-moore for the useful pushback here.

I think the better move is to close this PR and tackle the actual problem more directly: making the `qbft_manager` test harness comprehensible, without introducing another thin production abstraction.

I opened #946 with a concrete design and phased plan:
- extract a `QbftTestRuntime` for the runtime plumbing
- rename/reshape `QbftTester` into an explicit in-memory cluster harness
- simplify the scenario API so the tests read like cluster scenarios

Curious what you think about that direction on the issue.

Issue #924, opened 2026-04-02
Add root proposer-duty span and top-level checkpoints in validator_store


diegomrsantos:
## Summary
Add one canonical root proposer-duty span in \`validator_store\` and propagate it through the top-level proposer flow.

## Motivation
Anchor currently has local task spans and some duty-specific spans, but proposer signing does not have one canonical root span that ties the whole duty together. Without that, we cannot correlate pre-QBFT timing, QBFT timing, post-consensus timing, and final outcome for the same proposer duty.

## Relation to SSV
This is the highest-value slice of what [feat(traces): OTeL Traces implementation (duty flow)](https://github.com/ssvlabs/ssv/pull/2076) did for SSV's duty flow. We do not need the entire SSV rollout. We do need the equivalent of a proposer-duty parent context.

## Scope
Instrument the top-level proposer path around \`validator_store::sign_block\` and related top-level calls so that every proposer duty emits exactly one root span with a stable shape.

Minimum fields on the root span:
- \`slot\`
- \`validator_pubkey\`
- \`validator_index\` when available
- \`current_slot\`
- \`block_kind\` (\`full\` or \`blinded\`)
- final \`outcome\`

Minimum root-level checkpoints:
- duty entry
- pre-QBFT handoff
- consensus returned
- block signed
- publish path selected
- duty success / failure

## Out of Scope
- detailed QBFT round instrumentation
- detailed signature collector instrumentation
- changing timeout or proposal behavior

## Acceptance Criteria
- each proposer duty emits one root span
- failures are recorded with an explicit outcome and reason
- child work in the proposer path receives the root span context
- no high-cardinality metric labels are introduced as part of this work

## References
- sigp/anchor#758
- [feat(traces): OTeL Traces implementation (duty flow)](https://github.com/ssvlabs/ssv/pull/2076)
- [docs: (mev-considerations) update round-change-based estimates](https://github.com/ssvlabs/ssv/pull/2416)

Issue #919, opened 2026-04-02
Define proposer-duty observability conventions and measurement model


diegomrsantos:
## Summary
Define the proposer-duty observability model for Anchor before implementing instrumentation across the proposer path.

## Motivation
We want to measure the proposer duty critical path end to end before changing QBFT or proposer behavior. Right now Anchor has useful local spans and metrics, but it does not have one stable observability model for proposer duties. If we skip this step, each instrumentation PR will invent its own span names, event names, and metric fields, and the result will be hard to query and compare over time.

## Relation to SSV
This is the proposer-focused equivalent of the conventions and naming discipline introduced by `ssvlabs/ssv#2076`, but intentionally narrower. The goal is not to copy SSV's full OTel rollout. The goal is to define a stable proposer-duty measurement model for Anchor.

## Scope
Define:
- the canonical root span for proposer duties
- required root span fields
- child span / checkpoint event names
- the outcome taxonomy (`success`, `timeout`, `error`, etc.)
- the low-cardinality metric naming rules
- which fields belong in logs/spans versus Prometheus labels

Expected checkpoint vocabulary should cover at least:
- `duty_start`
- `pre_qbft_ready`
- `qbft_start`
- `proposal_received`
- `round_timeout`
- `round_advanced`
- `prepare_quorum`
- `commit_quorum`
- `consensus_decided`
- `post_consensus_start`
- `post_consensus_threshold`
- `signature_reconstructed`
- `publish_start`
- `publish_end`
- `duty_done`
- `duty_failed`

## Out of Scope
- implementing the instrumentation itself
- choosing an OTel exporter or backend
- changing proposer/QBFT behavior

## Acceptance Criteria
- there is one written proposer-duty observability model in the issue thread or linked design note
- later implementation issues can reference the agreed span names, fields, and metrics
- the design explicitly calls out high-cardinality fields that must stay out of Prometheus labels

## References
- `sigp/anchor#758`
- `ssvlabs/ssv#2076`
- `ssvlabs/ssv#2609`
- `ssvlabs/ssv#1825`

diegomrsantos:
## Proposer-Duty Observability Model

### Instrumentation Boundary

Anchor controls two trait methods called by Lighthouse's `BlockService`:
1. `randao_reveal(validator_pubkey, epoch)` — QBFT consensus for randao signature
2. `sign_block(validator_pubkey, block, current_slot)` — QBFT consensus for block + signature collection

[...]

PR #951, opened 2026-04-15
refactor: Creates error and proposer modules ahead of proposer-duty span

jnhsigmap:

A first PR related to #924.

This change simply moves `Error` type definition and `SpecificError` to a new module to allow common access and to avoid further cluttering `lib.rs` (ahead of [planned refactor](https://github.com/sigp/anchor/issues/896)).

## Change Overview (Required)

Moves moves `Error` type definition and `SpecificError` to `errors.rs` with required imports and removes these from `lib.rs`.

No real risk here. Imports of these is limited to `validator_store` and is contained to `lib.rs` for 'production' code.

## Validation (Required)

`cargo test --workspace`
`cargo clippy`
`cargo check`

Revert this commit - move error types back to `lib.rs` with dependencies. Remove `proposer.rs`.

- Create an error taxonomy from [original scope](https://github.com/sigp/anchor/issues/919#issuecomment-4177726189).
- Implement the root span and checkpoints in `sign_block`.
    - `outcome` in the root span will utilise the defined error taxonomy to report the outcome of block signing.


anchor/validator_store/src/proposer.rs
```diff
@@ -0,0 +1,27 @@
+use crate::errors::{Error, SpecificError};
+
+#[derive(Debug, Clone, Copy)]
```
shane-moore: more idiomatic to put the derive attribute below the comment

anchor/validator_store/src/proposer.rs
```diff
@@ -0,0 +1,27 @@
...
+/// Instrumentation taxonomy reported outcomes of signing a block.
+pub enum SignBlockOutcome {
+    Success,
+    Timeout,
+    Error,
```
shane-moore: `Error` reads like an error type at call sites, but this enum is a non-error type. consider `Failure` variant instead

anchor/validator_store/src/proposer.rs
```diff
@@ -0,0 +1,27 @@
+use crate::errors::{Error, SpecificError};
+
+#[derive(Debug, Clone, Copy)]
+/// Instrumentation taxonomy reported outcomes of signing a block.
```
shane-moore: maybe comment more like `/// Outcome of a block signing attempt, used for metrics labeling.`

anchor/validator_store/src/proposer.rs
```diff
@@ -0,0 +1,27 @@
...
+pub enum SignBlockOutcome {
+    Success,
+    Timeout,
+    Error,
+}
+
+#[expect(dead_code)]
```
shane-moore: I'd probably kill this whole file for now and add it once `SignBlockOutcome` is actually consumed, will then be easier to review how the enum is used

jnhsigmap: yeah, fair point, I'll include it in the follow up :slightly_smiling_face:

shane-moore (approved):

diegomrsantos:
@claude review this PR

diegomrsantos:
I'm struggling to understand why this needs to land ahead of [#924](https://github.com/sigp/anchor/issues/924), so I want to sanity-check the sequencing.

From the issue/design side, my understanding is:

- [#924](https://github.com/sigp/anchor/issues/924) is about adding the proposer-duty root span and checkpoints around `sign_block` in `validator_store`.
- [#919](https://github.com/sigp/anchor/issues/919) already defined the observability model, including the coarse top-level outcome taxonomy: `success`, `timeout`, `error`.
- exporter/backend choices like OTLP are out of scope for this milestone.

Given that, I'm struggling to see why this extraction needs to land first.

In particular:

- the current diff is mostly moving `Error` / `SpecificError` into `errors.rs`
- the proposer-specific follow-up type/module that seemed like the consumer of this extraction does not appear to be part of this PR anymore
- if the goal is just to classify proposer outcomes for spans/metrics, I don't yet see why that requires moving `SpecificError` out of `lib.rs` first

More concretely, I think there are two separate concerns here:

1. instrumentation outcome
2. detailed failure reason

Per [#919](https://github.com/sigp/anchor/issues/919), the top-level outcome sounds like it should stay coarse:

- `success`
- `timeout`
- `error`

Cases like `NotSynced`, `Slashable`, `SameData`, `GreaterThanCurrentSlot`, etc. read more like reasons/subcategories under `error` than peer outcomes alongside `success`.

So I think the main question is:

What exact code for [#924](https://github.com/sigp/anchor/issues/924) becomes materially simpler or safer if this PR lands first?

If there is a concrete follow-up that needs `errors.rs` immediately, that would help justify the split. Otherwise, this currently reads more like leftover prep work from a larger branch than a clear prerequisite for the proposer-duty instrumentation itself.

I'm not opposed to the extraction on its own, but I'd like to understand the dependency, because right now the connection to the milestone is not obvious to me.

jnhsigmap:
> I'm struggling to understand why this needs to land ahead of [#924](https://github.com/sigp/anchor/issues/924), so I want to sanity-check the sequencing.
> 
> From the issue/design side, my understanding is:
> 
> * [#924](https://github.com/sigp/anchor/issues/924) is about adding the proposer-duty root span and checkpoints around `sign_block` in `validator_store`.
> * [#919](https://github.com/sigp/anchor/issues/919) already defined the observability model, including the coarse top-level outcome taxonomy: `success`, `timeout`, `error`.
> * exporter/backend choices like OTLP are out of scope for this milestone.
> 
> Given that, I'm struggling to see why this extraction needs to land first.
> 
> In particular:
> 
> * the current diff is mostly moving `Error` / `SpecificError` into `errors.rs`
> * the proposer-specific follow-up type/module that seemed like the consumer of this extraction does not appear to be part of this PR anymore
> * if the goal is just to classify proposer outcomes for spans/metrics, I don't yet see why that requires moving `SpecificError` out of `lib.rs` first
> 
> More concretely, I think there are two separate concerns here:
> 
> 1. instrumentation outcome
> 2. detailed failure reason
> 
> Per [#919](https://github.com/sigp/anchor/issues/919), the top-level outcome sounds like it should stay coarse:
> 
> * `success`
> * `timeout`
> * `error`
> 
> Cases like `NotSynced`, `Slashable`, `SameData`, `GreaterThanCurrentSlot`, etc. read more like reasons/subcategories under `error` than peer outcomes alongside `success`.
> 
> So I think the main question is:
> 
> What exact code for [#924](https://github.com/sigp/anchor/issues/924) becomes materially simpler or safer if this PR lands first?
> 
> If there is a concrete follow-up that needs `errors.rs` immediately, that would help justify the split. Otherwise, this currently reads more like leftover prep work from a larger branch than a clear prerequisite for the proposer-duty instrumentation itself.
> 
> I'm not opposed to the extraction on its own, but I'd like to understand the dependency, because right now the connection to the milestone is not obvious to me.

It was purely to enable use of these types/enums in building of this milestone from other modules. `lib.rs` is quite large, I am aware of adding more functions, types, or any other definitions to that file as they increase its size and complexity.

If it needs to be `pub` as per claude's review then we can make it so. In absence of being able to submit a full detailed plan of my entire approach to this milestone, I was going to create dependents of this and `SignBlockOutcome` (as seen above, I will create this when I go on with further changes) to implement the taxonomy you outlined in the original spec.

Since the preference is for small PRs, this change alone is 81 lines without any of the instrumentation components that handle this root span. So this is how I intended to break it down.

diegomrsantos:
Thanks, that helps clarify the intent.

I think my view is still that the tradeoff here favors keeping the instrumentation work self-contained.

If `#924` only needs a small amount of additional structure, I'd rather accept some temporary growth in `validator_store::lib.rs` (or a small private module in the same PR) and review the proposer instrumentation end-to-end. If there were a real structural blocker, I'd be in favor of refactoring first, but I'm not seeing that here.

So from my side:

- `#924` feels like a focused instrumentation task, not a refactor task
- if it needs a small helper enum/classifier, that seems fine to introduce alongside the instrumentation itself
- if avoiding further growth in `lib.rs` is the goal, I think a private module in the same instrumentation PR would still address that
- `#896` still reads as a separate refactor concern unless there is a concrete blocker that makes the milestone awkward or unsafe to implement as-is

So I think I'd prefer to keep the milestone work focused on the proposer-duty instrumentation for now, and tackle the broader `validator_store` cleanup/refactor separately later.

Issue #958, opened 2026-04-15
test: add CommitteeMemberTest spec coverage


shane-moore:
## Parent

Part of #954 (split from PR #840).

## What

Add `CommitteeMemberTest` to the spec test suite. Mirrors Go's `CommitteeMemberTest` which validates a `SignedSSVMessage` then checks quorum and full-committee thresholds using deduplicated signer count.

## Reference implementation

Branch `test/simple-spec-test-types`, file: `anchor/spec_tests/src/types/committee_member.rs`

## Details

- Uses `RawSSVMessage` DTO (from `utils/dtos.rs`) since `SSVMessage` doesn't derive `Deserialize`
- Validates the signed message via pre-checks (empty sigs, null message) then `SignedSSVMessage::new()`
- Maps `SignedSSVMessageError` variants to Go integer error codes (reuses `error_codes` module)
- Checks `has_quorum` (unique signers >= 2f+1) and `full_committee` (unique signers == committee size)
- Uses `Vec<u64>` for operator IDs with manual conversion to `OperatorId` (no `Deserialize` on `OperatorId`)
- 4 fixtures: `committeemember.CommitteeMemberTest_committee_member_*.json`

## Changes needed

1. New file: `anchor/spec_tests/src/types/committee_member.rs`
2. Add module + re-export in `anchor/spec_tests/src/types/mod.rs`
3. Add dispatch arm `"committeemember.CommitteeMemberTest"` in `anchor/spec_tests/src/runner/types_dispatcher.rs`

Issue #957, opened 2026-04-15
test: add BeaconDepositDataSpecTest spec coverage


shane-moore:
## Parent

Part of #954 (split from PR #840).

## What

Add `BeaconDepositDataSpecTest` to the spec test suite. Mirrors Go's `GenerateETHDepositData()` which validates deposit signing root computation.

## Reference implementation

Branch `test/simple-spec-test-types`, file: `anchor/spec_tests/src/types/beacon_deposit_data.rs`

## Details

- Parses validator public key, withdrawal credentials, and fork version from fixture
- Constructs a `DepositMessage` with `ChainSpec::mainnet().max_effective_balance`
- Computes domain via `spec.compute_domain(Domain::Deposit, fork_version, Hash256::ZERO)`
- Validates `signing_root` matches expected value
- Fixture validator PK has no `0x` prefix (test prepends it)
- 1 fixture: `beacon.DepositDataSpecTest_deposit_data_root_and_ssz.json`

## Changes needed

1. New file: `anchor/spec_tests/src/types/beacon_deposit_data.rs`
2. Add module + re-export in `anchor/spec_tests/src/types/mod.rs`
3. Add dispatch arm `"beacon.DepositDataSpecTest"` in `anchor/spec_tests/src/runner/types_dispatcher.rs`

Issue #956, opened 2026-04-15
test: add DutySpecTest spec coverage


shane-moore:
## Parent

Part of #954 (split from PR #840).

## What

Add `DutySpecTest` to the spec test suite. Mirrors Go's `MapDutyToRunnerRole()` which maps `BeaconRole` constants to runner duty role integers.

## Reference implementation

Branch `test/simple-spec-test-types`, file: `anchor/spec_tests/src/types/duty.rs`

## Details

- Maps `BeaconRole` to duty role: attester/sync_committee -> 0, proposer -> 2, aggregator/sync_committee_contribution -> 6, validator_registration -> 4, voluntary_exit -> 5, unknown -> -1
- Fixture `beacon_role` is a `u64`, converted to `BeaconRole` via SSZ roundtrip
- No new dependencies or utils needed
- 8 fixtures: `duty.DutySpecTest_duty_map_*.json`

## Changes needed

1. New file: `anchor/spec_tests/src/types/duty.rs`
2. Add module + re-export in `anchor/spec_tests/src/types/mod.rs`
3. Add dispatch arm `"duty.DutySpecTest"` in `anchor/spec_tests/src/runner/types_dispatcher.rs`

Issue #955, opened 2026-04-15
test: add SSZSpecTest spec coverage


shane-moore:
## Parent

Part of #954 (split from PR #840).

## What

Add `SSZSpecTest` to the spec test suite. Mirrors Go's `SSZSpecTest.Run()` which verifies withdrawals `HashTreeRoot` via `ProposerConsensusData`.

## Reference implementation

Branch `test/simple-spec-test-types`, file: `anchor/spec_tests/src/types/ssz_spec_test.rs`

## Details

- Decodes `ProposerConsensusData` from SSZ bytes
- Tries blinded block then full block decode (same order as Go)
- Validates `withdrawals_root` matches expected hash
- Handles BLS validation errors gracefully (fixtures use synthetic BLS that Lighthouse rejects)
- Uses `decode_blinded_block` / `decode_block_contents` methods on `ProposerConsensusData`
- 1 fixture: `ssz.SSZSpecTest_ssz_withdrawals_marshalling.json`

## Changes needed

1. New file: `anchor/spec_tests/src/types/ssz_spec_test.rs`
2. Add module + re-export in `anchor/spec_tests/src/types/mod.rs`
3. Add dispatch arm `"ssz.SSZSpecTest"` in `anchor/spec_tests/src/runner/types_dispatcher.rs`

PR #959, opened 2026-04-15
test: SszSpecTest

shane-moore:

Anchor is missing the `SSZSpecTest` type from `ssv-spec`, leaving SSZ-level parity with the Go reference unverified for proposer consensus data. Regressions in withdrawals merkleization or `ProposerConsensusData` decoding are currently uncovered by the spec-test suite.

Part of the spec-test expansion in #954 (split from #840); sub-issue #955.

## Change Overview (Required)

Adds an `SSZSpecTest` runner that mirrors Go's `SSZSpecTest.Run()`: decode `ProposerConsensusData` from SSZ, extract the beacon block (blinded first, then full), and assert the execution-payload withdrawals root matches the fixture's expected root. Wired into the type dispatcher under `"ssz.SSZSpecTest"` with one fixture.

Intentional non-changes:
- Test-harness only; no production paths touched.
- `ExpectedErrorCode` is not asserted, matching the upstream Go runner. The generator emits only success fixtures today; Anchor-only enforcement would diverge from the reference contract.

The Capella fixture uses placeholder BLS values that Lighthouse rejects during SSZ decode (`BLST_BAD_ENCODING`) but Go's `fastssz` accepts. So the default `cargo test -p spec_tests` path fails decode and never reaches the withdrawals-root assertion; it is kept only for parity with the current `ssv-spec` runner. The meaningful coverage path is `--features fake_crypto`, which skips BLS validation and verifies the withdrawals root end-to-end. #938 tracks making the `fake_crypto` invocation explicit in the Makefile / CI. Upstream fixture fix is [ssvlabs/ssv-spec#622](https://github.com/ssvlabs/ssv-spec/issues/622); once bumped, the fallback becomes dead code (cleanup #953).

## Validation (Required)

- `cargo test -p spec_tests`: 55 passed (parity path).
- `cargo test -p spec_tests --features fake_crypto`: 55 passed (full withdrawals-root verification).
- `cargo clippy`, `cargo +nightly fmt --check`: clean.

## Rollback

N/A, test-only.

## Additional Info

Remaining spec test types from #954: #956, #957, #958.

### DIFF
--- a/anchor/spec_tests/src/runner/types_dispatcher.rs
+++ b/anchor/spec_tests/src/runner/types_dispatcher.rs
@@ -100,6 +100,11 @@ fn dispatch_fixture_by_prefix(prefix: &str, path: &Path, contents: &str) -> Disp
             DispatchOutcome::SkippedKnown
         }
 
+        // SSZ merkleization tests
+        "ssz.SSZSpecTest" => {
+            DispatchOutcome::Executed(run_test::<types::SSZSpecTest>(path, contents))
+        }
+
         // TODO(spec-tests): Add more test types here as they are implemented.
         // This arm will be replaced with panic!() once all test types are added.
         _ => {
--- a/anchor/spec_tests/src/types/mod.rs
+++ b/anchor/spec_tests/src/types/mod.rs
@@ -11,6 +11,7 @@ mod signed_ssv_msg;
 mod signed_ssv_msg_encoding;
 mod ssv_message_encoding;
 mod ssv_msg;
+mod ssz_spec_test;
 
 pub use aggregator_committee_consensus_data_encoding::*;
 pub use beacon_vote_encoding::*;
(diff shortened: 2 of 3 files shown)


shane-moore:
@claude please review this pr

shane-moore:
@claude On (1): I agree in the abstract, but I think it is outside the current fixture contract for this test type. `SSZSpecTest` here is specifically a withdrawals test, and there are no withdrawals pre-Capella, so pre-Capella fixtures would not be valid inputs as currently defined. The upstream `ssv-spec` generator also only emits a Capella fixture today, so I do not think it is worth complicating the error handling in this PR.

I also agree with your other two observations: the lack of `ExpectedErrorCode` handling is intentionally preserving upstream parity, and the degraded non-`fake_crypto` path is already called out in the PR description, with #938 tracking the intended `fake_crypto` Makefile / CI path.

PR #966, opened 2026-04-20
fix(qbft): align quorum rule with ssv-spec (2f + 1)

shane-moore:

Anchor's QBFT quorum used `N - f`, diverging from ssv-spec and Go-SSV (`2f + 1`). The formulas coincide at canonical SSV sizes (`N = 3f + 1`: `{4, 7, 10, 13}`) and diverge elsewhere. `message_validator::compute_quorum_size` already used `2f + 1`, so Anchor's two quorum rules disagreed internally on non-canonical input.

## Change Overview

- `ssv_types`: `quorum_size` now `2f + 1`; new `is_valid_committee_size` helper.
- `qbft::config`: `ConfigBuilder::build` rejects non-canonical sizes; removes the old `[2f + 1, N - f]` range validation; `quorum_size()` is a derived getter (no cached field).
- `message_validator`: duplicate `compute_quorum_size` wrapper deleted; call site uses the shared helper.
- `eth::util::validate_operators`: inline canonical check replaced with the shared helper.

Non-canonical committee sizes are now rejected at `ConfigBuilder::build` (previously accepted); the quorum formula changes from `N - f` to `2f + 1` at those sizes. No production impact: the SSV Network contract already enforces canonical sizes on-chain.

## Validation

`cargo test` + `cargo clippy --all-targets -- -D warnings` + `make cargo-fmt-check`: clean on all touched crates.

## Rollback

`git revert` reverses all changes cleanly. No on-wire, config, or storage impact.

## Blockers / Dependencies

Unblocks #958 (CommitteeMemberTest spec test).

<details>
<summary>Background: spec references, defense-in-depth, why <code>N - f</code> was wrong</summary>

### Why `N - f` was wrong

`N - f` had no comment, design note, or spec citation. It coincides with `2f + 1` at canonical sizes but differs elsewhere (e.g., `N=6`: `N - f = 5`, `2f + 1 = 3`). Since `message_validator::compute_quorum_size` already used `2f + 1`, Anchor had two quorum rules that agreed on all production inputs but would silently disagree on anything non-canonical.

### Spec references

- ssv-spec `types/operator.go:27-33`: `CommitteeMember.GetQuorum` returns `2 * FaultyNodes + 1`; `HasQuorum` compares signers against that.
- Go-SSV `protocol/v2/types/ssvshare.go:228-267`: `ComputeQuorumAndPartialQuorum` returns `(f*2 + 1, f + 1)`; `ValidCommitteeSize` accepts `(N-1) % 3 == 0 && f in [1, 4]`.
- SSV Network contract (`ssvlabs/ssv-network` at `stage-v2.0.0-upgrade7`): `ValidatorLib.validateOperatorsLength` enforces canonical sizes at `registerValidator`.

### Defense-in-depth

Non-canonical sizes are rejected at three layers: the SSV Network contract on-chain, eth event ingestion (`eth::util::validate_operators`), and the qbft config builder. Production is already all-canonical; this PR makes the invariant explicit in Rust as well as on-chain.

### Intentionally unchanged

- `get_f` formula and its `saturating_sub` guard.
- `Cluster::get_f` (used by `validator_store` for the BLS Lagrange reconstruction threshold; separate contract, same value at canonical sizes).
- On-wire behaviour; event format; runtime semantics at canonical sizes.

</details>

### DIFF
--- a/anchor/common/qbft/src/config.rs
+++ b/anchor/common/qbft/src/config.rs
@@ -221,9 +221,10 @@ where
             return Err(ConfigBuilderError::OperatorNotParticipant);
         }
 
-        // Validate `quorum_size`: must fall in `[2f + 1, quorum_size(N)]`.
+        // Validate `quorum_size`: must fall in `[2f + 1, N − f]`.
         let f = get_f(committee_size);
-        if self.quorum_size < f * 2 + 1 || self.quorum_size > quorum_size(committee_size) {
+        let max_quorum = quorum_size(committee_size);
+        if self.quorum_size < f * 2 + 1 || self.quorum_size > max_quorum {
             return Err(ConfigBuilderError::InvalidQuorumSize);
         }
 
--- a/anchor/common/qbft/src/config.rs
+++ b/anchor/common/qbft/src/config.rs
@@ -1,7 +1,7 @@
 use std::{fmt::Debug, time::Duration};
 
 use indexmap::IndexSet;
-use ssv_types::{OperatorId, Round, get_f, quorum_size};
+use ssv_types::{OperatorId, Round, is_valid_committee_size, quorum_size};
 
 use super::error::ConfigBuilderError;
 use crate::qbft_types::{DefaultLeaderFunction, InstanceHeight, LeaderFunction};
@@ -15,7 +15,6 @@ where
     instance_height: InstanceHeight,
     round: Round,
     committee_members: IndexSet<OperatorId>,
-    quorum_size: usize,
     round_time: Duration,
     max_rounds: usize,
     leader_fn: F,
@@ -43,7 +42,7 @@ impl<F: Clone + LeaderFunction> Config<F> {
     }
 
     pub fn quorum_size(&self) -> usize {
-        self.quorum_size
+        quorum_size(self.committee_members.len())
     }
 
     /// How long the round will last
@@ -70,7 +69,6 @@ impl<F: Clone + LeaderFunction> Config<F> {
             round: builder.round,
             round_time: builder.round_time,
             max_rounds: builder.max_rounds,
-            quorum_size: builder.quorum_size,
             leader_fn: builder.leader_fn.clone(),
         }
     }
@@ -92,7 +90,6 @@ where
     round: Round,
     round_time: Duration,
     max_rounds: usize,
-    quorum_size: usize,
 }
 
 impl<F> ConfigBuilder<F>
@@ -104,8 +101,6 @@ where
         instance_height: InstanceHeight,
         committee_members: IndexSet<OperatorId>,
     ) -> Self {
-        let default_quorum = quorum_size(committee_members.len());
-
         ConfigBuilder {
             operator_id,
             instance_height,
@@ -113,7 +108,6 @@ where
             round: Round::default(),
             round_time: Duration::new(2, 0),
             max_rounds: 4,
-            quorum_size: default_quorum,
             leader_fn: F::default(),
         }
     }
@@ -129,8 +123,6 @@ where
         committee_members: IndexSet<OperatorId>,
         leader_fn: F,
     ) -> Self {
-        let default_quorum = quorum_size(committee_members.len());
-
         ConfigBuilder {
             operator_id,
             instance_height,
@@ -138,7 +130,6 @@ where
             round: Round::default(),
             round_time: Duration::new(2, 0),
             max_rounds: 4,
-            quorum_size: default_quorum,
             leader_fn,
         }
     }
@@ -164,7 +155,7 @@ where
     }
 
     pub fn quorum_size(&self) -> usize {
-        self.quorum_size
+        quorum_size(self.committee_members.len())
     }
 
     pub fn max_rounds(&self) -> usize {
@@ -221,11 +212,9 @@ where
             return Err(ConfigBuilderError::OperatorNotParticipant);
         }
 
-        // Validate `quorum_size`: must fall in `[2f + 1, N − f]`.
-        let f = get_f(committee_size);
-        let max_quorum = quorum_size(committee_size);
-        if self.quorum_size < f * 2 + 1 || self.quorum_size > max_quorum {
-            return Err(ConfigBuilderError::InvalidQuorumSize);
+        // Validate `quorum_size`
+        if !is_valid_committee_size(committee_size) {
+            return Err(ConfigBuilderError::InvalidCommitteeSize);
         }
 
         // Validate `max_rounds`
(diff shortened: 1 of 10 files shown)


shane-moore:
@claude, please take a look

shane-moore:
@claude, resolved your two points in last commit. please take a look

anchor/common/ssv_types/src/lib.rs
```diff
@@ -26,6 +26,20 @@ pub use types::{Epoch, Slot};
 pub const RSA_SIGNATURE_SIZE: usize = 256;
 pub const MAX_SIGNATURES: usize = 13;
 
+/// Maximum Byzantine/faulty members a committee of `members` can tolerate:
+/// `f = ⌊(N − 1) / 3⌋`. Returns 0 for empty committees.
+pub fn get_f(members: usize) -> usize {
```
jnhsigmap, line 31: given that these are quite important, would it be worth adding a unit test for each of these functions here as a means to avoid any unintentional mishaps or do the spec tests impose sufficient testing on consensus outcomes?

anchor/message_validator/src/lib.rs
```diff
@@ -1059,8 +1060,7 @@ pub fn sync_committee_period(
 }
 
 pub(crate) fn compute_quorum_size(committee_size: usize) -> usize {
```
jnhsigmap: to me this function is semantically related to `get_f` and `quorum_size`. Would you consider also moving it to `ssv_types` with a clear description of its difference to `quorum_size` and why it exists? If its not in the scope of this change (you wanted primarily to remove duplicated implementations), maybe in a subsequent PR?

anchor/common/ssv_types/src/lib.rs
shane-moore: good call! added some unit tests since the spec tests cant differentiate between the two formulas, `get_f` and `quorum_size`. it uses both and does an assert later

also, we can exercise non-canonical quorum sizes in the unit tests like committee size `3`

anchor/message_validator/src/lib.rs
shane-moore: leaning toward keeping it in `message_validator` for now. `compute_quorum_size` has exactly one production caller (`consensus_message::validate_consensus_message_semantics`) and is `pub(crate)`. promoting it to `ssv_types` buys semantic co-location with `quorum_size` rather than shared reuse. If a second caller shows up, let's revisit then

anchor/common/qbft/src/config.rs
```diff
@@ -221,9 +221,10 @@ where
             return Err(ConfigBuilderError::OperatorNotParticipant);
         }
 
-        // Validate `quorum_size`: must fall in `[2f + 1, quorum_size(N)]`.
+        // Validate `quorum_size`: must fall in `[2f + 1, N − f]`.
```
diegomrsantos: From where is this requirement?

anchor/common/ssv_types/src/lib.rs
```diff
@@ -37,3 +51,48 @@ where
...
+            (13, 4), // canonical SSV
+        ];
+        for (n, expected) in cases {
+            assert_eq!(get_f(n), expected, "get_f({n})");
+        }
+    }
+
+    // Regression targets: silent `2f + 1` substitution for `N − f`.
+    #[test]
+    fn quorum_size_is_n_minus_f() {
```
diegomrsantos: Nice to add coverage here. I think these expectations should be updated to match the upstream quorum rule instead of locking in the current behavior.

Go SSV and `ssv-spec` define quorum as `2f + 1`, so if `quorum_size()` is the shared helper, the cases here should assert that directly. For example: `N=3 -> 1`, `N=4 -> 3`, `N=6 -> 3`, `N=7 -> 5`.

As written, this test suite cements the wrong contract in `ssv_types`.

anchor/common/ssv_types/src/lib.rs
```diff
@@ -26,6 +26,20 @@ pub use types::{Epoch, Slot};
 pub const RSA_SIGNATURE_SIZE: usize = 256;
 pub const MAX_SIGNATURES: usize = 13;
 
+/// Maximum Byzantine/faulty members a committee of `members` can tolerate:
+/// `f = ⌊(N − 1) / 3⌋`. Returns 0 for empty committees.
+pub fn get_f(members: usize) -> usize {
+    members.saturating_sub(1) / 3
+}
+
+/// Default QBFT quorum: `N − f`. Equivalent to `2f + 1` when `N = 3f + 1`
```
diegomrsantos: I think `quorum_size()` should match the upstream Go/spec quorum rule rather than preserve Anchor’s local `N - f` default.

Current Go SSV and `ssv-spec` model quorum as `2f + 1`, and `message_validator::compute_quorum_size()` already does that in this repo. Defining the shared `ssv_types::quorum_size()` helper as `N - f` makes the canonical helper disagree with both upstream and existing Anchor validation logic.

I think this should be:

```rust
pub fn quorum_size(committee_size: usize) -> usize {
    get_f(committee_size) * 2 + 1
}
```

anchor/common/qbft/src/config.rs
diegomrsantos: To elaborate a bit: this is not just a comment issue, the code is actually accepting any `quorum_size` in the interval `[2f + 1, N - f]`.

I don’t think that matches the upstream contract. Go SSV / `ssv-spec` define quorum as exactly `2f + 1`, not as a range of valid thresholds. So this builder is preserving an Anchor-local notion that multiple quorum values are acceptable for the same committee size.

Also, there does not seem to be a public `with_quorum_size(...)` setter anymore, which makes this even harder to reason about: why is quorum treated as an independently validated field here instead of being derived from the committee size?

I’d expect this to enforce one exact quorum rule for the committee, not a lower/upper bound interval.

anchor/message_validator/src/lib.rs
diegomrsantos: Once `ssv_types::quorum_size()` is corrected to match Go/`ssv-spec`, I think this local helper should go away and the call sites here should use the shared helper directly.

Right now the actual quorum formula is still duplicated in `message_validator`, which leaves us with two places to keep in sync for the same protocol rule.

anchor/common/qbft/src/config.rs
diegomrsantos: We should also validate `committee_size` itself here to match the upstream Go/SSV contract.

Go SSV does not treat arbitrary committee sizes as valid QBFT input; it expects canonical `N = 3f + 1` sizes. Right now this builder validates quorum against `committee_size`, but it does not validate that the committee size is canonical in the first place.

If we want Go/spec parity, this check should live here at the config/build boundary rather than implicitly allowing non-canonical committee sizes through.

anchor/common/ssv_types/src/lib.rs
shane-moore: Done. `quorum_size()` is now `2f + 1`, and the test asserts exactly the values you called out:

```
(3, 1), (4, 3), (6, 3), (7, 5), (10, 7), (13, 9)
```

Plus a `quorum_size_composes_with_get_f` property test across `0..=MAX_SIGNATURES` so a future inline refactor can't silently decouple the two.

anchor/common/ssv_types/src/lib.rs
shane-moore: Applied verbatim. `quorum_size` is now `get_f(committee_size) * 2 + 1`, with a doc comment citing ssv-spec's `CommitteeMember.GetQuorum` and Go-SSV's `ComputeQuorumAndPartialQuorum`.

anchor/message_validator/src/lib.rs
shane-moore: Done. `compute_quorum_size` and its test are gone. Its one call site in `consensus_message::validate_consensus_message_semantics` now calls `ssv_types::quorum_size` directly. Single authoritative quorum formula in the workspace.

shane-moore:
@claude, please take a look

anchor/common/qbft/src/config.rs
shane-moore: Good pushes - all three points here addressed together:

**Range check (`[2f + 1, N - f]`):** no documented rationale, and since there was never a public `with_quorum_size(...)` setter, the interval was unreachable in practice. Removed entirely.

**Quorum derivation:** dropped the cached `quorum_size` field from `Config`/`ConfigBuilder`. `quorum_size()` is now a derived getter from `committee_members.len()`, routed through the shared `ssv_types::quorum_size` helper (aligned with ssv-spec's `2f + 1`). It can't drift from the committee size anymore.

**Canonical committee size validation:** added at the config/build boundary via `ssv_types::is_valid_committee_size` (canonical `N = 3f + 1`, `f` in `1..=4`), returning a new `InvalidCommitteeSize` variant. Also consolidated `eth::util::validate_operators` onto the same helper (it had its own inline check), so canonical-size policy lives in exactly one place. Three-layer defense: SSV Network contract -> eth event ingestion -> qbft config build.

diegomrsantos:
Thanks for all the updates here.

One meta point: I don’t think `refactor` is the right framing for this PR anymore. It now changes the shared quorum rule, enforces canonical committee sizes in QBFT config, and removes the old range-based quorum validation. This looks like a fix, not just a restructuring, so I think the title/description should reflect that.

diegomrsantos (approved):
Thanks a lot!

Issue #920, opened 2026-04-02
Instrument proposer pre-QBFT fetch/build boundary timing


diegomrsantos:
## Summary
Instrument the pre-QBFT proposer path at Anchor-owned boundaries so we can quantify how much slot budget is spent before QBFT starts.

## Motivation
A major suspected bottleneck is fetch-before-consensus latency. SSV has an open proposer-liveness issue for the same class of problem in [proposer-runner: QBFT liveness choked by Beacon node](https://github.com/ssvlabs/ssv/issues/1825) and an open spec direction in [Do not pass value to instance](https://github.com/ssvlabs/ssv-spec/pull/371). Anchor appears exposed to the same class of risk, but we should measure it first.

## Relation to SSV
This is directly related to SSV's open proposer liveness work, but this issue is measurement-only. We are not adopting the async-fetch design yet. We are gathering the data needed to decide whether that class of change is warranted in Anchor.

## Scope
Add instrumentation around the Anchor-owned proposer boundaries leading into QBFT, including:
- when proposer signing starts
- when the unsigned block is available to Anchor's proposer path
- when we hand off into QBFT / `decide_abstract_block`
- remaining slot budget at QBFT start

This issue should stay on the Anchor side first. If dependency-side fetch/build timing in pinned Lighthouse remains too opaque after this work, we can open a follow-up issue.

## Out of Scope
- patching the pinned Lighthouse dependency
- changing block-fetch, relay, or proposal-selection behavior
- changing timeout values

## Acceptance Criteria
- proposer duties expose a measurable `pre_qbft` duration from Anchor-owned boundaries
- we can compute or log the remaining slot budget at QBFT start
- the instrumentation is enough to tell whether the dominant pre-QBFT cost is visible from Anchor's side
- any remaining Lighthouse-side blind spot is documented explicitly

## References
- [proposer-runner: QBFT liveness choked by Beacon node](https://github.com/ssvlabs/ssv/issues/1825)
- [Do not pass value to instance](https://github.com/ssvlabs/ssv-spec/pull/371)
- [docs: (mev-considerations) update round-change-based estimates](https://github.com/ssvlabs/ssv/pull/2416)

jnhsigmap:
what is your view of when proposer signing starts? Is this still contained to `sign_block` or is `randao_reveal` technically when the clock starts for anchor's proposer? It appears the main concern for this issue is the slot budget remaining when `sign_block` is entered, and more importantly at the `decide_instance` call?

[...]

PR #967, opened 2026-04-21
feat: Adds root proposer-duty span and top level checkpoints to sign_block in validator_store

jnhsigmap:

Proposed solution for #924. This change adds a root instrumentation span to the `sign_block` function in `validator_store`.

Follows [observability taxonomy](https://github.com/sigp/anchor/issues/919#issuecomment-4177726189) outlined in #919, omitting components that over complicating the approach as was highlighted in my first attempt.

## Change Overview (Required)

New module `instrumentation` created to store simple enum encapsulating the initial outcome taxonomy for the block signing process. A module within `instrumentation` was provided to group the various checkpoints that are instrumented in this change as a lightweight way to represent the universe of which components of `sign_block` are being tracked.

Checkpoints marked:
- duty entry
- pre-QBFT handoff
- consensus returned
- block signed
- publish path selected
- duty success / failure

Root span components:
`block_type`: full or blinded.
`cluster_size`: Correlates the number of operators in the quorum and number of messages to process to the duration of the round.
`clock_slot`: Slot number (1-32).
`proposal_matched`: Indicator of whether this operator was the round leader. This will produce metrics for number of rounds where leadership is obtained and relative differences in processing time when not a leader.
`outcome`: Indicated by `success`, `timeout`, or `error`.
`block_slot`: Block slot.
`validator_pubkey`: Public key of this validator.
`validator_index`: Index of this validator.

Detail from the spec not implemented _"failures are recorded with an explicit outcome and reason"_. This was not possible without more granular error mapping and complexity and it appears that ssvlabs didn't do this in a structured way in their OTeL implementation.

This wouldn't be able to be done without defying the taxonomy stated in the spec, or creating a verbose exhaustive mapping of all error cases to an outcome reason:
```
...
          SpecificError::SignatureCollectionFailed(_) => "signature_collection_failed",                                                                                   
          SpecificError::ArithError(_) => "arith_error",                                                                                                                  
          SpecificError::QbftError(_) => "qbft_error",                                                                                                                    
          SpecificError::Timeout => "timeout",  
...
```

Risks are limited as the span and traces lead nowhere without a collector and do not add much overhead. The new enums are 1:1 mappings of outcome or checkpoint to the variant. This should not create problems without being caught clearly in review.

Trade-offs were that this was just implemented in-place without more careful organising of interface separation and 

## Validation (Required)

`cargo test --workspace`
`cargo clippy`
`cargo check`

It can be safely reverted. This would remove the trace and checkpoints and the new `instrumentation` module.

### DIFF
--- a/anchor/validator_store/src/instrumentation.rs
+++ b/anchor/validator_store/src/instrumentation.rs
@@ -1,25 +1,10 @@
-use derive_more::Display;
-
 use crate::{Error, SpecificError};
 
 /// Outcome of a block signing attempt, used for metrics labeling.
-#[derive(Debug, Display)]
-pub enum SignBlockOutcome {
-    #[display("success")]
-    Success,
-    #[display("timeout")]
-    Timeout,
-    #[display("error")]
-    Failed,
-}
-
-impl SignBlockOutcome {
-    pub fn from_result<T>(result: &Result<T, Error>) -> Self {
-        match result {
-            Ok(_) => Self::Success,
-            Err(Error::SpecificError(SpecificError::Timeout)) => Self::Timeout,
-            Err(_) => Self::Failed,
-        }
+pub fn from_result<T>(result: &Result<T, Error>) -> &'static str {
+    match result {
+        Ok(_) => "success",
+        Err(_) => "failed",
     }
 }
 
@@ -31,4 +16,25 @@ pub mod checkpoints {
     pub const PUBLISH_BLOCK: &str = "publish_block";
     pub const DUTY_COMPLETED: &str = "duty_completed";
     pub const DUTY_FAILED: &str = "duty_failed";
+}
+
+/// Common reasons for block signing failures.
+pub fn failure_reason(error: &Error) -> &'static str {
+    match error {
+        Error::SpecificError(SpecificError::ClusterLiquidated) => "cluster_liquidated",
+        Error::SpecificError(SpecificError::InvalidQbftData(_)) => "invalid_qbft_data",
+        Error::SpecificError(SpecificError::MissingIndex) => "missing_index",
+        Error::SpecificError(SpecificError::NotSynced) => "not_synced",
+        Error::SpecificError(SpecificError::QbftError(_)) => "qbft_error",
+        Error::SpecificError(SpecificError::SignatureCollectionFailed(_)) => {
+            "signature_collection_failed"
+        }
+        Error::SpecificError(SpecificError::SlotClock) => "slot_clock",
+        Error::SpecificError(SpecificError::Timeout) => "timeout",
+        Error::Slashable(_) => "slashable",
+        Error::SameData => "same_data",
+        Error::SpecificError(_) => "other_specific_error",
+        Error::UnknownPubkey(_) => "unknown_pubkey",
+        _ => "other_error",
+    }
 }
--- a/anchor/validator_store/src/instrumentation.rs
+++ b/anchor/validator_store/src/instrumentation.rs
@@ -1,0 +1,47 @@
+use crate::{Error, SpecificError};
+
+/// Outcome of a block signing attempt, used for metrics labeling.
+#[derive(Debug)]
+pub enum SignBlockOutcome {
+    Success,
+    Timeout,
+    Failed,
+}
+
+impl SignBlockOutcome {
+    pub fn as_str(&self) -> &'static str {
+        match self {
+            Self::Success => "success",
+            Self::Timeout => "timeout",
+            Self::Failed => "error",
+        }
+    }
+    pub fn from_result<T>(result: &Result<T, Error>) -> Self {
+        match result {
+            Ok(_) => Self::Success,
+            Err(Error::SpecificError(SpecificError::Timeout)) => Self::Timeout,
+            Err(_) => Self::Failed,
+        }
+    }
+}
+
+#[derive(Debug)]
+pub enum BlockSigningCheckpoints {
+    DutyEntry,
+    PreConsensusHandoff,
+    ConsensusReturned,
+    BlockSigned,
+    PublishConsensus,
+}
+
+impl BlockSigningCheckpoints {
+    pub fn as_str(&self) -> &'static str {
+        match self {
+            Self::DutyEntry => "duty_entry",
+            Self::PreConsensusHandoff => "pre_consensus_handoff",
+            Self::ConsensusReturned => "consensus_returned",
+            Self::BlockSigned => "block_signed",
+            Self::PublishConsensus => "publish_consensus",
+        }
+    }
+}
--- a/anchor/validator_store/src/instrumentation.rs
+++ b/anchor/validator_store/src/instrumentation.rs
@@ -29,7 +29,7 @@ impl SignBlockOutcome {
 pub enum BlockSigningCheckpoints {
     DutyEntry,
     PreConsensusHandoff,
-    ConsensusReturned,
+    ConsensusDecided,
     BlockSigned,
     PublishConsensus,
 }
@@ -38,8 +38,8 @@ impl BlockSigningCheckpoints {
     pub fn as_str(&self) -> &'static str {
         match self {
             Self::DutyEntry => "duty_entry",
-            Self::PreConsensusHandoff => "pre_consensus_handoff",
-            Self::ConsensusReturned => "consensus_returned",
+            Self::PreConsensusHandoff => "qbft_start",
+            Self::ConsensusDecided => "consensus_decided",
             Self::BlockSigned => "block_signed",
             Self::PublishConsensus => "publish_consensus",
         }
--- a/anchor/validator_store/src/instrumentation.rs
+++ b/anchor/validator_store/src/instrumentation.rs
@@ -1,7 +1,7 @@
 use crate::{Error, SpecificError};
 
 /// Outcome of a block signing attempt, used for metrics labeling.
-pub fn from_result<T>(result: &Result<T, Error>) -> &'static str {
+pub fn outcome_from_result<T>(result: &Result<T, Error>) -> &'static str {
     match result {
         Ok(_) => "success",
         Err(_) => "failed",
@@ -24,6 +24,7 @@ pub fn failure_reason(error: &Error) -> &'static str {
         Error::SpecificError(SpecificError::ClusterLiquidated) => "cluster_liquidated",
         Error::SpecificError(SpecificError::InvalidQbftData(_)) => "invalid_qbft_data",
         Error::SpecificError(SpecificError::MissingIndex) => "missing_index",
+        Error::SpecificError(SpecificError::NoDataAgreed) => "no_data_agreed",
         Error::SpecificError(SpecificError::NotSynced) => "not_synced",
         Error::SpecificError(SpecificError::QbftError(_)) => "qbft_error",
         Error::SpecificError(SpecificError::SignatureCollectionFailed(_)) => {
--- a/anchor/validator_store/src/instrumentation.rs
+++ b/anchor/validator_store/src/instrumentation.rs
@@ -1,21 +1,19 @@
+use derive_more::Display;
+
 use crate::{Error, SpecificError};
 
 /// Outcome of a block signing attempt, used for metrics labeling.
-#[derive(Debug)]
+#[derive(Debug, Display)]
 pub enum SignBlockOutcome {
+    #[display("success")]
     Success,
+    #[display("timeout")]
     Timeout,
+    #[display("error")]
     Failed,
 }
 
 impl SignBlockOutcome {
-    pub fn as_str(&self) -> &'static str {
-        match self {
-            Self::Success => "success",
-            Self::Timeout => "timeout",
-            Self::Failed => "error",
-        }
-    }
     pub fn from_result<T>(result: &Result<T, Error>) -> Self {
         match result {
             Ok(_) => Self::Success,
@@ -25,23 +23,12 @@ impl SignBlockOutcome {
     }
 }
 
-#[derive(Debug)]
-pub enum BlockSigningCheckpoints {
-    DutyEntry,
-    PreConsensusHandoff,
-    ConsensusDecided,
-    BlockSigned,
-    PublishConsensus,
-}
-
-impl BlockSigningCheckpoints {
-    pub fn as_str(&self) -> &'static str {
-        match self {
-            Self::DutyEntry => "duty_entry",
-            Self::PreConsensusHandoff => "qbft_start",
-            Self::ConsensusDecided => "consensus_decided",
-            Self::BlockSigned => "block_signed",
-            Self::PublishConsensus => "publish_consensus",
-        }
-    }
+pub mod checkpoints {
+    pub const DUTY_ENTRY: &str = "duty_entry";
+    pub const PRE_CONSENSUS_HANDOFF: &str = "pre_consensus_handoff";
+    pub const CONSENSUS_DECIDED: &str = "consensus_decided";
+    pub const BLOCK_SIGNED: &str = "block_signed";
+    pub const PUBLISH_BLOCK: &str = "publish_block";
+    pub const DUTY_COMPLETED: &str = "duty_completed";
+    pub const DUTY_FAILED: &str = "duty_failed";
 }
--- a/anchor/validator_store/src/lib.rs
+++ b/anchor/validator_store/src/lib.rs
@@ -80,8 +80,6 @@ use validator_store::{
     ValidatorStore,
 };
 
-use crate::instrumentation::{SignBlockOutcome, checkpoints};
-
 /// Number of epochs of slashing protection history to keep.
 ///
 /// This acts as a maximum safe-guard against clock drift.
@@ -2331,7 +2329,7 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> ValidatorStore
         );
         let future = async {
             info!(
-                checkpoint = checkpoints::DUTY_ENTRY,
+                checkpoint = instrumentation::checkpoints::DUTY_ENTRY,
                 "Proposer block signing duty entered"
             );
             let result = async {
@@ -2360,7 +2358,7 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> ValidatorStore
                 };
 
                 info!(
-                    checkpoint = checkpoints::PRE_CONSENSUS_HANDOFF,
+                    checkpoint = instrumentation::checkpoints::PRE_CONSENSUS_HANDOFF,
                     "Handing block to consensus process"
                 );
 
@@ -2369,7 +2367,7 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> ValidatorStore
                     .await?;
 
                 info!(
-                    checkpoint = checkpoints::CONSENSUS_DECIDED,
+                    checkpoint = instrumentation::checkpoints::CONSENSUS_DECIDED,
                     "Block consensus completed successfully"
                 );
 
@@ -2391,7 +2389,7 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> ValidatorStore
                 }?;
 
                 info!(
-                    checkpoint = checkpoints::BLOCK_SIGNED,
+                    checkpoint = instrumentation::checkpoints::BLOCK_SIGNED,
                     "Block threshold signature completed"
                 );
 
@@ -2408,7 +2406,7 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> ValidatorStore
                                     signed_blinded_block.signature().clone(),
                                 );
                                 info!(
-                                    checkpoint = checkpoints::PUBLISH_BLOCK,
+                                    checkpoint = instrumentation::checkpoints::PUBLISH_BLOCK,
                                     "Publishing full reconstructed block as leader"
                                 );
                                 Ok(SignedBlock::Full(PublishBlockRequest::new(
@@ -2417,7 +2415,7 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> ValidatorStore
                                 )))
                             } else {
                                 info!(
-                                    checkpoint = checkpoints::PUBLISH_BLOCK,
+                                    checkpoint = instrumentation::checkpoints::PUBLISH_BLOCK,
                                     "Publishing blinded block as leader"
                                 );
                                 Ok(SignedBlock::Blinded(signed_blinded_block))
@@ -2425,7 +2423,7 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> ValidatorStore
                         } else {
                             tracing::Span::current().record("proposal_matched", false);
                             info!(
-                                checkpoint = checkpoints::PUBLISH_BLOCK,
+                                checkpoint = instrumentation::checkpoints::PUBLISH_BLOCK,
                                 "Publishing blinded block not as leader"
                             );
                             // Someone else's proposal won, return blinded
@@ -2435,7 +2433,7 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> ValidatorStore
                     SignedBlock::Full(signed_block) => {
                         tracing::Span::current().record("proposal_matched", false);
                         info!(
-                            checkpoint = checkpoints::PUBLISH_BLOCK,
+                            checkpoint = instrumentation::checkpoints::PUBLISH_BLOCK,
                             "Publishing full block directly"
                         );
                         Ok(SignedBlock::Full(signed_block))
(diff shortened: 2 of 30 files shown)


shane-moore:
@claude, please review

anchor/validator_store/src/instrumentation.rs
```diff
@@ -0,0 +1,47 @@
...
+
+impl SignBlockOutcome {
+    pub fn as_str(&self) -> &'static str {
+        match self {
+            Self::Success => "success",
+            Self::Timeout => "timeout",
+            Self::Failed => "error",
+        }
+    }
+    pub fn from_result<T>(result: &Result<T, Error>) -> Self {
```
diegomrsantos: Given the rest of the proposer observability milestone (#920/#921/#923/#924), I think a shared instrumentation module makes sense, but I am not convinced the enum wrappers do. The reusable part here is the classification policy in from_result: a helper like fn outcome_from_result<T>(...) -> &'static str would keep one canonical mapping for success/timeout/error without the extra enum -> string step. Similarly, the checkpoint vocabulary could probably just live here as const &str values.

anchor/validator_store/src/lib.rs
```diff
@@ -2346,39 +2395,67 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> ValidatorStore
...
 
-        run_and_update_metrics(
+        let result = run_and_update_metrics(
             BLOCK_LOG_NAME,
             &validator_metrics::SIGNED_BLOCKS_TOTAL,
             future,
         )
-        .await
+        .await;
+        span.record("outcome", SignBlockOutcome::from_result(&result).as_str());
```
diegomrsantos: I think this is mixing two different concerns: instrumenting the future gives the async body the root span context, but this line is only recording a span field, not emitting a terminal checkpoint event. Because the instrumented body has several early ? exits, an error path can leave after the last successful checkpoint without ever producing an explicit duty_failed or duty_done event, and there is still no failure reason. I think the terminal success or failure checkpoint should be emitted inside the instrumented future before it returns, for example by logging duty_failed with outcome and reason on Err and duty_done with outcome=success on Ok.

anchor/validator_store/src/lib.rs
```diff
@@ -2307,12 +2310,48 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> ValidatorStore
...
+                checkpoint = BlockSigningCheckpoints::DutyEntry.as_str(),
+                "Block signing duty entered"
+            );
+
             if !*self.is_synced.borrow() {
                 return Err(Error::SpecificError(SpecificError::NotSynced));
             }
             let (validator, cluster) = self.get_validator_and_cluster(validator_pubkey)?;
 
+            let cluster_size = cluster.cluster_members.len() as u64; // tracing::Value implemented for i64 and u64.
```
shane-moore: i think we can remove the cast here since `tracing::Value` is implemented for `usize` directly

anchor/validator_store/src/lib.rs
diegomrsantos: The practical shape I had in mind is to keep both the terminal checkpoint and the final span attribute inside the instrumented future, so every early `?` path still converges on one explicit end-of-duty marker:

```rust
let future = async {
    info!(checkpoint = DUTY_ENTRY);

    let result = async {
        // existing body with ? paths
        Ok(value)
    }
    .await;

    tracing::Span::current().record("outcome", outcome_from_result(&result));

    match &result {
        Ok(_) => info!(checkpoint = DUTY_DONE, outcome = "success"),
        Err(err) => info!(
            checkpoint = DUTY_FAILED,
            outcome = outcome_from_result(&result),
            reason = failure_reason(err),
        ),
    }

    result
}
.instrument(span.clone());
```

That gives you both pieces: the root span still carries the final outcome, and the trace also gets a timestamped terminal checkpoint instead of just stopping at the last successful phase.

anchor/validator_store/src/lib.rs
```diff
@@ -2307,12 +2310,48 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> ValidatorStore
         block: UnsignedBlock<E>,
         current_slot: Slot,
     ) -> Result<SignedBlock<E>, Error> {
+        let block_type = match &block {
```
shane-moore: i think simpler would be:

 ```rust
let (block_type, block_slot) = match &block {
     UnsignedBlock::Full(FullBlockContents::BlockContents(contents)) => {
         ("full", contents.block.slot())
     }
   UnsignedBlock::Full(FullBlockContents::Block(block)) => ("full", block.slot()),
    UnsignedBlock::Blinded(block) => ("blinded", block.slot()),
};
```

File: anchor/validator_store/src/lib.rs [hunk 5]
```diff
@@ -2307,12 +2310,48 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> ValidatorStore
...
+        };
+        let block_slot = match block {
+            UnsignedBlock::Full(FullBlockContents::BlockContents(ref contents)) => {
+                contents.block.slot()
+            }
+            UnsignedBlock::Full(FullBlockContents::Block(ref block)) => block.slot(),
+            UnsignedBlock::Blinded(ref block) => block.slot(),
+        };
+
+        // Root span
```
shane-moore: I was curious if this is actually the root span, so I looked into Lighthouse's block service, which drives the Anchor validator store. Turns out it's not, and the name also collides.
>
> Lighthouse wraps this call twice before it reaches us. First the outer duty cycle at [`[block_service.rs:406-410](https://github.com/sigp/lighthouse/blob/4b3a9d3/validator_client/validator_services/src/block_service.rs#L406-L410)`](https://github.com/sigp/lighthouse/blob/4b3a9d3/validator_client/validator_services/src/block_service.rs#L406-L410):
>
> ```rust
> #[instrument(
>     name = "block_proposal_duty_cycle",
>     skip_all,
>     fields(%slot, ?validator_pubkey)
> )]
> async fn get_validator_block_and_publish_block(...) { ... }
> ```
>
> Then an explicit wrap right at the call site, [`[block_service.rs:346-350](https://github.com/sigp/lighthouse/blob/4b3a9d3/validator_client/validator_services/src/block_service.rs#L346-L350)`](https://github.com/sigp/lighthouse/blob/4b3a9d3/validator_client/validator_services/src/block_service.rs#L346-L350):
>
> ```rust
> let res = self
>     .validator_store
>     .sign_block(*validator_pubkey, unsigned_block, slot)
>     .instrument(info_span!("sign_block"))
>     .await;
> ```
>
> So the actual captured tree this PR produces is:
>
> ```
> block_proposal_duty_cycle   (Lighthouse, fields: slot, validator_pubkey)
> └── sign_block              (Lighthouse, fields: empty)
>     └── sign_block          (Anchor, fields: block_type, cluster_size, ...)
> ```
>
> Two `sign_block` spans sit directly nested. A `span.name = "sign_block"` query in Tempo / Jaeger / any trace backend returns both, and tooling that collapses by name has to disambiguate by fields.
>
> #919 actually named the right thing for this exact reason. Diego's resolution comment ([[2026-04-02](https://github.com/sigp/anchor/issues/919#issuecomment-4177726189)](https://github.com/sigp/anchor/issues/919#issuecomment-4177726189)) sketches the proposer-duty root as `proposer_sign_block`, not `sign_block`. The `proposer_` prefix is the namespace that keeps Anchor's duty root distinct from the generic Lighthouse span.
>
> Given that, the `// Root span` comment above the declaration is also off. It restates what `info_span!` already says, and "root" here is #924's prose sense, not tracing's structural sense. After the rename the comment is visibly stale, so it can just go

anchor/validator_store/src/lib.rs
```diff
@@ -2346,39 +2395,67 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> ValidatorStore
...
+                        info!(
+                            checkpoint = BlockSigningCheckpoints::PublishConsensus.as_str(),
+                            "Publishing blinded block not as leader"
+                        );
                         // Someone else's proposal won, return blinded
                         Ok(SignedBlock::Blinded(signed_blinded_block))
                     }
                 }
-                SignedBlock::Full(signed_block) => Ok(SignedBlock::Full(signed_block)),
+                SignedBlock::Full(signed_block) => {
```
shane-moore: Suggest `span.record("proposal_matched", false)` here, matching the Blinded `else` at line 2431. Leaving it `Empty` is indistinguishable from "bug forgot to record" on a dashboard.
>
> Reasoning: Anchor always strips to blinded before QBFT (`lib.rs:2355-2365`, `blinded_block: BeaconBlock<E, BlindedPayload<E>>`), but the SSV spec permits full-block proposals (Go reference at `ssv-spec/ssv/proposer.go:91-105` marshals the BN's return directly — no stripping — and `ssv-spec/ssv/spectest/all_tests.go:142-145` has explicit mixed-format tests). So this arm fires when a non-stripping operator (non-MEV Go SSV, etc.) is the leader — i.e., we weren't. That's also why @claude [`proposer_matched = true` suggestion](https://github.com/sigp/anchor/pull/967#issuecomment-4289699153) doesn't fit.

anchor/validator_store/src/lib.rs
```diff
@@ -2307,12 +2310,48 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> ValidatorStore
...
+            UnsignedBlock::Full(FullBlockContents::Block(ref block)) => block.slot(),
+            UnsignedBlock::Blinded(ref block) => block.slot(),
+        };
+
+        // Root span
+        let span = info_span!(
+            "sign_block",
+            block_type,
+            cluster_size = tracing::field::Empty,
+            current_slot = current_slot.as_u64(),
```
shane-moore: Up to you on this one: consider renaming the span field `current_slot` → `clock_slot` (keep the parameter name `current_slot` on the function signature — that matches Lighthouse's `ValidatorStore` trait).
>
> On the span, `slot` and `current_slot` look like near-duplicates to anyone reading a trace. Naming the clock variant explicitly makes the two signals unambiguous in dashboards:
>
> - `block_slot` — the block's target slot (fixed when the BN produced it)
> - `clock_slot` — the `SlotClock::now()` view at `sign_block` entry

anchor/validator_store/src/lib.rs
```diff
@@ -2346,39 +2395,67 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> ValidatorStore
...
+                        );
                         // Someone else's proposal won, return blinded
                         Ok(SignedBlock::Blinded(signed_blinded_block))
                     }
                 }
-                SignedBlock::Full(signed_block) => Ok(SignedBlock::Full(signed_block)),
+                SignedBlock::Full(signed_block) => {
+                    info!(
+                        checkpoint = BlockSigningCheckpoints::PublishConsensus.as_str(),
+                        "Published full block directly"
```
shane-moore: Nit: The block hasn't been published yet at any of these log points -> `sign_block` returns a `SignedBlock` and the caller does the network publish after return. Normalize to `"Publishing full block directly"` for consistency and accuracy.

anchor/validator_store/src/lib.rs
jnhsigmap: ah ok thanks, I was still unsure what this meant. from what I read this actually should never occur? Does it mean that our block was identical to another proposer's?

shane-moore: So when `sign_block` is called, one of our managed validators is the block proposer for the slot, but that's separate from who's the QBFT round leader inside the ssv cluster. A go-ssv node could be the qbft round leader, and they could propose a full block.  We accept it, co-sign it, and land in this branch arm.  And in this scenario, we weren't the qbft leader, so `proposal_matched = false`

anchor/validator_store/src/lib.rs
jnhsigmap: done :heavy_check_mark:

anchor/validator_store/src/lib.rs
jnhsigmap: done :heavy_check_mark:

anchor/validator_store/src/lib.rs
jnhsigmap: I have opted for `block_slot` and `clock_slot` for clarity. good suggestion :heavy_check_mark:

anchor/validator_store/src/lib.rs
jnhsigmap: changed to `proposer_sign_block` :heavy_check_mark:

anchor/validator_store/src/lib.rs
jnhsigmap: done :heavy_check_mark:

anchor/validator_store/src/instrumentation.rs
jnhsigmap: Interesting yeah I did consider creating constants initially but am aware of ensuring that the groupings are clear and the semantics are retained. I have created a module called `checkpoints` that has the best of both worlds which is also relatively open-closed. respectfully, I disagree that this extends to `SignBlockOutcome`, as I deliberately chose that to retain exhaustiveness of variants and to enforce it given the taxonomy agreed upon. The `enum` still being used as an `enum` when the `outcome` is conveyed. To address the verbosity, I have included `derive_more` from the workspace level to reduce lines of code as `to_str` was over-complicated. The runtime impact should be negligible, however I've now added `derive_more` to `validator_store`.

jnhsigmap:
@claude please provide another review

anchor/validator_store/src/lib.rs
jnhsigmap: Yep this makes sense. I wasn't completely aware that the two clients have implemented this differently. I think I got a different interpretation from claude when I was initially reviewing that

anchor/validator_store/src/lib.rs
jnhsigmap: Makes sense, I've tried to get closer to your example here ^.

I did mention the added complexity in determining the failure reason which could be beyond the scope of a shallow set of checkpoints. Was this something you wanted in this change? In my head, it would look like a large manual retrieval of error variants given the type of error obtained. Otherwise what ssv had done, which appears to be the actual error text (cardinality?).

jnhsigmap:
> **Claude encountered an error** —— [View job](https://github.com/sigp/anchor/actions/runs/24977819895)
> 
> ### Code Review: PR #967 (Re-review)
> * [x]  Gather context and read changed files
> * [x]  Review `instrumentation.rs`
> * [x]  Review `lib.rs` changes
> * [x]  Review `Cargo.toml` changes
> * [x]  Post review findings
> 
> Previous issues **#1** (outcome recorded outside span) and **#2** (missing `proposal_matched` in full-block branch) are fixed. Span renamed to `proposer_sign_block`, field renamed to `clock_slot`, log messages corrected. Good progress.
> 
> ### Remaining Issues
> **1. `DUTY_FAILED` checkpoint doesn't include the error — Diego's feedback not addressed** `validator_store/src/lib.rs:2451`
> 
> Diego's [review comment](https://github.com/sigp/anchor/pull/967#discussion_r2289455437) specifically requested that the failure checkpoint include a `reason` field. Currently:
> 
> ```rust
> Err(_) => info!(checkpoint = checkpoints::DUTY_FAILED, outcome = &outcome),
> ```
> 
> The error is discarded. While `run_and_update_metrics` logs the error separately (line 1862), that log event isn't correlated with the checkpoint event in a trace. Adding `error = ?e` to the checkpoint would make the failure self-contained and queryable:
> 
> ```rust
> Err(e) => info!(
>     checkpoint = checkpoints::DUTY_FAILED,
>     outcome = &outcome,
>     error = ?e,
> ),
> ```
> 
> This is a low-effort change that closes the observability gap Diego flagged.
> 
> **2. Unnecessary `span.clone()`** `validator_store/src/lib.rs:2455`
> 
> ```rust
> .instrument(span.clone());
> ```
> 
> `span` is not used after this line, so the `.clone()` allocates a reference count bump for no reason. Should be:
> 
> ```rust
> .instrument(span);
> ```
> 
> [Fix this →](https://claude.ai/code?q=In%20anchor%2Fvalidator_store%2Fsrc%2Flib.rs%20at%20line%202455%2C%20change%20%60.instrument(span.clone())%60%20to%20%60.instrument(span)%60%20since%20span%20is%20not%20used%20after%20this%20point.%20Branch%3A%20feat%2F924%2Fproposer-duty-span&repo=sigp/anchor)
> 
> **3. `SameData` maps to `"error"` outcome — potentially misleading** `validator_store/src/instrumentation.rs:17-23`
> 
> `SignBlockOutcome::from_result` catches `Timeout` specifically but sends all other errors (including `Error::SameData`) through `Self::Failed` → `"error"`. `SameData` means the validator already signed this data — it's a safety guard, not a failure. Dashboards querying `outcome = "error"` would include these benign events.
> 
> This is a design decision rather than a bug — the current approach is defensible if `SameData` is rare for proposer duties, but worth noting. If you want to distinguish it, a `SameData` variant or folding it into `Success` would be options.
> 
> ### Summary
> The core instrumentation structure is sound: the outcome is now recorded inside the instrumented future, `proposal_matched` is set in all branches, naming follows the taxonomy. The main open item is **#1** — including the error in the `DUTY_FAILED` checkpoint, which was explicitly requested in review and is a one-line addition.
> 
> [View job](https://github.com/sigp/anchor/actions/runs/24977819895)

Point 1 - still being discussed.
Point 2 - addressed :heavy_check_mark:
Point 3 - at this stage, I am following the initial taxonomy outlined. so will leave this be for now.

anchor/validator_store/src/lib.rs
```diff
@@ -2307,71 +2310,149 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> ValidatorStore
...
-                SignedBlock::Full(signed_block) => Ok(SignedBlock::Full(signed_block)),
             }
-        };
+            .await;
+
+            let outcome = tracing::field::display(SignBlockOutcome::from_result(&result));
+            tracing::Span::current().record("outcome", &outcome);
+            match &result {
+                Ok(_) => info!(checkpoint = checkpoints::DUTY_COMPLETED, outcome = &outcome),
+                Err(_) => info!(checkpoint = checkpoints::DUTY_FAILED, outcome = &outcome),
```
diegomrsantos: This terminal failure checkpoint still drops the error. #924 asks for failures to include an explicit `outcome` and `reason`, and without that the trace can only show `timeout` or `error`, not why the duty failed.

Could this bind the error and record a fixed reason field here? Raw error text would probably create too many distinct values, but a helper like this would keep the checkpoint queryable:

```rust
Err(err) => info!(
    checkpoint = checkpoints::DUTY_FAILED,
    outcome = &outcome,
    reason = failure_reason(err),
),
```

anchor/validator_store/src/lib.rs
```diff
@@ -2307,71 +2310,149 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> ValidatorStore
...
+                        .await
+                    }
+                }?;
+
+                info!(
+                    checkpoint = checkpoints::BLOCK_SIGNED,
+                    "Block threshold signature completed"
+                );
+
+                match signed_block {
```
diegomrsantos: This branch makes `sign_block` much harder to read. The function already has the consensus handoff, signing, span updates, checkpoints, and terminal outcome handling in one place, and this nested publish path adds another match plus two levels of branching.

Could we pull the publish path decision into a small helper that returns the final `SignedBlock` plus the `proposal_matched` value? Then this function can record the span field and emit the `PUBLISH_BLOCK` checkpoint in a flatter way, while the helper owns the full block reconstruction and blinded fallback logic.

anchor/validator_store/src/lib.rs
diegomrsantos: Yeah, I think we can keep this much smaller than a full mapping of every possible error. I do not mean raw error text either, because that would create too many different `reason` values.

What I had in mind is a small fixed helper that groups failures into broad categories. That gives the `duty_failed` checkpoint enough information to be useful in traces, without making the PR much bigger. For example:

```rust
pub fn failure_reason(error: &Error) -> &'static str {
    match error {
        Error::SpecificError(SpecificError::Timeout) => "timeout",
        Error::SpecificError(SpecificError::NotSynced) => "not_synced",
        Error::SpecificError(SpecificError::MissingIndex) => "missing_index",
        Error::SpecificError(SpecificError::SignatureCollectionFailed(_)) => {
            "signature_collection_failed"
        }
        Error::SpecificError(SpecificError::QbftError(_)) => "qbft_error",
        Error::SpecificError(SpecificError::InvalidQbftData(_)) => "invalid_qbft_data",
        Error::SpecificError(SpecificError::NoDataAgreed) => "no_data_agreed",
        Error::Slashable(_) => "slashable",
        Error::SameData => "same_data",
        Error::UnknownPubkey(_) => "unknown_pubkey",
        Error::SpecificError(_) => "other_specific_error",
        _ => "other_error",
    }
}
```

Then the terminal checkpoint can bind the error and record that fixed reason:

```rust
Err(err) => info!(
    checkpoint = checkpoints::DUTY_FAILED,
    outcome = &outcome,
    reason = failure_reason(err),
),
```

That should satisfy the issue requirement without trying to make the reason perfect in this PR.

anchor/validator_store/src/instrumentation.rs
diegomrsantos: I get the point about keeping the allowed outcomes explicit, but I do not think the enum is buying us much exhaustiveness in this PR. The only real use is:

```rust
let outcome = tracing::field::display(SignBlockOutcome::from_result(&result));
```

So we look at the `Result`, convert it to `SignBlockOutcome`, and immediately turn it into display text for tracing. Nothing later matches on `SignBlockOutcome`, no caller receives it, and no logic depends on the variants. A helper that returns `&'static str` would give basically the same safety for this use.

I am fine keeping the enum if you prefer it, but I would not add `derive_more` just to avoid a tiny manual method here. Either a small `as_str()` method or an `outcome_from_result(&result) -> &'static str` helper seems enough. The extra dependency does not feel justified by how little the enum is doing.

jnhsigmap: Perhaps if we removed `Error::SpecificError(SpecificError::Timeout)` from `outcome` it would appear less arbitrary? I note that it is the most common and expected error variant of `sign_block`, but I actually think it would be simpler to report `success` or `failed` in the `outcome` attribute and add a failure reason to the `sign_block` span. A binary outcome makes more sense and makes querying this detail easier later.

The `timeout` variant is always caught as an error variant and reported in `failure_reason`. The check at the end of the `future` block is the same check of `result` for both `outcome` and `checkpoint`.

That collapses to a simple function (that should now never change) depending on `result` that is agnostic to the type of error and removes the visual and logical aid of an `enum`
```rust
pub fn from_result<T>(result: &Result<T, Error>) -> &'static str {
    match result {
        Ok(_) => "success",
        Err(_) => "failed",
    }
}
```

The resulting branch at the conclusion of the duty would then be:
```rust
let outcome = instrumentation::from_result(&result);
tracing::Span::current().record("outcome", outcome);
match &result {
    Ok(_) => info!(checkpoint = checkpoints::DUTY_COMPLETED, outcome = &outcome),
    Err(err) => {
        let failure_reason = instrumentation::failure_reason(err);
        info!(
            checkpoint = checkpoints::DUTY_FAILED,
            outcome = &outcome,
            failure_reason = &failure_reason
        );
        tracing::Span::current().record("failure_reason", failure_reason);
    }
}
```

What are your thoughts?

anchor/validator_store/src/lib.rs
jnhsigmap: nice, thanks this worked well :heavy_check_mark:

anchor/validator_store/src/lib.rs
jnhsigmap: addressed [here](https://github.com/sigp/anchor/pull/967#discussion_r3147800291)

anchor/validator_store/src/lib.rs
jnhsigmap: yeah I agree it's quite hard to follow. Do you think that this belongs in a separate PR though? It's a refactor and the current solution to this issue is still correct with or without this change.

I think handling the `proofs_and_blobs` object from the first `match block {...` requires some thought as it's only used on one of the branches in `match signed_block {...`. I can create an issue and open a separate PR for this.

diegomrsantos:
I think this is the right general base for later timing work: `sign_block` gets one `proposer_sign_block` span, and the async body is instrumented under it, so future QBFT, signature collection, and block assembly timing can share the same duty context.

The main thing I would change here is the log level. These checkpoints are useful markers for development and performance investigation, but they are not operator action items on a successful duty. Operators mostly need to know whether duties are succeeding, timing out, or failing. Emitting several `info!` events for every successful proposer duty feels too noisy, especially for larger operators.

Could we make the successful checkpoints `trace!` and keep the terminal failure checkpoint visible, probably as `warn!` with the fixed `failure_reason`?

One small schema issue too: if we record `failure_reason` on the root span, the span should declare `failure_reason = tracing::field::Empty`; otherwise that final `record("failure_reason", ...)` is a no op.

diegomrsantos:
I pushed a small cleanup that tries to keep this PR focused while addressing the readability and log level concerns.

The root `proposer_sign_block` span is still an `info_span!`, because creating the span does not emit a normal info log line by itself, and we want the parent duty context to exist under the default info filter.

The successful checkpoints are now `trace!`, since they are mostly timing markers for development and performance investigation rather than operator action items. The terminal failure checkpoint is now `warn!` and still carries the fixed `failure_reason`.

I also pulled the publish decision into `select_publish_block`, so `sign_block` keeps the main duty flow visible and emits one structured publish checkpoint with `publish_path`, instead of carrying the nested publish branches inline.

@jnhsigmap @shane-moore does this direction look right to you?

diegomrsantos:
@claude review this PR

diegomrsantos (approved):

shane-moore (approved):
great work y'all!

Direction looks right to me. The trace! checkpoints + info_span! root is the right split: fmt layers stay quiet at default operator filters, but once a tracing-opentelemetry layer lands the checkpoints attach as SpanEvents on the info-level root span and become queryable as phase deltas. select_publish_block plus a single structured publish_path field is much easier to follow than four branch-specific log lines.

PR #968, opened 2026-04-21
test: committee member test

### LINKED ISSUE #958: test: add CommitteeMemberTest spec coverage

shane-moore:

- Closes #958: adds the `CommitteeMemberTest` spec test from ssv-spec, validating Anchor's quorum and signed-message semantics against the Go reference.
- Worth doing now because the helper it depends on (`ssv_types::quorum_size`) was just aligned with Go's `2f + 1` formula in #966, so a spec test of the surrounding contract naturally fits.
- Anchor has no `CommitteeMember` type, but both contracts the Go test exercises (`SignedSSVMessage.Validate` and `CommitteeMember.HasQuorum`) map directly to production helpers (`SignedSSVMessage::new` and `ssv_types::quorum_size`).
- Relevant links: issue #958; ssv-spec `types/spectest/tests/committeemember/test.go`; depended on #966 (now merged).

## Change Overview (Required)

- One new spec-test type registered with the dispatcher: deserializes the Go fixture, runs it through the production `SignedSSVMessage` constructor for error-code parity, then evaluates quorum via the production helper and full-committee via deduplicated signer count.
- Read order: start with the new test module to see fixture shape and assertions; the diff in `signed_ssv_msg.rs` is a pure DRY refactor that promotes the existing error-code mapper into a shared utility now used by both spec tests.
- What intentionally did not change: production code (test-harness only); the crate's `Result<_, i64>` error-code convention; the older spec test's signature-padding policy (committee-member fixtures ship fixed 256-byte sigs and use strict size checks instead).

- Blast radius: spec-tests crate only; no on-wire, runtime, config, or storage impact.
- Strict signature size (`<[u8; 256]>::try_from`) rejects non-256-byte sigs with `EMPTY_SIGNATURE`. All four current fixtures ship exactly-256-byte RSA sigs (verified); a future fixture with shorter sigs would need to adopt the existing padding helper.
- The shared error-code mapper is now used by two call sites, so any drift between Anchor's error variants and Go's error codes surfaces in both spec tests at once.

## Validation (Required)

- `cargo test -p spec_tests --lib`: dispatcher passes; four committee-member fixtures registered and exercised (`has_quorum`, `has_quorum_3f1`, `no_quorum_duplicate`, `quorum_with_duplicate`).
- Fixture properties verified: signature sizes all `[256]`; committee sizes all `N=4`.

- Test-harness only; revert the commit to remove. No config, data, or operational impact.

- Originally blocked on #966; now merged. Branch is rebased onto current `unstable` and reduced to a single commit.

- The Go fixture's `FaultyNodes` field is deserialized and asserted equal to `ssv_types::get_f(committee.len())` before the quorum check. This catches drift between Anchor's `get_f` and Go's `ComputeF` with an explicit error message, rather than letting it surface only through a `has_quorum` mismatch.

### DIFF
--- a/anchor/spec_tests/src/runner/types_dispatcher.rs
+++ b/anchor/spec_tests/src/runner/types_dispatcher.rs
@@ -45,6 +45,9 @@ fn dispatch_fixture_by_prefix(prefix: &str, path: &Path, contents: &str) -> Disp
         }
 
         // Validation tests
+        "committeemember.CommitteeMemberTest" => {
+            DispatchOutcome::Executed(run_test::<types::CommitteeMemberTest>(path, contents))
+        }
         "signedssvmsg.SignedSSVMessageTest" => {
             DispatchOutcome::Executed(run_test::<types::SignedSSVMessageTest>(path, contents))
         }
--- a/anchor/spec_tests/src/types/committee_member.rs
+++ b/anchor/spec_tests/src/types/committee_member.rs
@@ -50,22 +50,13 @@ pub struct CommitteeMemberTest {
 
 impl SpecTest for CommitteeMemberTest {
     fn run(&self) -> Result<(), String> {
-        // Arrange + Act: build the `SignedSSVMessage` through the production constructor,
-        // capturing either success or a mapped Go error code.
         let actual_code = self
             .build_signed_message()
             .err()
             .unwrap_or(error_codes::NO_ERROR);
+        error_codes::assert_error_code(self.expected_error_code, actual_code)?;
 
-        // Assert: validation error code matches the fixture's expectation.
-        if actual_code != self.expected_error_code {
-            return Err(format!(
-                "Expected error code {}, got {actual_code}",
-                self.expected_error_code,
-            ));
-        }
-
-        // Assert: quorum thresholds use deduplicated signers (matches Go's
+        // Quorum thresholds use deduplicated signers (matches Go's
         // `GetUniqueMessageSignersCount` behavior).
         let unique_signers: HashSet<u64> = self.message.operator_ids.iter().copied().collect();
         let unique_count = unique_signers.len();
--- a/anchor/spec_tests/src/types/committee_member.rs
+++ b/anchor/spec_tests/src/types/committee_member.rs
@@ -1,0 +1,131 @@
+use std::collections::HashSet;
+
+use serde::{Deserialize, de::IgnoredAny};
+use ssv_types::{OperatorId, message::SignedSSVMessage, quorum_size};
+
+use crate::{
+    SpecTest,
+    utils::{
+        deserializers::{deserialize_base64_list, deserialize_base64_or_null},
+        dtos::RawSSVMessage,
+        error_codes,
+    },
+};
+
+/// Fixture container for the signed message in `CommitteeMemberTest`.
+#[derive(Debug, Deserialize)]
+#[serde(rename_all = "PascalCase")]
+struct FixtureMessage {
+    #[serde(deserialize_with = "deserialize_base64_list")]
+    signatures: Vec<Vec<u8>>,
+    #[serde(rename = "OperatorIDs")]
+    operator_ids: Vec<u64>,
+    #[serde(rename = "SSVMessage")]
+    ssv_message: RawSSVMessage,
+    #[serde(deserialize_with = "deserialize_base64_or_null", default)]
+    full_data: Option<Vec<u8>>,
+}
+
+/// Only the length of `Committee` is exercised; quorum is derived from committee size
+/// via `ssv_types::quorum_size` (matches production QBFT behavior, see struct docs).
+#[derive(Debug, Deserialize)]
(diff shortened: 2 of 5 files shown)


shane-moore:
@claude, please take a look

shane-moore:
@claude, thanks for the careful review. Quick response on each point.

### Correctness: quorum-on-validation-failure

I think this one's based on a Go control-flow misread. The test at `test.go:42-55`:

```go
func (test *CommitteeMemberTest) Run(t *testing.T) {
    err := test.Message.Validate()
    tests.AssertErrorCode(t, test.ExpectedErrorCode, err)

    numSigners := test.GetUniqueMessageSignersCount()

    require.Equal(t, test.ExpectedHasQuorum, test.CommitteeMember.HasQuorum(numSigners))
    require.Equal(t, test.ExpectedFullCommittee, (len(test.CommitteeMember.Committee) == numSigners))
}
```

There's no `if err != nil { return }`. `tests.AssertErrorCode` (`tests/error.go:13`) either passes silently (test continues) or calls `require.NoError`/`require.Failf`, which invoke `t.FailNow()` to halt the test. So when validation fails AND the error code matches expected, Go proceeds unconditionally to the quorum/full-committee checks.

Empirical confirmation: `no_quorum_duplicate.go` sets `ExpectedErrorCode = NonUniqueSignerErrorCode` together with `ExpectedHasQuorum = false`, `ExpectedFullCommittee = false`. Go runs all three assertions on this validation-failing fixture; Anchor matches.

Adding the `if actual_code == NO_ERROR` gate would actually make Anchor less faithful to Go (skipping quorum where Go runs it), so leaving the control flow as-is.

### Minor observation 1 (`Vec<IgnoredAny>`)

Agreed that overhead is negligible in test code. Worth flagging that the most natural alternative (a custom `Vec::<IgnoredAny>::deserialize(d).map(|v| v.len())` deserializer) still walks the same array tokens, so it wouldn't recover the parse cost. Keeping `Vec<IgnoredAny>` as the canonical serde idiom for "shape only."

### Minor observation 2 (strict 256-byte sigs)

Confirmed: all four committee fixtures ship exactly-256-byte RSA sigs. If a future fixture ships shorter, we'll mirror `prepare_signatures` from `signed_ssv_msg.rs`. Already documented in the PR's risks section.

Thanks again for the review.

jnhsigmap (dismissed):
It matches SSV's implementation and tests the same as their spec (without the structural/reflection check at the end) 👍

diegomrsantos:
I think we should deserialize the fixture `FaultyNodes` field and assert that it matches `ssv_types::get_f(committee.len())`.

Production go-ssv derives faulty-node tolerance from committee size, for example through `ComputeF(len(share.Committee))`, and derives quorum from committee size as well. So using `committee.len()` in Anchor is aligned with production behavior.

The reason I think we should still read `FaultyNodes` here is completeness: the spec-test fixture includes this field, and checking it makes the harness account for all quorum-related fixture data instead of leaving one serialized value unused.

I also opened https://github.com/ssvlabs/ssv-spec/issues/624 upstream to ask whether the fixture helper should use the same production helper directly, since the current fixture expression is correct but indirect.

shane-moore:
Done in 577f172 — `faulty_nodes` is now deserialized and asserted against `ssv_types::get_f(committee.len())` before the quorum check. Failure produces an explicit `FaultyNodes mismatch: fixture=X, get_f(N)=Y` message, which makes the drift attribution cleaner than letting it propagate through `has_quorum` only.

Updated the PR description's "Additional Info" section to reflect the new check (no longer "intentionally not deserialized").

diegomrsantos (approved):
Thanks!

PR #973, opened 2026-04-22
fix(eth): replay duplicate operator keys

### LINKED ISSUE #972: eth: historical sync rejects hex-encoded OperatorAdded keys and later aborts on OperatorRemoved
## Description

Historical sync rejects a valid `OperatorAdded` payload variant on Hoodi, skips inserting the operator, and later aborts on the matching `OperatorRemoved`.

The specific sequence is:

- block `48960`, tx `0xcc9805bba190e0131869fd90a420e034954eb46f997732415b30e986b494a2fb` emits `OperatorAdded(15, owner=0x25916caa0db559bc7f21850cfe678dc9f273a8d7, ...)`
- Anchor logs `Invalid event: Failed to construct operator: Unable to decode base64 PEM data: Invalid symbol 0, offset 0`
- block `48998`, tx `0x6c2a63f8a827f12c4043d5587edf57f49b42f850690594f363799f9a55b98207` emits `OperatorRemoved(15)`
- sync aborts with `Failed to remove operator: NotFound("Operator with id 15 not in database")`

The root cause is that Anchor still has a brittle operator-key decoding path:

- it only unwraps `OperatorAdded.publicKey` when the raw field length is exactly `704`
- otherwise it passes the bytes straight into the base64 PEM parser

That works for older wrapped-base64 payloads, but the Hoodi `OperatorAdded(15)` event is ABI-wrapped and its inner string is ASCII hex, not base64. After proper ABI unpacking, that hex decodes to the same RSA PEM as operator `14`.

This also exposed a broader parity issue: the SSV contract enforces operator uniqueness on raw `bytes` (`keccak256(publicKey)`), not on a canonical RSA key. On Hoodi, operator ids `14` and `15` were both registered by the same owner with different encodings of the same underlying RSA key, and their lifetimes overlapped until operator `14` was removed at block `48996`.

## Version

- `unstable` at `3dc2508bd` contains the bug in the operator path
- operator-specific replay evidence was captured on `1e7ca8b8f` because `unstable` currently halts earlier on `#969`
- `rustc 1.94.0 (4a4ef493e 2026-03-02)`

## Present Behaviour

- Anchor decodes `OperatorAdded.publicKey` using a length-based special case in `anchor/eth/src/event_processor.rs`
- `OperatorAdded(15)` is classified as malformed and skipped because the parser expects base64 PEM
- Anchor still advances `max_operator_id_seen` to `15`
- `OperatorRemoved(15)` later becomes a `Database` error and aborts historical sync

Before `#351`, this same sequence would not have halted sync. Anchor logged any per-log failure as malformed and continued processing the block. In practice that means it would have skipped `OperatorAdded(15)`, then skipped `OperatorRemoved(15)`, and continued syncing.

That was not correct behavior. It only masked the bug by committing progress while silently dropping contract-permitted operator state. `#351` exposed the inconsistency; it did not create it.

For upstream parity:

- the contract accepts arbitrary `bytes` for `registerOperator(bytes publicKey, ...)` and emits them unchanged
- `ssv-go` always ABI-unpacks `OperatorAdded.publicKey` with `method() returns (bytes)` before storing it
- `ssv-go` stores the unpacked string as-is; it does not normalize base64 vs hex into a canonical PEM before checking uniqueness

## Expected Behaviour

Anchor should not wedge historical sync on this sequence.

At minimum, it should mirror the upstream parse path for `OperatorAdded.publicKey` by ABI-unpacking the field instead of relying on the `704`-byte special case.

After that, Anchor needs an explicit policy for operator identity:

- if Anchor follows the contract literally, raw encoded bytes are the identity, so the current `UNIQUE(public_key)` / canonical PEM assumption is too strict
- if Anchor follows canonical RSA-key identity, then the second registration on Hoodi is malformed and both the add and the later remove need to be classified consistently at the event boundary instead of turning the remove into a batch-fatal database error

## Steps to resolve

1. Replace the `704`-byte special case with unconditional ABI unpacking for `OperatorAdded.publicKey`, matching `ssv-go`.
2. Decide whether Anchor models operator identity by raw on-chain bytes or by canonical decoded RSA key.
3. Make the add/remove classification consistent with that choice so a skipped `OperatorAdded` cannot later wedge sync via `OperatorRemoved`.
4. Keep `AbortBatch` for real internal failures; do not go back to the old “log and continue” masking behavior.

diegomrsantos:

- Closes #972.
- Historical replay can encounter `OperatorAdded` events whose nested `publicKey` bytes either resolve to a canonical RSA key Anchor already knows under another operator id, or use operator-key payloads Anchor cannot interpret at all. Before this PR, either case could leave Anchor without committed operator state for that id and later abort replay on `OperatorRemoved`.
- This is worth fixing now because `#351` correctly made replay inconsistencies batch-fatal instead of silently continuing with partial state. The underlying operator replay bug already existed; it is now visible as a hard sync stop.
- Relevant links: #972.

## Change Overview (Required)

- `eth` now ABI-unpacks operator `publicKey`, accepts the two on-chain encodings already observed during replay, normalizes them to a canonical RSA key, and classifies canonical duplicates before any DB writes.
- `database` now persists a skipped-operator marker keyed by operator id so a later `OperatorRemoved` for a previously skipped add is treated as an expected skip instead of `NotFound`.
- That skipped-marker policy is intentionally broader than canonical duplicates alone: it also covers operator-key payloads Anchor cannot interpret, so replay can consume the add/remove lifecycle symmetrically instead of wedging on the later remove.
- Start with `event_processor.rs` for the replay behavior, then the small DB helper and migration changes, then the regression tests.
- This does not change validator handling, upstream contract semantics, or Anchor's operator identity model. Anchor still treats the canonical decoded RSA key as the operator identity, not the raw submitted bytes.

- Main trade-off: Anchor remains intentionally stricter than the contract and raw-bytes client model. If upstream decides that differently encoded raw values should be distinct operators, or that uninterpretable payloads should be preserved differently, Anchor will still treat canonical duplicates and unparseable operator keys as malformed.
- Blast radius is limited to operator add/remove replay. Normal operator registration and unrelated event paths are unchanged.
- Mitigations: duplicate classification happens before insert side effects, skipped ids are durably recorded and consumed on remove, and the regression tests cover both duplicate-key and malformed-add lifecycle shapes.

## Validation (Required)

- `cargo fmt --all --check`
- `cargo clippy -p eth -p database --tests -- -D warnings`
- `cargo test -p database --features test-utils test_skipped_operator_add_round_trip -- --nocapture`
- `cargo test -p eth test_wrapped_hex_duplicate_operator_add_is_skipped_and_later_remove_does_not_abort -- --nocapture`
- `cargo test -p eth test_malformed_operator_add_is_skipped_and_later_remove_does_not_abort -- --nocapture`
- `cargo test -p eth event_processor::tests -- --nocapture`
- `cargo test -p eth --tests`
- Manual Hoodi replay on the updated PR branch completed historical sync, entered live sync, and continued processing new blocks.

- Safe code revert.
- Data impact is limited to the unreleased `V2` migration addition for skipped replay markers. If this branch is tested locally and then reverted before release, recreate local DBs created with this `V2` checksum.

- N/A

- This PR only makes Anchor replay resilient to the historical operator encodings and malformed operator-key payloads already observed in replay.
- Any broader upstream identity or normalization change should be handled separately in the SSV contract and client repos.

### DIFF
--- a/anchor/database/src/operator_operations.rs
+++ b/anchor/database/src/operator_operations.rs
@@ -1,5 +1,6 @@
 use base64::prelude::*;
-use rusqlite::{Transaction, params};
+use openssl::{pkey::Public, rsa::Rsa};
+use rusqlite::{OptionalExtension, Transaction, params};
 use ssv_types::{Operator, OperatorId};
 use tracing::trace;
 
@@ -49,6 +50,58 @@ impl NetworkDatabase {
             ])?;
         state_updates.insert_operator(operator.to_owned(), is_own_operator);
         Ok(())
+    }
+
+    /// Record that an `OperatorAdded` log was intentionally skipped after consuming its
+    /// monotonically-increasing operator id.
+    pub fn insert_skipped_operator_add_tx(
+        &self,
+        id: OperatorId,
+        reason: &str,
+        tx: &Transaction<'_>,
+    ) -> Result<(), DatabaseError> {
+        tx.prepare_cached(sql_operations::INSERT_SKIPPED_OPERATOR_ADD)?
+            .execute(params![id, reason])?;
+        Ok(())
+    }
+
+    /// Delete a previously recorded skipped `OperatorAdded` marker.
+    pub fn delete_skipped_operator_add_tx(
+        &self,
+        id: OperatorId,
+        tx: &Transaction<'_>,
+    ) -> Result<bool, DatabaseError> {
+        let rows = tx
+            .prepare_cached(sql_operations::DELETE_SKIPPED_OPERATOR_ADD)?
+            .execute(params![id])?;
+        Ok(rows > 0)
+    }
+
+    /// Check if an `OperatorAdded` log for this id was previously skipped.
+    pub fn was_operator_add_skipped_tx(
+        &self,
+        id: OperatorId,
+        tx: &Transaction<'_>,
+    ) -> Result<bool, DatabaseError> {
+        tx.prepare_cached(sql_operations::GET_SKIPPED_OPERATOR_ADD_REASON)?
+            .query_row(params![id], |row| row.get::<_, String>(0))
+            .optional()
+            .map(|reason| reason.is_some())
+            .map_err(DatabaseError::from)
+    }
+
+    /// Resolve any operator id that already owns this canonical public key, including soft-deleted
+    /// rows that still occupy the schema-level uniqueness slot.
+    pub fn get_any_operator_id_by_public_key_tx(
+        &self,
+        public_key: &Rsa<Public>,
+        tx: &Transaction<'_>,
+    ) -> Result<Option<OperatorId>, DatabaseError> {
+        let encoded = BASE64_STANDARD.encode(public_key.public_key_to_pem()?);
+        tx.prepare_cached(sql_operations::GET_ANY_OPERATOR_ID)?
+            .query_row(params![encoded], |row| row.get(0))
+            .optional()
+            .map_err(DatabaseError::from)
     }
 
     /// Delete an operator in the active transaction and queue the matching state update.
--- a/anchor/database/src/operator_operations.rs
+++ b/anchor/database/src/operator_operations.rs
@@ -78,6 +78,7 @@ impl NetworkDatabase {
     }
 
     /// Check if an `OperatorAdded` log for this id was previously skipped.
+    #[cfg(any(test, feature = "test-utils"))]
     pub fn was_operator_add_skipped_tx(
         &self,
         id: OperatorId,
--- a/anchor/database/src/tests/operator_tests.rs
+++ b/anchor/database/src/tests/operator_tests.rs
@@ -158,4 +158,43 @@ mod operator_database_tests {
                 .is_err()
         )
     }
+
+    #[test]
+    fn test_skipped_operator_add_round_trip() {
+        let fixture = InMemoryTestFixture::new_empty();
+        let mut conn = fixture.db.connection().unwrap();
+        let tx = conn.transaction().unwrap();
+
+        assert!(
+            !fixture
+                .db
+                .was_operator_add_skipped_tx(OperatorId(7), &tx)
+                .expect("Failed to query skipped operator marker")
+        );
+
+        fixture
+            .db
+            .insert_skipped_operator_add_tx(OperatorId(7), "duplicate key", &tx)
+            .expect("Failed to insert skipped operator marker");
+
+        assert!(
+            fixture
+                .db
+                .was_operator_add_skipped_tx(OperatorId(7), &tx)
+                .expect("Failed to query skipped operator marker")
+        );
+
+        assert!(
+            fixture
+                .db
+                .delete_skipped_operator_add_tx(OperatorId(7), &tx)
+                .expect("Failed to delete skipped operator marker")
+        );
+        assert!(
+            !fixture
+                .db
+                .was_operator_add_skipped_tx(OperatorId(7), &tx)
+                .expect("Failed to query skipped operator marker")
+        );
+    }
 }
--- a/anchor/database/src/tests/operator_tests.rs
+++ b/anchor/database/src/tests/operator_tests.rs
@@ -161,10 +161,12 @@ mod operator_database_tests {
 
     #[test]
     fn test_skipped_operator_add_round_trip() {
+        // Arrange: start from an empty in-memory database and an operator id with no marker.
         let fixture = InMemoryTestFixture::new_empty();
         let mut conn = fixture.db.connection().unwrap();
         let tx = conn.transaction().unwrap();
 
+        // Act/Assert: the marker should be absent before insertion.
         assert!(
             !fixture
                 .db
@@ -172,11 +174,13 @@ mod operator_database_tests {
                 .expect("Failed to query skipped operator marker")
         );
 
+        // Act: insert a skipped-operator marker.
         fixture
             .db
             .insert_skipped_operator_add_tx(OperatorId(7), "duplicate key", &tx)
             .expect("Failed to insert skipped operator marker");
 
+        // Assert: the inserted marker is visible through the typed helper.
         assert!(
             fixture
                 .db
@@ -184,12 +188,15 @@ mod operator_database_tests {
                 .expect("Failed to query skipped operator marker")
         );
 
+        // Act: delete the skipped-operator marker.
         assert!(
             fixture
                 .db
                 .delete_skipped_operator_add_tx(OperatorId(7), &tx)
                 .expect("Failed to delete skipped operator marker")
         );
+
+        // Assert: the marker is gone again after deletion.
         assert!(
             !fixture
                 .db
--- a/anchor/eth/src/event_processor.rs
+++ b/anchor/eth/src/event_processor.rs
@@ -1,6 +1,7 @@
 use std::sync::Arc;
 
 use alloy::{primitives::Address, rpc::types::Log, sol_types::SolEvent};
+use base64::prelude::*;
 use bls::PublicKeyBytes;
 use database::{NetworkDatabase, PendingStateUpdates, SlashingProtection};
 use indexmap::IndexSet;
@@ -380,24 +381,29 @@ impl EventProcessor {
             .set_max_operator_id_seen_tx(operatorId, tx, state_updates)
             .map_err(|e| ExecutionError::Database(e.to_string()))?;
 
-        let data = publicKey.as_ref();
-
-        // If the data is 704 bytes, remove the ssv encoding. Else, just parse the key
-        let data = if data.len() == 704 {
-            let mut data = &data[64..];
-            // while there is a 0 at the end of the data, remove it
-            while let [rest @ .., 0] = data {
-                data = rest;
+        let operator = match parse_operator_public_key(publicKey.as_ref(), operator_id, owner) {
+            Ok(operator) => operator,
+            Err(reason) => {
+                self.db
+                    .insert_skipped_operator_add_tx(operator_id, &reason, tx)
+                    .map_err(|e| ExecutionError::Database(e.to_string()))?;
+                return Err(ExecutionError::InvalidEvent(reason));
             }
-            data
-        } else {
-            data
         };
 
-        // Construct the Operator and insert it into the database
-        let operator = Operator::new(data, operator_id, owner).map_err(|e| {
-            ExecutionError::InvalidEvent(format!("Failed to construct operator: {e}"))
-        })?;
+        if let Some(existing_operator_id) = self
+            .db
+            .get_any_operator_id_by_public_key_tx(&operator.rsa_pubkey, tx)
+            .map_err(|e| ExecutionError::Database(e.to_string()))?
+        {
+            let reason =
+                format!("Operator public key already exists as operator {existing_operator_id}");
+            self.db
+                .insert_skipped_operator_add_tx(operator_id, &reason, tx)
+                .map_err(|e| ExecutionError::Database(e.to_string()))?;
+            return Err(ExecutionError::InvalidEvent(reason));
+        }
+
         self.db
             .insert_operator_tx(&operator, tx, state_updates)
             .map_err(|e| {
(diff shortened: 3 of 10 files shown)


anchor/eth/src/event_processor.rs
```diff
@@ -426,6 +432,19 @@ impl EventProcessor {
...
 
+        if self
+            .db
+            .was_operator_add_skipped_tx(operator_id, tx)
+            .map_err(|e| ExecutionError::Database(e.to_string()))?
+        {
+            self.db
+                .delete_skipped_operator_add_tx(operator_id, tx)
+                .map_err(|e| ExecutionError::Database(e.to_string()))?;
+            return Err(ExecutionError::SkippedEvent(format!(
```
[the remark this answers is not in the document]
diegomrsantos: @shane-moore there seems to be an issue with claude. Maybe it's using an outdated code to review the PR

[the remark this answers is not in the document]
shane-moore: interesting! will take a look man

diegomrsantos:
@claude review this PR

anchor/eth/src/event_processor.rs
```diff
@@ -380,24 +381,29 @@ impl EventProcessor {
...
 
-        let data = publicKey.as_ref();
-
-        // If the data is 704 bytes, remove the ssv encoding. Else, just parse the key
-        let data = if data.len() == 704 {
-            let mut data = &data[64..];
-            // while there is a 0 at the end of the data, remove it
-            while let [rest @ .., 0] = data {
-                data = rest;
+        let operator = match parse_operator_public_key(publicKey.as_ref(), operator_id, owner) {
```
shane-moore, line 384: > In `process_operator_added`, any `parse_operator_public_key(...)` failure records a row in `skipped_operator_adds` before returning `InvalidEvent` (`anchor/eth/src/event_processor.rs:384-390`). That is broader than the canonical-duplicate case handled later at `:394-404`; it also covers malformed / unsupported operator-key payloads. `process_operator_removed` then treats any existing marker for that `operator_id` as an expected skip and consumes it without inspecting the stored reason (`:435-442`).
>
> I can see the argument for this broader policy: the contract accepts arbitrary `bytes`, so if Anchor cannot interpret a registered operator, it may still need to consume the lifecycle symmetrically (`OperatorAdded` skipped -> matching `OperatorRemoved` skipped) rather than hard-stopping replay until Anchor is patched.
>
> If that is the intended policy, could you make it explicit in the PR description? Right now the description mostly justifies the canonical-duplicate replay case from `#972`, so this broader parse-failure behavior reads as wider than the stated scope. If it is intentional, I’d also want a regression test for a non-duplicate parse failure followed by `OperatorRemoved`. Otherwise, marker creation / consumption should be restricted to the duplicate-key case, or reason-classed so only duplicate markers suppress the later remove.

anchor/eth/tests/integration.rs
```diff
@@ -71,6 +71,101 @@ async fn test_operator_added_event_processing() {
     verify_operator_stored(&test.processor, OperatorId(operator_id));
 }
 
+#[tokio::test]
+async fn test_wrapped_hex_duplicate_operator_add_is_skipped_and_later_remove_does_not_abort() {
```
shane-moore, line 75: the regression test exercises `process_block_logs` but not the multi-block `process_logs` outer loop.
>
> The test calls `process_logs` three times with a single log each (`integration.rs:99, :116, :144`). In production both call sites pass multi-block batches: historical sync at `sync.rs:486` with `BATCH_SIZE = 10000`, and the live path at `sync.rs:737` during catchup. The #972 scenario itself was found through Hoodi historical replay, which is always multi-block.
>
> The outer loop at `event_processor.rs:94-113` uses `flush_current_block_if_any(...)?` — if block N flushes with `AbortBatch`, any subsequent blocks' logs already buffered in the same `Vec<Log>` are dropped before their `process_block_logs` runs. The existing three-single-call shape cannot exercise this continuation path; it verifies that a per-block transaction commits after a `SkippedEvent`, but not that later blocks in the same `process_logs` call survive.
>
> Could we add a single multi-block-batch test alongside the existing one? Something like:
>
> ```rust
> process_logs(vec![
>     add_op1 @ block 12345,
>     add_op2_hex_dup @ block 12346,  // becomes skip marker
>     remove_op2 @ block 12347,        // consumes marker, SkippedEvent
>     add_op3 @ block 12348,           // key assertion: this survives
> ], true, 12348)?;
>
> verify_operator_stored(OperatorId(3));
> assert_eq!(last_processed_block, 12348);
> ```
>
> This mirrors the shape historical sync actually drives and asserts that block 12348's log survives a non-abort at block 12347 — the invariant #972's fix hinges on at production scale.

anchor/database/src/tests/operator_tests.rs
```diff
@@ -158,4 +158,43 @@ mod operator_database_tests {
                 .is_err()
         )
     }
+
+    #[test]
+    fn test_skipped_operator_add_round_trip() {
```
shane-moore, line 163: the new `skipped_operator_adds` table is exercised on fresh DBs by `test_skipped_operator_add_round_trip`, but the manual-V1 adoption tests in `metadata_tests.rs` only assert other V2 effects (`max_operator_id_seen`, validator index, migration history). A small insert/read against `skipped_operator_adds` in `test_manual_v1_adoption` or `test_stamped_v1_restart_recovers_and_applies_v2` would make the bridge + V2 coverage more complete.

anchor/database/src/operator_operations.rs
```diff
@@ -51,6 +52,58 @@ impl NetworkDatabase {
...
+        tx: &Transaction<'_>,
+    ) -> Result<bool, DatabaseError> {
+        let rows = tx
+            .prepare_cached(sql_operations::DELETE_SKIPPED_OPERATOR_ADD)?
+            .execute(params![id])?;
+        Ok(rows > 0)
+    }
+
+    /// Check if an `OperatorAdded` log for this id was previously skipped.
+    pub fn was_operator_add_skipped_tx(
```
shane-moore, line 82: `was_operator_add_skipped_tx` (`operator_operations.rs:81`) and `GET_SKIPPED_OPERATOR_ADD_REASON` (`sql_operations.rs:33`) have no production caller — repo-wide grep shows the three call sites in `test_skipped_operator_add_round_trip` at `tests/operator_tests.rs:171, 183, 196`, and production's only presence-check is `delete_skipped_operator_add_tx`'s `bool` return at `event_processor.rs:437`. Right now they read as production API but are really test-only.
>
> Rather than deleting them, could you gate both behind `#[cfg(any(test, feature = "test-utils"))]`? The round-trip test relies on `was_operator_add_skipped_tx`'s honest `Result<bool, DatabaseError>` typing to panic loudly on SQL errors via `.expect(...)`. Repointing the test at `queries::get_skipped_operator_reason` would regress that — its `.ok()` at `tests/utils.rs:482` silently maps SQL errors to `None`, and two of the three `.is_some()` / `.is_none()` checks in the test would false-pass on a broken DB (expected-absent outcomes happen to match `None`). Gating keeps the honest typing the test depends on while removing it from the production surface.
>
> Small side note for a follow-up: `tests/utils.rs:363-364` defines a private `GET_SKIPPED_OPERATOR_REASON` with byte-identical SQL to `sql_operations::GET_SKIPPED_OPERATOR_ADD_REASON`. After the gating, both are test-only — worth consolidating to one constant, but not load-bearing for this PR.

anchor/eth/tests/integration.rs
diegomrsantos: Good suggestion. I don’t think it is required for this PR, since `#973` is about the operator add/remove replay classification itself, and that behavior is already covered directly plus validated in live replay.

I opened `#975` to track the broader test hardening separately. I think that follow-up should be more generic than this operator path: we already cover the `AbortBatch` boundary, but we do not yet have symmetric multi-block `process_logs` continuation coverage showing that later blocks survive `SkipMalformed`, `SkipExpected`, and `SkipAmbiguous` in the same batch.

anchor/eth/src/event_processor.rs
diegomrsantos: Addressed now.

I made the broader policy explicit in the PR description: skipped markers are intentional not only for canonical duplicate keys, but also for operator-key payloads Anchor cannot interpret, so replay can consume the add/remove lifecycle symmetrically instead of wedging on the later remove.

I also added a direct regression for the non-duplicate parse-failure case: `test_malformed_operator_add_is_skipped_and_later_remove_does_not_abort`.

anchor/eth/src/event_processor.rs
```diff
@@ -960,3 +976,181 @@ impl EventProcessor {
...
+fn abi_word_to_usize(word: &[u8]) -> Option<usize> {
+    if word.len() != 32 || word[..24].iter().any(|byte| *byte != 0) {
+        return None;
+    }
+
+    usize::try_from(u64::from_be_bytes(word[24..].try_into().ok()?)).ok()
+}
+
+#[cfg(test)]
+mod tests {
```
shane-moore, line 1055: the 5 tests added here cover the happy path plus two negative branches. A few more one-line tests would close the remaining parser-helper gaps:
>
> **`abi_decode_single_dynamic_bytes`**
> - A too-short input (e.g. `&[0u8; 32]`) — rejected by the first length guard.
> - A length-mismatch buffer — e.g. 96 bytes total but a declared `len = 40`, so `64 + padded_len(40) = 128 ≠ 96` — rejected by the second length guard.
>
> **`decode_hex_encoded_operator_pem`** (all four return `Ok(None)` today, none are tested):
> - Non-UTF8 bytes (e.g. `&[0xff, 0xfe]`).
> - Empty input.
> - Odd-length hex.
> - Non-hex ASCII (e.g. `b"zzzz"`).
>
> Plus one happy-path test with a `0x` prefix to pin the `strip_prefix` behavior.

anchor/eth/tests/common/mod.rs
```diff
@@ -42,6 +42,35 @@ pub fn create_valid_rsa_public_key_bytes() -> Bytes {
...
+
+    encoded[31] = 32;
+    encoded[56..64].copy_from_slice(&(public_key.len() as u64).to_be_bytes());
+    encoded[64..64 + public_key.len()].copy_from_slice(public_key);
+
+    Bytes::from(encoded)
+}
+
+/// Generate a valid operator key wrapped in the nested ABI string encoding used on-chain.
+pub fn create_wrapped_base64_operator_public_key_bytes() -> Bytes {
```
shane-moore: Defined but never called so think we can delete

anchor/database/src/tests/operator_tests.rs
diegomrsantos: Addressed now.

I added a direct `skipped_operator_adds` insert/read round-trip to both manual-`V1` bridge paths in `metadata_tests.rs`:
- `test_manual_v1_adoption`
- `test_stamped_v1_restart_recovers_and_applies_v2`

So the V2 adoption coverage now proves that the bridge path materializes a writable `skipped_operator_adds` table as well, not just the other V2 effects.

anchor/eth/tests/common/mod.rs
diegomrsantos: Addressed now. `create_wrapped_base64_operator_public_key_bytes` was unused, so I removed it rather than keeping dead test helper surface around.

anchor/database/src/operator_operations.rs
diegomrsantos: Addressed now.

I gated the genuinely test-only pieces behind `#[cfg(any(test, feature = "test-utils"))]`:
- `was_operator_add_skipped_tx`
- `GET_SKIPPED_OPERATOR_ADD_REASON`

I kept the stronger `Result<bool, DatabaseError>` helper for the round-trip test, and I also removed the duplicate test SQL by switching `tests/utils.rs` to reuse `sql_operations::GET_SKIPPED_OPERATOR_ADD_REASON`.

anchor/eth/src/event_processor.rs
diegomrsantos: Addressed now.

I added the compact helper coverage we discussed:
- `decode_hex_encoded_operator_pem_accepts_0x_prefixed_hex_payload`
- `decode_hex_encoded_operator_pem_returns_none_for_non_hex_like_inputs` (empty, odd-length, non-hex ASCII, non-UTF8)
- `abi_decode_single_dynamic_bytes_rejects_length_mismatch_buffer`

I left out the separate too-short-input case since it would only duplicate the first length guard without covering a different behavior branch.

shane-moore (approved):
Really nice work here. You tracked down a real replay bug, made the broader skip-marker policy explicit, and backed it up with solid regression coverage for both the duplicate-key and malformed-key paths. LGTM!

PR #976, opened 2026-04-24
fix(keysplit): replace unsafe unwrap with proper error handling in build_db

purelualight:
Keep this concise. Use bullets where helpful. Write `N/A` if not applicable.

There some  unsafe `.unwrap()` calls  in code.

## Change Overview (Required)

Replaced unsafe `.unwrap()` calls in the `build_db()` function with proper error handling using `Result` and `map_err`, providing meaningful error messages when RSA modulus or exponent extraction fails.

## Validation (Required)

- Examples: targeted tests, manual repro, benchmark/profiling result, before/after behavior, spec-parity check.


purelualight:
image

PR #977, opened 2026-04-24
fix(keysplit): replace unsafe unwrap with proper error handling in build_db

purelualight:
Keep this concise. Use bullets where helpful. Write `N/A` if not applicable.

There some unsafe .unwrap() calls in code.

## Change Overview (Required)

Replaced unsafe .unwrap() calls in the build_db() function with proper error handling using Result and map_err, providing meaningful error messages when RSA modulus or exponent extraction fails.

## Validation (Required)

- Examples: targeted tests, manual repro, benchmark/profiling result, before/after behavior, spec-parity check.

### DIFF
--- a/anchor/keysplit/src/split.rs
+++ b/anchor/keysplit/src/split.rs
@@ -44,7 +44,7 @@ pub fn onchain_split<'a>(
     secret_keys: impl IntoIterator<Item = &'a SecretKey>,
 ) -> Result<Vec<Split<KeyShare>>, KeysplitError> {
     // Construct DB and perform sync
-    let db = build_db();
+    let db = build_db()?;
     let mut syncer =
         SsvEventSyncer::new_keysplit(db.clone(), onchain.rpc, global_config.ssv_network);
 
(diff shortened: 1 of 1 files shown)


diegomrsantos:
Thanks for the contribution. The direction makes sense, but this needs a few changes before it can be reviewed for merge.

Please retarget the PR to `unstable`; regular contributions should not target `stable` directly. Also, `build_db()` still has fallible operations that can panic, especially `NetworkDatabase::new(...).expect(...)`, so the change does not fully remove the panic path in this setup code. Please propagate that error as `KeysplitError::Database(...)` instead.

There are also formatting issues in the changed hunk, and the PR description still has template prompt text in several sections. Please run formatting and fill in the validation/risk/rollback sections with the actual details for this change.

shane-moore:
you did not fill out all of the pr description categories or mark a section as N/A if N/A. also notice how your pr is targeting `stable` branch whereas pretty much every other open pr is against `unstable`

shane-moore:
Thanks for the PR, going to close.

The `.expect()` calls here are intentional invariant assertions. `Rsa::generate`, `BigNum::to_owned`, and `from_public_components` on freshly-generated components are logically infallible, and the expect messages document that. `keysplit` is a one-shot CLI, so a panic and `Err` surface identically to the user.

PR #982, opened 2026-04-29
refactor(signature_collector): extract state struct

shane-moore:

- This prepares the signature collector for the follow-up #933 work without mixing behavior changes into that PR.
- The current collector task keeps its state as several mutable locals inside the async receive loop, which makes the quorum/reconstruction behavior harder to test directly.
- This is worth doing now because #933 needs focused state-machine tests for verification/fallback paths.
- Related: #933.

## Change Overview (Required)

- Extracted the per-signature collector state into `SignatureCollectorState`.
- The async `signature_collector` task is now a thin receive loop that delegates each message to `state.process(...)`.
- Reconstruction logic moved into `SignatureCollectorState::try_reconstruct`, preserving the existing threshold/combine/notify behavior.
- Added direct state tests for quorum notification, conflicting thresholds, late notifier registration, and multiple waiting notifiers.
- Suggested reading order: start in `signature_collector/src/lib.rs` at `signature_collector`, then review `SignatureCollectorState`, then the tests in `signature_collector/src/tests.rs`.
- Intentionally did not add signature verification, fallback eviction, DB access, metrics, or validator pubkey plumbing. Those remain in the follow-up #933 PR.

- Main risk is accidental behavior drift while moving logic out of the receive loop.
- Mitigation: this PR keeps the same message handling, same threshold conflict behavior, same duplicate-share behavior, and same reconstruction path.
- The tests drive the extracted state directly so the refactor has focused coverage without adding test-only production hooks.
- Trade-off: this adds a small internal struct before the feature PR, but keeps the later behavioral diff smaller and easier to review.

## Validation (Required)

- `cargo fmt --package signature_collector --check`
- `cargo test -p signature_collector`
- `git diff --cached --check`

- Revert this PR. There are no config, database, or operational changes.

- N/A

- Follow-up PR will implement post-reconstruction signature verification with fallback for #933.

### DIFF
--- a/anchor/signature_collector/src/tests.rs
+++ b/anchor/signature_collector/src/tests.rs
@@ -31,7 +31,7 @@ fn register_notifier(
     threshold: u64,
 ) -> oneshot::Receiver<Arc<Signature>> {
     let (notify, rx) = oneshot::channel();
-    let outcome = state.process(CollectorMessageKind::RegisterNotifier { notify, threshold });
+    let outcome = state.register_request(notify, threshold);
     assert!(
         outcome.is_continue(),
         "register_notifier should not break the collector"
@@ -44,10 +44,7 @@ fn feed_partial_sig(
     operator_id: OperatorId,
     signature: Signature,
 ) {
-    let outcome = state.process(CollectorMessageKind::PartialSignature {
-        operator_id,
-        signature: Box::new(signature),
-    });
+    let outcome = state.add_partial_signature(operator_id, signature);
     assert!(
         outcome.is_continue(),
         "feed_partial_sig should not break the collector"
@@ -87,10 +84,7 @@ fn state_breaks_on_conflicting_thresholds() {
     let _first_rx = register_notifier(&mut state, THRESHOLD);
 
     let (second_notify, mut second_rx) = oneshot::channel();
-    let outcome = state.process(CollectorMessageKind::RegisterNotifier {
-        notify: second_notify,
-        threshold: THRESHOLD + 1,
-    });
+    let outcome = state.register_request(second_notify, THRESHOLD + 1);
 
     assert!(
         outcome.is_break(),
--- a/anchor/signature_collector/src/tests.rs
+++ b/anchor/signature_collector/src/tests.rs
@@ -153,3 +153,51 @@ fn state_notifies_all_registrants_on_reconstruction() {
     expect_signature(&mut rx_b, "registrant b should be notified");
     expect_signature(&mut rx_c, "registrant c should be notified");
 }
+
+/// Partial signatures that arrive before any `RegisterNotifier` is buffered.
+/// Reconstruction triggers when a notifier and the threshold arrives.
+#[test]
+fn state_buffers_shares_arriving_before_first_notifier_register() {
+    let shares = split_random_master();
+    let mut state = SignatureCollectorState::default();
+
+    for (op_id, sk) in &shares[..THRESHOLD as usize] {
+        feed_partial_sig(&mut state, *op_id, sk.sign(SIGNING_ROOT));
+    }
+
+    assert!(
+        state.full_signature.is_none(),
+        "State should not reconstruct without a threshold"
+    );
+
+    let mut result_rx = register_notifier(&mut state, THRESHOLD);
+    expect_signature(
+        &mut result_rx,
+        "Notifier should receive the signature reconstructed from buffered shares",
+    );
+}
+
+/// Partial signatures that arrive after reconstruction are silently dropped:
+/// no panic, no Break, no re-buffering of the late share. This is the common
+/// case in production: slow operators' shares routinely arrive after a fast
+/// majority has already reconstructed the signature.
+#[test]
+fn state_drops_partial_signatures_after_reconstruction() {
+    let shares = split_random_master();
+    let mut state = SignatureCollectorState::default();
+
+    let mut rx = register_notifier(&mut state, THRESHOLD);
+    for (op_id, sk) in &shares[..THRESHOLD as usize] {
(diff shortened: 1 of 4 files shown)


shane-moore:
@claude, please take a look

shane-moore:
@claude, thanks for the close read.

I do not think Issue 1 is a behavior change. In the original implementation, the reconstruction check ran after the full `match`, not only after `PartialSignature`:

```rust
match message.kind {
    CollectorMessageKind::RegisterNotifier { .. } => {
        ...
        threshold = Some(new_threshold);
    }
    CollectorMessageKind::PartialSignature { .. } => { ... }
}

if let Some(threshold) = threshold
    && signature_share.len() as u64 >= threshold
{
    ...
}
```

So the old shares-before-notifier flow was:

1. partial signatures arrive and are stored while `threshold` is still `None`;
2. the first notifier arrives and sets `threshold`;
3. the same loop iteration immediately reaches the threshold check and reconstructs if enough shares are already present.

The extracted `process()` + `try_reconstruct()` preserves that same sequence.

For Issue 2, agreed that the helper message is generic, but it does not silently swallow `Break`: it asserts `outcome.is_continue()`, so an unexpected break fails the test immediately. I do not think that needs a change in this refactor PR.

anchor/signature_collector/src/tests.rs
```diff
@@ -0,0 +1,155 @@
...
+    let mut rx_c = register_notifier(&mut state, THRESHOLD);
+
+    for (op_id, sk) in &shares[..THRESHOLD as usize] {
+        feed_partial_sig(&mut state, *op_id, sk.sign(SIGNING_ROOT));
+    }
+
+    expect_signature(&mut rx_a, "registrant a should be notified");
+    expect_signature(&mut rx_b, "registrant b should be notified");
+    expect_signature(&mut rx_c, "registrant c should be notified");
+}
```
jnhsigmap, line 149: I had a great time exploring this one. Conveniently found a few points that I can raise when we implement #933.

I'm curious whether it would be worth including a test for the case where shares arrive from other operators before there is any notifiers registered. I'm just reversing the order of messages here.

I assume this can be caused by unexpected latency, randomness when receiving messages from other nodes, or client-side issues that are eventually recovered from?

```suggestion
}

/// Partial signatures that arrive before any `RegisterNotifier` is buffered.
/// Reconstruction triggers when a notifier and the threshold arrives.
#[test]
fn state_buffers_shares_arriving_before_first_notifier_register() {
    let shares = split_random_master();
    let mut state = SignatureCollectorState::default();

    for (operator_id, secret_key) in &shares[..THRESHOLD as usize] {
        feed_partial_sig(&mut state, *operator_id, secret_key.sign(SIGNING_ROOT));
    }

    assert!(
        state.full_signature.is_none(),
        "State should not reconstruct without a threshold"
    );

    let mut result_rx = register_notifier(&mut state, THRESHOLD);
    expect_signature(
        &mut result_rx,
        "Notifier should receive the signature reconstructed from buffered shares",
    );
}

```

looks good though 👍

shane-moore: Added it, thanks. You're right on the cause that `receive_partial_signatures` and `sign_and_collect` are independent entry points into `get_or_spawn`, so a remote share can land before any local notifier registers. The buffering branch (`try_reconstruct` early-returning when `threshold` is `None`) had no coverage

Following the same line of thought I added the symmetric case too where a late `PartialSignature` arriving after reconstruction (slow operators landing after a fast majority finished). That covers the `if self.full_signature.is_some()` early-return in the `PartialSignature` arm, which was the other uncovered ordering branch

Both pass

jnhsigmap: cool, otherwise looks good to me :slightly_smiling_face:

jnhsigmap (dismissed):

diegomrsantos (changes requested):
I would request one small API cleanup before this lands.

The state extraction is useful, but `SignatureCollectorState::process(CollectorMessageKind)` still exposes the async channel format as the direct state API. `CollectorMessageKind` is a good transport enum for the receive loop, but the state machine itself is handling typed operations: register a collection request with a threshold, and add a partial signature.

Could we make the recv loop the only place that matches on `CollectorMessageKind`, and expose typed state methods instead?

```rust
state.register_request(notify, threshold)
state.add_partial_signature(operator_id, signature)
```

That would remove the need for tests to construct channel messages, avoid leaking details like `Box<Signature>` into direct state callers, and make the current threshold coupling clearer. Registering a waiter is also how the reconstruction threshold enters the state, so naming this as a request/config path is easier to reason about before #933 adds verification and fallback behavior.

While looking at this, I also noticed a larger design concern: the collector still mixes request handling, partial signature ingestion, per-signature state, and async runtime boundaries. I opened #995 to track that broader cleanup separately. I am not asking for that larger redesign in this PR; the requested change here is just the typed state API above.

shane-moore:
Applied the typed-API change in `ab7da9e`. Recv loop is the only place matching on `CollectorMessageKind`, state exposes `register_request(notify, threshold)` and `add_partial_signature(op_id, sig)`, and `Box<Signature>` does not leak into the state API or tests. All 6 tests still green.

On #995: agreed it stays separate. We'll continue with the existing #931 stack first (#933 is the next PR, then #934), then come back to #995's broader cleanup, per your framing on the issue ("after the current behavior-focused stack is in place"). Happy to discuss timing or design direction on #995 whenever useful.

diegomrsantos (approved):
Thanks, this addresses my requested change. The recv loop now owns the `CollectorMessageKind` dispatch, and the state tests call the typed API directly without constructing channel messages.

The broader collector cleanup can stay in #995. This PR looks good to me now.

Issue #953, opened 2026-04-15
chore: remove fake_crypto from spec tests once ssv-spec fixtures are fixed


shane-moore:
## Context

PR #942 adds proposer consensus-data spec tests that require `fake_crypto` to run full assertions. Without it, Lighthouse's SSZ decoder rejects the invalid BLS points in ssv-spec's Capella/Deneb proposer fixtures, and the tests fall back to a path that skips block root, block SSZ, and blinded state checks.

The root cause is fixture quality in ssv-spec: the proposer fixtures use placeholder BLS values (`0x000102...` pubkeys, `0x606162...` signatures) copied from go-eth2-client codec tests. These are correct length but not valid compressed BLS curve points. Go accepts them because its decoder only checks hex+length.

## Upstream issue

https://github.com/ssvlabs/ssv-spec/issues/622

## Action

Once ssv-spec #622 is resolved and we bump the submodule:
1. Remove the `fake_crypto` feature from `spec_tests/Cargo.toml`
2. Remove the BLS fallback path in `ConsensusDataProposerTest` (`is_bls_validation_error` handling in `consensus_data_proposer.rs`)
3. Remove the BLS fallback path in `SSZSpecTest` (`is_bls_validation_error` handling in `ssz_spec_test.rs`)
4. Remove the BLS fallback in `can_decode_block` (`is_bls_validation_error` handling in `encoding_helpers.rs`)
5. Remove the BLS fallback in `ProposerConsensusDataTest` (uses `can_decode_block` in `proposer_consensus_data.rs`)
6. Remove the `is_bls_validation_error` helper from `encoding_helpers.rs`
7. Drop any CI targets that run spec tests with `--features fake_crypto`

## Discussion

- Diego's review: https://github.com/sigp/anchor/pull/942#issuecomment-4252156723
- Follow-up: https://github.com/sigp/anchor/pull/942#issuecomment-4252469671
- Analysis: https://github.com/sigp/anchor/pull/942#issuecomment-4253449773

PR #983, opened 2026-04-30
ci: run spec_tests with fake_crypto

### LINKED ISSUE #938: Add a Makefile target for spec tests with fake_crypto coverage

shane-moore:

- PR #942 added `ConsensusDataProposerTest` whose success-path assertions (block blinded state, block root, block SSZ roundtrip) only run when `bls/fake_crypto` is enabled. Without it, success fixtures fall through to a weaker BLS-rejection branch because Lighthouse validates BLS during SSZ decode while Go's `fastssz` does not.
- Today nothing in the Makefile or CI runs `spec_tests` with `fake_crypto`, so the branch can stay green while bypassing the strongest assertions of the new proposer test.
- Closes #938.

## Change Overview

Test-infra only, two additions:

- **Makefile**: new `test-spec-tests` / `nextest-spec-tests` targets that run `spec_tests` with `--features fake_crypto`; `test-full` extended to include them. Mirrors Lighthouse's `ef_tests` pattern.
- **CI**: new `spec-tests-fake-crypto` job in `test-suite.yml` that runs `make nextest-spec-tests`, gated by `test-suite-success`. Runs in parallel with the existing release-tests job.

What did **not** change: no production code; no edits to the BLS-fallback heuristic in `consensus_data_proposer.rs` or `is_bls_validation_error`. Both paths remain valuable — the workspace test run continues to exercise the realistic BLS-rejection branch, and the new job covers the full-coverage path.

- Adds a CI job with ~1–2 min compile + run; runs in parallel so it does not extend the critical path materially.
- `fake_crypto` cannot be enabled at the workspace level (the `anchor` root crate does not declare it), so the target is scoped to `-p spec_tests`.
- Trade-off vs folding inline into `release-tests-ubuntu`: separate job preserves independent failure visibility and parallel execution; cost is ~30–60s of duplicate Rust setup, accepted.

## Validation

- `make nextest-spec-tests`: 1 test, 114 fixtures, 0.06s runtime (after ~1m48s cold compile). Success fixtures now exercise `verify_block` (block root + SSZ roundtrip) instead of falling through to the BLS-rejection branch.
- `cargo nextest run --release -p spec_tests` (no `fake_crypto`): still passes via the existing BLS-fallback path — confirms the workspace test run is unaffected.
- `.github/scripts/check_success_job.sh ./.github/workflows/test-suite.yml test-suite-success` reports `COMPLETENESS CHECK PASSED`; `yq` parses the workflow YAML cleanly.

## Rollback

N/A — test-infrastructure-only change.

## Blockers / Dependencies

None.

### DIFF
--- a/.claude/rules/commands.md
+++ b/.claude/rules/commands.md
@@ -12,6 +12,7 @@ Use project Make targets by default.
 - `make test` - all tests, release mode (standard)
 - `make test-debug` - all tests, debug mode
 - `make nextest-release` / `make nextest-debug` - nextest runner
+- `make test-spec-tests` / `make nextest-spec-tests` - `spec_tests` with `fake_crypto` (full proposer block coverage)
 - `cargo test -p <crate>` - specific crate
 - `make check-benches` - compile benchmarks without running
 
--- a/.github/workflows/test-suite.yml
+++ b/.github/workflows/test-suite.yml
@@ -101,6 +101,26 @@ jobs:
     - name: Show cache stats
       if: env.SELF_HOSTED_RUNNERS == 'true'
       run: sccache --show-stats
+  spec-tests-fake-crypto:
+    name: spec-tests-fake-crypto
+    needs: [check-labels]
+    if: needs.check-labels.outputs.skip_ci != 'true'
+    runs-on: ubuntu-22.04
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        submodules: 'recursive'
+    - name: Get latest version of stable Rust
+      if: env.SELF_HOSTED_RUNNERS == 'false'
+      uses: moonrepo/setup-rust@v1
+      with:
+          cache: false
+          cache-target: release
+          bins: cargo-nextest
+      env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    - name: Run spec_tests with fake_crypto
+      run: make nextest-spec-tests
   check-fmt:
     name: check-fmt
     runs-on: ubuntu-22.04
@@ -224,6 +244,7 @@ jobs:
       'check-labels',
       'release-tests-ubuntu',
       'debug-tests-ubuntu',
+      'spec-tests-fake-crypto',
       'check-fmt',
       'check-code',
       'check-msrv',
--- a/Makefile
+++ b/Makefile
@@ -97,6 +97,15 @@ coverage:
 coverage-html:
 	cargo llvm-cov nextest --workspace --features "$(TEST_FEATURES)" --html
 
+# Runs `spec_tests` with `fake_crypto` for full proposer block coverage
+# (workspace run already covers the BLS-rejection path).
+test-spec-tests:
+	cargo test --release -p spec_tests --features fake_crypto
+
+# Same as `test-spec-tests`, using nextest.
+nextest-spec-tests:
+	cargo nextest run --release -p spec_tests --features fake_crypto
+
 # Runs cargo-fmt (linter).
 cargo-fmt:
 	cargo +$(PINNED_NIGHTLY) fmt --all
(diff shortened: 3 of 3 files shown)


shane-moore:
@claude, please take a look

diegomrsantos:
Stepping back from the mechanics of this PR — I want to flag a pattern I'd like us to push back on as a team.

When we hit friction with an upstream dependency or spec, our default has been to add a workaround inside Anchor: a feature flag, a heuristic, a fallback branch, a bypass in CI. Each one is individually reasonable, but they accumulate. The result is that the real fix — upstream, or in the shared contract — loses urgency, and we carry the complexity indefinitely.

This PR is a concrete example. The underlying issue is [ssv-spec#622](https://github.com/ssvlabs/ssv-spec/issues/622): the Capella/Deneb fixtures use placeholder BLS bytes that aren't valid compressed points. Anchor's decoder is correct to reject them. \`fake_crypto\` lets us sidestep that, and the \`is_bls_validation_error\` string match keeps the non-\`fake_crypto\` path green. Both paths will stay green regardless of whether the upstream fixture ever gets fixed, and we now have a CI job committing us to maintaining the bypass.

I'd like us to try, before adding a workaround, to:
1. Confirm there's an open upstream issue and push on it.
2. Consider fixing the problem at the boundary (e.g. regenerating or patching fixtures locally with valid inputs) so production code paths stay honest.
3. Treat the workaround as temporary with a clear removal condition, not as the new steady state.

For this specific PR, I'm not blocking — the diff is small and the intent is good. But I'd like to see us revisit ssv-spec#622 and set a bar for when this CI job comes back out, rather than treat it as permanent infrastructure.

shane-moore:
Wanted to follow up on this directly — opened [ssv-spec#627](https://github.com/ssvlabs/ssv-spec/pull/627) against `spec-test-ci` to fix the upstream root cause for #622. Capella/Deneb proposer fixtures get rewritten so every BLS-typed field holds a valid compressed point (derived from an existing test SK; non-BLS fields like KZG commitments/proofs are preserved byte-for-byte via a typed-walk discovery). A new `TestProposerFixturesBLSDecodable` in ssv-spec gates the regen so this can't regress upstream.

The removal condition for the `fake_crypto` bypass and the `is_bls_validation_error` string match is already tracked in #953: once #627 lands and the spec-tests JSONs are regenerated against it, the Anchor submodule bump should let real `blst` decode the Capella/Deneb proposer fixtures without either escape hatch. I've updated #953 to name PR #627 as the explicit trigger and listed the seven follow-up deletions, so the bypass has a concrete removal path rather than living as steady state.

PR #984, opened 2026-04-30
test: AggregatorCommitteeConsensusDataTest

shane-moore:

- Anchor's spec test suite lacks coverage for ssv-spec's `AggregatorCommitteeConsensusDataTest`. The companion `EncodingTest` for the same type is already on `unstable` (PR #835), so SSZ encode/decode parity is checked, but the 9 distinct error paths in `AggregatorCommitteeDataValidator::do_validation` are uncovered.
- Continues the type spec test rollout that PR #942 (proposer consensus data) and PR #968 (committee member) are part of. After this lands, only `maxmsgsize.StructureSizeTest` remains for full parity, after which the dispatcher's `_` arm can be replaced with `panic!()`.

## Change Overview

One new validation test type wired through the existing dispatcher:

1. **`AggregatorCommitteeConsensusDataTest`** — Mirrors Go's `AggregatorCommitteeConsensusData.Validate()` by exercising Anchor's production `AggregatorCommitteeDataValidator::do_validation` and mapping its 9 typed error variants to ssv-spec's integer error codes (71–78, 84). Drift between Anchor's validator and ssv-spec's would surface as a fixture mismatch.

Reading order:

- `types/aggregator_committee_consensus_data.rs` (new) — small test struct mirroring `proposer_consensus_data.rs`'s shape from PR #942.
- `utils/dtos.rs` — new `RawAssignedAggregator` and `RawAggregatorCommitteeConsensusData` DTOs with `TryFrom` impls into the production types. `SyncCommitteeContribution<E>` is deserialized directly through Lighthouse's existing serde derive — no DTO sub-type for the sync side. List fields use `Option<Vec<_>>` because Go marshals empty slices as JSON `null`.
- `utils/error_codes.rs` — 9 new constants and `aggregator_committee_validation_error_code` mapper. The match is exhaustive on `AggregatorCommitteeValidationError`, so a new variant on the production enum will fail to compile here rather than silently mapping to `UNMAPPED_ERROR_CODE`.
- `utils/deserializers.rs` — small additions (`deserialize_hex`, `deserialize_base64_list_or_null`) reused for the new DTO.
- `runner/types_dispatcher.rs` — one new match arm. `parse_validator_index` was extracted in `dtos.rs` and now backs `RawDuty`, `RawPartialSignatureMessage`, and `RawAssignedAggregator`.

What did **not** change: production validation logic. Anchor's `do_validation` is the source of truth and the test is the consumer. No production type gained `serde` derives.

- **`fake_crypto` not required.** Selection-proof zeros short-circuit via `bls::Signature::deserialize`'s `NONE_SIGNATURE` handling; real proofs are valid blst points; aggregated attestations stay as raw SSZ bytes until decoded inside `do_validation`. All 16 new fixtures pass under both default and `--features fake_crypto`.
- **DTO conversion failure → `UNMAPPED_ERROR_CODE`** (matches the established `committee_member.rs` pattern). Loses the underlying error message on failure, but every fixture is well-formed today; a TryFrom failure indicates a real bug or fixture/spec drift and `assert_error_code` reports the unmapped sentinel loudly.
- **Variant order in the mapper** is aligned with `AggregatorCommitteeValidationError`'s declaration order in `consensus.rs`, so future variant changes are easy to audit.

## Validation

- `cargo test -p spec_tests --release` — 74 fixtures executed (was 58 on `unstable`, +16 matches the count of `aggregatorcommitteeconsensusdata.AggregatorCommitteeConsensusDataTest_*.json` fixtures).
- `cargo test -p spec_tests --release --features fake_crypto` — same 74 pass.
- End-to-end sanity: temporarily flipping an `ExpectedErrorCode` in a fixture caused the runner to report the mismatch correctly; fixture restored.

## Rollback

N/A — test-only changes, no runtime impact.

### DIFF
--- a/anchor/spec_tests/src/runner/types_dispatcher.rs
+++ b/anchor/spec_tests/src/runner/types_dispatcher.rs
@@ -68,6 +68,13 @@ fn dispatch_fixture_by_prefix(prefix: &str, path: &Path, contents: &str) -> Disp
             DispatchOutcome::Executed(run_test::<types::ProposerConsensusDataTest>(path, contents))
         }
 
+        // Aggregator-committee consensus data validation tests
+        "aggregatorcommitteeconsensusdata.AggregatorCommitteeConsensusDataTest" => {
+            DispatchOutcome::Executed(run_test::<types::AggregatorCommitteeConsensusDataTest>(
+                path, contents,
+            ))
+        }
+
         // Proposer block data extraction tests
         "consensusdataproposer.ProposerSpecTest" => {
             DispatchOutcome::Executed(run_test::<types::ConsensusDataProposerTest>(path, contents))
--- a/anchor/spec_tests/src/types/aggregator_committee_consensus_data.rs
+++ b/anchor/spec_tests/src/types/aggregator_committee_consensus_data.rs
@@ -1,0 +1,47 @@
+use serde::Deserialize;
+use ssv_types::consensus::AggregatorCommitteeDataValidator;
+use types::MainnetEthSpec;
+
+use crate::{
+    SpecTest,
+    utils::{
+        dtos::RawAggregatorCommitteeConsensusData,
+        error_codes::{self, aggregator_committee_validation_error_code},
+    },
+};
+
+/// Mirrors Go's `AggregatorCommitteeConsensusData.Validate()`.
+///
+/// Exercises the production `AggregatorCommitteeDataValidator::do_validation` so any
+/// drift between Anchor's validator and ssv-spec's is caught by the fixtures.
+#[derive(Debug, Deserialize)]
+#[serde(rename_all = "PascalCase")]
+pub struct AggregatorCommitteeConsensusDataTest {
+    #[serde(rename = "AggCommConsensusData")]
+    consensus_data: RawAggregatorCommitteeConsensusData,
+    expected_error_code: i64,
+}
+
+impl SpecTest for AggregatorCommitteeConsensusDataTest {
+    fn run(&self) -> Result<(), String> {
+        let actual_code = match self.validate() {
+            Ok(()) => error_codes::NO_ERROR,
+            Err(code) => code,
+        };
+        error_codes::assert_error_code(self.expected_error_code, actual_code)
+    }
+}
+
+impl AggregatorCommitteeConsensusDataTest {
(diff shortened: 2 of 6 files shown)


shane-moore:
@claude, please take a look

shane-moore:
@claude, fair point about the pattern consistency note. implemented in https://github.com/sigp/anchor/pull/985

diegomrsantos:
Small follow-up: this should be easy to make more robust, and it can stay entirely inside the spec-test adapter.

In `RawAssignedAggregator::try_from`, the adapter parses `SelectionProof` into a real `bls::Signature`, but `AggregatorCommitteeDataValidator::do_validation()` does not read `selection_proof`. That means a future ssv-spec fixture with malformed proof bytes could fail while constructing the Rust DTO, before the test reaches the validation path the fixture is meant to cover. The current fixtures are fine, so I do not think this blocks the PR.

One way to keep this focused on `Validate()` parity is to use a known placeholder signature for this validation-only conversion and make that intent explicit in the helper name:

```rust
fn to_assigned_aggregator_for_validation(&self) -> Result<AssignedAggregator, String> {
    Ok(AssignedAggregator {
        validator_index: parse_validator_index(&self.validator_index)?,
        selection_proof: bls::Signature::empty(),
        committee_index: self.committee_index,
    })
}
```

Then the adapter still checks the fields that `Validate()` uses, without adding BLS proof validity as an implicit precondition for these fixtures.

shane-moore:
Thanks @diegomrsantos, applied your suggestion in 510ffff.

Replaced the `TryFrom` impl with `to_assigned_aggregator_for_validation()` and substituted `bls::Signature::empty()` so BLS proof validity is no longer an implicit precondition for these fixtures. Kept the change inside the spec-test adapter, with no production type changes.

Verified locally: `cargo test -p spec_tests --release` and the `--features fake_crypto` variant both pass; clippy + fmt clean.

diegomrsantos (approved):
Thanks for the quick follow-up. I re-reviewed 510ffff and the validation adapter now avoids making BLS proof validity an implicit precondition while still preserving the fields used by `AggregatorCommitteeDataValidator::do_validation()`.

The PR remains test-only, the spec tests pass in normal and fake-crypto modes, and I do not have any blocking concerns. Approved.

PR #988, opened 2026-05-04
chore: ignore hickory-proto audit advisories

shane-moore:

CI's `audit-CI` step fails on two advisories against `hickory-proto 0.25.2`, pulled in transitively via `libp2p-dns`:

- [RUSTSEC-2026-0118](https://rustsec.org/advisories/RUSTSEC-2026-0118.html) — NSEC3 unbounded loop
- [RUSTSEC-2026-0119](https://rustsec.org/advisories/RUSTSEC-2026-0119.html) — O(n²) name-compression CPU DoS

Both clear once libp2p picks up hickory ≥ 0.26.1 (tracked upstream in [libp2p/rust-libp2p#6395](https://github.com/libp2p/rust-libp2p/pull/6395)). Until then, CI is blocked. Mirrors [sigp/lighthouse#9257](https://github.com/sigp/lighthouse/pull/9257).

Anchor tracking issue (reachability analysis, removal condition, review date): #989

## Change Overview

Adds `--ignore` flags for both advisories to the `audit-CI` target, plus a TODO comment pointing at the tracking issue so the ignores get removed when the libp2p bump lands.

Verified non-reachability via `cargo tree -e features -i hickory-proto`. The only `hickory-proto` features active in Anchor's resolved graph are `futures-io`, `std`, `tokio`; `dnssec-ring`/`dnssec-aws-lc-rs` are not enabled because `libp2p-dns` declares `hickory-resolver` with `default-features = false, features = ["system-config"]`.

- **RUSTSEC-2026-0118** (NSEC3 unbounded loop in `DnssecDnsHandle`) — unreachable. The advisory states the bug requires `dnssec-ring` or `dnssec-aws-lc-rs` plus DNSSEC validation configured. Neither feature is enabled, so `DnssecDnsHandle` is `cfg`'d out and the vulnerable code is not compiled.
- **RUSTSEC-2026-0119** (O(n²) name-compression in encoder) — not practically reachable. The bug is in DNS message encoding and triggers on messages with many compressible names. `libp2p-dns` uses hickory as a stub client: it encodes single-question outbound queries (one hostname per multiaddr) and decodes responses. Stub queries don't construct the multi-name messages the bug needs.

Tracking issue #989 captures the removal condition and a 2026-05-25 review date so the ignores don't go stale.

## Validation

`make audit-CI` exits 0 locally; remaining output is the standard non-fatal `unmaintained`/`yanked` warnings.

## Rollback

Revert the Makefile change.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -144,8 +144,12 @@ audit: install-audit audit-CI
 install-audit:
 	cargo install --force cargo-audit
 
+# TODO: drop the hickory-proto ignores once libp2p picks up hickory >= 0.26.1
+# (libp2p/rust-libp2p#6395). Hickory is a transitive dep of libp2p, so the bump
+# happens by updating the libp2p source in workspace Cargo.toml (version bump
+# and/or the [patch] rev), then refreshing the lockfile.
 audit-CI:
-	cargo audit
+	cargo audit --ignore RUSTSEC-2026-0118 --ignore RUSTSEC-2026-0119
 
 # Runs `cargo vendor` to make sure dependencies can be vendored for packaging, reproducibility and archival purpose.
 vendor:
--- a/Makefile
+++ b/Makefile
@@ -144,10 +144,9 @@ audit: install-audit audit-CI
 install-audit:
 	cargo install --force cargo-audit
 
-# TODO: drop the hickory-proto ignores once libp2p picks up hickory >= 0.26.1
-# (libp2p/rust-libp2p#6395). Hickory is a transitive dep of libp2p, so the bump
-# happens by updating the libp2p source in workspace Cargo.toml (version bump
-# and/or the [patch] rev), then refreshing the lockfile.
+# Tracked in sigp/anchor#989. Drop the hickory-proto ignores once libp2p ships
+# a release with hickory >= 0.26.1 (libp2p/rust-libp2p#6395) and we bump our
+# libp2p dep. See the tracking issue for the reachability analysis.
 audit-CI:
 	cargo audit --ignore RUSTSEC-2026-0118 --ignore RUSTSEC-2026-0119
 


shane-moore:
@claude, please take a look

diegomrsantos:
I’m not comfortable merging this as a bare `cargo audit --ignore` change.

This PR teaches CI to ignore two DoS advisories, but the follow-up is only a Makefile TODO pointing at an upstream libp2p PR. If that upstream PR stalls or changes shape, Anchor has no local issue, owner, or review date to make sure these ignores are removed.

The reachability explanation also needs to be more precise. Anchor enables libp2p `dns` and wraps the transport with `libp2p::dns::tokio::Transport::system`, so DNS is part of the runtime path. My current read is that the specific Hickory paths behind these advisories do not appear reachable with Anchor’s enabled features and usage, but that reasoning should be written down in the PR.

Can we please add an Anchor tracking issue, a concrete removal condition, a review date, and a short note explaining the reachability analysis for `RUSTSEC-2026-0118` and `RUSTSEC-2026-0119`?

I’m fine with a temporary CI unblock, but I don’t think we should merge an open-ended ignore for security advisories.

shane-moore:
Thanks for the pushback. Done:

- Tracking issue: #989, with reachability analysis, removal condition, and a 2026-05-25 review date.
- Makefile comment now points at #989 instead of just the upstream libp2p PR.
- PR description rewritten to drop the loose "DNSSEC paths not exercised" framing and replace it with the verified analysis from #989.

Question on enforcing the review date — do you prefer:
- a GitHub issue milestone (date-tagged, repo-visible, no active notification), or
- me holding it on my own Google calendar (lower repo overhead, invisible to others)?

diegomrsantos:
Thanks for updating this.

Please keep the follow-up visible in the repo. Assign #989 to yourself; a private calendar reminder is not enough because nobody else can see it or check whether the ignore is still needed.

One small wording caveat: the active runtime path I verified is via `libp2p-dns`. `libp2p-mdns` may appear in the lockfile, but it is not enabled in the Anchor `network` crate feature graph. I would avoid wording that implies every lockfile path is active.

Also, do we know why libp2p/rust-libp2p#6395 has not merged yet? Before we merge an ignore locally, I would like to know whether the upstream bump is just waiting on review/CI or whether there is a real blocker that could keep this around longer.

shane-moore:
Thanks. Addressing each:

1. Assigning #989 to myself.

2. The PR description and #989 already say "via `libp2p-dns`" only (no `libp2p-mdns`); the reachability paragraph is libp2p-dns–specific (its declared `default-features = false, features = ["system-config"]`). I'll keep that discipline.

3. libp2p#6395 is genuinely blocked, not just review/CI lag. Verified against hickory 0.26.1 source:
   - `hickory_resolver::name_server` is now a private module (`mod name_server;` at `crates/resolver/src/lib.rs:190`, was `pub mod` in 0.25), so libp2p-dns's imports of `name_server::TokioConnectionProvider` and `name_server::ConnectionProvider` no longer resolve.
   - `Ipv4Lookup` / `Ipv6Lookup` / `TxtLookup` were removed in favour of a unified `Lookup`.
   - `ResolveError` / `ResolveErrorKind` are no longer exported from the crate root.

   These are exactly the import paths used at Anchor's pinned `defcaf1a78` rev (`transports/dns/src/lib.rs:62, 107–115`), and they match the `unresolved import` / `module name_server is private` errors in 6395's CI logs for `Test libp2p-dns`. The other workspace test/MSRV/clippy/semver failures cascade from the same root cause.

   Status of any actual port work: no human activity on #6395 since dependabot opened it on 2026-04-27 (today's events were a dependabot auto-rebase, not a human). No competing PR exists — searched libp2p PRs for "hickory"; no other open work. The libp2p tracking issues for these advisories (#6411, #6412) have zero maintainer comments. libp2p's umbrella release cadence is also slow (~5–6 months between v0.55.0 → v0.56.0).

So upstream is realistically weeks-to-months away from a fix. The reachability analysis in #989 is the load-bearing safety argument; the review date is for re-evaluating whether the ignore is still warranted, not "expect upstream to have shipped." I can update #989 with the upstream-blocker status if helpful.

diegomrsantos:
Thanks, that clarifies the upstream state.

Given that libp2p/rust-libp2p#6395 is blocked on real `libp2p-dns` porting work, I am not comfortable approving this as-is. This no longer looks like a short-lived local ignore while an upstream bump is about to land.

The facts that concern me are:

- libp2p/rust-libp2p#6395 is a Dependabot PR with failing `libp2p-dns` checks against hickory `0.26.1`.
- The failures are API breakage, not flaky CI: `name_server` is private now, `TokioConnectionProvider` no longer resolves from the old path, the typed lookup aliases are gone, and `ResolveError` / `ResolveErrorKind` are no longer exported the same way.
- libp2p/rust-libp2p#6411 and #6412 are open with no assignee, labels, or maintainer discussion.

The reachability analysis here may be correct for Anchor’s current feature graph, but it is still a local, manual argument for ignoring two DoS advisories. Before merging this, I would want to see upstream treated as active work: at minimum the libp2p issues should be updated with the blocker, and ideally there should be a maintainer-owned fix or port PR rather than an unattended Dependabot bump.

For now I am going to hold approval.

shane-moore:
got it,  posted the upstream blocker note on https://github.com/libp2p/rust-libp2p/pull/6395#issuecomment-4374376830

diegomrsantos (approved):
Approving this as a temporary workaround to unblock CI. I opened https://github.com/libp2p/rust-libp2p/pull/6418 to address the underlying hickory/libp2p compatibility issue upstream. My preference is that we get the libp2p PR merged upstream and remove these audit ignores before cutting an Anchor release.

PR #996, opened 2026-05-06
ci: add coverage workflow

diegomrsantos:

Anchor runs Rust tests in CI with `cargo nextest`, but there is no coverage workflow. This adds a free coverage path that uses the same test runner shape as the existing CI jobs and publishes a standard lcov report for Codecov.

Evidence: `cargo-llvm-cov` supports `nextest` directly, and a targeted smoke run produced an lcov report for the `version` crate.

## Change Overview (Required)

- Add `make coverage` for generating `lcov.info` through `cargo llvm-cov nextest`.
- Add `make coverage-html` for local HTML reports.
- Add a dedicated `coverage` GitHub Actions workflow that installs `cargo-nextest` and `cargo-llvm-cov`, runs the Make target, uploads `lcov.info` as an artifact, and uploads the report to Codecov.

Reviewers can start with `.github/workflows/coverage.yml`, then check the two Make targets. This does not add coverage thresholds or change the existing test-suite workflow.

The main risk is extra CI runtime from an instrumented test run. Keeping this in a separate workflow makes that cost visible and easy to tune independently.

The Codecov upload is non-blocking for the first version, so missing Codecov-side setup will not block CI. The lcov artifact still gives reviewers access to the generated report.

## Validation (Required)

- The new Make coverage target expands to the expected `cargo-llvm-cov nextest` command.
- The workflow YAML parses successfully.
- A targeted coverage smoke run against the `version` crate passed and produced an lcov report.
- The repo commit hook completed successfully, including formatting, clippy, and Cargo.toml sorting checks.

Full workspace coverage was not run locally.

Revert this PR to remove the workflow and Make targets. There is no runtime, schema, or configuration impact.

Codecov may need repository or organization setup for OIDC uploads before hosted reporting appears in the Codecov UI.

After the first few reports establish a baseline, a follow-up can decide whether to add Codecov status checks or explicit coverage thresholds.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - stable
+      - unstable
       - 'pr/*'
   pull_request:
     types: [opened, synchronize, reopened, ready_for_review]
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -1,0 +1,56 @@
+name: coverage
+
+on:
+  push:
+    branches:
+      - stable
+      - 'pr/*'
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
+  merge_group:
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+env:
+  CARGO_INCREMENTAL: 0
+  CARGO_TERM_COLOR: always
+  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+  RUSTFLAGS: "-D warnings"
+
+jobs:
+  coverage:
+    name: coverage
+    if: github.event_name != 'pull_request' || !contains(github.event.pull_request.labels.*.name, 'skip-ci')
+    runs-on: ubuntu-22.04
+    permissions:
+      contents: read
+      id-token: write
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        submodules: 'recursive'
+    - name: Get latest version of stable Rust
+      uses: moonrepo/setup-rust@v1
+      with:
+        cache: false
+        channel: stable
+        bins: cargo-nextest, cargo-llvm-cov
+    - name: Generate coverage report
+      run: make coverage
+    - name: Upload lcov report artifact
+      uses: actions/upload-artifact@v4
+      with:
+        name: lcov-report
+        path: lcov.info
+        if-no-files-found: error
+    - name: Upload coverage to Codecov
+      uses: codecov/codecov-action@v6
+      with:
+        files: lcov.info
+        flags: rust
+        name: anchor-rust
+        use_oidc: true
+        fail_ci_if_error: false
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: tests
+.PHONY: tests coverage coverage-html
 
 GIT_TAG := $(shell git describe --tags --candidates 1)
 BIN_DIR = "bin"
@@ -88,6 +88,14 @@ test-debug:
 # vectors, using nextest.
 nextest-debug:
 	cargo nextest run --workspace --features "$(TEST_FEATURES)"
+
+# Generates an lcov coverage report using the nextest runner.
+coverage:
+	cargo llvm-cov nextest --workspace --features "$(TEST_FEATURES)" --lcov --output-path lcov.info
+
+# Generates a local HTML coverage report in `target/llvm-cov/html`.
+coverage-html:
+	cargo llvm-cov nextest --workspace --features "$(TEST_FEATURES)" --html
 
 # Runs cargo-fmt (linter).
 cargo-fmt:


diegomrsantos:
@claude review this PR

diegomrsantos:
Thanks @claude, addressed the actionable item in 22db53e by adding `unstable` to the coverage workflow push branches, so Codecov can get post-merge baseline uploads from the development branch.

I left the other notes as-is: coverage needs debug info, `workflow_dispatch` is useful for manual reruns, debug-mode coverage is the expected Rust setup, and `CARGO_TERM_COLOR` is harmless.

diegomrsantos:
One note on the duplicate test run from this coverage workflow: coverage cannot reuse the existing `debug-tests-ubuntu` result because the code has to be rebuilt and run with coverage instrumentation to produce `lcov.info`.

There are two reasonable Rust CI patterns here. The current PR uses the lower-risk one: keep coverage in a separate job, so the existing debug and release test jobs stay unchanged. That is easy to review, but it means the debug test suite also runs under coverage instrumentation.

The cleaner steady-state would be to fold coverage into `debug-tests-ubuntu`: keep `release-tests-ubuntu` as the normal release-mode signal, have the debug job run the coverage-instrumented nextest pass, upload `lcov.info` from that job, and remove the separate coverage workflow. That avoids running the debug tests twice, but it does mean the debug test signal is coverage-instrumented rather than a plain debug run.

My preference is to keep this PR as the simple first step unless reviewers want the broader CI refactor now. A follow-up can merge coverage into the debug test job once we agree that is the intended CI shape.

shane-moore:
@macladson @jimmygchen, could use some thoughts on this since you've been maintaining lh CI for some time. This PR would add `cargo-llvm-cov` + `Codecov` in a separate workflow to anchor. Has lh considered this and rejected it, or just not gotten to it? What would make it worth the CI minutes for you?

macladson:
We did make a decision a while ago to forgo coverage checks for Lighthouse, This was years ago so I forget the exact reasoning but I think the pushback was related to coverage mostly just providing a "false sense of security". Not every code path requires the same level of scrutiny/testing as others, especially in complex applications. 

We ended up trialing it in our libraries, (SSZ, etc) and tbh I don't think is has provided huge value, but it also hasn't hurt to have. I'm not anti-coverage and I would probably support a selective coverage CI check in Lighthouse, but this hasn't been something we have discussed in a long time.

diegomrsantos:
Thanks, that context helps. I agree with the concern if coverage is treated as a quality score or a substitute for judgment. That is not what I’m trying to introduce here.

The value I see is more limited and practical: coverage tells us which parts of Anchor are not exercised by automated tests at all. It does not tell us that covered code is correctly tested, and it should not imply that all code paths need the same level of scrutiny. It is just an early signal for test gaps and review discussion.

That matches how Google frames coverage in their [Code Coverage Best Practices](https://testing.googleblog.com/2020/08/code-coverage-best-practices.html): high coverage does not guarantee test quality, but low coverage means areas of the product are going untested, and the useful discussion is often around the lines that are not covered. Google’s [Code Coverage at Google](https://research.google/pubs/code-coverage-at-google/) paper also describes using coverage at review/change-set level to make the data actionable. Martin Fowler makes the same distinction in [Test Coverage](https://martinfowler.com/bliki/TestCoverage.html): coverage is useful for finding untested parts of a codebase, but weak as a numeric statement about test quality.

For this PR, I intentionally kept the signal non-blocking. Codecov upload does not fail CI, and there is no threshold gate. That should avoid the failure mode where teams chase a percentage. Once we have actual Anchor data, we can decide whether to keep it separate, make it selective, or fold it into `debug-tests-ubuntu` to avoid running the debug tests twice.

So I’d frame this PR as adding visibility, not adding a quality guarantee. If the signal turns out noisy or too expensive, we can tune the CI shape, but I think having the baseline is still a useful first step.

shane-moore (approved):
LGTM. Mac's "false sense of security" concern is fair as a general worry, but this PR seems to handle it via being advisory only and no merge gate. Happy to revisit the shape (folding into the debug job) once we have data

jimmygchen:
> Has lh considered this and rejected it, or just not gotten to it?

@shane-moore Both actually - we've considered adding coverage but the line/function coverage metric isn't really useful for us, as our test suite is dominated by integration/e2e tests. So right now we're seeing 80+% coverage just because we run full beacon chains in many tests, and  it just means 80% of the code is executed but doesn't tell us if we're making any right assertions anywhere.

We do want to understand how well our tests cover production code though. 
I think line coverage + [mutation testing](https://hackmd.io/@jimmygchen/B1w-ACdRZe) could give us a better understanding of our real coverage, but we haven't gotten there yet.

diegomrsantos:
That distinction makes sense. Line coverage still tells you what production code you execute at all, and low coverage is a genuine signal. But you're right that high global numbers don't mean behavior is well specified. In a system where full beacon chains run as tests, a lot of code gets hit transitively, so covered and well checked drift pretty far apart.

There are really two separate questions here: what code do we even reach, and would the tests actually fail if that code were wrong. Coverage answers the first reasonably well and the second only weakly. The second is much more about assertion quality than raw execution.

On mutation testing, I'd be cautious about making a full-suite run the default next step. Naive whole-suite mutation is expensive. It's probably better to scope it to changed code and filter unproductive mutants aggressively.

What seems more valuable as a first step is using coverage to spot unexecuted or newly changed code, carving critical logic into smaller units that are easier to observe in isolation, and adding focused tests with stronger assertions around that logic. Once you have that, mutation on those specific modules or on diffs gives you much more signal for the cost.

PR #999, opened 2026-05-07
feat: instruments proposer pre-QBFT fetch in randao_reveal step 

jnhsigmap:

This change is the first of 2 PRs addressing #920 by adding an instrumentation span and start/end checkpoints to the `randao_reveal` function in `validator_store`.

It fulfils the following objectives:
- Checkpoints explicitly when `randao_reveal` starts and ends.
- Provides an indication of the slot budget remaining at the conclusion of `randao_reveal` ahead of Lighthouse fetching a beacon node block prior to signing.
- Allows reliable quantification of the time spent between `randao_reveal` and `sign_block` while Lighthouse performs that task.

This is worth doing now because a span and shallow checkpoints were added in #967 to the body of the main proposer duty signing flow. This change is required to begin reporting actionable information for pre-QBFT timing and slot budget expenditure.

## Change Overview (Required)
Another span was added to `randao_reveal` that instruments the pre-sign flow prior to Lighthouse performing the block fetch from the beacon node.

Specifics:
- Constants for entry, completion, and failure added to `checkpoints` module where existing checkpoints are defined.
- Relevant failure reasons from `randao_reveal` added to `failure_reason` function `match` body (+4).
- Helper functions `get_cluster_size` and `determine_slot_elapsed_ms` created for DRY implementation.
- Checkpoints added at randao reveal entry and upon conclusion of signature collection from committee peers.
- Slot budget also calculated post-signature collection and added to the trace.
- Trivial replacement of `cluster.cluster_members.len();` in `sign_block` with `get_cluster_size` call.

The span `proposer_randao_reveal` tracks the following values:
- `cluster_size`
- `clock_slot`
- `slot_elapsed_ms`
- `signing_epoch`
- `failure_reason`
- `outcome`
- `validator_pubkey`
- `validator_index`

What intentionally did not change?
- No changes were made to the block sign flow in `sign_block` as this was deferred to a follow-up to keep this PR concentrated on pre-block fetch.
- Decided against instrumenting within `decide_abstract_block` where in the case we receive a full-block from a non-Anchor operator, the block is still converted to a blinded block before entering `decide_abstract_block`. Otherwise the `as_ssz_bytes` function contains some work that could be expensive and worth considering.

  - Similar in nature to #967 as traces added to `randao_reveal` function body with some light refactoring, however these are largely no-op and otherwise should not add any risk.
  - Not necessarily a trade-off, but no major refactoring or improvements are made here, just instrumentation and light cleanup.
  - By small PRs to `unstable`, which will eventually be tested for correctness when traces are generated.

## Validation (Required)

`cargo test --workspace`
`cargo clippy`
`cargo check`

It can be safely reverted. This would remove the trace and checkpoints, helper functions, and changes to instrumentation module.

Next steps in follow up PR are 
  Specifically, in sign_block the following additions are needed:
1. Add `slot_elapsed_ms` to the `proposer_sign_block` span declaration.
2. Record it at `DUTY_ENTRY` for _"time spent before block was available for signing"_ (i.e. randao + Lighthouse BN fetch).
3. Add `slot_remaining_ms` (or `pre_qbft_elapsed_ms`) at pre consensus handoff checkpoint to record the remaining budget at QBFT start.

### DIFF
--- a/anchor/validator_store/src/lib.rs
+++ b/anchor/validator_store/src/lib.rs
@@ -2,7 +2,6 @@ mod instrumentation;
 pub mod metadata_service;
 mod metrics;
 pub mod registration_service;
-
 use std::{
     collections::{HashMap, HashSet},
     fmt::Debug,
@@ -61,7 +60,7 @@ use tokio::{
     sync::{Barrier, RwLock, watch},
     time::{Instant, sleep},
 };
-use tracing::{Instrument, debug, error, info, info_span, trace, warn};
+use tracing::{Instrument, Span, debug, error, field, info, info_span, trace, warn};
 use types::{
     AbstractExecPayload, Address, AggregateAndProof, AggregateAndProofBase,
     AggregateAndProofElectra, Attestation, AttestationBase, AttestationElectra, BeaconBlock,
@@ -2249,14 +2248,14 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> ValidatorStore
     ) -> Result<Signature, Error> {
         let span = info_span!(
             "proposer_randao_reveal",
-            cluster_size = tracing::field::Empty,
-            clock_slot = tracing::field::Empty,
-            slot_elapsed_ms = tracing::field::Empty,
+            cluster_size = field::Empty,
+            clock_slot = field::Empty,
+            slot_elapsed_ms = field::Empty,
             signing_epoch = signing_epoch.as_u64(),
-            failure_reason = tracing::field::Empty,
-            outcome = tracing::field::Empty,
+            failure_reason = field::Empty,
+            outcome = field::Empty,
             validator_pubkey = %validator_pubkey,
-            validator_index = tracing::field::Empty,
+            validator_index = field::Empty,
         );
         let future = async {
             trace!(
@@ -2265,7 +2264,7 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> ValidatorStore
             );
             let result = async {
                 let clock_slot = self.slot_clock.now().ok_or(SpecificError::SlotClock)?;
-                tracing::Span::current().record("clock_slot", clock_slot.as_u64());
+                Span::current().record("clock_slot", clock_slot.as_u64());
 
                 let domain_hash = self.get_domain(signing_epoch, Domain::Randao);
                 let signing_root = signing_epoch.signing_root(domain_hash);
@@ -2273,11 +2272,11 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> ValidatorStore
                 let (validator, cluster) = self.get_validator_and_cluster(validator_pubkey)?;
 
                 if let Some(validator_idx) = validator.index {
-                    tracing::Span::current().record("validator_index", *validator_idx);
+                    Span::current().record("validator_index", *validator_idx);
                 }
 
                 let cluster_size = cluster.cluster_members.len();
-                tracing::Span::current().record("cluster_size", cluster_size);
+                Span::current().record("cluster_size", cluster_size);
 
                 self.collect_signature(
                     PartialSignatureKind::RandaoPartialSig,
@@ -2294,13 +2293,13 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> ValidatorStore
 
             match determine_slot_elapsed_ms(&self.slot_clock) {
                 Some(ms) => {
-                    tracing::Span::current().record("slot_elapsed_ms", ms);
+                    Span::current().record("slot_elapsed_ms", ms);
                 }
                 None => trace!("slot_elapsed_ms unavailable: clock returned None"),
             }
 
             let outcome = instrumentation::outcome_from_result(&result);
-            tracing::Span::current().record("outcome", outcome);
+            Span::current().record("outcome", outcome);
             match &result {
                 Ok(_) => trace!(
                     checkpoint = instrumentation::checkpoints::RANDAO_REVEAL_COMPLETED,
@@ -2313,7 +2312,7 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> ValidatorStore
                         outcome = &outcome,
                         failure_reason = &failure_reason
                     );
-                    tracing::Span::current().record("failure_reason", failure_reason);
+                    Span::current().record("failure_reason", failure_reason);
                 }
             }
 
@@ -2381,14 +2380,14 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> ValidatorStore
         let span = info_span!(
             "proposer_sign_block",
             block_type,
-            cluster_size = tracing::field::Empty,
+            cluster_size = field::Empty,
             clock_slot = current_slot.as_u64(),
-            failure_reason = tracing::field::Empty,
-            proposal_matched = tracing::field::Empty,
-            outcome = tracing::field::Empty,
+            failure_reason = field::Empty,
+            proposal_matched = field::Empty,
+            outcome = field::Empty,
             block_slot = block_slot.as_u64(),
             validator_pubkey = %validator_pubkey,
-            validator_index = tracing::field::Empty,
+            validator_index = field::Empty,
         );
         let future = async {
             trace!(
@@ -2402,10 +2401,10 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> ValidatorStore
                 let (validator, cluster) = self.get_validator_and_cluster(validator_pubkey)?;
 
                 let cluster_size = cluster.cluster_members.len();
-                tracing::Span::current().record("cluster_size", cluster_size);
+                Span::current().record("cluster_size", cluster_size);
 
                 if let Some(validator_idx) = validator.index {
-                    tracing::Span::current().record("validator_index", *validator_idx);
+                    Span::current().record("validator_index", *validator_idx);
                 }
 
                 let (blinded_block, local_full_block) = match block {
@@ -2457,8 +2456,7 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> ValidatorStore
 
                 let publish_decision =
                     select_publish_block(signed_block, &blinded_block, local_full_block);
-                tracing::Span::current()
-                    .record("proposal_matched", publish_decision.proposal_matched);
+                Span::current().record("proposal_matched", publish_decision.proposal_matched);
                 trace!(
                     checkpoint = instrumentation::checkpoints::PUBLISH_BLOCK,
                     publish_path = publish_decision.publish_path,
@@ -2470,7 +2468,7 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> ValidatorStore
             .await;
 
             let outcome = instrumentation::outcome_from_result(&result);
-            tracing::Span::current().record("outcome", outcome);
+            Span::current().record("outcome", outcome);
             match &result {
                 Ok(_) => trace!(
                     checkpoint = instrumentation::checkpoints::DUTY_COMPLETED,
@@ -2483,7 +2481,7 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> ValidatorStore
                         outcome = &outcome,
                         failure_reason = &failure_reason
                     );
-                    tracing::Span::current().record("failure_reason", failure_reason);
+                    Span::current().record("failure_reason", failure_reason);
                 }
             }
             result
--- a/anchor/validator_store/src/lib.rs
+++ b/anchor/validator_store/src/lib.rs
@@ -157,6 +157,13 @@ fn get_cluster_size(cluster: &Cluster) -> usize {
     cluster.cluster_members.len()
 }
 
+fn determine_slot_elapsed_ms(slot_clock: &impl SlotClock) -> Result<u128, SpecificError> {
+    let duration = slot_clock
+        .millis_from_current_slot_start()
+        .ok_or(SpecificError::SlotClock)?;
+    Ok(duration.as_millis())
+}
+
 pub struct AnchorValidatorStore<
     T: SlotClock + 'static,
     E: EthSpec,
@@ -2253,6 +2260,7 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> ValidatorStore
             clock_slot = clock_slot.as_u64(),
             slot_elapsed_ms = tracing::field::Empty,
             signing_epoch = signing_epoch.as_u64(),
+            failure_reason = tracing::field::Empty,
             outcome = tracing::field::Empty,
             validator_pubkey = %validator_pubkey,
             validator_index = tracing::field::Empty,
@@ -2288,6 +2296,10 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> ValidatorStore
             }
             .await;
 
+            if let Ok(millis_from_slot_start) = determine_slot_elapsed_ms(&self.slot_clock) {
+                tracing::Span::current().record("slot_elapsed_ms", millis_from_slot_start);
+            }
+
             let outcome = instrumentation::outcome_from_result(&result);
             tracing::Span::current().record("outcome", outcome);
             match &result {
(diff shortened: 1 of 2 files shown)


diegomrsantos:
This PR adds observability around `AnchorValidatorStore::randao_reveal`, including the new `proposer_randao_reveal` span, randao-specific checkpoints, failure reasons, and slot timing fields. I think the current implementation has a failure-path gap, and the underlying issue is the boundary of `run_and_update_metrics`, not only the placement of one line.

In the current diff, `randao_reveal` reads the slot clock before creating the span and before passing the future into `run_and_update_metrics`:

```rust
let clock_slot = self.slot_clock.now().ok_or(SpecificError::SlotClock)?;

let span = info_span!(...);

let future = async {
    ...
}
.instrument(span);

run_and_update_metrics(..., future).await
```

If `slot_clock.now()` returns `None`, the method returns immediately. That means this path skips the new `proposer_randao_reveal` span, skips `RANDAO_REVEAL_FAILED`, skips `failure_reason = "slot_clock"`, and also skips the existing randao signing metric update. On `unstable`, the same clock read happened inside the future passed to `run_and_update_metrics`, so the error still flowed through the shared accounting path.

The broader design issue is that `run_and_update_metrics(action: Future)` only observes the future it receives. Callers can still do fallible setup before constructing that future, and a normal Rust `?` there bypasses instrumentation and metrics. That makes the API fragile because the correct behavior depends on everyone remembering exactly where `?` is allowed.

A more robust shape would be to invert the control: the shared helper should create the observed boundary and run the fallible body itself. In other words, instead of passing an already-built future to `run_and_update_metrics`, pass a closure that contains all fallible work. The helper calls the closure, awaits the returned future, records checkpoints/metrics from the completed `Result`, and only then returns the `Result` to the caller.

For example, the helper could look roughly like this:

```rust
struct SigningObservation {
    metric: &'static LazyLock<validator_metrics::Result<IntCounterVec>>,
    entered_checkpoint: &'static str,
    completed_checkpoint: &'static str,
    failed_checkpoint: &'static str,
}

async fn observe_signing<T, Action, Fut, After>(
    span: tracing::Span,
    observation: SigningObservation,
    action: Action,
    after_action: After,
) -> Result<T, Error>
where
    Action: FnOnce() -> Fut,
    Fut: Future<Output = Result<T, Error>>,
    After: FnOnce(&Result<T, Error>),
{
    async move {
        trace!(checkpoint = observation.entered_checkpoint);

        let result = action().await;

        after_action(&result);
        record_signing_result(&observation, &result);
        update_signing_metrics(observation.metric, &result);

        result
    }
    .instrument(span)
    .await
}
```

Then `randao_reveal` can keep ordinary Rust error propagation inside the observed body:

```rust
let span = info_span!(
    "proposer_randao_reveal",
    cluster_size = tracing::field::Empty,
    clock_slot = tracing::field::Empty,
    slot_elapsed_ms = tracing::field::Empty,
    signing_epoch = signing_epoch.as_u64(),
    validator_pubkey = %validator_pubkey,
    validator_index = tracing::field::Empty,
);

observe_signing(
    span,
    SigningObservation {
        metric: &metrics::SIGNED_RANDAO_REVEALS_TOTAL,
        entered_checkpoint: RANDAO_REVEAL_ENTERED,
        completed_checkpoint: RANDAO_REVEAL_COMPLETED,
        failed_checkpoint: RANDAO_REVEAL_FAILED,
    },
    || async {
        let clock_slot = self.slot_clock.now().ok_or(SpecificError::SlotClock)?;
        tracing::Span::current().record("clock_slot", clock_slot.as_u64());

        let domain_hash = self.get_domain(signing_epoch, Domain::Randao);
        let signing_root = signing_epoch.signing_root(domain_hash);

        tracing::Span::current().record("cluster_size", cluster.cluster_members.len());
        if let Some(validator_idx) = validator.index {
            tracing::Span::current().record("validator_index", *validator_idx);
        }

        self.collect_signature(
            PartialSignatureKind::RandaoPartialSig,
            Role::Proposer,
            &validator,
            &cluster,
            signing_root,
            clock_slot,
        )
        .await
    },
    |_| {
        if let Ok(slot_elapsed_ms) = determine_slot_elapsed_ms(&self.slot_clock) {
            tracing::Span::current().record("slot_elapsed_ms", slot_elapsed_ms);
        }
    },
)
.await
```

The important property is that the fallible operation body becomes the observed boundary. `?` is safe inside the closure because it returns to `observe_signing`, not past it. The helper then records the completion/failure checkpoint, updates the signing metric, and returns the same result.

This would also let us retire or narrow `run_and_update_metrics`. The metric classification logic is useful, but the current “run this already-built future and then update metrics” API is what makes this class of bug easy to reintroduce.

jnhsigmap:
Nice spot! I like the approach, my only question would be should we correct the API error boundary here and merge this, and then create a separate issue to implement your approach in both `sign_block` and `randao_reveal`? I can see the bug doesn't otherwise exist in the proposer flow once this is fixed here.

There may be an opportunity to turn the nested closure structure into a single callable that wraps the `work` in metric collection and instrumentation while still guarding that boundary by not allowing fallible work to be implemented outside of it. 

I've just moved `self.slot_clock.now()` back inside `result`.

diegomrsantos:
Agreed. Moving `self.slot_clock.now()` back inside `result` fixes the concrete failure path in this PR, so I don't think this needs to block on the broader helper/API cleanup.

I opened #1008 for the follow-up: https://github.com/sigp/anchor/issues/1008. That issue covers making the observed signing boundary harder to bypass across both `randao_reveal` and `sign_block`, and narrowing or replacing `run_and_update_metrics` so future callers can use `?` naturally inside the signing body without accidentally skipping checkpoints or metrics.

diegomrsantos:
@claude review this PR

jnhsigmap:
> **Claude finished @diegomrsantos's task in 1m 20s** —— [View job](https://github.com/sigp/anchor/actions/runs/25563314591)
> 
> ### PR Review
> * [x]  Read changed files
> * [x]  Analyze changes vs `origin/unstable`
> * [x]  Post review feedback
> 
> Scope is tight and matches the PR description: a `proposer_randao_reveal` span, enter/complete/fail checkpoints, slot-elapsed measurement, and a few additional `failure_reason` mappings. Latest commit (`050f2e7`) correctly moves `self.slot_clock.now()` back inside the observed body, addressing the earlier failure-path gap — `SlotClock` errors now flow through `RANDAO_REVEAL_FAILED`, `failure_reason`, and `run_and_update_metrics` as intended. Deferring the broader helper/API cleanup to #1008 is the right call for keeping this PR reviewable.
> 
> ### Observations
> **1. `slot_elapsed_ms` recording swallows the failure reason on `SlotClock` error** — `anchor/validator_store/src/lib.rs:2300-2302`
> 
> ```rust
> if let Ok(millis_from_slot_start) = determine_slot_elapsed_ms(&self.slot_clock) {
>     tracing::Span::current().record("slot_elapsed_ms", millis_from_slot_start);
> }
> ```
> `determine_slot_elapsed_ms` returns `Err(SpecificError::SlotClock)` when `millis_from_current_slot_start()` yields `None`, and the `if let Ok(..)` discards that. In the (rare) case where the clock fails here but the inner `result` was `Ok`, `slot_elapsed_ms` is silently absent on the span with no indication why. Minor: consider at least a `trace!`/`debug!` when this falls through, or dropping the error wrapping since the caller can't differentiate.
> 
:white_check_mark: 
I've added a `slot_elapsed_ms` unavailable `trace!` call here.

> **2. `determine_slot_elapsed_ms` error type is effectively unused** — `anchor/validator_store/src/lib.rs:159-164`
> 
> The helper returns `Result<u128, SpecificError>`, but the only caller in this PR matches `Ok` and discards the error. Since the failure mode is `None` -> `SlotClock`, `Option<u128>` would express intent more precisely and avoid constructing an error value that's never propagated. Not blocking, but worth a small simplification if `sign_block` also adopts this helper.

Fair enough, I considered pushing back here, however Lighthouse also swallow the error information with the signature `millis_from_current_slot_start(&self) -> Option<Duration>`. Considering we don't act on the failure reason either and the `let clock_slot = self.slot_clock.now().ok_or(SpecificError::SlotClock)?;` would fail if the slot clock was not able to be read at the beginning of the future, perhaps swallowing it is fine here.

Given that, I think the only failure case 'ignored' here would be if the slot clock becomes unreadable during the function call.

> **3. `get_cluster_size` is a thin wrapper over `.len()`** — `anchor/validator_store/src/lib.rs:156-158`
> 
> ```rust
> fn get_cluster_size(cluster: &Cluster) -> usize {
>     cluster.cluster_members.len()
> }
> ```
Fair enough, I have removed it.

> The DRY argument is weak for a single-field `.len()` — two call sites don't really justify an abstraction, and it adds a level of indirection when reading the hot path. Inlining `cluster.cluster_members.len()` would be equally readable. This is a style call; defer to project taste.
> 
> **4. Checkpoint ordering on the success path** — `anchor/validator_store/src/lib.rs:2300-2310`
> 
> `slot_elapsed_ms` is recorded _before_ the `RANDAO_REVEAL_COMPLETED` trace emits. That's fine, but note the completion trace itself doesn't include `slot_elapsed_ms` as a field — it's only on the span. If dashboards/queries rely on log-line fields rather than span attributes, they'll miss it. Consider whether `slot_elapsed_ms` should be inlined into the `trace!` for completion/failure as well (it already feeds the span via `record`).
> 
No other `trace!` calls add metadata beyond what is relevant to the call site. I don't think this is necessary in this case. This can be queried from the span.

> **5. `failure_reason` additions** — `anchor/validator_store/src/instrumentation.rs:27-45`
> 
> The four new variants (`ArithError`, `DataTooLarge`, `KeyShareDecryptionFailed`, `ValidatorClusterMismatch`) are fine, but none of them are actually reachable from `randao_reveal`'s code path based on my reading (the inner body only calls `get_domain`, `signing_epoch.signing_root`, `get_validator_and_cluster`, and `collect_signature`). That's not a problem — the function is shared with `sign_block` — but it's worth confirming these variants are intended for the broader signing surface, not specifically this PR's flow, so reviewers don't expect them to fire here.

I believe they are
- `ArithError` and `KeyShareDecryptionFailed` through `collect_signature`.
- `ValidatorClusterMismatch` from `get_validator_and_cluster`.
- `DataTooLarge` accessible from `sign_block`.

> **6. `validator_index` recording vs `MissingIndex`** — `anchor/validator_store/src/lib.rs:2280-2282`
> 
> `validator.index` is recorded only when `Some`. If a validator is missing its index, `collect_signature` may eventually return `MissingIndex` and we'll see `failure_reason = "missing_index"` without `validator_index` on the span. That's the correct behavior — just flagging that the absence itself is meaningful signal when reading traces.
> 
> ### Nits
> * `signing_epoch = signing_epoch.as_u64()` is recorded eagerly on span construction (good), but `clock_slot` is `Empty` at construction and populated later inside the future. Mixed eager/lazy population is fine; make sure the PR description's listing of tracked fields matches the actual span declaration (it does).
> * The `warn!` for `RANDAO_REVEAL_FAILED` is probably appropriate given this is in the validator's critical path, but it will be noisy if the node is unsynced or a cluster is liquidated (both normal-ish states). Consider whether `info!` is more appropriate for expected failure modes, with `warn!` reserved for unexpected ones. This can be tuned later based on field experience.
> 
> ### Nothing blocking
> The fix to move `slot_clock.now()` inside `result` resolves the concrete correctness issue raised earlier. The remaining items are style/precision notes. Follow-up #1008 is the right venue for the larger `observe_signing`/`run_and_update_metrics` restructuring. • `feat/920/pre-qbft`

jnhsigmap:
@claude, please provide a follow up review

anchor/validator_store/src/lib.rs
```diff
@@ -2241,23 +2247,79 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> ValidatorStore
...
+
+            match determine_slot_elapsed_ms(&self.slot_clock) {
+                Some(ms) => {
+                    tracing::Span::current().record("slot_elapsed_ms", ms);
+                }
+                None => trace!("slot_elapsed_ms unavailable: clock returned None"),
+            }
+
+            let outcome = instrumentation::outcome_from_result(&result);
+            tracing::Span::current().record("outcome", outcome);
```
shane-moore: would probably import `Span` at the top since used in a few different places as opposed to fully qualified path each time

jnhsigmap:
> **Claude finished @jnhsigmap's task in 58s** —— [View job](https://github.com/sigp/anchor/actions/runs/25652244595)
> 
> ### Follow-up Review
> * [x]  Fetch latest changes and diff against `origin/unstable`
> * [x]  Read updated `validator_store/src/lib.rs` and `instrumentation.rs`
> * [x]  Verify author responses to prior review observations
> * [x]  Post follow-up review
> 
> Re-reviewed against `origin/unstable`. The three changes since the prior review are applied as described:
> 
> * `get_cluster_size` wrapper removed; `cluster.cluster_members.len()` is inlined at the call site (`anchor/validator_store/src/lib.rs:2279-2280`).
> * `determine_slot_elapsed_ms` now returns `Option<u128>` instead of `Result<u128, SpecificError>` (`anchor/validator_store/src/lib.rs:156-160`). The caller uses a `match` and falls through to a `trace!` when the clock is unavailable (`anchor/validator_store/src/lib.rs:2295-2300`).
> * Prior author responses on items 4–6 and the nits are reasonable; no further follow-up required in this PR.
> 
> ### Residual observations
> **1. `trace!` fall-through event is not attached to a checkpoint** — `anchor/validator_store/src/lib.rs:2299`
> 
> ```rust
> None => trace!("slot_elapsed_ms unavailable: clock returned None"),
> ```
> 
> Minor: every other `trace!`/`warn!` in the observed body carries a `checkpoint = ...` field, which is what downstream queries key off. The unavailability event is a free-text log only; it won't be discoverable by checkpoint-name filters. Since the span still records `slot_elapsed_ms` as `Empty` when this branch fires, the signal is already recoverable from absence-of-field, so this is a logging taste call rather than a correctness issue.
> 
> **2. `outcome` is recorded on the span twice for the `Err` path** — `anchor/validator_store/src/lib.rs:2303` and `anchor/validator_store/src/lib.rs:2316`
> 
> ```rust
> tracing::Span::current().record("outcome", outcome);        // line 2303
> ...
> Err(err) => {
>     let failure_reason = instrumentation::failure_reason(err);
>     warn!(
>         checkpoint = instrumentation::checkpoints::RANDAO_REVEAL_FAILED,
>         outcome = &outcome,
>         failure_reason = &failure_reason
>     );
>     tracing::Span::current().record("failure_reason", failure_reason);
> }
> ```
> 
> Not a bug — `record` on `outcome` happens once before the match, and the `warn!` event carries `outcome` as an event field (separate from span fields). Worth confirming the dashboard consumer expects `outcome` from the span attribute, not from the `warn!`/`trace!` event payload, so the two sources don't drift. Same note applies to `failure_reason` (event on the `warn!`, recorded on span after).
> 
> **3. Ordering: `slot_elapsed_ms` record happens before `outcome`/`failure_reason` are known** — `anchor/validator_store/src/lib.rs:2295-2316`
> 
> Reading top-to-bottom: elapsed-ms is recorded, then outcome, then failure_reason. Fine for the span (fields are set-once per span regardless of order), but note that the success `trace!(RANDAO_REVEAL_COMPLETED)` and failure `warn!(RANDAO_REVEAL_FAILED)` events do _not_ carry `slot_elapsed_ms` in their payloads. If you want the elapsed-ms to appear next to the completion checkpoint in flat log output (not just the span export), adding `slot_elapsed_ms = ?...` to those macros is cheap. The author pushed back on this in prior review ("can be queried from the span") — that stance is consistent if the tracing pipeline is span-aware. Flagging again only because it's the same question that's likely to come up for `sign_block` follow-up.
> 
> **4. `failure_reason` recording is skipped in early-panic/drop scenarios** — `anchor/validator_store/src/lib.rs:2261-2322`
> 
> The instrumentation body is an async block that awaits `collect_signature`. If the future is dropped (task cancellation, timeout outside the body), the `match &result` arm never executes and the span closes with `failure_reason = Empty` and `outcome = Empty`. That's the same behavior as `sign_block` today and consistent with the existing pattern, so it's not a regression — but it's exactly the class of failure mode #1008 is intended to address. No action for this PR.
> 
> ### Nothing blocking
> Changes since the previous review are focused, and the remaining notes are all cosmetic or deferred to the `observe_signing`/helper refactor in #1008. Looks good to merge once CI is green. • `feat/920/pre-qbft`

I'll just respond to these
1. The `slot_elapsed_ms unavailable` `trace!` call is not congruent with any particular checkpoint and does not warrant adding one.
2. I looked at this for a while, but came to the conclusion that it provides some utility. `outcome` is recorded in the trace event and the parent span to allow clear correlation of logs in a particular failure or success case. I vote to keep this.
3. Valid but providing again in the trace events will not carry any useful information as the difference will be in the order of milliseconds. No heavy operations between recording at the span level and again when determining `result`.
4. More of a refactor scope, can be looked at in #1008.

Nothing blocking here for me

anchor/validator_store/src/lib.rs
jnhsigmap: done :slightly_smiling_face: also did the same with `tracing::field`.

shane-moore (approved):
LGTM

- `slot_clock.now()` is back inside the observed `result` body, so the `SlotClock` failure path flows through `RANDAO_REVEAL_FAILED`, `failure_reason`, and `run_and_update_metrics` as intended
- Bot-review actionables are applied: `determine_slot_elapsed_ms` returns `Option`, `get_cluster_size` wrapper removed, `slot_elapsed_ms unavailable` trace added on the `None` branch, and `Span` / `field` imported at the top
- agree with deferring `run_and_update_metrics` to issue #1008

Issue #950, opened 2026-04-14
research: Anchor interoperability with ssv-dkg ceremonies


shane-moore:
## Problem

Anchor operators should be able to participate in DKG ceremonies alongside Go SSV operators using the existing [ssv-dkg](https://github.com/ssvlabs/ssv-dkg) tooling. Before building anything, we need to understand exactly what (if anything) needs to change in Anchor to make this work.

## Context

- DKG (Distributed Key Generation) lets a committee of operators collaboratively generate a BLS validator keypair without any single party seeing the full private key
- [ssv-dkg](https://github.com/ssvlabs/ssv-dkg) is a standalone process, completely separate from the ssv-node operator software. It communicates over HTTP, not P2P. The only coupling to the operator node is the RSA private key file on disk.
- A DKG ceremony can be initiated by a non-operator (the "initiator"). The chosen operators must each be running an `ssv-dkg` instance that has access to their RSA private key.
- After the ceremony, operators don't receive shares directly from DKG. The initiator gets `keyshares.json` and submits it to the SSV smart contract. Each operator's ssv-node then reads its encrypted share from the `ValidatorAdded` contract event and decrypts it with its RSA private key.

### Phase 1: Research ssv-dkg

Study the [ssv-dkg repo](https://github.com/ssvlabs/ssv-dkg) and [dkg-spec](https://github.com/ssvlabs/dkg-spec) to understand:

1. **Operator RSA key usage**: How does `ssv-dkg` running on an operator's machine load and use their RSA private key? What file format does it expect? What operations does it perform with it (signing, decryption, both)?
2. **Ceremony flow**: What HTTP endpoints does an operator's `ssv-dkg` instance expose? What messages are exchanged between initiator and operators, and between operators?
3. **Key format expectations**: What RSA key file format does `ssv-dkg` expect operators to provide? How does this compare to how Anchor stores operator keys today? (Start with `anchor/common/operator_key/`)

### Phase 2: Gap analysis

After understanding the above, determine what changes (if any) Anchor needs so that an operator running Anchor can also run `ssv-dkg` and participate in ceremonies with Go SSV operators. Consider:

- Key file format compatibility
- Any shared state or configuration between the operator node and the DKG process
- Whether Anchor needs code changes, or just documentation/operational guidance

### Phase 3: Implementation plan

Create a new issue with a concrete implementation plan based on findings. This should detail:

- What changes are needed in Anchor (if any)
- Estimated scope/complexity
- Any risks or open questions

## References

[...]

PR #1001, opened 2026-05-07
fix(eth): force WS reconnect when live block stream goes idle

### LINKED ISSUE #993: eth: live_sync silently stalls when EL WebSocket subscription goes stale (no timeout on stream.next())
## Summary

Anchor's SSV event sync silently halts after the EL WebSocket subscription goes stale. No error is raised, no log line is emitted, the binary stays alive, but `metadata.block_number` stops advancing. New on-chain events (e.g. `ValidatorAdded` for newly-registered validators) are never observed until Anchor is restarted.

Root cause is at [`anchor/eth/src/sync.rs:686`](https://github.com/sigp/anchor/blob/unstable/anchor/eth/src/sync.rs#L686):

```rust
while let Some(block_header) = stream.next().await {
    // ... process block
}
```

`stream.next().await` has no timeout. If the underlying WebSocket connection is silently dropped (proxy idles it out, NAT keepalive too lax, server-side stale connection), the stream neither yields a new item nor returns `None` — the await blocks indefinitely. The outer reconnect loop only runs after the inner `while` exits, which never happens.

`try_sync`'s top-level recovery loop in `sync()` only catches errors that propagate via `?`. A silent hang is invisible to it, so neither `troubleshoot_ws` nor a reconnect is ever attempted.

## Observed symptom (real reproduction during #950 e2e)

Anchor (commit on `unstable` branch) running on Hoodi against `wss://colo.sigp-dev.net/hoodi-ee/...`:

```
2026-04-25T05:33:54  DEBUG Processing new block  block_number=2686629  relevant_block=2686621
... (live_sync running normally for hours) ...
2026-04-25T05:38:59  ERROR Sync failed, attempting recovery  e=RpcError("Error fetching logs: ...")
2026-04-25T05:39:00  INFO  Starting historical sync
2026-04-25T05:39:01  INFO  Syncing all events  start_block=2686621 end_block=2686645
2026-04-25T05:39:14  DEBUG Processing new block  block_number=2686654  relevant_block=2686646
... (silence for ~2 days; binary stays alive but block sync emits nothing) ...
```

After the recovery hop, `live_sync` was clearly running again (the `Processing new block` line is from `live_sync`, not `historical_sync`). It then went silent for ~48 hours while:

- duty scheduler kept ticking
- BN connection kept reporting healthy
- `metadata.block_number` stayed frozen at `2686646`
- chain head advanced ~15,000 blocks past that
- a `ValidatorAdded` for our test validator at block `2701496` was never observed

Restarting Anchor immediately resumed sync (a new WS subscription is created in the outer `loop`). Switching to `stable` branch with a fresh data-dir resynced cleanly from genesis.

## Why the existing recovery doesn't catch this

In `live_sync()`:

```rust
loop {
    let mut stream = self.ws_client.subscribe_blocks().await
        .map_err(|e| ExecutionError::WsError(...))?
        .into_stream();
    info!("Successfully subscribed to block stream");

    while let Some(block_header) = stream.next().await {     // ← can block forever
        // ... process block
    }

    error!("WebSocket stream ended, reconnecting...");
}
```

If `subscribe_blocks` itself fails, the `?` propagates and `try_sync`'s outer loop catches it and runs `troubleshoot_ws`. But once a subscription succeeds and the stream silently goes idle, none of the recovery machinery runs.

## Proposed fix

Wrap `stream.next()` in a timeout that's well above the expected block interval (12s on Ethereum). On timeout, drop the stream and re-enter the outer `loop`, which already re-subscribes:

```rust
const STREAM_IDLE_TIMEOUT: Duration = Duration::from_secs(60);

loop {
    let mut stream = self.ws_client.subscribe_blocks().await?.into_stream();
    info!("Successfully subscribed to block stream");

    loop {
        match tokio::time::timeout(STREAM_IDLE_TIMEOUT, stream.next()).await {
            Ok(Some(block_header)) => { /* process */ }
            Ok(None) => break,                                    // stream ended cleanly
            Err(_) => {
                warn!(
                    timeout_s = STREAM_IDLE_TIMEOUT.as_secs(),
                    "WebSocket idle past threshold, forcing reconnect"
                );
                break;
            }
        }
    }

    error!("WebSocket stream ended or idled out, reconnecting...");
}
```

Optional follow-ups:

- Surface the timeout duration as a CLI / config flag so operators on slow chains can tune it.
- Add a metric counter for forced-reconnects so this stall mode is observable.
- Consider periodic heartbeat read (e.g. `eth_blockNumber` HTTP poll) as a liveness check on the WS connection — but the timeout-based fix is sufficient for the primary symptom.

## Reproduction

The natural failure was non-deterministic (we observed it once during a multi-day Hoodi run).

## Severity / scope

- **Branch:** observed on `unstable`. 
- **Impact:** any operator whose EL WebSocket connection goes stale (any flaky proxy / NAT / load balancer between Anchor and the EL) will silently miss `OperatorAdded`, `ValidatorAdded`, `ValidatorRemoved`, etc. — meaning they fall out of sync with the SSV contract until the binary is restarted. New validators assigned to them won't be served.
- **Not destructive** — restart fully recovers (sync resumes from `metadata.block_number`).

## Context

Encountered while running the mixed-client interop e2e for #950 — see [the linked comment](https://github.com/sigp/anchor/issues/950#issuecomment-4383863318) for the broader experiment, of which this stall was an incidental finding.

### LINKED ISSUE #1004: refactor(eth): consolidate `EXECUTION_SYNC_STATUS` gauge writes in sync recovery handler
## Motivation

The observability gauge `anchor_execution_sync_status` (declared at `anchor/eth/src/metrics.rs:6-11`, "0 = down, 1 = up") is meant to reflect whether the SSV event syncer is actively consuming blocks. Today it does not — it stays at `1` during recovery on multiple error paths.

## Current state — gauge writers (`upstream/unstable`)

Three writers, none in the recovery handler:

- `anchor/eth/src/sync.rs:155` — `set_gauge(0)` in `SsvEventSyncer::new()` (one-shot at construction).
- `anchor/eth/src/sync.rs:670` — `set_gauge(1)` at top of `live_sync()` (every entry).
- `anchor/eth/src/sync.rs:747` — `set_gauge(0)` after the inner stream loop, only reached when `stream.next()` returns `None` (clean stream end).

## Current state — `live_sync` exit paths

| # | Path | Reaches `set_gauge(0)`? | Recovery |
|---|------|---|---|
| a | `subscribe_blocks().await?` failure | **No** — `?` skips :747 | `sync()` matches `WsError`, runs `troubleshoot_ws` |
| b | `Ok(None)` clean stream end | **Yes** | re-subscribes within `live_sync` |
| c | Inner-loop `?`-propagation: `fetch_logs?`, `set_block_timestamps?`, `process_logs?` | **No** | `sync()` matches variant, runs corresponding troubleshoot |
| d | (After #1001) Idle-timeout `Err(WsError)` | **No** | `sync()` matches `WsError`, runs `troubleshoot_ws` |

Only path (b) resets the gauge. Paths (a), (c), (d) leave the gauge at `1` for the entire recovery window (`troubleshoot_*` retries with backoff up to 30s, plus `historical_sync` replay on success).

`sync()`'s recovery handler at sync.rs:262-277 correctly flips the *functional* signal `is_synced.send_replace(false)` at line 240, but never touches the gauge. The functional signal is consumed by `message_receiver`, `message_sender`, `validator_store`, and `client/notifier` — those services correctly see "not synced" during recovery. Only the Prometheus gauge lies.

## Proposed change

Make `sync()`'s recovery handler the single source of truth for the gauge:

```rust
// anchor/eth/src/sync.rs around line 240
Err(e) => {
    error!(?e, "Sync failed, attempting recovery");
    self.is_synced.send_replace(false);
    metrics::set_gauge(&metrics::EXECUTION_SYNC_STATUS, 0);  // NEW
    match e { ... }
}
```

Then drop the scattered writes:

- Remove `set_gauge(0)` at sync.rs:155 — gauge defaults to 0 anyway.
- Remove `set_gauge(0)` at sync.rs:747 — `sync()` now owns the reset on error exits, and the `Ok(None)` path falls through to the next `subscribe_blocks` attempt within the same `live_sync` invocation.
- Keep `set_gauge(1)` at sync.rs:670 — "we successfully entered live mode" flip, still correct.

## Scope

1. Add `set_gauge(EXECUTION_SYNC_STATUS, 0)` in `sync()`'s `Err(e) =>` arm at sync.rs:262.
2. Remove redundant `set_gauge(0)` at sync.rs:155.
3. Remove redundant `set_gauge(0)` at sync.rs:747.
4. Verify `keysplit_sync()` (calls `historical_sync` directly) still produces sensible gauge values.

## Acceptance criteria

- [ ] `EXECUTION_SYNC_STATUS` is `0` whenever recovery is in flight.
- [ ] `EXECUTION_SYNC_STATUS` is set to `1` exactly once per successful `live_sync` entry.
- [ ] Scattered `set_gauge` calls outside the recovery handler are removed.
- [ ] No behavioural change to `is_synced` watch channel propagation.

petarjuki7:

- `live_sync` awaits `stream.next()` on the EL WebSocket subscription with no timeout. If the underlying connection silently goes stale (proxy idles it out, NAT keepalive too lax, server-side stale connection), the stream neither yields a new item nor returns `None`, and the await blocks indefinitely. SSV event sync stops advancing, no error is logged, and `sync()`'s recovery handler never runs.
- Reproduced live during the mixed-client e2e for #950: Anchor sat at the same `metadata.block_number` for ~48h on Hoodi and never observed the on-chain `ValidatorAdded` for our test validator. Restarting Anchor immediately resumed sync.
- Static analysis trace + reproduction details: #993.
- Closes #993.
- Closes #1004.

## Change Overview

- Extract `next_block_or_timeout` helper that wraps `stream.next()` in `tokio::time::timeout(STREAM_IDLE_TIMEOUT, ...)`.
- New constant `STREAM_IDLE_TIMEOUT = 60s` (5x Ethereum slot time, generous enough to ride out transient slowness, tight enough to catch a stale subscription within ~1 minute).
- `live_sync` calls the helper. On `Ok(Some(header))` processes the block (existing logic unchanged). On `Ok(None)` (clean stream end) breaks to re-subscribe on the same `ws_client`, preserves existing behavior. On `Err(_)` (idle timeout) bumps `EXECUTION_CONNECTION_ERRORS["ws_idle_timeout"]` and returns `WsError`, which routes through `sync()`'s existing `troubleshoot_ws` recovery for a full client rebuild with backoff.

- **Risk:** spurious reconnects on healthy connections during chains with > 60s effective inter-block periods. Mitigation: 60s is 5x Ethereum's 12s slot, well past any realistic transient slowness; if needed, the constant is trivial to lift to a CLI/config flag in a follow-up.
- **Trade-off:** extra reconnect adds a brief gap during which live events aren't observed; on resume, `historical_sync` is invoked first and replays missed blocks before live mode resumes. Net behavior is "eventually consistent within seconds" rather than "permanently stalled."
- **Blast radius:** scoped to `live_sync` recovery path; existing `historical_sync`, event-processing, and DB code paths are untouched.

## Validation

- Three new unit tests in `live_sync_tests` covering idle, yielding, and ended streams. Use `tokio::test(start_paused = true)` virtual-time so a 60s+ wait runs in sub-millisecond wall time. The idle test wraps the helper call in an outer `timeout` guard so a future regression panics with a clear message instead of hanging the suite.
- Verified the regression test fails meaningfully when the helper is reverted to `Ok(stream.next().await)` (panic with `timeout guard missing?` rather than infinite hang).
- `cargo test -p eth --lib`, 19/19 pass (debug + release).
- `cargo clippy -p eth --all-targets --no-deps -- -D warnings`, clean.
- `cargo fmt -p eth -- --check`, clean.
- The original failure mode was reproduced on Hoodi during the #950 e2e; this fix would have surfaced as a clean reconnect rather than a 48h silent stall.

## Rollback

Revert this commit. Pure code change, no migrations, no config impact, no DB schema change. Operators on the prior behavior would just regain the silent-stall vulnerability, a pure functional regression.

- The `STREAM_IDLE_TIMEOUT` value could later be exposed as a CLI/config flag if operators want to tune it for slower chains; not done here per "simple solutions first."
- The `ws_idle_timeout` metric label is intentionally distinct from `troubleshoot_ws`'s `websocket` label so dashboards can attribute reconnects to silent-stall versus subscribe-time failures.

### DIFF
--- a/anchor/eth/src/sync.rs
+++ b/anchor/eth/src/sync.rs
@@ -179,8 +179,6 @@ impl SsvEventSyncer {
         );
         debug!("Created event processor - done");
 
-        metrics::set_gauge(&metrics::EXECUTION_SYNC_STATUS, 0);
-
         Ok(Self {
             rpc_client,
             ws_client,
@@ -265,6 +263,7 @@ impl SsvEventSyncer {
                 Err(e) => {
                     error!(?e, "Sync failed, attempting recovery");
                     self.is_synced.send_replace(false);
+                    metrics::set_gauge(&metrics::EXECUTION_SYNC_STATUS, 0);
 
                     match e {
                         ExecutionError::WsError(e) => {
@@ -773,7 +772,6 @@ impl SsvEventSyncer {
 
             // If we get here, the stream ended (likely due to disconnect)
             error!("WebSocket stream ended, reconnecting...");
-            metrics::set_gauge(&metrics::EXECUTION_SYNC_STATUS, 0);
         }
     }
 }
--- a/anchor/eth/src/sync.rs
+++ b/anchor/eth/src/sync.rs
@@ -85,19 +85,22 @@ const FOLLOW_DISTANCE: u64 = 8;
 // before assuming the WebSocket has gone stale and forcing a full reconnect.
 const STREAM_IDLE_TIMEOUT: Duration = Duration::from_secs(60);
 
-// Awaits the next item from `stream`, returning `Err(WsError)` if no item arrives
-// within `STREAM_IDLE_TIMEOUT`. Caller is responsible for any metric updates.
+// Sole producer of the idle-timeout `WsError` so a unit test can pin the variant.
+fn record_idle_timeout() -> ExecutionError {
+    metrics::inc_counter_vec(&metrics::EXECUTION_CONNECTION_ERRORS, &["ws_idle_timeout"]);
+    ExecutionError::WsError(format!(
+        "WebSocket idle for more than {}s, forcing full reconnect",
+        STREAM_IDLE_TIMEOUT.as_secs()
+    ))
+}
+
 async fn next_block_or_timeout<S>(stream: &mut S) -> Result<Option<Header>, ExecutionError>
 where
     S: Stream<Item = Header> + Unpin,
 {
-    match timeout(STREAM_IDLE_TIMEOUT, stream.next()).await {
-        Ok(item) => Ok(item),
-        Err(_) => Err(ExecutionError::WsError(format!(
-            "WebSocket idle for more than {}s, forcing full reconnect",
-            STREAM_IDLE_TIMEOUT.as_secs()
-        ))),
-    }
+    timeout(STREAM_IDLE_TIMEOUT, stream.next())
+        .await
+        .map_err(|_| record_idle_timeout())
 }
 
 // Connection timeout duration
@@ -707,16 +710,9 @@ impl SsvEventSyncer {
             info!("Successfully subscribed to block stream");
 
             loop {
-                let block_header = match next_block_or_timeout(&mut stream).await {
-                    Ok(Some(header)) => header,
-                    Ok(None) => break,
-                    Err(e) => {
-                        metrics::inc_counter_vec(
-                            &metrics::EXECUTION_CONNECTION_ERRORS,
-                            &["ws_idle_timeout"],
-                        );
-                        return Err(e);
-                    }
+                let block_header = match next_block_or_timeout(&mut stream).await? {
+                    Some(header) => header,
+                    None => break,
                 };
 
                 // Guard against integer underflow when calculating relevant_block.
--- a/anchor/eth/src/sync.rs
+++ b/anchor/eth/src/sync.rs
@@ -8,16 +8,21 @@ use alloy::{
     eips::BlockNumberOrTag,
     primitives::Address,
     providers::{Provider, ProviderBuilder, RootProvider, WsConnect},
-    rpc::types::{Filter, Log, SyncStatus},
+    rpc::types::{Filter, Header, Log, SyncStatus},
     sol_types::SolEvent,
     transports::{RpcError, TransportErrorKind},
 };
 use database::{NetworkDatabase, SlashingProtection};
-use futures::{FutureExt, StreamExt, stream::FuturesOrdered};
+use futures::{FutureExt, Stream, StreamExt, stream::FuturesOrdered};
 use reqwest::Url;
 use sensitive_url::SensitiveUrl;
 use ssv_network_config::SsvNetworkConfig;
-use tokio::{select, sync::watch, task::spawn_blocking, time::Duration};
+use tokio::{
+    select,
+    sync::watch,
+    task::spawn_blocking,
+    time::{Duration, timeout},
+};
 use tracing::{debug, error, info, instrument, trace, warn};
 
 use crate::{
@@ -75,6 +80,25 @@ const MAX_BACKOFF_MS: u64 = 30_000; // Don't wait longer than 30 seconds
 
 // Block follow distance
 const FOLLOW_DISTANCE: u64 = 8;
+
+// Maximum time to wait for the next block on the live `subscribe_blocks` stream
+// before assuming the WebSocket has gone stale and forcing a full reconnect.
+const STREAM_IDLE_TIMEOUT: Duration = Duration::from_secs(60);
+
+// Awaits the next item from `stream`, returning `Err(WsError)` if no item arrives
+// within `STREAM_IDLE_TIMEOUT`. Caller is responsible for any metric updates.
+async fn next_block_or_timeout<S>(stream: &mut S) -> Result<Option<Header>, ExecutionError>
+where
+    S: Stream<Item = Header> + Unpin,
+{
+    match timeout(STREAM_IDLE_TIMEOUT, stream.next()).await {
+        Ok(item) => Ok(item),
+        Err(_) => Err(ExecutionError::WsError(format!(
+            "WebSocket idle for more than {}s, forcing full reconnect",
+            STREAM_IDLE_TIMEOUT.as_secs()
+        ))),
+    }
+}
 
 // Connection timeout duration
 pub const CONNECT_TIMEOUT: u64 = 10;
@@ -682,8 +706,19 @@ impl SsvEventSyncer {
 
             info!("Successfully subscribed to block stream");
 
-            // If we have a connection, continuously stream in blocks
-            while let Some(block_header) = stream.next().await {
+            loop {
+                let block_header = match next_block_or_timeout(&mut stream).await {
+                    Ok(Some(header)) => header,
+                    Ok(None) => break,
+                    Err(e) => {
+                        metrics::inc_counter_vec(
+                            &metrics::EXECUTION_CONNECTION_ERRORS,
+                            &["ws_idle_timeout"],
+                        );
+                        return Err(e);
+                    }
+                };
+
                 // Guard against integer underflow when calculating relevant_block.
                 if block_header.number < self.network.ssv_contract_block + FOLLOW_DISTANCE {
                     warn!(
@@ -695,11 +730,9 @@ impl SsvEventSyncer {
                     continue;
                 }
 
-                // Block we are interested in is the current block number - follow distance
                 let relevant_block = block_header.number - FOLLOW_DISTANCE;
 
-                // If the relevant block was already processed, do not process it again. This can
-                // happen if `block_header.number` was seen before due to a reorg.
+                // Skip blocks already processed (e.g. due to a reorg replaying numbers).
                 let last_processed_block =
                     self.event_processor.db.state().get_last_processed_block();
                 if relevant_block <= last_processed_block {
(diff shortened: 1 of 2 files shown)


shane-moore:
@claude, please take a look

anchor/eth/src/sync.rs
```diff
@@ -682,8 +706,19 @@ impl SsvEventSyncer {
...
 
-            // If we have a connection, continuously stream in blocks
-            while let Some(block_header) = stream.next().await {
+            loop {
+                let block_header = match next_block_or_timeout(&mut stream).await {
+                    Ok(Some(header)) => header,
+                    Ok(None) => break,
+                    Err(e) => {
+                        metrics::inc_counter_vec(
+                            &metrics::EXECUTION_CONNECTION_ERRORS,
```
shane-moore: maybe add comment:
```
Err(e) => {
    // Increments alongside the existing `["websocket"]` label that
    // troubleshoot_ws will emit — distinguishes silent-stall recoveries
    // from subscribe-time WS failures on dashboards.
    metrics::inc_counter_vec(
        &metrics::EXECUTION_CONNECTION_ERRORS,
        &["ws_idle_timeout"],
    );
    return Err(e);
}
```

shane-moore:
Looking again at test #1, the variant is **already** pinned by the existing `match` cascade — if the helper returned `Err(SyncError(...))` instead of `Err(WsError(...))`, the first arm wouldn't match and `Ok(other) => panic!(...)` would catch it. So the "lift to `assert!(matches!(...))`" reformulation doesn't add coverage.

The regression that actually escapes is a wrap inside `live_sync`'s `Err` arm itself, e.g.

```rust
Err(e) => return Err(ExecutionError::SyncError(format!("stream stalled: {e}"))),
```

That routes through `sync()`'s `_ => {}` at `sync.rs:251`, `troubleshoot_ws` never runs, and we silently never recover. The helper's tests can't catch this because they don't drive `live_sync`.

A small extraction makes the typed contract testable:

```rust
// near next_block_or_timeout
fn record_idle_timeout(err: ExecutionError) -> ExecutionError {
    metrics::inc_counter_vec(
        &metrics::EXECUTION_CONNECTION_ERRORS,
        &["ws_idle_timeout"],
    );
    err
}
```

Call site becomes:

```rust
Err(e) => return Err(record_idle_timeout(e)),
```

Plus one unit test:

```rust
#[test]
fn record_idle_timeout_preserves_ws_error_variant() {
    let input = ExecutionError::WsError("idle".into());
    let output = record_idle_timeout(input);
    assert!(
        matches!(output, ExecutionError::WsError(_)),
        "must return WsError so sync() routes to troubleshoot_ws; got {output:?}",
    );
}
```

diegomrsantos:
I think #1004 should be handled in PR #1001.

PR #1001 adds a new recoverable `WsError` path from inside `live_sync()`, and that path skips the current `EXECUTION_SYNC_STATUS = 0` write that only runs when the stream ends cleanly. Since `sync()` is already the common recovery point, it should own the gauge reset before matching on the error:

```rust
Err(e) => {
    error!(?e, "Sync failed, attempting recovery");
    self.is_synced.send_replace(false);
    metrics::set_gauge(&metrics::EXECUTION_SYNC_STATUS, 0);

    match e {
        ...
    }
}
```

Then the scattered reset writes can be removed where they are no longer the source of truth. That keeps the metric invariant simple: entering live sync marks execution sync up, and entering recovery marks it down.

petarjuki7:
Thanks for the catch, makes sense. Pushed a11ba2c6.

Went with a slightly different shape than your sketch: a producer (`fn() -> ExecutionError`) rather than a pass-through. The pass-through only guards changes inside the helper, if a future refactor inlines the wrap at the call site (`Err(e) => return Err(SyncError(...))`), the helper test still passes. The producer makes the variant live in exactly one place; regressing it requires either touching the producer (caught) or visibly deleting the `record_idle_timeout()` call from the site, which is loud in a diff.

`next_block_or_timeout` is now wired via `.map_err(|_| record_idle_timeout())`, so the existing `_returns_ws_error_when_stream_is_idle` test transitively pins the producer's variant, a wrong variant would fall through to the `Ok(other) => panic!(...)` arm. Skipped a separate direct test for that reason, happy to add one if you'd prefer.

diegomrsantos:
Thanks, `record_idle_timeout()` handles the `WsError` variant concern, but the `EXECUTION_SYNC_STATUS` issue from #1004 still looks open on the latest head.

The idle-timeout path now propagates through `next_block_or_timeout(...).await?`, so `live_sync()` still exits before the clean stream-end `set_gauge(..., 0)` runs. `sync()` then flips `is_synced` to false and enters recovery, but the Prometheus gauge can remain `1` during that recovery window.

Can we move the gauge reset into `sync()`'s recovery handler and remove the reset writes that are no longer the source of truth? That would keep the metric invariant consistent: live sync sets it up, recovery sets it down.

petarjuki7:
Fixed in 8d08fcc9. `sync()`'s recovery handler is now the single writer of `set_gauge(0)` for `EXECUTION_SYNC_STATUS`, removed the construction-time and clean-stream-end writes per the issue. Also merged latest unstable.

diegomrsantos (approved):
Thanks!

PR #1002, opened 2026-05-07
fix: prioritize urgent processor work

diegomrsantos:

The processor scheduler could select ready permitless work before ready urgent consensus work, even when a worker permit was immediately available. That made the permitless lane capable of delaying work that is supposed to have higher priority.

The regression coverage now captures the main priority case and the two permitless boundary cases: permitless-only work with capacity available, and permitless-only work while worker permits are saturated.

Relevant context: this PR fixes the scheduler selection order in the processor only. It does not redesign queue capacity, expiry, or network backpressure.

## Change Overview (Required)

- Adds focused processor tests for urgent consensus priority and permitless fallback behavior.
- Changes the outer receiver selection so the processor first acquires an available worker permit, then lets the existing permit-aware selector prefer urgent consensus.
- Keeps permitless fallback behavior for the case where no worker permit is available.
- Documents the transient permit acquisition when only permitless work is ready, including the assumption that the processor has a single receiver loop.

Reviewer reading order: start with the tests in `anchor/processor/src/receivers.rs`, then read the small selection-order change in the same file.

The main behavior change is scheduling priority when a worker permit is available. Urgent consensus now wins over permitless in that case. Permitless work still runs without a worker permit when all permits are occupied, so the existing bypass behavior is preserved for saturated worker pools.

When capacity exists but only permitless work is ready, the scheduler briefly acquires and releases a permit before returning permitless work. That is harmless with the current single receiver loop and is now documented near the selection logic.

## Validation (Required)

- Format check passed.
- Targeted processor lint check passed.
- Processor tests passed, including the new regression and boundary tests plus the existing max-worker test.
- The repo commit hook passed its formatter, clippy, and dependency sort checks before committing.

Revert the commits in this PR. There is no config, data, or migration impact.

N/A

This only fixes the immediate priority inversion. Separate follow-up work is still needed for broader overload policy, such as bounded per-key mailboxes, TTLs for stale work, and reserved capacity for critical classes.

### DIFF
--- a/anchor/processor/src/receivers.rs
+++ b/anchor/processor/src/receivers.rs
@@ -79,3 +79,60 @@ impl Receivers {
         })
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use std::sync::Arc;
+
+    use tokio::sync::{Semaphore, mpsc};
+
+    use super::{Receiver, Receivers};
+    use crate::{QueueKind, work::WorkItem};
+
+    const PERMITLESS_WORK_NAME: &str = "permitless_test_work";
+    const URGENT_WORK_NAME: &str = "urgent_consensus_test_work";
+
+    #[tokio::test]
+    async fn test_next_work_item_prioritizes_urgent_consensus_over_ready_permitless() {
+        // Arrange: both queues are ready, and urgent consensus has worker capacity available.
+        let (permitless_tx, permitless_rx) = mpsc::channel(1);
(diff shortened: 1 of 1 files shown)


diegomrsantos:
@claude review this PR

shane-moore (approved):
LGTM. The fix matches the inner select's already-documented intent (urgent wins when capacity allows), test 1 is a true regression test, and the transient-permit comment correctly scopes the single-loop assumption

Issue #986, opened 2026-04-30
research: end-to-end eager-attesting gap (head events not driving sign_attestations)


shane-moore:
## Issue Addressed

Follow-up to #806 — the unresolved part of comment-3857920948 (https://github.com/sigp/anchor/issues/806#issuecomment-3857920948). On a closer look the gap is broader than that comment described, so filing it end-to-end.

This is a research ticket for someone to confirm the gap, validate feasibility, and either spin out a follow-up implementation issue/PR or document why we shouldn't do it.

## Background

Lighthouse 8.1.0 ([sigp/lighthouse#7892](https://github.com/sigp/lighthouse/pull/7892)) introduced "earlier attestation broadcasting": the validator client's `AttestationService` races a beacon-node SSE `head` event against the 1/3-slot timer and calls `sign_attestations` on whichever fires first. Anchor's Lighthouse pin (`4b3a9d3`) includes this code, but on Anchor it's compiled in and dormant.

## Gap

There are three coupled reasons eager attesting doesn't reach the wire on Anchor today.

First, Anchor never opts into the LH head-monitor wiring. In `anchor/client/src/lib.rs`, `BeaconNodeFallback` is constructed without `set_head_send(...)` and `AttestationServiceBuilder` without `head_monitor_rx(...)`. Both default to `None` upstream, which means the SSE polling task is never spawned and the eager-attest race in `validator_services/src/attestation_service.rs` short-circuits to its timer-only branch. PR #813 was a clean dependency bump and didn't pick up the small wiring patch LH 8.1.0 added in its own VC binary.

Second, MetadataService's Phase 2 — the step that produces `VotingContext` from a fresh `attestation_data` fetch — runs on a fixed 1/3-slot timer in `anchor/validator_store/src/metadata_service.rs`. There is no head-event subscription anywhere in the metadata service.

Third, the two are intertwined. Even if the LH side were wired, Anchor's `sign_committee_attestations` in `anchor/validator_store/src/lib.rs` blocks on `get_voting_context(slot)` before any signing work, and that watch channel is filled only by Phase 2 at t ≈ 4 s. Fixing the LH side alone, or the MetadataService side alone, doesn't move attestation publishing earlier.

## Target behaviour

[...]

PR #1011, opened 2026-05-12
feat(client): wire LH beacon head monitor

petarjuki7:
First in a series of PRs addressing #986 (end-to-end eager-attesting gap). This one only wires the LH 8.1.0 beacon head-monitor; follow-ups will cover MetadataService Phase 2 and any related plumbing.

LH 8.1.0 ships an SSE-driven eager-attest path (`set_head_send` + `AttestationServiceBuilder::head_monitor_rx`) that Anchor pulls in via the `4b3a9d3` pin but never opts into. The SSE polling task is gated on `head_monitor_send.is_some()` in `start_fallback_updater_service` and is never spawned today. See #986 for the full gap analysis.

## Change Overview

- New `--disable-beacon-head-monitor` CLI flag, mirroring LH (default: enabled).
- New `Config::enable_beacon_head_monitor` (defaults to `true`).
- In `Client::run`, after `set_slot_clock` and before `start_fallback_updater_service`, conditionally build the `mpsc::channel::<HeadEvent>`, call `beacon_nodes.set_head_send(...)` on the primary `BeaconNodeFallback` only, and pass `Option<Mutex<Receiver>>` into `AttestationServiceBuilder::head_monitor_rx`.
- Wiring structure mirrors LH's `validator_client/src/lib.rs:402-411 + 529`.

- End-to-end eager attesting is still inert: `sign_committee_attestations` blocks on `get_voting_context`, which is filled by MetadataService Phase 2 on a fixed 1/3-slot timer. So this PR alone does not change attestation publish timing.
- New SSE stream per BN candidate (additional `/eth/v1/events?topics=head` subscription). Behind a flag and matches LH's default.

## Validation

- `make cargo-fmt-check`, `make lint` clean.
- `cargo test --release --workspace --exclude spec_tests` passes (spec_tests skip is a pre-existing missing-fixture issue, unrelated).

## Rollback

Revert the commit or pass `--disable-beacon-head-monitor` at runtime.

Follow-up PR(s) will race MetadataService Phase 2 against head events so `VotingContext` is published early and `sign_committee_attestations` can actually start sooner.

### DIFF
--- a/anchor/cli/src/cli.rs
+++ b/anchor/cli/src/cli.rs
@@ -537,6 +537,17 @@ pub struct Node {
 
     #[clap(
         long,
+        help = "Disable the beacon head monitor which tries to attest as soon as any of the \
+                configured beacon nodes sends a head event. Leaving the service enabled is \
+                recommended, but disabling it can lead to reduced bandwidth and more predictable \
+                usage of the primary beacon node (rather than the fastest BN).",
+        display_order = 0,
+        help_heading = FLAG_HEADER
+    )]
+    pub disable_beacon_head_monitor: bool,
+
+    #[clap(
+        long,
         help = "Disable slashing protection for all validator clients. DO NOT ENABLE THIS UNLESS YOU HAVE A MORE THAN SUFFICIENT REASON TO",
         hide = true,
         display_order = 0
--- a/anchor/cli/src/cli.rs
+++ b/anchor/cli/src/cli.rs
@@ -540,9 +540,7 @@ pub struct Node {
         help = "Disable the beacon head monitor which tries to attest as soon as any of the \
                 configured beacon nodes sends a head event. Leaving the service enabled is \
                 recommended, but disabling it can lead to reduced bandwidth and more predictable \
-                usage of the primary beacon node (rather than the fastest BN).",
-        display_order = 0,
-        help_heading = FLAG_HEADER
+                usage of the primary beacon node (rather than the fastest BN)."
     )]
     pub disable_beacon_head_monitor: bool,
 
--- a/anchor/client/src/config.rs
+++ b/anchor/client/src/config.rs
@@ -71,6 +71,8 @@ pub struct Config {
     pub prefer_builder_proposals: bool,
     /// Controls whether the latency measurement service is enabled
     pub disable_latency_measurement_service: bool,
+    /// Enables the beacon head monitor that reacts to head updates from connected beacon nodes.
+    pub enable_beacon_head_monitor: bool,
     /// Enable operator doppelgänger protection (blocks messages and monitors for twins)
     pub operator_dg: bool,
     /// Number of epochs to monitor for twins after grace period
@@ -124,6 +126,7 @@ impl Config {
             prefer_builder_proposals: false,
             gas_limit: 36_000_000,
             disable_latency_measurement_service: false,
+            enable_beacon_head_monitor: true,
             operator_dg: false,
             operator_dg_wait_epochs: 2,
             strict_mfp: false,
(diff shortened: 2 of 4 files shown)


shane-moore:
CI is failing because `FLAG_HEADER` was removed from `cli.rs` in 624ad33 ("adds next_help_heading, removes flag header"), which landed on `unstable` after this branch was cut. Headings moved from per-field `help_heading = FLAG_HEADER` to struct-level `#[clap(next_help_heading = "...")]`.

Fix: drop the `help_heading = FLAG_HEADER` line on the new `disable_beacon_head_monitor` field and also the `display_order = 0`. It'll inherit the "Payload Building Options" heading via the last `flatten`'d group on `Node`, matching the sibling `disable_latency_measurement_service`.

shane-moore:
@claude, please take a look. ignore the ci failures as i already left a comment for that on the pr

anchor/cli/src/cli.rs
```diff
@@ -535,6 +535,17 @@ pub struct Node {
...
+    #[clap(
+        long,
+        help = "Disable the beacon head monitor which tries to attest as soon as any of the \
+                configured beacon nodes sends a head event. Leaving the service enabled is \
+                recommended, but disabling it can lead to reduced bandwidth and more predictable \
+                usage of the primary beacon node (rather than the fastest BN).",
+        display_order = 0,
+        help_heading = FLAG_HEADER
+    )]
+    pub disable_beacon_head_monitor: bool,
```
shane-moore, line 545: `docs/docs/generated/cli-node-options.mdx` doesn't include the new
`--disable-beacon-head-monitor` flag. After the `FLAG_HEADER` fix lets
the workspace compile, run `make cli-reference-update` and commit the
diff.

#1020 adds `cli-reference-check` as a merge-gate workflow. Once it
lands, any push to #1011 with stale generated docs will block the
merge queue, so easier to regenerate now in the same PR

shane-moore:
@claude, thanks for the review. Responses inline:

**1.** Already covered by my [earlier comment](https://github.com/sigp/anchor/pull/1011#issuecomment-4499702286).

**2.** Prefer to keep `enable_beacon_head_monitor` on the config side to match LH. The PR mirrors LH's wiring end-to-end and LH's config field is also `enable_beacon_head_monitor` (positive form, see `validator_client/src/lib.rs:404`). Also, `config.rs:203` already uses the same `enable_X = !cli_args.disable_X` flip pattern (`upnp_enabled = !disable_upnp`), so the polarity isn't a new convention in this file.

**3.** Agreed, we should add a comment noting `MAX_HEAD_EVENT_QUEUE_LEN` mirrors LH's value at `validator_client/src/lib.rs:75`.

**4.** Agreed, we should add a comment noting the `Mutex` is required by `AttestationServiceBuilder::head_monitor_rx`'s signature.

**6.** Leaning skip. We don't log feature-flag state at startup for any other flag (`disable_latency_measurement_service`, `operator_dg`, `with_weighted_attestation_data`, `enable_high_validator_count_metrics`, etc.), so adding it here would be inconsistent and create a precedent to apply uniformly or not at all. The SSE subscription is observable on the BN side and the flag is visible via `--help` and startup args. Open to revisit as part of a broader "log feature-flag state" change.

**7.** `anchor/client/` and `anchor/cli/` have no `#[test]` functions or test infrastructure today. Singling out this one polarity flip for a test when the existing `upnp_enabled = !disable_upnp` flip at `config.rs:203` is untested would be arbitrary. Out of scope; if test infrastructure for the CLI/config layer lands later, both flips become candidates.

shane-moore:
lgtm! just pushed a commit to add comments and rebase unstable. for pr 1041, you'll probably want to cherry pick your new commits on top of this pr's commits since the rebase will have changed the history of this branch. then, force push to 1041.

shane-moore (approved):

PR #1012, opened 2026-05-13
feat: instruments proposer block sign duty pre-QBFT fetch

jnhsigmap:

A follow up to #999 addressing the remaining components of #920:
1. Adds `slot_elapsed_ms` to the `proposer_sign_block` span declaration.
2. Records it at `DUTY_ENTRY` for _"time spent before block was available for signing"_ (i.e. randao + Lighthouse BN fetch).
3. Adds `slot_elapsed_ms` at pre consensus handoff checkpoint to record the remaining budget at QBFT start.

Additionally, the current `match` statement determining recording of `slot_elapsed_ms` in `randao_reveal` (from #999) had scope for improvement as `.record` of `None` results in a no-op and `slot_elapsed_ms` remains `empty`. This enabled a more concise expression than the existing verbose `match` statement: 
```
match determine_slot_elapsed_ms(&self.slot_clock) {
    Some(ms) => {
        Span::current().record("slot_elapsed_ms", ms);
     }
     None => trace!("slot_elapsed_ms unavailable: clock returned None"),
}
```

becomes
```
Span::current().record(
    "slot_elapsed_ms",
    determine_slot_elapsed_ms(&self.slot_clock),
);
```
This is worth doing now to follow from instrumenting `randao_reveal` and complete the stated objective of #920 to instrument the pre-QBFT proposer path and quantify slot budget spent prior to QBFT starting.

## Change Overview (Required)

Specific changes:
- `slot_elapsed_ms` added to `sign_block` span.
- Added `slot_elapsed_ms` field to existing traces at duty entry and pre-consensus handoff checkpoints.
- Reported `slot_elapsed_ms` in the final span prior to the sign block future returning.

No new failure modes or behavioural changes were made as `determine_slot_elapsed_ms` returns `Option<u128>` and no side effects from the application's perspective. There were also no distinct trade offs made as this change was narrow and specific.

## Validation (Required)

`make cargo-fmt`
`cargo clippy`
`cargo check`

This change can be safely reverted without any impacts.

### DIFF
--- a/anchor/validator_store/src/lib.rs
+++ b/anchor/validator_store/src/lib.rs
@@ -2385,15 +2385,19 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> ValidatorStore
             failure_reason = field::Empty,
             proposal_matched = field::Empty,
             outcome = field::Empty,
+            slot_elapsed_ms = field::Empty,
             block_slot = block_slot.as_u64(),
             validator_pubkey = %validator_pubkey,
             validator_index = field::Empty,
         );
+
         let future = async {
             trace!(
                 checkpoint = instrumentation::checkpoints::DUTY_ENTRY,
+                slot_elapsed_ms = determine_slot_elapsed_ms(&self.slot_clock),
                 "Proposer block signing duty entered"
             );
+
             let result = async {
                 if !*self.is_synced.borrow() {
                     return Err(Error::SpecificError(SpecificError::NotSynced));
@@ -2420,6 +2424,7 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> ValidatorStore
 
                 trace!(
                     checkpoint = instrumentation::checkpoints::PRE_CONSENSUS_HANDOFF,
+                    slot_elapsed_ms = determine_slot_elapsed_ms(&self.slot_clock),
                     "Handing block to consensus process"
                 );
 
@@ -2466,6 +2471,13 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> ValidatorStore
                 Ok(publish_decision.signed_block)
             }
             .await;
+
+            match determine_slot_elapsed_ms(&self.slot_clock) {
+                Some(ms) => {
+                    Span::current().record("slot_elapsed_ms", ms);
+                }
+                None => trace!("slot_elapsed_ms unavailable: clock returned None"),
+            }
 
             let outcome = instrumentation::outcome_from_result(&result);
             Span::current().record("outcome", outcome);
--- a/anchor/validator_store/src/lib.rs
+++ b/anchor/validator_store/src/lib.rs
@@ -152,10 +152,10 @@ async fn run_committee_signing<T>(
     }
 }
 
-fn determine_slot_elapsed_ms(slot_clock: &impl SlotClock) -> Option<u128> {
+fn determine_slot_elapsed_ms(slot_clock: &impl SlotClock) -> Option<u64> {
     slot_clock
         .millis_from_current_slot_start()
-        .map(|d| d.as_millis())
+        .map(|d| d.as_millis() as u64)
 }
 
 pub struct AnchorValidatorStore<
@@ -2291,13 +2291,10 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> ValidatorStore
             }
             .await;
 
-            match determine_slot_elapsed_ms(&self.slot_clock) {
-                Some(ms) => {
-                    Span::current().record("slot_elapsed_ms", ms);
-                }
-                None => trace!("slot_elapsed_ms unavailable: clock returned None"),
-            }
-
+            Span::current().record(
+                "slot_elapsed_ms",
+                determine_slot_elapsed_ms(&self.slot_clock),
+            );
             let outcome = instrumentation::outcome_from_result(&result);
             Span::current().record("outcome", outcome);
             match &result {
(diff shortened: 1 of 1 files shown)


jnhsigmap:
@claude provide a review for this PR - also validate and provide confirmation that the following is true:

>   When `Option<T>` implements `Value`, it records the inner value if `Some`, and does nothing if `None` — the field simply won't appear in the event output. It doesn't render as `Some(1234)` or `None`. It either emits the bare `1234` or omits the field entirely.
> 
>   So the current code on line 2425 in anchor/validator_store/src/lib.rs:
> ```
>   trace!(
>       checkpoint = instrumentation::checkpoints::DUTY_ENTRY,
>       slot_elapsed_ms = determine_slot_elapsed_ms(&self.slot_clock),
>       "Proposer block signing duty entered"
>   );
> ```
>   Already produces:
>   - `slot_elapsed_ms = 1234` when the clock succeeds
>   - Field absent when the clock fails
>

jnhsigmap:
> **Claude finished @jnhsigmap's task in 2m 57s** —— [View job](https://github.com/sigp/anchor/actions/runs/25776310691)
> ### Review feedback
> Overall this is a clean, narrow instrumentation change that mirrors the precedent set by #999 for `randao_reveal` (same pattern at `anchor/validator_store/src/lib.rs:2253`/`:2294-2298`). Approving the _intent_; a couple of small observations below.
> 
> #### 1. Nit — the `Span::current().record(...)` final update could match the event-site pattern (or vice versa) — at `anchor/validator_store/src/lib.rs:2475-2480`
> ```rust
> match determine_slot_elapsed_ms(&self.slot_clock) {
>     Some(ms) => {
>         Span::current().record("slot_elapsed_ms", ms);
>     }
>     None => trace!("slot_elapsed_ms unavailable: clock returned None"),
> }
> ```
> 
> Because `Option<u128>: Value`, this could equivalently be:
> 
> ```rust
> Span::current().record("slot_elapsed_ms", determine_slot_elapsed_ms(&self.slot_clock));
> ```
> 
> `Span::record` on `None` is a no-op — the previously declared `field::Empty` slot simply stays empty. The current form is justified only by the extra `trace!("slot_elapsed_ms unavailable: clock returned None")` diagnostic on the `None` arm. That diagnostic is **reasonable** (and consistent with the randao_reveal block at `:2294-2298`), so I would keep it for symmetry. Just flagging that the asymmetry between "event sites use `slot_elapsed_ms = determine_slot_elapsed_ms(...)` directly" and "the final span record goes through a `match`" is deliberate, not missed simplification. No change requested.

Fair enough I agree here, it can be inferred that if the span does not have this field populated the call failed as this is occurring in only one place. It's unlikely to occur anyway. I have updated this and the call in `randao_reveal`.

> #### 2. Observation — slot rollover across the signing future
> `determine_slot_elapsed_ms` returns `millis_from_current_slot_start()` relative to **whatever slot the clock is in at the moment of the call**, not relative to `current_slot` / `block_slot` passed into `sign_block`. If the future runs long enough to cross a slot boundary (e.g. QBFT stall, BN stall before this span wraps), the final `slot_elapsed_ms` recorded on the span at `:2475` will be elapsed-into-the-_next_-slot, which can be small while the _actual_ wall-clock time from duty entry was large.
> 
> This matches how the existing `randao_reveal` instrumentation behaves, so it is consistent — but it means `slot_elapsed_ms` on the span is not a monotonically-increasing "budget consumed" number across the full lifetime of the duty. Callers interpreting this metric should be aware. If the downstream intent (per #920) is "how much of the slot budget was spent before QBFT", then the `DUTY_ENTRY` and `PRE_CONSENSUS_HANDOFF` event fields are the authoritative ones and the final span field is most useful as "elapsed-in-current-slot at duty completion", not "total duration". Worth a one-line doc comment near `determine_slot_elapsed_ms` or at the span declaration if this interpretation matters operationally.

I don't think this is worth accommodating as the `millis_from_current_slot_start` call is doing exactly its purpose of returning the duration of the current slot that has occurred, and it is known to users. A proposer duty that crosses a slot boundary is a failure anyway.

> #### 3. Nit — `slot_elapsed_ms` type is `u128`
> `Duration::as_millis` returns `u128`, which the fmt subscriber handles fine, but `u128` is awkward for some export paths (e.g. OpenTelemetry/Prometheus integrations that type-narrow numeric fields). A slot's worth of milliseconds fits comfortably in `u64` (even `u32`). Consider `.try_into().ok()` or `u64::try_from(d.as_millis()).ok()` inside `determine_slot_elapsed_ms` to future-proof against exporters that don't handle `u128` cleanly. Not a blocker.

I have addressed this as a forward-looking fix on the premise that the `Visit` trait in `tracing-core` falls back to `record_debug` and returns the value as a debug string if `record_u128` is not overridden by the collector. `record_u64` is implemented by OpenTelemetry and will not suffer from this side effect. `Option<u64>` delegates to `record_u64` and the conversion bears no material cost.

> #### 4. Minor — no test coverage for the new fields
> Per `.claude/rules/verification.md`, behavioral claims want evidence. The PR description notes `cargo check` / `cargo clippy` / `cargo fmt` were run, but no tests exercising the instrumentation. A small unit test using a `tracing-subscriber::fmt::TestWriter` or `tracing-test` to assert the field is present on `Some` and absent on `None` would lock in the `Option<u128>: Value` behavior against future tracing-core changes. Arguably overkill for a pure-observability change, but worth considering if this pattern is going to keep spreading (randao_reveal, sign_block, and presumably further duty types in the future).

Tests are not required at this stage.

> #### 5. No concerns on
> * Correctness of span/field declarations (`:2388` declares `slot_elapsed_ms = field::Empty`, which is required for the later `Span::current().record("slot_elapsed_ms", ms)` to actually emit — good).
> * No new failure modes; `determine_slot_elapsed_ms` is total.
> * No side effects, matching the PR description.
> * Consistency with the randao_reveal instrumentation precedent.
> 
> ### Summary
> The `Option<T>: Value` claim is **correct** for the subscribers in this workspace: `Some(1234)` emits as `1234`, `None` omits the field. The PR is a clean, low-risk instrumentation addition consistent with #999. Comments above are observations / nits, not blockers. · Branch: `feat/920/sign-block`

jnhsigmap:
@claude provide a subsequent review

jnhsigmap:
> 1. as u64 cast — safe in this domain, but worth a brief mention

We're protected by what is logically implemented here right? `millis_from_current_slot_start` returns as it says, which means it is bounded and we do not need to be concerned with overflow?

diegomrsantos:
One small cleanup for the PR description: it still says this adds `slot_remaining_ms` at the pre-consensus handoff checkpoint, but the diff records `slot_elapsed_ms` there instead.

I do not think the code needs to change for that. Recording elapsed milliseconds at handoff is enough to compute remaining slot budget from the slot duration. Please update the description so reviewers do not go looking for a separate `slot_remaining_ms` field.

shane-moore (approved):
lgtm!

PR #1021, opened 2026-05-15
fix(operator_key): accept go-ssv keystores with diacritic passwords

petarjuki7:

An operator migrating from go-ssv to Anchor reported `Failed to start Anchor: Key decryption failed`. Their password contained the Spanish character `ñ`; other operators with ASCII-only passwords migrated successfully.

Root cause: go-ssv pins `wealdtech/go-eth2-wallet-encryptor-keystorev4 v1.1.3`, whose `normPassphrase` NFKD-decomposes the password but keeps only the leading rune of each decomposition segment, dropping combining marks. For any password whose NFKD decomposition produces combining marks (`ñ`, `ü`, `é`, `ç`, …), the bytes fed into the KDF diverge from spec-compliant EIP-2335 NFKD. Anchor uses Lighthouse's `eth2_keystore`, which is spec-correct, so it cannot decrypt those keystores.

Upstream `wealdtech` acknowledged the bug in v1.4.1 by adding an alt-normalization fallback (spec-correct first, alt-norm second). go-ssv has not bumped past v1.1.3.

## Change Overview

- `EncryptedKey::decrypt` retries once with an alt-normalized password on `Error::InvalidPassword`. Other errors are surfaced as before.
- New private helper `alt_normalize_password` reproduces wealdtech v1.1.3's algorithm: NFKD then filter to canonical combining class 0.
- Encryption path is unchanged; new keystores produced by Anchor remain EIP-2335 compliant.
- Adds `unicode-normalization` as a direct workspace dep (already present transitively via `eth2_keystore`, same `0.1.25` version).

- **Surface**: only the decrypt path, only on `InvalidPassword`, single bounded retry.
- **Trade-off**: the fallback is intentionally lenient on decrypt to support migration. Encryption stays spec-compliant, so Anchor never produces non-spec keystores.
- **Secret handling**: the alt-normalized password lives in `Zeroizing<String>`; no intermediate allocations outside the wrapper.

## Validation

- Unit tests for `alt_normalize_password` covering ASCII, `ñ`, `café`.
- End-to-end fallback tests: simulate go-ssv keystores by encrypting with the alt-normalized form, then decrypting with the original NFC password (`café`, `piñataüber`).
- Negative test: a truly wrong password still returns `DecryptionError::Keystore(eth2_keystore::Error::InvalidPassword)`.
- `cargo test -p operator_key` — 46 pass, 0 fail.
- `make cargo-fmt-check`, `make lint` — clean.

## Rollback

Revert this commit. No config, data, or on-disk format change. Keystores produced before or after this PR remain readable both ways.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5550,6 +5550,7 @@ dependencies = [
  "serde_json",
  "tempfile",
  "thiserror 2.0.18",
+ "unicode-normalization",
  "zeroize",
 ]
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -172,6 +172,7 @@ tracing-subscriber = { version = "0.3.18", features = ["fmt", "env-filter"] }
 tree_hash = "0.12.0"
 tree_hash_derive = "0.12.0"
 typenum = "1.18"
+unicode-normalization = "0.1.25"
 zeroize = "1.8.1"
 
 [patch.crates-io]
--- a/anchor/common/operator_key/Cargo.toml
+++ b/anchor/common/operator_key/Cargo.toml
@@ -13,6 +13,7 @@ rpassword = { workspace = true }
 serde = { workspace = true }
 serde_json = { workspace = true }
 thiserror = { workspace = true }
+unicode-normalization = { workspace = true }
 zeroize = { workspace = true }
 
 [dev-dependencies]
--- a/anchor/common/operator_key/src/encrypted.rs
+++ b/anchor/common/operator_key/src/encrypted.rs
@@ -46,6 +46,7 @@ use eth2_keystore::{
 use openssl::{pkey::Private, rsa::Rsa};
 use serde::{Deserialize, Serialize};
 use thiserror::Error;
+use unicode_normalization::{UnicodeNormalization, char::canonical_combining_class};
 use zeroize::Zeroizing;
 
 use crate::{ConversionError, public};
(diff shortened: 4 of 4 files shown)


shane-moore:
@claude, please review

shane-moore:
@claude, walked through wealdtech v1.1.3 `norm.go` and eth2_keystore's pre-processing to validate these. Bottom line: the doc comment is slightly off, but the fallback is correct for every realistic input.

**1. Wealdtech equivalence.** `normPassphrase` does more than drop combining marks. It iterates NFKD segments, keeps the starter, and runs single-byte starters through `stripChars` (0x00–0x1F + 0x7F + 0x80–0x9F). The C1 entries are unreachable due to the `len(buf) == 1` guard, so effectively it strips C0 + DEL only. `alt_normalize_password` doesn't strip controls at all.

The practical convergence comes from `eth2_keystore::decrypt`, which strips all control chars via `char::is_control()` on both attempts (`keystore.rs:340,375`). Tracing a wealdtech-encrypted `"ca\x09fé"`: wealdtech stored `PBKDF2(b"cafe")`. Anchor primary normalizes to `"cafe\u{0301}"` and fails. Fallback alt-normalizes to `"ca\x09fe"`, eth2_keystore strips the tab, PBKDF2 input is `"cafe"`. Matches.

The only break is C1-control + combining-mark in one password (essentially impossible). So: tighten the helper doc to say "drops combining marks; control-char handling is left to eth2_keystore's spec filter on both attempts" and link the wealdtech source. So, we can tighten that up

**2. Sibling negative test.** It would only fire on a PBKDF2 collision. Skipping.

**3. Decrypt-only asymmetry in doc comment.** Fair, we should do that

**4-6.** Observations, no action.

shane-moore (approved):
lgtm! great job diagnosing this issue and improving operator UX via this fix

PR #1024, opened 2026-05-15
feat: fork-aware attestation deadline at Gloas

shane-moore:

Under Gloas, the unaggregated attestation deadline tightens from 1/3 slot (4000ms mainnet) to 25% slot (3000ms mainnet). Anchor's two QBFT/voting-context deadlines currently hardcode `slot_duration / 3`, so post-Gloas they would consume the full pre-Gloas window and miss the on-time threshold every slot.

## Change Overview

Two call sites switched to LH's fork-aware `ChainSpec::get_attestation_due::<E>(slot)`:

- `lib.rs:1483` (QBFT instance start in `sign_committee_attestations`): slot is given, direct call.
- `metadata_service.rs:219` (Phase 2 `VotingContext` wakeup): looks up the deadline for the *next* slot (`current_slot + 1`) so the tighter Gloas deadline applies from the first CStar slot rather than one slot late.

Unchanged: aggregate, sync-message, and contribution deadlines. LH has no `_gloas` variant for those

- Boundary precision via `current_slot + 1` in metadata_service. Without the `+ 1`, the first Gloas slot would see Phase 2 wake 999ms late (pre-Gloas deadline used while sleeping into a Gloas slot). Comment at the call site explains.

## Validation

- `cargo test -p anchor_validator_store --lib`: 33 tests pass, including the new boundary-anchoring test `attestation_due_switches_at_gloas_boundary`.
- `make cargo-fmt-check` clean.
- `make lint` clean.
- The boundary test acts as a tripwire: any future LH bump that changes the gating logic or default field values fails this test rather than silently shipping.

## Rollback

Revert the commit. Two files, no schema or state changes, no operational impact.

## Blockers / Dependencies
N/A

### DIFF
--- a/anchor/validator_store/src/lib.rs
+++ b/anchor/validator_store/src/lib.rs
@@ -1480,7 +1480,7 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> AnchorValidator
         let timer = metrics::start_timer_vec(&metrics::CONSENSUS_TIMES, &[metrics::BEACON_VOTE]);
         let timeout_mode = TimeoutMode::SlotTime {
             instance_start_time: self
-                .get_instant_in_slot(slot, self.spec.get_slot_duration() / 3)?,
+                .get_instant_in_slot(slot, self.spec.get_attestation_due::<E>(slot))?,
         };
 
         let completed = self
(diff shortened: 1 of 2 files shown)


shane-moore:
@claude, please take a look

shane-moore:
@claude, thanks for the review. Replying inline:

1. **lib.rs LGTM.** No action.

2. **Slot-clock read race.** Pre-existing pattern in the loop, considered during the local /simplify pass. Window is microseconds and self-corrects on the next iteration; you self-label it not a blocker. Skipping.

3. **Test magnitude assertions (4s/3s).** Deliberately not asserted. The current form asserts field equality (`pre == spec.unaggregated_attestation_due`, `post == spec.unaggregated_attestation_due_gloas`), which tests the gate. Coupling to specific magnitudes would fail on any legitimate LH tuning of mainnet defaults, which is the wrong direction for a tripwire. Skipping.

4. **Cross-check spec for other Gloas-tightened windows.** Already verified via LH source: only `unaggregated_attestation_due` has a `_gloas` field in `ChainSpec` at pinned rev `1a68631`. Spec-vs-LH divergence is an LH-tracking concern (M6.4), not this PR. Skipping.

jnhsigmap (approved):

jnhsigmap (commented):
looks good :slightly_smiling_face:

diegomrsantos:
### Title framing nit: `refactor:` vs `fix:`/`feat:`

`refactor:` reads as behavior-preserving, but this change shortens the unaggregated-attestation deadline by ~1s/slot at the two patched call sites once the Gloas fork activates. That's a real behavior flip at the boundary, gated by `ChainSpec::get_attestation_due::<E>(slot)`.

Suggest renaming to one of:
- `fix: use fork-aware attestation deadline at Gloas` — frames it as forward-compat correctness (pre-Gloas would otherwise miss the on-time threshold every slot post-fork)
- `feat: fork-aware attestation deadline at Gloas` — if you'd rather frame Gloas-readiness as a feature

CI accepts `refactor:`, so this is non-blocking — purely about future readers scanning history not being misled into thinking it's a no-op.

PR #1030, opened 2026-05-18
chore: add cstar example to built-in fork schedule configs

shane-moore:

Built-in `ssv_fork_schedule.yaml` files already document a commented `boole:` example but no `cstar:` equivalent.

## Change Overview

Adds a commented `cstar:` block mirroring the existing `boole:` example in all three built-in network configs (mainnet, holesky, hoodi). Comments only; no operational behavior changes.

N/A. YAML parser ignores comments; runtime behavior unchanged.

## Validation

- `cargo test -p ssv_network_config` (14 passed)
- `make cargo-fmt-check` and `make lint` clean

## Rollback

Revert the commit; no operational impact.

### DIFF
--- a/anchor/common/ssv_network_config/built_in_network_configs/holesky/ssv_fork_schedule.yaml
+++ b/anchor/common/ssv_network_config/built_in_network_configs/holesky/ssv_fork_schedule.yaml
@@ -6,3 +6,6 @@
 # boole:
 #   epoch: 12500
 #   domain_type: "00000002"
+# cstar:
+#   epoch: 13000
+#   domain_type: "00000003"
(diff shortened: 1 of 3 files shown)


shane-moore:
@claude, please take a look

shane-moore:
@claude, thanks for the review. On the optional follow-up to extract the example into a shared docs location: we'll leave it inline per file.

YAML has no include/ref mechanism, so the alternative is a comment pointing operators to external docs, which adds friction for someone who expects to see the format right where they're editing. Drift risk on inert comments is effectively zero, and when production values eventually land each network gets its own real epoch and domain values, so the three files are forecasted to diverge anyway, not deduplicate further.

jnhsigmap (approved):
looks good, given that the `ForkScheduleFile` is empty right now given the commented sections, this doesn't need optimization right now.

Issue #1032, opened 2026-05-18
feat: add Role::PTCCommittee with message-validator fork-gate


shane-moore:
## Goal

Add `Role::PTCCommittee` (wire byte `[7, 0, 0, 0]`, matches SIP-94 `BNRolePTCAttester = 7`) in `anchor/common/ssv_types/src/msgid.rs`. Bind the new role to `PartialSignatureKind::PostConsensus` (no new partial-sig kind variant). Wire the message-validator fork-gate, TTL, duty limit, partial-sig kind binding, and role-specific bounds so the workspace compiles. Insert a one-line `//todo(epbs)` stub in `qbft_manager` that the PTC routing issue replaces with real instance plumbing.

## Context

Per SIP-94 §3 ([ssvlabs/SIPs#94](https://github.com/ssvlabs/SIPs/pull/94)), PTC is a new committee-scoped duty added at the Gloas / EIP-7732 / ePBS fork. The wire surface needs the new role byte before any other PTC work can land. Adding `Role::PTCCommittee` to the `Role` enum breaks exhaustive matches in `message_validator` (`committee_info` dispatch, `message_lateness`, `duty_limit`, partial-sig kind binding, role bounds) and `qbft_manager` (committee-role dispatch) — those all live in this issue so the workspace stays green.

The new role is wire-incompatible with pre-CStar operators: pre-CStar nodes decode the byte as `Err(NoMatchingVariant)`. The message-validator gate is a defense-in-depth check that also rejects CStar-capable nodes if they receive a PTC message before their local `active_fork >= Fork::CStar` (e.g., near epoch boundary timing).

PTC is committee-scoped: one QBFT instance per cluster per slot batches all locally-assigned PTC validators. Routing is by `(Role, CommitteeId)` not `(Role, ValidatorPubKey)`, mirroring `Role::Committee` and `Role::AggregatorCommittee`. Per SIP-94 SC-2, the partial-sig kind is `PostConsensus` (PTC signs decided `PayloadAttestationData` after QBFT, structurally identical to `Role::Committee`'s post-consensus path); no new `PTCAttester` kind is introduced.

## Suggested approach

`ssv_types`:
[...]

PR #1033, opened 2026-05-18
feat: add Role::PTCCommittee with message-validator fork-gate

shane-moore:

- SIP-94 §3 ([ssvlabs/SIPs#94](https://github.com/ssvlabs/SIPs/pull/94)) adds the Payload Timeliness Committee duty at the Gloas / EIP-7732 / ePBS fork. Anchor needs a new `Role::PTCCommittee` wire surface before any other PTC work can land.
- Adding the variant breaks five exhaustive `match` sites in `message_validator` and `qbft_manager`; those edits ship together so the workspace stays green between PRs. The `qbft_manager` change is a one-line stub that the PTC instance routing PR replaces.
- The role is wire-incompatible with pre-CStar operators (pre-CStar nodes decode the byte as `Err(NoMatchingVariant)`). The message-validator fork-gate is defense-in-depth for CStar-capable nodes that might receive a PTC message before their local `active_fork >= Fork::CStar`.
- Issue: sigp/anchor#1032
- Milestone: [ePBS: Payload Timeliness Committee (PTC) Attestation](https://github.com/sigp/anchor/milestone/4)
- Spec: [consensus-specs Gloas `compute_ptc`](https://github.com/ethereum/consensus-specs/blob/4a4937bea332d72a55a76aaebcb97fbcdc189f69/specs/gloas/beacon-chain.md#L666)

## Change Overview

Diff shape: **+553 / -9 across 5 files. ~69 lines of production code, ~484 lines of tests (≈7:1).** Most of the test volume is `ValidationContext` construction boilerplate (matches the existing house style for full-pipeline integration tests in `partial_signature.rs`).

`ssv_types` — wire surface:
- `Role::PTCCommittee` variant: byte `[7, 0, 0, 0]` (matches SIP-94 `BNRolePTCAttester = 7`), `is_committee_role = true`, `duty_executor = Committee`, `max_round = Some(4)`.
- **No new `PartialSignatureKind` variant.** PTC is structurally a post-consensus signature (signed after QBFT decides `PayloadAttestationData`, same shape as `Role::Committee`'s post-consensus path). Reusing `PostConsensus` eliminates a wire-discriminant slot and ~30 LOC of test boilerplate. The role-to-kind binding lives in `partial_signature_type_matches_role`.

`message_validator` — fork-gate + bounds:
- `validate_role_for_fork`: pre-CStar reject mirroring the AggregatorCommittee/pre-Boole safety net.
- `message_lateness`: PTC bucketed in the long-TTL arm (committee-scoped messages need cross-cluster propagation time, even with mid-slot deadlines).
- `duty_limit`: `Some(min(slots_per_epoch, V))` where `V` = local validator count. PTC pool for slot S = union of beacon committees for slot S, so each validator is PTC-eligible in at most one slot per epoch. Distinct from Committee/AggregatorCommittee's `min(slots_per_epoch, 2*V)` — PTC has no sync-committee component and no attest+sync multiplier.
- `partial_signature_type_matches_role`: binds PTC to `PostConsensus` (folded into the existing Committee arm).
- Per-role bounds: new `Role::PTCCommittee` arm with `message_count <= V` and validator-index occurrence cap `1`. Structural maximum — PTC produces exactly one partial-sig per locally-assigned validator per slot. Distinct from Committee's caps (2V, 2) and AggregatorCommittee's (5V, 5).

`qbft_manager` — transient stub:
- Adds `Role::PTCCommittee` to the `Some(DutyExecutor::Validator(_))` invalid-combination group.
- New committee-role arm returns `Err(QbftError::RoleNotActive)` with a warn-log, marked `//todo(cstar)`. The PTC instance routing PR replaces both.

Tests delivered (10): 3 in `ssv_types`, 7 in `message_validator`. Each maps one-to-one to a design decision documented in the local milestone plan.

**Intentionally not changed.** `partial_sig.rs` and `message_counts.rs` (no new `PartialSignatureKind` variant). Slot-advancement-skip and validator-index-mismatch-skip behavior is inherited via `is_committee_role() == true` and structurally covered by existing AggregatorCommittee skip tests; adding PTC-specific copies would duplicate ~150 LOC with no new signal. The `ptc_is_committee_role` unit test asserts the boolean that drives both skips.

- **Wire-byte collision audit.** Anchor's `Role::AggregatorCommittee = [6, 0, 0, 0]` is Anchor-internal; PTC = 7 is upstream SIP-94. Manual audit for upstream go-ssv role conflict is required before merge. Byte 7 matches the SIP-assigned value and Anchor's byte 6 sits below the SIP allocation.
- **`qbft_manager` stub is transient.** PTC messages reaching `receive_data` are rejected with `RoleNotActive` + warn-log. Not a safety issue (no PTC QBFT instance exists yet — rejection is the correct response). The `//todo(cstar)` marker and warn-log must not survive the PTC instance routing PR.
- **`max_round = Some(4)` is timing-derived, not measured.** Follows from the 75% slot start and `QUICK_TIMEOUT = 2s`: round 1 finishes within slot, rounds 2-4 cover operator clock skew, rounds 5+ are dead weight. Retune only if Lighthouse moves the PTC service start time off 75% slot.

## Validation

- `cargo test -p ssv_types -p message_validator` — 64 + 83 tests pass (10 new PTC tests; 0 regressions).
- `make cargo-fmt-check`, `make lint` — green.
- Each of the 10 new tests targets a specific design decision (PostConsensus binding, fork gate, TTL bucketing, `duty_limit` formula, per-role bounds, committee-role semantics, `max_round` value, committee-style duty-executor).
- Production code comments at three non-obvious sites document the WHY for reviewers reading the diff cold: `lib.rs::message_lateness` (long-TTL rationale), `lib.rs::duty_limit` (PTC selection mechanics + consensus-specs permalink to `compute_ptc`), `partial_signature.rs::partial_signature_type_matches_role` (`PostConsensus` reuse).

## Rollback

- Single-commit revert. The variant is wire-incompatible with pre-CStar operators by design; pre-CStar peers reject the byte already. No data, config, or operational impact. Until CStar activates on a live SSV network, the new role is unreachable from any production code path.

## Blockers / Dependencies

- None for this PR.
- Follow-up issues (same milestone): PTC instance routing in `qbft_manager` (replaces `//todo(cstar)` stub), `PayloadAttestationVote` SSZ container + value checker, `metadata_service` PTC voting-context phase, `sign_payload_attestation` impl, Lighthouse `PayloadAttestationService` spawn.

## Additional Info

Base: `epbs`. The milestone bundles original M3.1 (ssv_types) + M3.7 (message-validator) because the exhaustive-match cascade made splitting them produce double-touched files. The follow-up issues land independently against this base.

### DIFF
--- a/anchor/common/ssv_types/src/msgid.rs
+++ b/anchor/common/ssv_types/src/msgid.rs
@@ -78,11 +78,11 @@ impl Role {
         match self {
             Role::Committee | Role::Aggregator | Role::AggregatorCommittee => Some(12),
             Role::Proposer | Role::SyncCommittee => Some(6),
-            // PTC duty starts at 75% slot with ~3s remaining; QUICK_TIMEOUT = 2s
-            // gives one useful round within the slot (round 1 expires at +2s).
-            // Rounds 2-4 (cumulative +4/+6/+8s) recover from operator start delay
-            // or round-1 message loss; rounds 5+ exceed any realistic inclusion
-            // window and are dead weight.
+            // PTC is expected near 75% of a 12s slot, leaving about 3s.
+            // With 2s quick round timeouts, only round 1 is likely to finish
+            // before slot end. Allow up to 4 rounds as a small local grace
+            // window for delayed starts or message loss; this is not a
+            // consensus-spec requirement.
             Role::PTCCommittee => Some(4),
             // These roles don't use QBFT consensus
             Role::ValidatorRegistration | Role::VoluntaryExit => None,
--- a/anchor/common/ssv_types/src/msgid.rs
+++ b/anchor/common/ssv_types/src/msgid.rs
@@ -21,6 +21,7 @@ pub enum Role {
     ValidatorRegistration,
     VoluntaryExit,
     AggregatorCommittee,
+    PTCCommittee,
 }
 
 impl From<Role> for [u8; 4] {
@@ -33,6 +34,7 @@ impl From<Role> for [u8; 4] {
             Role::ValidatorRegistration => [4, 0, 0, 0],
             Role::VoluntaryExit => [5, 0, 0, 0],
             Role::AggregatorCommittee => [6, 0, 0, 0],
+            Role::PTCCommittee => [7, 0, 0, 0],
         }
     }
 }
@@ -49,13 +51,15 @@ impl TryFrom<&[u8]> for Role {
             [4, 0, 0, 0] => Ok(Role::ValidatorRegistration),
             [5, 0, 0, 0] => Ok(Role::VoluntaryExit),
             [6, 0, 0, 0] => Ok(Role::AggregatorCommittee),
+            [7, 0, 0, 0] => Ok(Role::PTCCommittee),
             _ => Err(DecodeError::NoMatchingVariant),
         }
     }
 }
 
 impl Role {
-    /// Returns true if this role is a committee-based role (Committee or AggregatorCommittee).
+    /// Returns true if this role is a committee-based role (Committee, AggregatorCommittee, or
+    /// PTCCommittee).
     ///
     /// Committee roles handle multiple validators in batched operations and have relaxed
     /// validation rules compared to per-validator roles:
@@ -63,7 +67,10 @@ impl Role {
     /// - Skip slot advancement checks (allow processing "older" slots within 34-slot window)
     /// - Have different message count limits and validator index occurrence limits
     pub fn is_committee_role(self) -> bool {
-        matches!(self, Role::Committee | Role::AggregatorCommittee)
+        matches!(
+            self,
+            Role::Committee | Role::AggregatorCommittee | Role::PTCCommittee
+        )
     }
 
     pub fn max_round(self) -> Option<u64> {
@@ -71,6 +78,10 @@ impl Role {
         match self {
             Role::Committee | Role::Aggregator | Role::AggregatorCommittee => Some(12),
             Role::Proposer | Role::SyncCommittee => Some(6),
+            // PTC duty starts at 75% slot with ~3s remaining; `QUICK_TIMEOUT` = 2s
+            // gives one useful round before the slot-end deadline. Rounds 2-4
+            // cover clock skew between operators; rounds 5+ are dead weight.
+            Role::PTCCommittee => Some(4),
             // These roles don't use QBFT consensus
             Role::ValidatorRegistration | Role::VoluntaryExit => None,
         }
@@ -143,7 +154,7 @@ impl MessageId {
     pub fn duty_executor(&self) -> Option<DutyExecutor> {
         // which kind of executor we need to get depends on the role
         match self.role()? {
-            Role::Committee | Role::AggregatorCommittee => {
+            Role::Committee | Role::AggregatorCommittee | Role::PTCCommittee => {
                 self.0[24..].try_into().ok().map(DutyExecutor::Committee)
             }
             Role::Aggregator
@@ -306,5 +317,41 @@ mod tests {
             }
             None => panic!("Failed to extract duty executor"),
         }
+    }
+
+    #[test]
+    fn ptc_uses_committee_duty_executor() {
+        // PTCCommittee must use Committee-style duty executor (CommitteeId),
+        // not Validator-style (PublicKeyBytes)
+        let domain = DomainType([0, 0, 0, 0]);
+        let committee_id = CommitteeId::from(vec![OperatorId(100), OperatorId(200)]);
+        let duty_executor = DutyExecutor::Committee(committee_id);
+
+        let msg_id = MessageId::new(&domain, Role::PTCCommittee, &duty_executor);
+
+        assert_eq!(msg_id.role(), Some(Role::PTCCommittee));
+
+        match msg_id.duty_executor() {
+            Some(DutyExecutor::Committee(id)) => assert_eq!(id, committee_id),
+            Some(DutyExecutor::Validator(_)) => {
+                panic!("PTCCommittee should use Committee duty executor, not Validator")
+            }
+            None => panic!("Failed to extract duty executor"),
+        }
+    }
+
+    #[test]
+    fn ptc_max_round_is_four() {
+        // PTC duty starts at 75% slot; Some(4) is the deliberate cap distinct
+        // from Committee's Some(12). Regressions copying the Committee value
+        // would compile silently — this guards against that.
+        assert_eq!(Role::PTCCommittee.max_round(), Some(4));
+    }
+
+    #[test]
+    fn ptc_is_committee_role() {
+        // is_committee_role drives the validator-index-mismatch skip and
+        // slot-advancement skip in message_validator. PTC must qualify.
+        assert!(Role::PTCCommittee.is_committee_role());
     }
 }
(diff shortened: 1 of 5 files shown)


shane-moore:
@claude, please take a look. focus on the production code changes first and then the tests in another pass

shane-moore:
@claude, thanks for the thorough pass. Walking through:

**Table-driven Role encoding (msgid.rs):** Not pursuing. Two properties matter for `Role`: (1) compiler catches missing arms on variant add/remove, (2) wire bytes for existing variants stay stable across refactors (go-ssv interop). The current explicit-match pattern gives both by construction. Wire bytes live in match arm bodies, pinned to their variant binding, so enum reorders or renames cannot silently change what byte gets emitted. A const table or `repr(u8)`+`FromRepr` approach moves wire bytes into a separate structure (table entries or `= N` annotations) that someone could silently reshuffle during a reorder, with round-trip tests still passing because they verify internal consistency, not absolute bytes. Since `Role` is actively churning during the ePBS arc (`AggregatorCommittee` landed recently, PTC just now, ProposerPreferences planned for a later milestone), the trade weakens (2) more than it strengthens (1). Worth revisiting once the enum stabilizes.

**`message_lateness` PTC comment:** Considered, then reverted. The match arm covers 6 roles; a PTC-specific clarification was awkward there, and the TTL-vs-duty-deadline distinction applies equally to Committee/Aggregator/AggregatorCommittee. Function name `message_lateness` + variable `ttl` already convey "message ageing, not duty timing" without a comment.

**`duty_limit` sync-committee short-circuit confirmation:** Intentional. The Committee/AggregatorCommittee short-circuit gates a sync-committee membership lookup. PTC has no analog (PTC pool = union of beacon committees per `compute_ptc`, no sync multiplier), so there is no lookup to gate and the simpler arm is correct.

**Occurrence-counting helper extraction:** Agreed, filed as #1039. The triplication pre-dates this PR; PTC is the third instance but extracting in this PR expands scope outside the wire surface. Test boilerplate (your Pass 2 #2) bundled into the same issue.

**`warn!` -> `debug!` in qbft_manager stub:** Keeping `warn!`. Matches the `AggregatorCommittee` precedent at `qbft_manager/src/lib.rs:317`, and the upstream `error!` at `message_receiver/src/manager.rs:187` fires for every QBFT error including this stub's `RoleNotActive` anyway, so the dominant log noise source is not here. The whole arm goes away when routing lands.

**Fork gate on qbft_manager stub:** Already addressed by the expanded TODO in `364fc03`: `// TODO(cstar): wire PTC instance routing and add pre-CStar fork gate (mirror AggregatorCommittee arm above).` The routing follow-up will add both. Note this layer is defense-in-depth only; the message-validator fork gate at `lib.rs:862` already rejects pre-CStar PTC upstream of qbft_manager (single production caller, gated by `Validator::validate`).

**Test comment overstates ordering (Pass 2 #1):** Good catch, applied in `a0bc8f1`. Trimmed the wrong claim; kept the "V+1 distinct validator indices" hint since it explains why the loop is shaped that way. Comment now reads: `// V+1 distinct validator indices: verifies the structural message-count cap (V).`

Follow-up issue: #1039.

anchor/common/ssv_types/src/msgid.rs
```diff
@@ -49,28 +51,39 @@ impl TryFrom<&[u8]> for Role {
...
+            Role::Committee | Role::AggregatorCommittee | Role::PTCCommittee
+        )
     }
 
     pub fn max_round(self) -> Option<u64> {
         // as per https://github.com/ssvlabs/ssv/blob/6382d4b52ea5e0efd9378a5a00ef481f39d6234f/message/validation/consensus_validation.go#L370
         match self {
             Role::Committee | Role::Aggregator | Role::AggregatorCommittee => Some(12),
             Role::Proposer | Role::SyncCommittee => Some(6),
+            // PTC duty starts at 75% slot with ~3s remaining; QUICK_TIMEOUT = 2s
```
diegomrsantos: I don’t think this comment explains the `Some(4)` choice clearly enough. Starting PTC near 75% of a 12s slot leaves about 3s, so only round 1 is likely to complete before slot end with a 2s quick timeout. Rounds 2-4 are a local grace window rather than something derived from the protocol, and the current wording makes that sound more precise than it is.

Could we soften it to something like this?

```rust
// PTC is expected near 75% of a 12s slot, leaving about 3s.
// With 2s quick round timeouts, only round 1 is likely to finish
// before slot end. Allow up to 4 rounds as a small local grace
// window for delayed starts or message loss; this is not a
// consensus-spec requirement.
```

diegomrsantos:
I think these are the right behaviors to cover, but the new validation tests are much heavier than the rules they assert. Most of the PTC cases rebuild a full signed message, RSA key pair, fork schedule, clock, committee info, and operator key map just to exercise a small branch like `message_count > V`, duplicate validator-index counting, or `PartialSignatureKind` matching.

Could we factor the repeated setup into test-only helpers and keep only one or two full-pipeline PTC smoke tests? The count and duplicate-index cases should ideally read mostly as "build these partial sig messages, validate PTC bounds, expect this error." That would keep the coverage while making the rule under test much easier to review.

The TTL coverage could also be more targeted: use a receive time that would fail under the short proposer/sync TTL but pass under the long committee TTL, so the test proves PTC is intentionally in the committee-style TTL bucket rather than only proving one accepted and one rejected timestamp.

anchor/common/ssv_types/src/msgid.rs
shane-moore: Applied verbatim in 4d98a50.

shane-moore:
Pushed 4d98a50. The three small branches you flagged now read as pure-function asserts:

- `message_count > V` → `validate_ptc_committee_message_count(count, V)`, mirroring the existing `validate_aggregator_committee_message_count` shape that the small message-count tests at lines 519-554 already use.
- Duplicate validator-index → `validate_validator_index_occurrence_limit(messages, limit)`.
- `PartialSignatureKind` matching → direct asserts on the existing pure function `partial_signature_type_matches_role`. The integration mapping (predicate false → `PartialSignatureTypeRoleMismatch`) is already covered end-to-end by `test_partial_signature_message_with_invalid_type_for_role` at line 578, and PTC shares that arm.

TTL is now bucket-specific via `COMMITTEE_TTL_BUCKET_SLOTS = 20` (past proposer/sync TTL of 3, inside committee TTL of 34). Per your "rather than only proving one accepted and one rejected timestamp" framing I also dropped `test_ptc_committee_beyond_ttl_rejected`; the bucket-specific accepted test covers it and the generic `LateSlotMessage` upper-bound is exercised for other roles.

Remaining full-pipeline smoke tests: pre-CStar fork gate + bucket-specific TTL accepted. Two total, all using existing helpers rather than raw signing/key construction.

What I deliberately did not do: migrate the Committee and AggregatorCommittee inline arms (and the 152-LOC AggregatorCommittee occurrence test) to use the new helpers. That migration is a clean standalone refactor; doing it here mixes refactor with new code. The helpers ship with their first consumer in this PR and the rest moves in #1039.

Inline `msgid.rs:81` thread resolved separately with your suggested wording.

diegomrsantos (approved):
Approved. Re-reviewed after the latest changes; the PTC round-limit comment and validation test refactor address my remaining concerns.

PR #1041, opened 2026-05-20
feat(validator_store): drive Phase 2 from beacon head events

petarjuki7:
Second PR for #986. Depends on #1011 (LH head-monitor wiring). The diff will be clean once #1011 merges.

#1011 wires the LH 8.1.0 head monitor but the eager-attest path is inert end to end: `sign_committee_attestations` blocks on `get_voting_context(slot)`, which is only populated by MetadataService Phase 2 on a fixed timer. Even with the LH side firing early, attestation publishing still waits.

## Change Overview

* `Client::run` fans out the single LH head-event mpsc into two consumer channels via a relay task, one for LH `AttestationService` and one for a new `MetadataService` receiver. No LH changes.
* `MetadataService` Phase 2 waits out the current slot, then races `wait_for_head_event` against the fallback timer in a `tokio::select!`. First to fire triggers the voting-context update.
* When the head event wins, `update_voting_context` calls `run_on_candidate_index(event.beacon_node_index, ...)` directly, bypassing WAD. The returned block_root is verified against the event; on mismatch or RPC error we fall back to the WAD/first_success path.
* Two metrics: `anchor_metadata_service_voting_context_triggers_total{trigger}` and `anchor_metadata_service_voting_context_offset_seconds{trigger}` for hit rate and time-into-slot.

A couple of intentional choices worth flagging:

* The Phase 2 fallback timer uses `spec.get_unaggregated_attestation_due()` rather than `slot_duration / 3`, for parity with LH and so the value tracks any future fork that retunes it.
* The fan-out task's `send().await` is blocking. At expected head-event rates (one per BN per slot) the 1024 buffer is heavy overkill, so blocking is unreachable in practice. A single `warn!` fires if a downstream closes.

* `MetadataService` is `#[derive(Clone)]`, so the receiver field is `Option<Arc<Mutex<mpsc::Receiver<HeadEvent>>>>` rather than LH's bare `Option<Mutex<...>>`. The `Arc` keeps clones cheap.
* Block-root mismatch handling (head event says root X, BN returns root Y) treats the BN response as untrusted and falls back. Avoids publishing a `BeaconVote` that diverges from the head event the cluster may also be receiving.

## Validation

* `cargo test -p anchor_validator_store --lib metadata_service::tests` clean (21 passed).
* Three unit tests cover `wait_for_head_event`: matching-slot returns the event, stale-slot events are dropped, disabled monitor returns `None` immediately.
* End-to-end "did this move attestation publishing earlier" is a hypothesis until run on a real network. The two new metrics are what's needed to confirm.

## Rollback

Revert the commit set, or pass `--disable-beacon-head-monitor` from #1011 to short-circuit the whole path.

### DIFF
--- a/anchor/client/src/lib.rs
+++ b/anchor/client/src/lib.rs
@@ -341,14 +341,51 @@ impl Client {
 
         // Only the beacon_nodes are used for attestation duties, so proposer_nodes do not need a
         // head_send ref.
-        let head_monitor_rx = if config.enable_beacon_head_monitor {
-            let (head_monitor_tx, head_monitor_rx) =
-                mpsc::channel::<HeadEvent>(MAX_HEAD_EVENT_QUEUE_LEN);
-            beacon_nodes.set_head_send(Arc::new(head_monitor_tx));
-            Some(Mutex::new(head_monitor_rx))
-        } else {
-            None
-        };
+        let (attestation_head_monitor_rx, metadata_head_monitor_rx) =
+            if config.enable_beacon_head_monitor {
+                let (head_event_tx, mut head_event_rx) =
+                    mpsc::channel::<HeadEvent>(MAX_HEAD_EVENT_QUEUE_LEN);
+                let (attestation_service_tx, attestation_service_rx) =
+                    mpsc::channel::<HeadEvent>(MAX_HEAD_EVENT_QUEUE_LEN);
+                let (metadata_service_tx, metadata_service_rx) =
+                    mpsc::channel::<HeadEvent>(MAX_HEAD_EVENT_QUEUE_LEN);
+                beacon_nodes.set_head_send(Arc::new(head_event_tx));
+
+                executor.spawn(
+                    async move {
+                        while let Some(head_event) = head_event_rx.recv().await {
+                            let HeadEvent {
+                                beacon_node_index,
+                                slot,
+                                beacon_block_root,
+                            } = head_event;
+                            let to_attest = HeadEvent {
+                                beacon_node_index,
+                                slot,
+                                beacon_block_root,
+                            };
+                            let to_meta = HeadEvent {
+                                beacon_node_index,
+                                slot,
+                                beacon_block_root,
+                            };
+                            if attestation_service_tx.send(to_attest).await.is_err()
+                                || metadata_service_tx.send(to_meta).await.is_err()
+                            {
+                                warn!("head event fan-out: downstream closed, exiting relay");
+                                return;
+                            }
+                        }
+                    },
+                    "head_event_fanout",
+                );
+                (
+                    Some(Mutex::new(attestation_service_rx)),
+                    Some(Arc::new(Mutex::new(metadata_service_rx))),
+                )
+            } else {
+                (None, None)
+            };
 
         let beacon_nodes = Arc::new(beacon_nodes);
         start_fallback_updater_service::<_, E>(executor.clone(), beacon_nodes.clone())?;
@@ -674,7 +711,7 @@ impl Client {
             .beacon_nodes(beacon_nodes.clone())
             .executor(executor.clone())
             .chain_spec(spec.clone())
-            .head_monitor_rx(head_monitor_rx)
+            .head_monitor_rx(attestation_head_monitor_rx)
             .build()?;
 
         let preparation_service = PreparationServiceBuilder::new()
@@ -717,6 +754,7 @@ impl Client {
             spec.clone(),
             fork_schedule.clone(),
             config.with_weighted_attestation_data,
+            metadata_head_monitor_rx,
         );
 
         // We use `SLOTS_PER_EPOCH` as the capacity of the block notification channel, because
--- a/anchor/client/src/lib.rs
+++ b/anchor/client/src/lib.rs
@@ -18,7 +18,8 @@ use anchor_validator_store::{
     registration_service::RegistrationService,
 };
 use beacon_node_fallback::{
-    BeaconNodeFallback, CandidateBeaconNode, start_fallback_updater_service,
+    BeaconNodeFallback, CandidateBeaconNode, beacon_head_monitor::HeadEvent,
+    start_fallback_updater_service,
 };
 use config::Config;
 use database::{NetworkDatabase, OwnOperatorId};
@@ -45,7 +46,10 @@ use task_executor::TaskExecutor;
 use tokio::{
     net::TcpListener,
     select,
-    sync::{mpsc, mpsc::unbounded_channel},
+    sync::{
+        Mutex,
+        mpsc::{self, unbounded_channel},
+    },
     time::{Instant, interval, sleep},
 };
 use tracing::{debug, error, info, warn};
@@ -78,6 +82,7 @@ const HTTP_GET_DEBUG_BEACON_STATE_QUOTIENT: u32 = 4;
 const HTTP_GET_DEPOSIT_SNAPSHOT_QUOTIENT: u32 = 4;
 const HTTP_GET_VALIDATOR_BLOCK_TIMEOUT_QUOTIENT: u32 = 4;
 const HTTP_DEFAULT_TIMEOUT_QUOTIENT: u32 = 4;
+const MAX_HEAD_EVENT_QUEUE_LEN: usize = 1_024;
 
 pub struct Client {}
 
@@ -333,6 +338,17 @@ impl Client {
 
         beacon_nodes.set_slot_clock(slot_clock.clone());
         proposer_nodes.set_slot_clock(slot_clock.clone());
+
+        // Only the beacon_nodes are used for attestation duties, so proposer_nodes do not need a
+        // head_send ref.
+        let head_monitor_rx = if config.enable_beacon_head_monitor {
+            let (head_monitor_tx, head_monitor_rx) =
+                mpsc::channel::<HeadEvent>(MAX_HEAD_EVENT_QUEUE_LEN);
+            beacon_nodes.set_head_send(Arc::new(head_monitor_tx));
+            Some(Mutex::new(head_monitor_rx))
+        } else {
+            None
+        };
 
         let beacon_nodes = Arc::new(beacon_nodes);
         start_fallback_updater_service::<_, E>(executor.clone(), beacon_nodes.clone())?;
@@ -658,6 +674,7 @@ impl Client {
             .beacon_nodes(beacon_nodes.clone())
             .executor(executor.clone())
             .chain_spec(spec.clone())
+            .head_monitor_rx(head_monitor_rx)
             .build()?;
 
         let preparation_service = PreparationServiceBuilder::new()
--- a/anchor/client/src/lib.rs
+++ b/anchor/client/src/lib.rs
@@ -48,7 +48,7 @@ use tokio::{
     select,
     sync::{
         Mutex,
-        mpsc::{self, unbounded_channel},
+        mpsc::{self, error::TrySendError, unbounded_channel},
     },
     time::{Instant, interval, sleep},
 };
@@ -354,6 +354,8 @@ impl Client {
                 executor.spawn(
                     async move {
                         while let Some(head_event) = head_event_rx.recv().await {
+                            // LH's `HeadEvent` doesn't derive Clone, so rebuild for each
+                            // downstream. All fields are Copy so this is essentially free.
                             let HeadEvent {
                                 beacon_node_index,
                                 slot,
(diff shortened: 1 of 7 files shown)


anchor/client/src/lib.rs
```diff
@@ -334,6 +339,54 @@ impl Client {
...
+                                beacon_node_index,
+                                slot,
+                                beacon_block_root,
+                            };
+                            let to_meta = HeadEvent {
+                                beacon_node_index,
+                                slot,
+                                beacon_block_root,
+                            };
+                            if attestation_service_tx.send(to_attest).await.is_err()
```
shane-moore: Suggest try_send for the fan-out instead of send().await

Why

The current send().await on both downstreams creates an upstream-backpressure path that bites on cold start. AttestationService and MetadataService don't begin draining their receivers until after the SSV-sync gate + doppelganger wait (typically 15-60+ min on a fresh node). With downstream channels at 1024 cap and N BNs producing ~1 event/slot/BN, the relay blocks on send() after ~`204/Nminutes. That blockshead_event_rx.recv(), which blocks LH's head_monitor_send.send().await in [poll_head_event_from_beacon_nodes][lh-send]. The merged SSE stream then stops being polled, per-BN reqwest buffers fill, and BNs that enforce slow-consumer drops will close the connection. Head monitor returns Err, outer loop logs warn!("Head service failed retrying..."), restarts the whole subscription. Self-recovering but generates BN reconnect cycles and log spam every ~204/N` min during sync.

Proposal

match attestation_service_tx.try_send(to_attest) {
    Ok(()) => {}
    Err(TrySendError::Full(_)) => metrics::inc_counter_vec(
        &metrics::HEAD_EVENT_FANOUT_DROPS, &["attestation_service"],
    ),
    Err(TrySendError::Closed(_)) => {
        warn!("head event fan-out: attestation downstream closed");
        return;
    }
}
// same for metadata_service_tx

Why it's safe

Both downstreams already filter to current-slot events only (LH's poll_for_head_events, MetadataService's wait_for_head_event in this PR). Past-slot events are explicitly discarded by the consumer. Pre-sync events are past-slot by definition by the time consumers start, so dropping them in the relay is semantically equivalent to the consumer's existing behavior.

Steady-state post-sync: producer ≈ N events/slot, each consumer drains ≥1 per slot via its inner recv loop. Channel cannot fill under normal conditions; the counter surfaces any anomalies.

Tradeoffs

- Lossy under a degenerate consumer stall (multi-slot pause). The drop counter makes this observable rather than silent.
- One added metric (anchor_head_event_fanout_drops_total{downstream})

anchor/client/src/lib.rs
```diff
@@ -334,6 +339,54 @@ impl Client {
...
+                let (attestation_service_tx, attestation_service_rx) =
+                    mpsc::channel::<HeadEvent>(MAX_HEAD_EVENT_QUEUE_LEN);
+                let (metadata_service_tx, metadata_service_rx) =
+                    mpsc::channel::<HeadEvent>(MAX_HEAD_EVENT_QUEUE_LEN);
+                beacon_nodes.set_head_send(Arc::new(head_event_tx));
+
+                executor.spawn(
+                    async move {
+                        while let Some(head_event) = head_event_rx.recv().await {
+                            let HeadEvent {
```
shane-moore, line 360: The destructure-then-rebuild reads as intentional ceremony but is actually a
> workaround for LH's `HeadEvent` only deriving `Debug`. Worth a one-line comment
> so the next reader doesn't have to chase the type two crates away.
>
> ````suggestion
>                         while let Some(head_event) = head_event_rx.recv().await {
>                             // LH's `HeadEvent` doesn't derive Clone — rebuild for each downstream.
>                             let HeadEvent {
>                                 beacon_node_index,
>                                 slot,
>                                 beacon_block_root,
>                             } = head_event;
>                             let to_attest = HeadEvent {
>                                 beacon_node_index,
>                                 slot,
>                                 beacon_block_root,
>                             };
>                             let to_meta = HeadEvent {
>                                 beacon_node_index,
>                                 slot,
>                                 beacon_block_root,
>                             };
> ````

anchor/validator_store/src/metadata_service.rs
```diff
@@ -129,6 +132,40 @@ pub struct MetadataService<E: EthSpec, T: SlotClock + 'static> {
     spec: Arc<ChainSpec>,
     fork_schedule: Arc<ForkSchedule>,
     weighted_attestation_data: bool,
+    head_monitor_rx: Option<Arc<Mutex<mpsc::Receiver<HeadEvent>>>>,
```
shane-moore: nit/cleanup: consider dropping the `Arc<Mutex<>>` wrapper and passing the receiver
into `start_update_service` as a method arg instead.

Right now `head_monitor_rx: Option<Arc<Mutex<mpsc::Receiver<HeadEvent>>>>` is on
the struct because `MetadataService` is `#[derive(Clone)]` and the other Phase
clones need to compile against the field. But only Phase 2 ever reads the
receiver — Phase 1 and Phase 3 never touch it — so the `Arc<Mutex<>>` advertises
shared mutability that doesn't exist. The Mutex is held by exactly one task with
no contender; the Arc is one allocation that's never reused.

If `start_update_service` takes the receiver as a moved parameter, Phase 2's
async block owns it directly:
```
  pub fn start_update_service(
      self,
      mut head_monitor_rx: Option<mpsc::Receiver<HeadEvent>>,
  ) -> Result<(), String> {
      // ... Phase 1 spawn unchanged ...
      let self_clone_phase2 = self.clone();
      executor.spawn(async move {
          loop {
              // ... sleep to slot boundary ...
              let head_event = match head_monitor_rx.as_mut() {
                  Some(rx) => tokio::select! {
                      _ = sleep(due) => None,
                      event = wait_for_head_event(rx, &self_clone_phase2.slot_clock) => event,
                  },
                  None => { sleep(due).await; None }
              };
              // ... rest unchanged ...
          }
      }, "voting_context_service");
      // ... Phase 3 spawn unchanged ...
  }
```
Then `wait_for_head_event` becomes
`async fn wait_for_head_event(rx: &mut mpsc::Receiver<HeadEvent>, slot_clock: &T)
-> Option<HeadEvent>` (no Option, no .lock().await). `mpsc::Receiver::recv()` is
cancel-safe (tokio docs), so dropping the future under select! still strands
nothing.

Call site in `lib.rs` drops the `Arc::new(Mutex::new(...))` wrap:
  Some(metadata_service_rx)  // instead of Some(Arc::new(Mutex::new(...)))

LH's parallel field is `Option<Mutex<Receiver>>` (no Arc) only because LH uses an
`inner: Arc<Inner>` Clone pattern — the bare Mutex relies on the outer Arc that
MetadataService doesn't have. Method-arg sidesteps both the inner-Arc refactor
and the extra wrapper.

Net change: ~20-25 lines across these same two files, one trivial unit test
deleted (the None-helper case becomes a Phase 2 outer match, not unit-testable
without a full MetadataService construction).

anchor/validator_store/src/metadata_service.rs
```diff
@@ -129,6 +132,40 @@ pub struct MetadataService<E: EthSpec, T: SlotClock + 'static> {
...
+}
+
+/// Wait for a head event matching the current slot reported by `slot_clock`.
```
shane-moore, line 137: comment here describe past-slot behavior, but the check is `==` — future-slot
> events (clock skew, NTP step) also fall into the discard branch

````suggestion
> /// Wait for a head event matching the current slot reported by `slot_clock`.
> ///
> /// Drops events whose slot doesn't match the current slot and keeps waiting.
> /// Resolves to `None` if `receiver` is `None`, the channel is closed, or
> /// `slot_clock.now()` fails.
> ///
> /// Extracted as a free function so it can be unit-tested without constructing
> /// a full `MetadataService` (which requires a beacon node, duties service,
> /// validator store, etc.).

File: anchor/validator_store/src/metadata_service.rs [hunk 5]
```diff
@@ -129,6 +132,40 @@ pub struct MetadataService<E: EthSpec, T: SlotClock + 'static> {
...
+    loop {
+        match receiver.recv().await {
+            Some(head_event) => {
+                // Only return head events for the current slot - this ensures the
+                // block for this slot has been produced before triggering attestation.
+                let current_slot = slot_clock.now()?;
+                if head_event.slot == current_slot {
+                    return Some(head_event);
+                }
+                // Head event is for a previous slot, keep waiting
```
shane-moore: Consider the following comment for same reason as above:
// Head event slot doesn't match current; drop and keep waiting.

anchor/validator_store/src/metadata_service.rs
```diff
@@ -199,27 +238,67 @@ impl<E: EthSpec, T: SlotClock + 'static> MetadataService<E, T> {
...
+                            &metrics::METADATA_SERVICE_VOTING_CONTEXT_OFFSET_SECONDS,
+                            &[trigger],
+                            offset,
+                        );
+                    }
+
+                    if let Err(err) = self_clone_phase2.update_voting_context(head_event).await {
+                        error!(err, "Failed to update voting context")
+                    } else {
+                        trace!(%intended_slot, from_head_event, "Updated voting context");
```
shane-moore: Phase 1 and Phase 3 only log via `trace!(%slot, ...)` inside their publish functions — no outer-loop success trace. Phase 2 already has the same inner trace (`trace!(%slot, "Published VotingContext")` at the end of `update_voting_context`), so the outer `trace!(%intended_slot, from_head_event, ...)` is a redundant duplicate at the same event.

// After:
```
if let Err(err) = self_clone_phase2.update_voting_context(head_event).await {
    error!(err, "Failed to update voting context")
}
```

shane-moore:
@claude, please take a look. keep in mind all the comments i've already left and also ignore the commits that were already made in https://github.com/sigp/anchor/pull/1011 as they've already been commented on there

shane-moore:
@claude, thanks for the review. Responses inline.

**1.** Real race, benign outcome. `fetch_attestation_data_from_event`'s `block_root` check is the safety net: on N→N+1 crossing, either N+1's head is still N's block (eager succeeds, data consistent) or N+1 has a new block (mismatch fires, fallback runs `fetch_attestation_data(N+1)` with correct data). `get_voting_assignments(N+1)` waits on Phase 1's watch-channel publication (`validator_store/src/lib.rs:698-718`) and returns `Ok` only when `voting_assignments.slot == slot`; no "wrong attesters" path. The metric/trace attribution can be slightly off on the race, and the misleading `intended_slot` trace is the only artifact making it visible. We're dropping that trace in a follow-up patch. Not refactoring.

**2.** Possible but unlikely in production. `SystemTimeSlotClock::now()` only fails pre-genesis and `wait_for_genesis` gates startup; if it does happen, the next `slot_clock.now()` in `update_voting_context:431` errors loudly. Also: when `wait_for_head_event` returns `None`, `from_head_event = false` and `trigger = TRIGGER_TIMER`, not `head_event`. The metric pollution direction in the description is inverted. Not changing.

**3.** Bypass is intentional, documented at `metadata_service.rs:426-429`. Agreed the flag interaction isn't visible at the CLI level; will add a note to `--with-weighted-attestation-data` help text. Skipping eager-when-WAD defeats the optimization, the `block_root` check is a reasonable cross-check.

**4.** `get_unaggregated_attestation_due` is a precomputed-field accessor (Lighthouse `chain_spec.rs:867-871`). Trivial field read, per your own opt-out.

**5.** Real gap. Cross-phase precedent in `metadata_service.rs`: Phase 3 extracts and tests pure helpers because sort order is consensus-encoded (must match SSV-Go, lines 1204-1255). Phase 1 has no tests because logic is shallow. The eager-path classify falls closer to Phase 1's bar: one `==` check, two warn-and-return branches, false-negative regression is harmless (fallback runs anyway). Following the same convention. Open to a follow-up if the path turns out regression-prone.

**`MAX_HEAD_EVENT_QUEUE_LEN`:** Constant was introduced in #1011, sizing-rationale comment is on that PR's action items.

**Late wakeup:** Same `slot_clock.now()`-inside-publish-fn pattern in all three phases. Late wake means late publish, not wrong publish. No change.

petarjuki7:
Thank you @shane-moore for the review! Pushing fixes shortly. Two were earlier decisions I decided on but now makes sense to update them post review:

A. I considered try_send pre-PR and went with send().await on the read that the 1024 buffer was unreachable at steady-state rates. Cold-start window (sync gate + doppelganger wait) collapses that. Applying with the drop metric.

C. Went with Arc<Mutex<>> because of the Clone constraint. Didn't think about pulling the receiver out of the struct entirely. Cleaner, applying.

Rest is clear, applying as suggested.

shane-moore:
Recent push addresses all the inline feedback cleanly:

- `try_send` + `HEAD_EVENT_FANOUT_DROPS{downstream}` metric on the fan-out
- `HeadEvent` rebuild comment
- `Arc<Mutex<>>` removed; receiver passed as moved arg to `start_update_service`; disabled-case unit test correctly dropped along with it
- Past-slot comment and `wait_for_head_event` doc updated
- Redundant outer Phase 2 `trace!` + `intended_slot` capture removed

The WAD-bypass note on `--with-weighted-attestation-data` is a nice extra.

LGTM. Will formally approve once #1011 merges and the diff cleans up.

petarjuki7:
Thanks @shane-moore! Fixed merge conflicts in [e101626](https://github.com/sigp/anchor/pull/1041/commits/e101626e95e7f0f8c9f80757b3c6a00f1d6315ff)

shane-moore (approved):

Issue #1037, opened 2026-05-18
feat(validator_store): implement sign_payload_attestation


shane-moore:
## Goal

Implement the `sign_payload_attestation` LH trait method on `AnchorValidatorStore` in `anchor/validator_store/src/lib.rs`, modeled on `sign_committee_attestations` (around `lib.rs:1459`). The method runs committee-mode QBFT over `PayloadAttestationVote` constructed from LH's passed `PayloadAttestationData`, then signs the decided data with each local PTC-assigned validator's BLS share under `Domain::PTCAttester`. LH's `PayloadAttestationService` calls this method per validator at 75% slot; Anchor batches internally per cluster.

**Critical:** must NOT touch slashing protection. Per CR-1 of the SIP-94 security cross-cuts, PTC and proposer-preferences signing must not call `slashing_protection.check_*` or `slashing_protection.insert_*`. LH's reference implementation bypasses doppelganger and skips the slashing DB; Anchor must do the same.

## Context

Per SIP-94 §3 ([ssvlabs/SIPs#94](https://github.com/ssvlabs/SIPs/pull/94)), each PTC-assigned validator signs `PayloadAttestationData { beacon_block_root, slot, payload_present, blob_data_available }` under `DOMAIN_PTC_ATTESTER`. In SSV, a cluster runs one QBFT instance per slot over a stripped `PayloadAttestationVote` (per #1034), then each validator on PTC for the slot contributes a partial signature on the decided data. Partial sigs reconstruct into a full `PayloadAttestationMessage` that LH's `PayloadAttestationService` submits.

This issue brings together #1032 (role wire surface + `PostConsensus` binding), #1034 (`PayloadAttestationVote` + value checker), #1035 (qbft_manager routing + `QbftDecidable` impl), and #1036 / PR #1053 (publishes per-slot PTC validator set on `VotingAssignments` for batch sizing). The follow-up service-spawn issue wires LH's `PayloadAttestationService` to call this method.

## Suggested approach

Add a `sign_payload_attestation` method on `AnchorValidatorStore` modeled on `sign_committee_attestations` (`lib.rs:1459`). High-level flow:

1. Look up `(validator, cluster)` via `get_validator_and_cluster(validator_pubkey)` (`lib.rs:239+`).
2. Await `get_voting_assignments(slot)` to load the Phase-1 snapshot.
[...]

Issue #1035, opened 2026-05-18
feat(qbft_manager): wire PTC committee-scoped QBFT instances and fork gate


shane-moore:
## Goal

Replace #1032's `//todo(epbs)` stub at the `Some(Role::PTCCommittee)` arm in `qbft_manager/src/lib.rs::receive_data` with real PTC plumbing: a new `PTCCommitteeInstanceId`, a fourth instance `Map` on `QbftManager`, routing in `receive_data`, a `cleaner` entry, and the `QbftDecidable` impl for `PayloadAttestationVote`. Fork-gate mirroring the AggregatorCommittee/Boole+ pattern.

## Context

Per SIP-94 §3 ([ssvlabs/SIPs#94](https://github.com/ssvlabs/SIPs/pull/94)), PTC runs one QBFT instance per cluster per slot over `PayloadAttestationVote` (introduced in #1034). The `qbft_manager` currently routes `Role::PTCCommittee` messages to the placeholder stub introduced in #1032, which rejects with `RoleNotActive` + a warn-log + a `//todo(epbs)` marker. This issue replaces the stub with the real instance map plumbing: PTC joins `Committee`, `AggregatorCommittee`, and `Proposer` as the fourth committee-scoped instance kind.

`PayloadAttestationVote` (the QBFT decision type for PTC) needs a `QbftDecidable` impl so the manager can map `Role::PTCCommittee` + `CommitteeId` to a deterministic message id, in the same style as the existing `BeaconVote` and `AggregatorCommitteeConsensusData` impls.

## Suggested approach

`anchor/qbft_manager/src/lib.rs`:
- Add `PTCCommitteeInstanceId { committee: CommitteeId, instance_height: InstanceHeight }`, mirroring the existing `AggregatorCommitteeInstanceId`.
- Add a fourth `Map<PTCCommitteeInstanceId, PayloadAttestationVote>` field on `QbftManager`; initialize it in the constructor.
- In `receive_data`, replace the `Some(Role::PTCCommittee)` stub arm (currently `warn!` + `Err(QbftError::RoleNotActive)` + `//todo(epbs)` line comment) with real routing next to `Role::AggregatorCommittee`. Fork-gate with `if active_fork < Fork::CStar { return Err(QbftError::RoleNotActive); }`. The `//todo(epbs)` line comment and the placeholder warn-log must be gone after this PR.
- Add `.retain(...)` for the new map in `cleaner`, matching the TTL/eviction policy of the existing committee-scoped maps.
[...]

PR #1047, opened 2026-05-21
feat(ssv_types): add PayloadAttestationVote with QbftData and value checker

### LINKED ISSUE #1034: feat(ssv_types): add PayloadAttestationVote with QbftData and value checker
## Goal

Introduce `PayloadAttestationVote { beacon_block_root, payload_present, blob_data_available }` as an SSZ container in `anchor/common/ssv_types/src/consensus.rs`, modeled on `BeaconVote` at `consensus.rs:890-907`. Implement `QbftData` on it. Add `PayloadAttestationVoteValidator` implementing `QbftDataValidator<PayloadAttestationVote>` with a value check that returns `false` only for zero `beacon_block_root` — no comparison against any local beacon-node view, no slashing-protection writes.

## Context

Per SIP-94 §3 ([ssvlabs/SIPs#94](https://github.com/ssvlabs/SIPs/pull/94)), PTC's QBFT instance decides over a stripped `PayloadAttestationVote { beacon_block_root, payload_present, blob_data_available }` — `slot` is omitted because the QBFT instance pins it (`instance_height = slot.as_usize()`). After consensus, each PTC-assigned validator signs the full `PayloadAttestationData { beacon_block_root, slot, payload_present, blob_data_available }` (slot reconstructed from the duty) under `DOMAIN_PTC_ATTESTER`. See [consensus-specs Gloas validator duties](https://github.com/ethereum/consensus-specs/blob/dev/specs/gloas/validator.md#payload-attester-duties) and [beacon-chain.md containers](https://github.com/ethereum/consensus-specs/blob/dev/specs/gloas/beacon-chain.md).

Per SIP-94 SC-2, the value check rejects only zero `beacon_block_root`. Payload-status fields (`payload_present`, `blob_data_available`) are trusted from the QBFT leader and **NOT** compared against the local BN's view: operators' BNs may legitimately disagree on payload-arrival timing near the 75% deadline, so requiring local agreement would fail QBFT rounds. The duty also does not write to slashing protection (CR-1: PTC and proposer-preferences signing must not touch the block-signing DB).

## Suggested approach

`anchor/common/ssv_types/src/consensus.rs`:
- Add `PayloadAttestationVote` struct next to the existing `BeaconVote` (`consensus.rs:890-907`), with SSZ `Encode`/`Decode`/`TreeHash` derives.
- `QbftData` impl mirrors `BeaconVote`'s: `type Hash = Hash256; fn hash() -> Hash256 { sha256(self.as_ssz_bytes()) }` for cross-operator determinism.
- Add `PayloadAttestationVoteValidator` modeled on `AggregatorCommitteeDataValidator` at `consensus.rs:595-744`. The `validate(value, start_value) -> bool` impl returns `true` for any non-zero `beacon_block_root`; returns `false` when `value.beacon_block_root == Hash256::zero()`. Resist any temptation to add consistency checks against `start_value` or the local BN view — that would fail QBFT rounds near the 75% deadline and violates SC-2.

## Acceptance criteria

- `PayloadAttestationVote` SSZ encode/decode round-trip succeeds.
- `PayloadAttestationVote::hash()` is deterministic across operators (SHA-256 over SSZ bytes).
- `PayloadAttestationVoteValidator::validate` returns `false` for any `PayloadAttestationVote` with `beacon_block_root == Hash256::zero()`; returns `true` for any combination of non-zero `beacon_block_root` with `payload_present`/`blob_data_available` boolean values.
- The validator does NOT touch `slashing_protection.check_*`/`insert_*`, and does NOT compare against any local BN view (only `value.beacon_block_root` is inspected; `start_value` is ignored).

## Open questions

- Should `PayloadAttestationVoteValidator` carry an `EthSpec` generic parameter (matching `BeaconVoteValidator<E: EthSpec>` and `AggregatorCommitteeDataValidator<E: EthSpec>`)? PTC has no spec-dependent fields, so the generic likely drops, but confirm at PR time — there may be a downstream usage that requires it.

## Risks

- **Value-check structurally weaker than `BeaconVoteValidator`.** No slashability call, no observation-drift enforcement. This is intentional per SIP-94 SC-2. The cost of getting this wrong is high: adding any check against `start_value` or local BN observation will fail QBFT consensus during normal operation, not just adversarial cases. The cluster simply trusts the QBFT leader on payload-status fields; gossip REJECT at the BN catches dishonest values one layer up.

shane-moore:

Closes #1034. Next ePBS milestone item after PR #1033 (`Role::PTCCommittee` + message-validator handling).

Per [SIP-94](https://github.com/ssvlabs/SIPs/pull/94) §3 and the [Gloas validator spec](https://github.com/ethereum/consensus-specs/blob/dev/specs/gloas/validator.md#payload-attester-duties), PTC committees run QBFT each slot over a stripped `PayloadAttestationVote { beacon_block_root, payload_present, blob_data_available }` (slot omitted, pinned by the QBFT instance). After consensus, each PTC-assigned validator signs the full `PayloadAttestationData` (slot reconstructed from the duty) under `DOMAIN_PTC_ATTESTER`. This PR adds the SSZ container and value checker that the QBFT instance will use; downstream issues #1035 and #1037 wire the consumers.

## Change Overview

`anchor/common/ssv_types/src/consensus.rs` (single file, additive):

- **`PayloadAttestationVote`** — SSZ struct placed next to `BeaconVote`. `QbftData` impl with `sha256(self.as_ssz_bytes())` hash for cross-operator determinism. Mirrors `BeaconVote::hash` shape.
- **`PayloadAttestationVoteValidator`** + **`PayloadAttestationVoteValidationError`** — placed next to `BeaconVoteValidationError`. One rule: reject zero `beacon_block_root`. Uses the same `do_validation -> Result -> match Ok/Err -> warn -> bool` shape as `BeaconVoteValidator` and `AggregatorCommitteeDataValidator` for consistency.
- **6 unit tests** in the existing `mod tests` block under a new `═══ PayloadAttestationVote Tests ═══` section.

**Reading order:** start at the new `PayloadAttestationVote` struct (~L909); skim down to the validator (~L1205); the test block at end of `mod tests` mirrors the `AssignedAggregator` / `AggregatorCommittee` test patterns.

**Intentionally unchanged:**
- No production consumers wired (`qbft_manager` routing lands in #1035; `sign_payload_attestation` lands in #1037).
- No `QbftDecidable` impl — that lives in `qbft_manager` (#1035).
- The existing four `sha256(ssz_bytes)` hash impls (`ProposerConsensusData`, `AggregatorCommitteeConsensusData`, `BeaconVote`) — deferred standalone refactor; not mixed into this feature PR.

**Validator is structurally weaker than `BeaconVoteValidator`.** No slashing-protection call, no comparison of `payload_present` / `blob_data_available` against local BN view. This is intentional per SIP-94 SC-2: at the 75% slot deadline, operator BNs may legitimately disagree on envelope arrival, and a "match local view" gate would amplify normal BN drift into cluster-level QBFT deadlocks even with an honest leader. PTC is non-slashable, so the worst case from leader malice is a wasted cluster signature, bounded.

**Mitigation:** the value check matches the SIP as-written. An [open thread on SIP-94 line 138](https://github.com/ssvlabs/SIPs/pull/94#discussion_r3265815226) between Diego and Shane discusses whether to strengthen this; if the SIP eventually demands strict equality, a follow-up PR adds the second rule (the `do_validation -> Result` shape this PR uses is exactly the extension point).

**Validator boilerplate.** The single-rule check could be inlined in `validate` directly. The `do_validation` wrapper is preserved deliberately to match the sibling-validator precedent and keep tests calling the same shape (`do_validation` + `matches!`) used elsewhere in the file.

## Validation

- `cargo test -p ssv_types` — 86 unit + 2 doctests pass (6 new + 80 existing); no regressions.
- `make cargo-fmt && make cargo-fmt-check` — clean.
- `make lint` — clean (workspace-wide clippy).
- `cargo check --workspace` — clean.

Test coverage for the new code:
- SSZ roundtrip across all 4 `(payload_present, blob_data_available)` combinations.
- Wire-byte layout: fixed 34 bytes (32-byte root + 2 bools); position-by-position assertions.
- Hash determinism: same inputs → same hash (two independently constructed values, mirroring `SelectionProofBatchId` precedent); different root or flipped boolean → different hash.
- Validator zero-root rejection: `do_validation` returns `Err(ZeroBeaconBlockRoot)`.
- Validator non-zero acceptance: 4-combo loop documents that payload-status flags are not validated.
- Validator ignores `start_value`: trait-surface guard for the SC-2 trust-leader contract.

## Rollback

Single-file additive change. Revert removes the type, validator, error enum, and tests. No production consumers yet, so revert has zero behavior impact.

## Blockers / Dependencies

None for merge. Downstream consumers (under milestone #4):
- **#1035** `feat(qbft_manager): wire PTC committee-scoped QBFT instances and fork gate` — consumes `PayloadAttestationVote` + `PayloadAttestationVoteValidator`.
- **#1037** `feat(validator_store): implement sign_payload_attestation` — signs the decided value.

N/A

### DIFF
--- a/anchor/common/ssv_types/src/consensus.rs
+++ b/anchor/common/ssv_types/src/consensus.rs
@@ -906,6 +906,27 @@ impl QbftData for BeaconVote {
     }
 }
 
+#[derive(Clone, Debug, TreeHash, PartialEq, Eq, Encode, Decode)]
+#[cfg_attr(feature = "arbitrary-fuzz", derive(arbitrary::Arbitrary))]
+pub struct PayloadAttestationVote {
+    pub beacon_block_root: Hash256,
+    pub payload_present: bool,
+    pub blob_data_available: bool,
+}
+
+impl QbftData for PayloadAttestationVote {
+    type Hash = Hash256;
+
+    fn hash(&self) -> Self::Hash {
+        let bytes = self.as_ssz_bytes();
+
+        let mut hasher = Sha256::new();
+        hasher.update(bytes);
+        let hash: [u8; 32] = hasher.finalize().into();
+        Hash256::from(hash)
+    }
+}
+
 /// Identifies a batch of pre-consensus selection proofs for a committee.
 /// All operators compute the same hash for a given `(slot, committee_id)` pair, ensuring
 /// consistent batching across the network.
@@ -1179,6 +1200,48 @@ pub enum BeaconVoteValidationError {
     EpochMismatch(Box<EpochMismatch>),
     #[error("Attestation would be slashable: {0}")]
     SlashableAttestation(NotSafe),
+}
+
+#[derive(Error, Debug)]
+pub enum PayloadAttestationVoteValidationError {
+    #[error("Beacon block root is zero")]
+    ZeroBeaconBlockRoot,
+}
+
+/// Validator for `PayloadAttestationVote` during QBFT consensus.
+///
+/// Per SIP-94 SC-2, `payload_present` and `blob_data_available` are trusted
+/// from the QBFT leader and intentionally not compared against any local view.
+/// The only required check is that `beacon_block_root` is non-zero.
+#[derive(Debug, Default)]
+pub struct PayloadAttestationVoteValidator;
+
+impl QbftDataValidator<PayloadAttestationVote> for PayloadAttestationVoteValidator {
+    fn validate(
+        &self,
+        value: &PayloadAttestationVote,
+        _our_value: &PayloadAttestationVote,
+    ) -> bool {
+        match self.do_validation(value) {
+            Ok(_) => true,
+            Err(err) => {
+                warn!(%err, "Operator proposed invalid payload attestation vote");
+                false
+            }
+        }
+    }
+}
+
+impl PayloadAttestationVoteValidator {
+    pub fn do_validation(
+        &self,
+        value: &PayloadAttestationVote,
+    ) -> Result<(), PayloadAttestationVoteValidationError> {
+        if value.beacon_block_root.is_zero() {
+            return Err(PayloadAttestationVoteValidationError::ZeroBeaconBlockRoot);
+        }
+        Ok(())
+    }
 }
 
 #[cfg(test)]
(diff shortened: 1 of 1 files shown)


shane-moore:
@claude, please take a look

diegomrsantos:
Can you add `///` docs for `PayloadAttestationVote`, its fields, and `PayloadAttestationVoteValidationError`? A short description of the vote shape and the zero-root rejection would make the contract clearer where it is defined.

anchor/common/ssv_types/src/consensus.rs
```diff
@@ -1181,6 +1202,48 @@ pub enum BeaconVoteValidationError {
...
+/// from the QBFT leader and intentionally not compared against any local view.
+/// The only required check is that `beacon_block_root` is non-zero.
+#[derive(Debug, Default)]
+pub struct PayloadAttestationVoteValidator;
+
+impl QbftDataValidator<PayloadAttestationVote> for PayloadAttestationVoteValidator {
+    fn validate(
+        &self,
+        value: &PayloadAttestationVote,
+        _our_value: &PayloadAttestationVote,
```
diegomrsantos, line 1244: Can we treat this value-check behavior as still unsettled until the SIP-94 PTC discussion resolves? This PR bakes in the current model by ignoring `_our_value` and testing that divergent local values are accepted. That may be the right design, but the SIP discussion is still weighing leader observation vs local observation, especially because one `PayloadAttestationVote` can drive all local PTC validators for the slot.

I think we should either wait for that decision or make the PR text clear that this is provisional and may change with the SIP.

Relevant thread: https://github.com/ssvlabs/SIPs/pull/94#discussion_r3265815226

anchor/common/ssv_types/src/consensus.rs
```diff
@@ -2046,4 +2109,123 @@ mod tests {
...
+        // calling `validate(value, our_value)` with divergent values and asserting
+        // acceptance; structurally also guaranteed by `do_validation` only taking
+        // `value`.
+        let validator = PayloadAttestationVoteValidator;
+        let proposed =
+            create_payload_attestation_vote(Hash256::from_low_u64_be(0x1111), true, true);
+        let local_view =
+            create_payload_attestation_vote(Hash256::from_low_u64_be(0x2222), false, false);
+
+        assert!(validator.validate(&proposed, &local_view));
```
diegomrsantos, line 2250: I think this test is a bit too strong while the SIP-94 PTC value-check behavior is still being discussed.

`test_payload_attestation_vote_validator_ignores_start_value` does not only test the current implementation. It pins the rule that the validator must accept a proposed value even when the local value has different `payload_present`, `blob_data_available`, and `beacon_block_root` fields. That may be where the SIP lands, but it is the same leader-observation vs local-observation choice still under discussion.

Can we either remove this test for now, or reword it so it clearly documents current draft behavior rather than a final contract? If the intended rule is only about not comparing the payload-status booleans, I would also keep `beacon_block_root` the same in `local_view`; otherwise the test is pinning root mismatch as a separate accepted behavior too.

shane-moore:
> Can you add `///` docs for `PayloadAttestationVote`, its fields, and `PayloadAttestationVoteValidationError`? A short description of the vote shape and the zero-root rejection would make the contract clearer where it is defined.

done in [15769b1](https://github.com/sigp/anchor/commit/15769b136bfee6e31c50831e8e3546d01f371865): struct-level `///` on the stripped shape + post-consensus signing context, per-field docs on all three fields, summary `///` on `PayloadAttestationVoteValidationError`, and a variant-level `///` on `ZeroBeaconBlockRoot` explaining the zero-hash rejection.

anchor/common/ssv_types/src/consensus.rs
shane-moore: went with the "mark provisional" path in [15769b1](https://github.com/sigp/anchor/commit/15769b136bfee6e31c50831e8e3546d01f371865). validator doc-comment now opens with `Reflects [SIP-94 §3]` pinned to commit [7e8b5bd](https://github.com/ssvlabs/SIPs/blob/7e8b5bd6d4007682d8bd75b06a2f2ac7b617e9e5/sips/epbs_support.md#3-new-duty-payload-timeliness-committee-ptc-attestation), with a closing paragraph naming the `do_validation -> Result` extension point if the SIP later requires local-view equality on the payload-status booleans. paired with the test-side narrowing on the other thread, the code now only pins what §3 currently says.

intentionally stronger than a generic "may change" disclaimer: future readers can diff the live SIP against `7e8b5bd` to see exactly what's changed since this code was written. the §3 sentence i added in [0feaf57](https://github.com/ssvlabs/SIPs/commit/0feaf57ae75eeaf3aa370d9c3d3637774ddae91c) on the correlated-observations-within-a-cluster point is part of that pinned snapshot.

the type shape is independent of the rule and adding the second `do_validation` clause is a few-line follow-up if §3 flips.

anchor/common/ssv_types/src/consensus.rs
shane-moore: took the narrowing path in [15769b1](https://github.com/sigp/anchor/commit/15769b136bfee6e31c50831e8e3546d01f371865). renamed to `test_payload_attestation_vote_validator_accepts_status_flag_disagreement`, held `beacon_block_root` constant between `proposed` and `local_view`, and only the payload-status booleans differ. test comment is explicit: "Root-mismatch behavior is intentionally not pinned here; the trait surface (`do_validation` taking only `value`) and the zero-root rejection test cover the root contract."

so the test only pins what SC-2 of the pinned SIP-94 snapshot actually says (trust-leader on the payload-status booleans), not the broader "ignores all of start_value" framing the old test name implied.

diegomrsantos (approved):
Looks good now. The docs make the SIP-94 assumption explicit and pinned, and the status-flag disagreement test no longer pins root mismatch behavior.

PR #1052, opened 2026-05-21
fix: batch pre-Boole sync contribution proofs

### LINKED ISSUE #1050: v1.2.3: `InvalidPartialSignatureTypeCount` outbound-drops on multi-subnet sync-committee aggregators
## Summary

When a validator is selected as a sync-committee aggregator on K subcommittees in the same slot, Anchor's outbound message validator drops K-1 of the K `ContributionProofs` partial-signature messages with `err=InvalidPartialSignatureTypeCount`. The bug is internal to Anchor: the producer sends K separate `SignedSSVMessage` envelopes (one per subcommittee), while the validator correctly expects the canonical wire format of *one* envelope per `(validator, slot)` with K entries inside. Closing the loop on the spec, the validator is right and the producer is wrong.

## Reproduction

Observed on Anchor v1.2.3 (`22057222`) in a 4-of-4 SSV/Anchor mixed committee on `ssv-mini`, 10 validators, sync subnet count = 4, Fulu at genesis.

```
WARN Validation of outgoing message failed (Reject) err=InvalidPartialSignatureTypeCount {
  got: "pre-consensus, having MessageCounts { pre_consensus: 1, proposal: 0, prepare: 0, commit: 0, round_change: 0, post_consensus: 0 }"
}
```

Counts per Anchor node over 16 minutes:
- **2490** `InvalidPartialSignatureTypeCount` outbound-rejections — fires every slot at `slot_start + 8s` (the 2/3-of-slot mark, when the sync-committee aggregator path runs). 30 events per slot = 10 validators × (4 subnets − 1 success).
- **320** `Signing sync selection proof timed out` warnings concentrated in a 7-slot warmup window (first timeout 12:44:43Z, last 12:46:07Z) — direct downstream consequence: the K-1 dropped messages prevent the cluster from reaching `2f+1` quorum until the gossip mesh fully peers.

After the cluster warms up, sync contributions submit successfully (`INFO Successfully published sync contributions subnet=0 ... num_signers=112` from slot 22 onwards, 216 such events per node) thanks to the other peers' spec-bundled envelopes covering the lost subnets, but Anchor still drops 30 of its own messages per slot indefinitely.

## Root cause

`MessageCounts::validate_partial_signature_message` in [`anchor/message_validator/src/message_counts.rs`](https://github.com/sigp/anchor/blob/v1.2.3/anchor/message_validator/src/message_counts.rs):

```rust
const MAX_MESSAGES_PER_ROUND: u8 = 1;

pub fn validate_partial_signature_message(&self, messages: &PartialSignatureMessages) -> Result<(), ValidationFailure> {
    match messages.kind {
        PartialSignatureKind::RandaoPartialSig
        | PartialSignatureKind::SelectionProofPartialSig
        | PartialSignatureKind::ContributionProofs   // ← multi-subnet kind
        | PartialSignatureKind::ValidatorRegistration
        | PartialSignatureKind::VoluntaryExit => {
            if self.pre_consensus >= MAX_MESSAGES_PER_ROUND {
                return Err(ValidationFailure::InvalidPartialSignatureTypeCount { ... });
            }
        }
        ...
    }
}
```

The counter is indexed per `(MessageId, slot, signer)` — and for outbound validation, where the signer is always the local Anchor operator, this reduces to one envelope per `(validator, slot)`. The validator enforces one envelope, but Anchor's `validator_store` path produces K envelopes (one per subcommittee), driven by Lighthouse's `SyncCommitteeService` calling `produce_sync_selection_proof` once per `(validator, subnet)` per slot.

The spec wire format is one envelope per `(validator, slot)` bundling K `PartialSignatureMessage` entries:
- [`ssv-spec/ssv/sync_committee_aggregator.go:346-369`](https://github.com/ssvlabs/ssv-spec/blob/v1.2.2/ssv/sync_committee_aggregator.go#L346-L369) — reference runner builds one envelope, accumulates K entries in `Messages`, broadcasts once.
- [`ssv-spec/types/testingutils/ssv_msgs_sync_committee_contribution.go:79-113`](https://github.com/ssvlabs/ssv-spec/blob/v1.2.2/types/testingutils/ssv_msgs_sync_committee_contribution.go#L79-L113) (the `contributionProofMsg` helper) — same bundling (loop appends K messages, single envelope constructed at lines 107-111).
- [SSV-Go runner](https://github.com/ssvlabs/ssv/blob/0fa3b4a64/protocol/v2/ssv/runner/sync_committee_contribution.go#L514-L553) — same bundling. (SSV's incoming envelopes never trigger any Anchor-side rejection in our test logs.)

Anchor's *validator* is consistent with the spec; Anchor's *producer* is not.

## Related work

Closely-analogous closed issue [#324](https://github.com/sigp/anchor/issues/324) (`ExcessiveDutyCount`), fixed in [#328](https://github.com/sigp/anchor/pull/328). #328 took a validator-side approach ("only count the first message of a new duty"). The fix didn't generalize to `ContributionProofs`. This is the same class of bug as #324, surfacing for a different `PartialSignatureKind` than the one #328 patched.

## Proposed fix

Two approaches that would each close the bug; the team may prefer one for consistency with #328's precedent.

**Validator-side (matches #328's precedent, easier to land):** Change `MessageCounts.pre_consensus` to a per-`(validator, slot, subcommittee_index)` counter for the `ContributionProofs` kind only — or equivalently, accept up to `SYNC_COMMITTEE_SUBNET_COUNT` `ContributionProofs` envelopes per `(validator, slot)` instead of 1. Smallest change set; same shape as #328's "only count the first message of a new duty" approach.

**Producer-side (spec-aligned, larger refactor):** Modify Anchor's sync-committee aggregator path so K subcommittee selection proofs are accumulated into a single `PartialSignatureMessages` envelope per `(validator, slot)` and broadcast once. The aggregation point would be in `SignatureCollectorManager` or at a new per-slot accumulator boundary, since `produce_sync_selection_proof` itself is called externally by Lighthouse's `SyncCommitteeService` (pinned via `Cargo.lock` at `git+https://github.com/sigp/lighthouse?rev=e3ee7feb`) once per `(validator, subnet)` and can't naively change shape without crossing the `validator_services` boundary. This matches the spec and matches SSV-Go's behaviour, but is structurally more invasive.

Either is acceptable from the cross-client correctness standpoint. I lean toward whichever is consistent with #328's precedent.

## Side observations (out of scope, flagged for visibility)

While analyzing the outgoing-validation failures, two more producer-vs-validator inconsistencies surface — possibly worth separate issues:

- `err=SlotAlreadyAdvanced` — **996 (node 0) / 1006 (node 1)** outbound rejections in the same 16-minute window. Likely a race between Anchor's producer building a message for a recently-passed slot and the outbound validator's slot-advancement check.
- `err=DecidedWithSameSigners` — **24 (node 0) / 26 (node 1)** outbound rejections. Edge case; low impact.

## References
- `MessageCounts`: [`anchor/message_validator/src/message_counts.rs`](https://github.com/sigp/anchor/blob/v1.2.3/anchor/message_validator/src/message_counts.rs)
- Outbound drop path: [`anchor/message_sender/src/network.rs:133-149`](https://github.com/sigp/anchor/blob/v1.2.3/anchor/message_sender/src/network.rs#L133-L149)
- Sync-committee aggregator caller: [`anchor/validator_store/src/lib.rs:1337-1383`](https://github.com/sigp/anchor/blob/v1.2.3/anchor/validator_store/src/lib.rs#L1337-L1383) (the 8s deadline at line 1356)
- ssv-spec reference runner: [`ssv-spec/ssv/sync_committee_aggregator.go:346-369`](https://github.com/ssvlabs/ssv-spec/blob/v1.2.2/ssv/sync_committee_aggregator.go#L346-L369)
- SSV-Go runner: [`protocol/v2/ssv/runner/sync_committee_contribution.go:514-553`](https://github.com/ssvlabs/ssv/blob/0fa3b4a64/protocol/v2/ssv/runner/sync_committee_contribution.go#L514-L553)
- Related: [#324](https://github.com/sigp/anchor/issues/324), [#328](https://github.com/sigp/anchor/pull/328)

diegomrsantos:

Fixes #1050.

Before Boole, Lighthouse calls `produce_sync_selection_proof` once per sync committee subnet. When the same validator is assigned to multiple sync subnets in the same slot, Anchor was producing one SSV `ContributionProofs` partial-signature envelope per subnet.

SSV treats `ContributionProofs` as a pre-consensus message, so peers expect one envelope per signer and slot. The Go SSV producer batches the per-subnet selection-proof roots into one `ContributionProofs` envelope. This PR now fixes the producer shape instead of relaxing message validation. The final diff keeps the validator-side single-envelope validation behavior unchanged.

## Change Overview (Required)

- Adds a single-validator batching mode to `signature_collector` for cases where one validator signs multiple roots that must be sent in one outgoing partial-signature message.
- Uses that batching mode for pre-Boole sync selection proof production.
- Derives the expected batch size from `VotingAssignments` and checks that the requested subnet belongs to the validator's assigned sync subnets for the slot.
- Dedupe local batch entries by `(validator_index, signing_root)` so duplicate signing calls for the same subnet/root cannot fill the batch early.
- Retains a completed batch marker until normal slot cleanup so a retry after the envelope is sent cannot create a second `ContributionProofs` envelope for the same validator/slot batch.
- Leaves the Boole committee batching path unchanged.
- Adds focused regression coverage at both layers: the validator-store producer chooses batching for a multi-subnet validator, rejects unassigned subnet requests before calling the collector, and the signature collector emits one decoded `ContributionProofs` envelope with all unique per-subnet roots through the normal `sign_and_collect` path.

The main risk is waiting on the wrong local batch size if the voting assignments are stale or inconsistent with the request. The code now derives the count from the cached `VotingAssignments`, confirms the validator is assigned to the requested subnet, stores the expected count in the collector batch, and still uses the existing within-slot timeout.

The local batch key uses the slot/committee batch id together with the validator public key, rather than the signing root, because each subnet has a different signing root but must be sent in the same `ContributionProofs` envelope.

If a later call reaches the same batch with a different expected count, the collector logs the mismatch and does not send a malformed partial-signature envelope.

No config, storage, or migration changes are required.

## Validation (Required)

I validated the changed producer and collector paths with focused tests. The validator-store tests cover multi-subnet batching and the unassigned-subnet rejection before the collector is called. The signature collector tests cover the normal `sign_and_collect` path and decode the outbound `ContributionProofs` envelope to confirm it contains one message with the expected unique roots. The message-validator tests also pass, which confirms this PR keeps the existing single-envelope validation behavior unchanged.

Formatting, dependency ordering, whitespace checks, and clippy for the changed crates passed locally. The commit hook also ran workspace formatting, workspace clippy, and cargo-sort checks successfully.

Revert these commits to restore the previous per-subnet producer behavior. No config, storage, or migration rollback is needed.

N/A

N/A

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7223,6 +7223,7 @@ dependencies = [
  "fork",
  "futures",
  "message_sender",
+ "parking_lot",
  "processor",
  "rand 0.9.2",
  "slot_clock",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5561,6 +5561,7 @@ dependencies = [
  "serde_json",
  "tempfile",
  "thiserror 2.0.18",
+ "unicode-normalization",
  "zeroize",
 ]
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7213,17 +7213,20 @@ dependencies = [
 name = "signature_collector"
 version = "0.1.0"
 dependencies = [
+ "async-channel 1.9.0",
  "bls",
  "bls_lagrange",
  "dashmap",
  "database",
  "ethereum_ssz",
  "fork",
+ "futures",
  "message_sender",
  "processor",
  "rand 0.9.2",
  "slot_clock",
  "ssv_types",
+ "task_executor",
  "thiserror 2.0.18",
  "tokio",
  "tracing",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -174,6 +174,7 @@ tracing-subscriber = { version = "0.3.18", features = ["fmt", "env-filter"] }
 tree_hash = "0.12.0"
 tree_hash_derive = "0.12.0"
 typenum = "1.18"
+unicode-normalization = "0.1.25"
 zeroize = "1.8.1"
 
 [patch.crates-io]
--- a/anchor/cli/src/cli.rs
+++ b/anchor/cli/src/cli.rs
@@ -619,7 +619,10 @@ pub struct Node {
         help = "Enable parallel querying and scoring of attestation data across multiple beacon \
                 nodes. When enabled, Anchor queries all configured beacon nodes simultaneously \
                 and selects the attestation data with the highest score based on checkpoint \
-                epochs and head block proximity. Only useful with multiple beacon nodes.",
+                epochs and head block proximity. Only useful with multiple beacon nodes. \
+                Note: WAD is bypassed when the beacon head monitor (enabled by default, see \
+                --disable-beacon-head-monitor) wins the eager-attest race for a slot; the \
+                firing BN is queried directly. WAD still runs on the fallback path.",
         display_order = 0
     )]
     pub with_weighted_attestation_data: bool,
--- a/anchor/cli/src/cli.rs
+++ b/anchor/cli/src/cli.rs
@@ -537,6 +537,15 @@ pub struct Node {
 
     #[clap(
         long,
+        help = "Disable the beacon head monitor which tries to attest as soon as any of the \
+                configured beacon nodes sends a head event. Leaving the service enabled is \
+                recommended, but disabling it can lead to reduced bandwidth and more predictable \
+                usage of the primary beacon node (rather than the fastest BN)."
+    )]
+    pub disable_beacon_head_monitor: bool,
+
+    #[clap(
+        long,
         help = "Disable slashing protection for all validator clients. DO NOT ENABLE THIS UNLESS YOU HAVE A MORE THAN SUFFICIENT REASON TO",
         hide = true,
         display_order = 0
--- a/anchor/client/src/config.rs
+++ b/anchor/client/src/config.rs
@@ -71,6 +71,8 @@ pub struct Config {
     pub prefer_builder_proposals: bool,
     /// Controls whether the latency measurement service is enabled
     pub disable_latency_measurement_service: bool,
+    /// Enables the beacon head monitor that reacts to head updates from connected beacon nodes.
+    pub enable_beacon_head_monitor: bool,
     /// Enable operator doppelgänger protection (blocks messages and monitors for twins)
     pub operator_dg: bool,
     /// Number of epochs to monitor for twins after grace period
@@ -124,6 +126,7 @@ impl Config {
             prefer_builder_proposals: false,
             gas_limit: 36_000_000,
             disable_latency_measurement_service: false,
+            enable_beacon_head_monitor: true,
             operator_dg: false,
             operator_dg_wait_epochs: 2,
             strict_mfp: false,
@@ -281,6 +284,7 @@ pub fn from_cli(mut cli_args: Node, global_config: GlobalConfig) -> Result<Confi
 
     config.impostor = cli_args.impostor.map(OperatorId);
     config.disable_latency_measurement_service = cli_args.disable_latency_measurement_service;
+    config.enable_beacon_head_monitor = !cli_args.disable_beacon_head_monitor;
 
     // Operator doppelgänger protection
     config.operator_dg = cli_args.operator_dg;
--- a/anchor/client/src/lib.rs
+++ b/anchor/client/src/lib.rs
@@ -18,7 +18,8 @@ use anchor_validator_store::{
     registration_service::RegistrationService,
 };
 use beacon_node_fallback::{
-    BeaconNodeFallback, CandidateBeaconNode, start_fallback_updater_service,
+    BeaconNodeFallback, CandidateBeaconNode, beacon_head_monitor::HeadEvent,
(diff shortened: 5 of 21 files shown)


shane-moore:
Superseded by #1201, merged 2026-08-05 (`581bde12`), now in both `unstable` and `epbs`.

#1201 covers the same producer path but corrects the batching semantics. This PR deduped local batch entries by `(validator_index, signing_root)`; go-ssv and ssv-spec require one entry per raw sync-committee position, so multiplicity must be preserved rather than deduped. Positions `[0, 1, 128]` are two unique subnets but three envelope entries.

Closing in favour of the merged work. #1050 remains open only because the `Closes` keyword did not fire against the default branch.

PR #1053, opened 2026-05-22
feat(validator_store): publish PTC validator set on VotingAssignments

### LINKED ISSUE #1036: feat(validator_store): add PTC voting context phase to metadata_service
## Goal

Add a new metadata-service phase loop in `anchor/validator_store/src/metadata_service.rs` that pre-fetches `PayloadAttestationData` from the BN shortly before LH's `PayloadAttestationService` starts QBFT at 75% slot. The loop publishes the fetched context on a watch channel that `sign_payload_attestation` (follow-up issue) reads as its QBFT initial value. Modeled on the existing Phase-2 voting context loop in `metadata_service.rs`.

## Context

Per SIP-94 §3 ([ssvlabs/SIPs#94](https://github.com/ssvlabs/SIPs/pull/94)), PTC requires one QBFT instance per cluster per slot. Each operator needs an initial value for that instance; LH's `PayloadAttestationService` fetches per-operator inside its own loop, which gives Anchor no place to stage that initial value at the metadata layer.

This phase produces the operator-local initial value by pre-fetching `PayloadAttestationData` through that operator's `BeaconNodeFallback`. Honest operators can still disagree on `payload_present` / `blob_data_available` near the 75% deadline because their BNs observe envelope and blob data at slightly different times; cluster coherence comes from QBFT deciding one value, not from the metadata fetch itself. Mirrors the §2 attestation pattern (per-operator pre-fetch into `metadata_service`).

When the BN has not yet seen a block for the slot, the honest behavior is to skip the duty (consensus-specs Gloas `validator.md`: "If the validator has not seen any beacon block for the assigned slot, do not submit"). The phase surfaces this to `sign_payload_attestation` via a sentinel on the watch channel.

## Suggested approach

`anchor/validator_store/src/metadata_service.rs`:
- Add a new phase loop (Phase-4-style) that fires shortly before `chain_spec.get_payload_attestation_due()` (default close to 75% slot, with margin for a BN round-trip; implementer tunes).
- Fetch via `BeaconNodeFallback::first_success(|bn| bn.get_validator_payload_attestation_data(slot))`. Falls back to the next BN on failure.
- Publish to a new `PayloadAttestationContext` watch channel carrying `beacon_block_root`, `payload_present`, `blob_data_available`.
- When the BN returns no block for the slot, publish a "skip duty" sentinel that the follow-up `sign_payload_attestation` recognizes (sentinel shape is an open question, see below).
- Phase-1/2/3 loops should be untouched.

## Acceptance criteria

- New phase loop fires at the configured slot fraction (close to LH's 75% PTC service start, with margin for BN round-trip).
- BN fetch via `BeaconNodeFallback::first_success`; falls back to the next BN on failure.
- Watch channel publishes a `PayloadAttestationContext` per slot.
- "No block seen" BN response surfaces as a "skip duty" sentinel on the watch channel.
- Existing Phase-1/2/3 loops unaffected.
- Integration tests using a mocked BN:
  - `ptc_phase_publishes_on_bn_success`: BN returns `PayloadAttestationData`; slot clock advances; assert the watch channel publishes within slot.
  - `ptc_phase_publishes_skip_on_bn_none`: BN returns nothing for the slot; assert the watch channel publishes the "skip duty" sentinel.

## Open questions

- **"Skip duty" sentinel shape.** Options include `Option<PayloadAttestationContext>` (None = skip), an enum like `PtcSlotState { Vote(ctx), Skip }`, or a flag on the context struct. Decide before publishing the type so the follow-up `sign_payload_attestation` PR can pattern-match cleanly.
- **Phase ordering against existing loops.** Phases 1/2/3 fire at different slot fractions; Phase 4 needs to fire before 75%. Confirm against the existing loop structure in `metadata_service.rs` whether the new phase runs as an independent task or sequences after Phase 3.
- **BN-None vs BN-error distinction.** SIP-94 references the validator-spec "skip if no block seen" rule. Check the LH BN client (`eth2/src/lib.rs::get_validator_payload_attestation_data`) for whether "no block" surfaces as `None`, an explicit error variant, or both — the sentinel handling depends on it.

## Risks

- **Timing tuning.** Fetch too early and envelope arrival is uncertain; too late and the cluster's QBFT round budget shrinks below the 75% deadline. Default ~70% slot is a starting point; measure on ssv-mini before locking.
- **BN fallback failure.** If all configured BNs fail for the slot, the cluster cannot vote PTC. Missed reward only, not slashing. Surface via a metric counter so operators can spot a degraded BN fleet.

shane-moore:

Closes #1036. Next ePBS milestone item after PR #1033 (`Role::PTCCommittee` + message-validator handling), alongside PR #1047 (`PayloadAttestationVote` SSZ container).

Per [SIP-94](https://github.com/ssvlabs/SIPs/pull/94) §3 and the [Gloas validator spec](https://github.com/ethereum/consensus-specs/blob/dev/specs/gloas/validator.md#payload-attester-duties), each PTC-assigned validator signs `PayloadAttestationData` under `DOMAIN_PTC_ATTESTER`. In SSV, a cluster runs one QBFT instance per slot and each locally-assigned PTC validator contributes a partial signature on the decided data. The downstream `sign_payload_attestation` path (#1037) needs the per-cluster partial-signature batch size at signing time, mirroring how `sign_committee_attestations` uses `voting_message_count_for_committee` at \`lib.rs:1534-1537\` for the attestation batch.

This PR publishes that count input on \`VotingAssignments\` so #1037 can size its batch with the same closure-based filter pattern.

## Change Overview

Additive change across four files, +122/-3. All work lives in \`anchor/validator_store\`.

- **\`VotingAssignments::ptc_validators: Vec<ValidatorIndex>\`** — new field populated at slot start during Phase 1, mirroring \`attesting_validators\` exactly.
- **\`ptc_signature_count_for_committee(F)\`** — count helper on \`VotingAssignments\`, closure-based filter matching sibling \`voting_message_count_for_committee\`. PTC stays in a separate helper because the QBFT instance is separate from attestation.
- **Fork-gated wire-up in \`update_voting_assignments\`** — first \`>= Fork::CStar\` site in the codebase; mirrors the existing \`>= Fork::Boole\` precedent at \`metadata_service.rs:501\`. Pre-CStar slots get an empty vec.
- **Pure helper \`build_ptc_validators\`** — intersects LH's epoch-cached \`PtcMap\` with the local validator set; filters liquidated clusters via \`.is_ok()\` on \`get_validator_and_cluster\` (same pattern as the existing Phase-2/3 builder at \`metadata_service.rs:421-428\`).
- **\`METADATA_SERVICE_PTC_VALIDATORS\` gauge** — sibling of existing attester/sync gauges.
- **Empty-assignments counter tightened** — now requires \`attester_count == 0 && sync_count == 0 && ptc_count == 0\` so a PTC-only slot post-CStar isn't miscounted as empty.

**Reading order:** start at \`validator_store/src/lib.rs\` (new field + count helper on \`VotingAssignments\`); then \`metadata_service.rs\` (fork-gated lookup + the pure helper in the helper bay); unit tests in the existing \`#[cfg(test)] mod tests\` block mirror \`test_aggregators_sorted_by_validator_index\`.

**Intentionally unchanged:**
- No new metadata-service phase, no new watch channel, no new spawned task, no BN fetch. Issue #1036 was rescoped after re-reading the attestation precedent at \`lib.rs:1461-1497\`: LH's \`sign_payload_attestation(pubkey, data)\` passes every field of \`PayloadAttestationVote\`, so #1037 can build the QBFT initial value directly without a metadata-level BN fetch. The honest-skip case is filtered by LH itself before the trait is invoked.
- Phase-1/2/3 loops untouched outside the single Phase-1 insert.
- \`voting_message_count_for_committee\` and the attestation count path unchanged.

**Slot-start lookup cost.** Adds ~30 O(1) hashmap reads per slot for a typical operator (1-10 PTC duties, three \`database.state()\` reads each via \`get_validator_and_cluster\`). Cheaper than the per-validator \`SyncSubnetId::compute_subnets_for_sync_committee::<E>(...)\` already running in Phase 1. The alternative (separate 9s-fraction fetch phase) was rejected because PTC duties are epoch-stable and the source is an in-memory \`PtcMap\` read.

**Flat vec vs per-cluster grouping.** \`ptc_validators\` is flat across clusters, mirroring \`attesting_validators\`. The attestation precedent at \`lib.rs:1534-1537\` shows the flat-vec + closure-filter pattern is sufficient for per-cluster batch sizing; #1037 mirrors it. If a non-count consumer ever emerges, a refactor is one field-shape change with the call site as the regression test.

**Empty-assignments counter behavior.** Tightening to include \`ptc_count == 0\` makes the counter strictly more accurate — it fires only when no duties exist at the slot. Not noisier during early CStar rollout; the prior condition would have falsely counted PTC-only slots as empty post-fork.

## Validation

- \`cargo test -p anchor_validator_store\` — 36/36 pass (3 new PTC tests + 33 existing).
- \`make cargo-fmt && make cargo-fmt-check\` — clean.
- \`make lint\` — clean.
- \`cargo check --workspace\` — clean.

New test coverage (pure-helper tests, no tokio runtime, no \`MetadataService\` instance):
- \`test_build_ptc_validators_empty_input\` — empty input returns empty Vec.
- \`test_build_ptc_validators_all_local\` — all-\`true\` predicate retains every duty with correct \`u64 → usize\` conversion.
- \`test_build_ptc_validators_filters_liquidated\` — predicate returns false for one pubkey (simulating \`ClusterLiquidated\`); excluded duty is dropped.

## Rollback

Additive change on a metadata-only struct. Revert removes the field, helper method, fork-gated lookup, pure helper, gauge, and tests. Pre-CStar behavior is unchanged (empty vec until fork activation), so revert pre-CStar is a no-op. Post-CStar revert removes the input #1037 reads, but #1037 has not landed yet, so no downstream impact.

## Blockers / Dependencies

None for merge. Downstream consumer (milestone #4):
- **#1037** \`feat(validator_store): implement sign_payload_attestation\` — first consumer of \`ptc_validators\`. Will mirror \`sign_committee_attestations\`'s shape and use \`ptc_signature_count_for_committee\` to size the partial-sig batch per cluster.

N/A

### DIFF
--- a/anchor/validator_store/src/lib.rs
+++ b/anchor/validator_store/src/lib.rs
@@ -2014,7 +2014,7 @@ impl VotingAssignments {
     {
         self.ptc_validators
             .iter()
-            .filter(|idx| is_in_committee(idx))
+            .filter(|&idx| is_in_committee(idx))
             .count()
     }
 }
@@ -3342,6 +3342,17 @@ mod tests {
         // Voting message: 2 attesting + 2 sync = 4
         let message_count = voting_assignments.voting_message_count_for_committee(all_in_committee);
         assert_eq!(message_count, 4);
+    }
+
+    #[test]
+    fn test_ptc_signature_count_with_filter() {
+        let mut voting_assignments = create_test_voting_assignments(vec![], vec![]);
+        voting_assignments.ptc_validators =
+            vec![ValidatorIndex(1), ValidatorIndex(2), ValidatorIndex(3)];
+
+        let in_committee = |idx: &ValidatorIndex| matches!(idx.0, 1 | 3);
+        let count = voting_assignments.ptc_signature_count_for_committee(in_committee);
+        assert_eq!(count, 2);
     }
 
     #[tokio::test]
--- a/anchor/validator_store/src/lib.rs
+++ b/anchor/validator_store/src/lib.rs
@@ -1927,8 +1927,7 @@ pub struct VotingAssignments {
     /// Sync committee validators mapped to their subnet IDs.
     /// A validator may participate in multiple subnets.
     pub sync_validators_by_subnet: HashMap<ValidatorIndex, HashSet<SyncSubnetId>>,
-    /// The indices of local validators with a PTC duty in this slot. Empty when the operator has
-    /// no local PTC duties in this slot.
+    /// The indices of local validators with a PTC duty in this slot.
     pub ptc_validators: Vec<ValidatorIndex>,
 }
 
--- a/anchor/validator_store/src/lib.rs
+++ b/anchor/validator_store/src/lib.rs
@@ -1927,6 +1927,9 @@ pub struct VotingAssignments {
     /// Sync committee validators mapped to their subnet IDs.
     /// A validator may participate in multiple subnets.
     pub sync_validators_by_subnet: HashMap<ValidatorIndex, HashSet<SyncSubnetId>>,
+    /// The indices of local validators with a PTC duty in this slot. Empty when the operator has
+    /// no local PTC duties in this slot.
+    pub ptc_validators: Vec<ValidatorIndex>,
 }
 
 impl VotingAssignments {
@@ -1998,6 +2001,21 @@ impl VotingAssignments {
         }
 
         count
+    }
+
+    /// Counts expected PTC partial signatures for a given cluster.
+    ///
+    /// PTC runs a separate QBFT instance from attestation/sync, so its batch
+    /// size is tallied independently rather than being folded into
+    /// `voting_message_count_for_committee`.
+    pub fn ptc_signature_count_for_committee<F>(&self, is_in_committee: F) -> usize
+    where
+        F: Fn(&ValidatorIndex) -> bool,
+    {
+        self.ptc_validators
+            .iter()
+            .filter(|idx| is_in_committee(idx))
+            .count()
     }
 }
 
@@ -3201,6 +3219,7 @@ mod tests {
                     )
                 })
                 .collect(),
+            ptc_validators: Vec::new(),
         }
     }
 
@@ -3355,6 +3374,7 @@ mod tests {
             attesting_validators: vec![ValidatorIndex(1)],
             attesting_committees: HashMap::new(),
             sync_validators_by_subnet: HashMap::new(),
+            ptc_validators: Vec::new(),
         };
         tx.send_replace(Some(Arc::new(voting_assignments)));
 
@@ -3377,6 +3397,7 @@ mod tests {
             attesting_validators: vec![],
             attesting_committees: HashMap::new(),
             sync_validators_by_subnet: HashMap::new(),
+            ptc_validators: Vec::new(),
         };
         tx.send_replace(Some(Arc::new(voting_assignments)));
 
--- a/anchor/validator_store/src/metadata_service.rs
+++ b/anchor/validator_store/src/metadata_service.rs
@@ -312,9 +312,7 @@ impl<E: EthSpec, T: SlotClock + 'static> MetadataService<E, T> {
             })
             .unwrap_or_default();
 
-        // Get PTC validators (CStar+ only). Duties are read from LH's epoch-cached
-        // `PtcMap`; the per-pubkey check is the same `get_validator_and_cluster`
-        // shape used by the Phase-2/3 consensus-data builder.
+        // Get PTC validators
         let epoch = slot.epoch(E::slots_per_epoch());
         let ptc_validators = if self.fork_schedule.active_fork(epoch) >= Fork::CStar {
             let duties = self.duties_service.get_ptc_duties_for_slot(slot);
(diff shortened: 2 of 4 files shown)


shane-moore:
@claude, please take a look

shane-moore:
@claude, please take a look

shane-moore:
@claude, thanks. On each:

**1.** Keeping `pub` on `build_ptc_validators` to match the sibling pure helpers in this file (`sort_aggregators_by_validator_index`, `sort_contributors_by_signing_root_then_validator_index`, `filter_aggregators_with_attestations`, `filter_contributors_with_contributions`), all `pub` with zero out-of-crate callers, same as the parent struct's fields. C-STRUCT-PUBLIC is satisfied either way; no behavioral difference in a workspace-internal crate.

**2.** I don't think there's a real gap here. `database.state()` is a `watch::Sender::borrow()`, i.e. a nanosecond-scale read on the watch RwLock; 1-10 borrows per slot is microseconds. The "cache it" alternative is actively worse: holding a `Ref<NetworkState>` across the closure would block every `send_modify`/`send_if_modified` on the state channel (`database/src/lib.rs:310,316`), which the index-sync path uses. Serial cheap borrows are the lock-friendly shape. Attestation does this for hundreds of validators per slot without trouble; no flamegraph or metric has flagged any of the three sites.

**3.** No `update_voting_assignments`-level branch test exists for the Boole precedent at `:517`, and more broadly every test in `metadata_service.rs` is a pure-helper unit test. The orchestrator is untested end-to-end across the board, not just for fork branches. Adding scaffolding (mock `DutiesService`/`Database`/`SlotClock` plus a tokio runtime) belongs in its own coverage-investment PR that exercises the whole orchestrator (Phases 1/2/3 + Boole + CStar together), not as a one-off branch test here.

diegomrsantos (approved):

shane-moore:
closing for now since we're going to move to a non-qbft ptc approach

Issue #1043, opened 2026-05-20
ci: claude review bot uses fork's stale base branch when computing PR diff


shane-moore:
## Problem

When `claude-pr-review` or `claude-mentions` runs on a PR from a fork, the bot computes the "what changed in this PR" diff using `git diff origin/<base-branch>...HEAD`. Inside the workflow's checkout, `origin` points at the **fork** (because `actions/checkout` is configured with `repository: <fork>` for fork PRs). So `origin/<base-branch>` resolves to the fork's local copy of the base branch, which may be arbitrarily stale.

Result: every upstream commit that landed on `sigp/anchor:<base-branch>` since the fork last synced gets falsely attributed to the PR.

## Evidence

The trigger for this issue is the bot review on PR #1042: https://github.com/sigp/anchor/pull/1042#issuecomment-4502135924

That PR's actual diff (per `gh pr diff 1042`) is 8 lines: 4 GitHub Action version bumps across 2 files.

The bot's review described a roughly +50 / −50 line diff including a model swap (`sonnet-4-5` → `opus-4-7`), addition of `ACTIONS_STEP_DEBUG: true`, reordering of steps, switch of checkout `ref` from `head.ref` to `head.sha`, removal of `track_progress: true`, and several other changes. **None of these were in the PR.** They were all upstream changes from PRs #705 (Nov 2025), #845, #974, #997 that had landed on `sigp/anchor:stable` over the last 6-9 months but were absent from the contributor's fork's stale `stable` branch:

- `sigp/anchor:stable` HEAD: `ae4ca2de` (current)
- `shane-moore/anchor:stable` HEAD: `3feea1e0` ("Release 1.1.0", ~9 months stale)

## Where it bites

| Workflow | Trigger | Checkout source | Problem severity |
| --- | --- | --- | --- |
| `claude-mentions.yml` | `issue_comment` on PR from fork | Fork (per `repository:` override) | High — `origin/<base>` = fork's local copy |
| `claude-mentions.yml` | `issue_comment` on PR from same-org branch | `sigp/anchor` | None |
| `claude-pr-review.yml` | `pull_request_target` | `sigp/anchor` (no override) | Indirect — `origin/<base>` may not be locally fetched at all under `fetch-depth: 1` |

[...]

PR #1042, opened 2026-05-20
ci: bump Node20-deprecated actions in `stable`-loaded workflows (Phase 1 of #1040)

shane-moore:
## Problem, Evidence, and Context (Required)

GitHub is deprecating the Node.js 20 runtime on Actions runners:

- **2026-06-02** — Node20-declared actions get force-migrated to the Node24 runtime.
- **2026-09-16** — Node20 runtime is removed from runners entirely.

See [GitHub's announcement](https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/) and the full audit in #1040.

The audit identified two operational groups of workflows that need action bumps. This PR handles **Phase 1**: the subset that *must* land on `stable` because their triggers (`pull_request_target` / `issue_comment` / `pull_request_review_comment`) load workflow YAML from the **default branch of the base repository**, not from the PR base ref. Per [GitHub's docs](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request_target):

> "This event runs in the context of the default branch of the base repository, rather than in the context of the merge commit, as the `pull_request` event does."

A change landed on `unstable` would therefore have zero effect on these workflows' runs until the next release-merge. Historical precedent for landing this class of fix on `stable`: #430, #705, #830, #844, #845, #974, #997.

Phase 2 (the remaining workflows — `coverage`, `docker`, `docs`, `linkcheck`, `local-testnet`, `release`, `test-suite`) will be a separate PR against `unstable`.

## Change Overview (Required)

Six action version bumps across three workflows:

| File | Action | From | To |
| --- | --- | --- | --- |
| `claude-mentions.yml` | `actions/checkout` | `@v4` | `@v6` |
| `claude-mentions.yml` | `actions/create-github-app-token` | `@v2` | `@v3` |
| `claude-mentions.yml` | `aws-actions/configure-aws-credentials` | `@v4` | `@v6` |
| `claude-mentions.yml` | `anthropics/claude-code-action` | `@v1.0.52` | `@v1.0.127` |
| `claude-pr-review.yml` | (same four bumps) | (same) | (same) |
| `pr-checks.yml` | `amannn/action-semantic-pull-request` | `@v5` | `@v6` |
| `pr-checks.yml` | `marocchino/sticky-pull-request-comment` | `@v2` | `@v3` (×2 occurrences) |

[...]

PR #1054, opened 2026-05-22
fix(ci): claude-mentions fork PR checkout for action v1.0.127

shane-moore:

PR #1042 bumped `anthropics/claude-code-action` from `v1.0.52` to `v1.0.127`. v1.0.127 added fork-PR fetch logic that assumes `origin` is the base repo:

```ts
// claude-code-action v1.0.127, src/github/operations/branch.ts:175-184
if (prData.isCrossRepository) {
  execGit(["fetch", "origin", `--depth=${fetchDepth}`,
           `pull/${entityNumber}/head:${branchName}`]);
}
```

`claude-mentions.yml` predated this: when `is_fork == 'true'`, it pre-pointed `origin` at the **fork** via the `actions/checkout` `repository:` override. The fork doesn't have `refs/pull/N/head` (that namespace lives only on the base repo), so the action's fetch fails for every fork PR.

Live failure on PR #1053 (fork PR from `shane-moore/anchor`, [run 26296833180](https://github.com/sigp/anchor/actions/runs/26296833180/job/77411552034)):

```
PR #1053 is from a fork, fetching via refs/pull/1053/head...
fatal: couldn't find remote ref pull/1053/head
##[error]Action failed with error: Command failed:
  git fetch origin --depth=20 pull/1053/head:feat/ptc-voting-assignments
```

`claude-pr-review.yml` already runs the correct pattern (origin = base repo, app-token embedded before checkout; see its own `lines 63-68` comment block). This PR brings `claude-mentions.yml` in line with it.

Related: issue #1043 (separate stale-fork-base diff misattribution bug). Its proposed prompt fix is intentionally kept out of this PR to split by blast radius.

## Change Overview (Required)

One file: `.github/workflows/claude-mentions.yml`.

- Remove the `Get PR info for fork support` step. Its four outputs were consumed only by the now-removed `repository:` and `ref:` overrides; no other steps reference them.
- Move `Generate GitHub App token` to run **before** `Checkout repository` so the token is embedded in `origin` for the action's PR-ref fetch (matches the ordering and rationale documented in `claude-pr-review.yml:63-68`).
- Simplify `Checkout repository` to drop both `repository:` and `ref:` overrides. With defaults, `origin = sigp/anchor` and v1.0.127's `git fetch origin pull/N/head:branch` succeeds because that ref exists on the base repo.

End state structurally mirrors `claude-pr-review.yml:63-82`.

Intentionally unchanged:
- All other action version bumps from #1042.
- `claude-pr-review.yml` (already correct).
- `.github/prompts/review.md` (deferred; tied to #1043).
- Behavior for non-PR issue comments and for same-repo PRs.

- **Blast radius**: workflow-only, gated by `@claude` mention plus org-membership check.
- **Same-repo PR path**: unchanged in effect. v1.0.127's non-fork branch (`isCrossRepository: false`) does `git fetch origin <branchName>` against `origin = sigp/anchor`, which works in both old and new flavors.
- **Non-PR issue comments**: `actions/checkout` defaults to the base repo at the default branch, and `setupBranch()` skips the PR-fetch path when `isPR == false`.
- **`pull_request_review_comment` events**: `GITHUB_REF` is `refs/pull/N/merge` (resolvable from `sigp/anchor`), then the action does its own `pull/N/head` fetch. Should work; only fully empirical post-merge.
- **App-token scope**: the same app-token + checkout ordering is already used by `claude-pr-review.yml` for fork PR ref fetches, so the GitHub App installation on `sigp/anchor` already has the required permissions.

## Validation (Required)

- **Source trace** (workflow YAML is not unit-testable):
  - `setupBranch()` runs before `configureGitAuth()` in `claude-code-action/src/modes/tag/index.ts:66, 83`.
  - `configureGitAuth()` rewrites `origin` to `context.repository.owner/repo` (base repo, derived from `GITHUB_REPOSITORY` env) in `src/github/operations/git-config.ts:82`.
  - Therefore the workflow's `actions/checkout` token must already be valid for fork PR ref reads on the base repo at the moment `setupBranch()` runs. This PR ensures that.
- `actionlint` passes on the changed file. The remaining SC2086 warnings are pre-existing in untouched shell blocks.
- **Post-merge live check** (mirrors #1043's acceptance pattern):
  1. Open a small fork PR (e.g. a comment-only README tweak from a personal fork).
  2. Mention `@claude` on it.
  3. Confirm the run succeeds (no `fatal: couldn't find remote ref pull/N/head`).

## Rollback

Revert this single commit. No data, config, or operational state involved. Restores the prior (broken) fork-PR behavior but is otherwise neutral.

## Blockers / Dependencies

None. Targets `stable`. `issue_comment` and `pull_request_review_comment` triggers load this workflow from the base repo's default branch, so the fix only takes effect once merged to `stable` (same rule that applied to PRs #830, #844, #845, #974, #997, #1042).

### DIFF
--- a/.github/workflows/claude-mentions.yml
+++ b/.github/workflows/claude-mentions.yml
@@ -40,26 +40,12 @@ jobs:
                     exit 1
                   fi
 
-            - name: Get PR info for fork support
-              if: steps.check.outputs.is_member == 'true' && github.event.issue.pull_request
-              id: pr-info
-              run: |
-                  PR_DATA=$(gh api repos/${{ github.repository }}/pulls/${{ github.event.issue.number }})
-                  echo "pr_head_owner=$(echo "$PR_DATA" | jq -r '.head.repo.owner.login')" >> $GITHUB_OUTPUT
-                  echo "pr_head_repo=$(echo "$PR_DATA" | jq -r '.head.repo.name')" >> $GITHUB_OUTPUT
-                  echo "pr_head_ref=$(echo "$PR_DATA" | jq -r '.head.ref')" >> $GITHUB_OUTPUT
-                  echo "is_fork=$(echo "$PR_DATA" | jq -r '.head.repo.fork')" >> $GITHUB_OUTPUT
-              env:
-                  GH_TOKEN: ${{ github.token }}
-
-            - name: Checkout repository
-              if: steps.check.outputs.is_member == 'true'
-              uses: actions/checkout@v6
-              with:
-                  repository: ${{ github.event.issue.pull_request && steps.pr-info.outputs.is_fork == 'true' && format('{0}/{1}', steps.pr-info.outputs.pr_head_owner, steps.pr-info.outputs.pr_head_repo) || github.repository }}
-                  ref: ${{ github.event.issue.pull_request && steps.pr-info.outputs.pr_head_ref || github.ref }}
-                  fetch-depth: 0
-
+            # Generate the app token before checkout so it can be used for
+            # git operations. `claude-code-action` calls `setupBranch()` (which
+            # fetches PR refs via `git fetch origin pull/N/head:...`) before
+            # `configureGitAuth()`, so the token embedded in `origin` by
+            # `actions/checkout` must already have permission to fetch fork
+            # PR refs.
             - name: Generate GitHub App token
               if: steps.check.outputs.is_member == 'true'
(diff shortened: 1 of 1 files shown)


jnhsigmap (approved):
I verified the stated change vector and this looks fine to me, well caught :)

Issue #1067, opened 2026-05-28
Proposer QBFT boundary-layer instrumentation


jnhsigmap:
## Description

Implement concretely the instrumentation taxonomy from #1066 into the `qbft_instance()` loop, emitting metrics (Prometheus) and tracing events (OTLP, currently no-op without a collector) at QBFT lifecycle boundaries.

Aim to satisfy all acceptance criteria of #921 without altering consensus semantics.

Sub-issue of #921. Depends on #1066.

## Present Behaviour

The proposer QBFT instance loop (`qbft_manager/src/instance.rs`) runs without emitting any metrics or structured tracing about round advances, handoff budget, outcome, or decided round.

## Expected Behaviour

This owns implementation of:
- `anchor_proposer_qbft_decided_round` histogram records the round on which each proposer instance decided or timed out.
- `anchor_proposer_round_advance_total` counter (labelled by reason: `timeout`, `f_plus_1_rc`, `rc_quorum`, `future_proposal`) tracks round advances.
- `anchor_proposer_qbft_handoff_budget_seconds` histogram records slot budget remaining at QBFT start. Enables distinguishing late-start from round-churn.
- `anchor_proposer_qbft_outcome_total` counter (labelled: `decided`, `max_round_timeout`, `channel_closed`) tracks instance outcomes.
- `anchor_proposer_qbft_duration_seconds` histogram records total instance duration.
- A `proposer_qbft_instance` tracing span with structured checkpoint events covers the full lifecycle.
- All observation is external (boundary-layer) with no changes to `common/qbft` consensus logic.

## Implementation Detail

1. Add `handoff_budget_ms: Option<u64>` field to `QbftInitialization`.
2. Modify `qbft_instance()` to snapshot `(state, round)` before/after each recv arm and emit metrics + tracing events for proposer instances.
3. Add `handoff_budget_ms` parameter to `decide_instance` signature and `ConsensusDecider` trait.
4. Compute `handoff_budget_ms` in `validator_store` proposer path using `slot_clock.slot_duration()` and `determine_slot_elapsed_ms()`; pass `None` for non-proposer callers.

## Acceptance criteria mapping

| #921 criterion | How this PR satisfies it |
|---|---|
| Record decided/timed-out round | `PROPOSER_QBFT_DECIDED_ROUND` histogram |
| Round timeout and round-advance visible | `PROPOSER_ROUND_ADVANCE_TOTAL` counter + checkpoint tracing events |
[...]

Issue #1066, opened 2026-05-28
Proposer QBFT instrumentation taxonomy and API surface


jnhsigmap:
## Description

Define the instrumentation vocabulary (enums, metrics, classifier) and minimal API surface needed for proposer QBFT lifecycle observability. This is the first of 2 sub-issues - the additive and zero-runtime-risk foundation that #1067 aims to implement concretely.

Sub-issue of #921.

## Present Behaviour

`qbft_manager` has no metrics module, no round-advance taxonomy, and no way to query `InstanceState` from outside the `common/qbft` crate without pattern-matching on `Completed`.

## Expected Behaviour

This issue owns implementation of:
- `Qbft` exposes a `pub fn state() -> InstanceState` getter.
- `qbft_manager` has a `metrics.rs` declaring 5 Prometheus metrics (histograms + counters).
- `qbft_manager` has an `instrumentation.rs` with `RoundAdvanceReason`, `RecvArmTag` enums and a pure `classify_round_advance()` function.
- All new symbols compile but are unused (no behavioral change).

## Implementation Detail

1. Add `pub fn state()` one-line getter to `common/qbft/src/lib.rs`.
2. Add `metrics = { workspace = true }` dependency to `qbft_manager/Cargo.toml`.
3. Create `qbft_manager/src/metrics.rs` with `LazyLock<Result<...>>` metric definitions.
4. Create `qbft_manager/src/instrumentation.rs` with enums, classifier, and inline unit tests.
5. Declare `mod instrumentation; mod metrics;` in `qbft_manager/src/lib.rs`.

## Scope

**In scope:** Type definitions, metric registrations, classifier logic, metric taxonomy, unit tests.

**Out of scope:** Wiring metrics into the QBFT instance loop (addressed in #1067).

PR #1068, opened 2026-05-29
feat(qbft_manager): add instrumentation taxonomy and metrics surface

jnhsigmap:

- Anchor's proposer QBFT path lacks structured observability: when consensus stalls or rounds advance, operators and maintainers have no visibility of the cause. Debugging round-change cascades, understanding bottlenecks, and explaining timeouts are currently difficult.
- This is **PR 1 of 2** for the QBFT instrumentation work tracked in #1066. It introduces new components and API modifications without altering the control flow of `qbft_instance`. This includes taxonomy, metrics, and a specific bare-state getter on `Qbft`. A follow up change tracking #1067 will address concrete implementation.
- Splitting the work into 2 PRs introduces a pure-additive review (no behavior change to audit) that is isolated from higher-risk recv-loop changes.
- It is worth doing now because we have already added instrumentation to pre-QBFT flow in #924 and #920.

## Change Overview (Required)

- New `qbft_manager::instrumentation` module: checkpoint string constants, `RoundAdvanceReason` and `RecvArmTag` enums, and a pure `classify_round_advance` function. Classification uses new `InstanceStateKind` enum as a distinguisher between the four round-advance causes:
  - timeout (recv arm)
  - f+1 round-change
  - a justified future-round proposal
  - a round-change quorum
- New `qbft_manager::metrics` module: five Prometheus statics (decided round, total duration, handoff budget, outcome counter, round-advance counter) following the existing `eth/src/metrics.rs` `LazyLock<Result<...>>` pattern. This makes these metrics available to operators in addition to considering OTLP inclusions for us.
- Public read-only `Instance::state_kind()` getter on the QBFT instance. This is the only leakage of this change into `common/qbft`. Required so the boundary layer in the follow up change (#1067) can observe before/after state without changing internal consensus logic. We still keep the QBFT internal API untouched behaviourally with only a new type and this read-only method that allows `proposal_root` payloads to not be included in bare state machine comparison.
- **Intentionally not changed**: `qbft::Instance` recv loop, message handlers, state transitions, or any behavior. No call sites are added to the new symbols in this PR.

- **Risk**: Main risks only arise from delayed implementation of #1067. If we follow this right away then that is largely mitigated and the surface introduced here will be used.
- **Trade-off**: `state_kind()` getter widens `Qbft`'s public surface by one read-only accessor and a new type. Chosen over passing `Discriminant<InstanceState>` through internal hooks because it's simpler and keeps the boundary layer purely observational.

## Validation (Required)

- 12 simple tests in `instrumentation::tests` cover new `classify_round_advance` function across all 4 reason branches, both recv arms, and the no-advance / decreased-round defensive cases.
  - Additionally, unit tests implemented that drive a `Qbft` instance through each round-advance path, and feed observed state into classifier function.
- `make cargo-fmt-check`
- `make lint`
- `cargo test -p qbft -p qbft_manager --release`

- Pure-additive change with no runtime impact. Revert is simple with no overhead.

- File-scope `#![expect(dead_code, reason = ...)]` on both new modules to keep lint clean while symbols are unused.

### DIFF
--- a/anchor/qbft_manager/src/instrumentation.rs
+++ b/anchor/qbft_manager/src/instrumentation.rs
@@ -26,9 +26,6 @@ pub enum RoundAdvanceReason {
     /// The node observed f+1 round-change messages from peers at a higher
     /// round.
     FPlusOneRoundChange,
-    /// A full quorum of round-change messages was observed, achieving
-    /// round-change consensus.
-    RoundChangeQuorum,
     /// A justified proposal for a higher round arrived, pulling the instance
     /// forward to catch up with the leader's round (no round change occurred).
     FutureRoundProposal,
@@ -39,7 +36,6 @@ impl RoundAdvanceReason {
         match self {
             Self::Timeout => "timeout",
             Self::FPlusOneRoundChange => "f_plus_1_rc",
-            Self::RoundChangeQuorum => "rc_quorum",
             Self::FutureRoundProposal => "future_proposal",
         }
     }
@@ -57,16 +53,18 @@ pub enum RecvArmTag {
 /// Classify a round advance observed at the boundary layer.
 ///
 /// Consumes the payload-free `InstanceStateKind` of the after-state so the classifier depends
-/// only on the state variant without concern for state attributes.
+/// only on the state variant without concern for state attributes. Returns `None` if the round
+/// did not actually advance and `Some(reason)` when it did.
 ///
-/// `Message` match arm after-state kind captures how the round advanced:
+/// `Message`-arm after-states coincide with a
+/// round advance:
 /// - `SentRoundChange`: f+1 peers at a higher round pulled us forward.
-/// - `Prepare`: a justified future-round proposal arrived (we caught up to the leader, no round
-///   change occurred).
-/// - anything else (e.g. `AwaitingProposal`, `RoundChangeConsensus`): a full round-change quorum
-///   was reached.
+/// - `Prepare`: a justified future-round proposal arrived.
 ///
-/// Returns `None` if the round did not actually advance and `Some(reason)` when it did.
+/// `AwaitingProposal` and `RoundChangeConsensus` can appear on the `Message` arm when a
+/// round-change quorum completes, but the round has *already* advanced on the prior f+1
+/// message (`f+1 < 2f+1` for all canonical committees), so `after_round == before_round`.
+/// This variant is functionally unreachable.
 pub fn classify_round_advance(
     _before_state: InstanceStateKind,
     after_state: InstanceStateKind,
@@ -83,7 +81,10 @@ pub fn classify_round_advance(
         RecvArmTag::Message => match after_state {
             InstanceStateKind::SentRoundChange => RoundAdvanceReason::FPlusOneRoundChange,
             InstanceStateKind::Prepare => RoundAdvanceReason::FutureRoundProposal,
-            _ => RoundAdvanceReason::RoundChangeQuorum,
+            InstanceStateKind::AwaitingProposal
+            | InstanceStateKind::RoundChangeConsensus
+            | InstanceStateKind::Commit
+            | InstanceStateKind::Complete => return None,
         },
     };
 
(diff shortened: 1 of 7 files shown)


jnhsigmap:
@claude, please take a look

jnhsigmap:
@claude, in response to review points 1-3:

1. Added `RoundAdvanceReason::FutureRoundProposal` variant as suggested. `classify_round_advance` distinguishes in the `RecvArmTag::Message` match arm between a `Prepare` message and us sending a round change message (`SentRoundChange`). New test `future_proposal_when_after_state_is_prepare` added to test for `AwaitingProposal` to `Prepare` state transition.

2. `decided_round`: `linear_buckets(1.0, 1.0, 12)` created for one bucket per round with some headroom. `duration_seconds`: explicit buckets to 30.0. ~8s timeout (round_time 2s × max_rounds 4). `handoff_budget_seconds`: explicit buckets to 12.0 (slot duration).

3. The most contentious as we are adding more code to `common/qbft`. `InstanceStateKind` enum created to provide the state without the payload to maintain the pure state-machine result that was required in the `qbft_manager` call. In other words, renamed the getter `state()` to `state_kind()` returning `InstanceStateKind`. This was intuitive to me to keep the payload out of sight for simple state comparisons. This removed `std::mem::Discriminant`/`discriminant` as we can now match on `InstanceStateKind` everywhere.

Regarding 3, this leans on the bounds set in #921, however I think this is a cleaner and more idiomatic implementation that keeps inference from `qbft_manager` simple an interpretable.

Noted 4. no changes made here.

jnhsigmap:
@claude, Provide a follow up review given these changes. Ensure our PR title, body, comments, and diffs are also aligned.

Edit: That was a retry before the above review, he kept failing before :thinking:

anchor/qbft_manager/src/instrumentation.rs
```diff
@@ -0,0 +1,235 @@
...
+    if after_round <= before_round {
+        return None;
+    }
+
+    let reason = match recv_arm {
+        RecvArmTag::RoundEnd => RoundAdvanceReason::Timeout,
+        RecvArmTag::Message => match after_state {
+            InstanceStateKind::SentRoundChange => RoundAdvanceReason::FPlusOneRoundChange,
+            InstanceStateKind::Prepare => RoundAdvanceReason::FutureRoundProposal,
+            _ => RoundAdvanceReason::RoundChangeQuorum,
```
shane-moore: Minor hardening: can we make this match explicit? `AwaitingProposal | RoundChangeConsensus => RoundChangeQuorum`, and `Commit | Complete => return None`. Those states are not reachable for a message-driven round advance today, but the wildcard would silently count them as `rc_quorum` if that changes.

shane-moore:
Pure-additive surface looks correct, and the earlier points (future-round classification, buckets, payload-free getter) are cleanly resolved.

One ask before merge: add a correspondence test in this PR. The 8 unit tests hand-feed `InstanceStateKind` values into `classify_round_advance` and assert it returns what it was written to return, so they can't catch a wrong assumption about what the state machine actually emits. That's not hypothetical: the `FutureRoundProposal` miscategorization was caught by reading `received_propose`, not by a test. A test that constructs a real `Qbft`, drives each advance path (`end_round` timeout, f+1, quorum, future-round proposal), and feeds the observed `state_kind()` + `get_round()` into `classify_round_advance` closes that gap. It's behavior-inert (no recv-loop wiring), so it fits the zero-runtime-risk scope, and it's writable now with the existing infra (public `Qbft` API plus the message construction already in `qbft_manager` tests, keeping round-change justifications at `data_round=0`). One thing to assert explicitly: the quorum branch lands in `AwaitingProposal` after `start_round`, not `RoundChangeConsensus`. The recv-loop integration and metric-increment test belongs with #1067.

jnhsigmap:
> Pure-additive surface looks correct, and the earlier points (future-round classification, buckets, payload-free getter) are cleanly resolved.
> 
> One ask before merge: add a correspondence test in this PR. The 8 unit tests hand-feed `InstanceStateKind` values into `classify_round_advance` and assert it returns what it was written to return, so they can't catch a wrong assumption about what the state machine actually emits. That's not hypothetical: the `FutureRoundProposal` miscategorization was caught by reading `received_propose`, not by a test. A test that constructs a real `Qbft`, drives each advance path (`end_round` timeout, f+1, quorum, future-round proposal), and feeds the observed `state_kind()` + `get_round()` into `classify_round_advance` closes that gap. It's behavior-inert (no recv-loop wiring), so it fits the zero-runtime-risk scope, and it's writable now with the existing infra (public `Qbft` API plus the message construction already in `qbft_manager` tests, keeping round-change justifications at `data_round=0`). One thing to assert explicitly: the quorum branch lands in `AwaitingProposal` after `start_round`, not `RoundChangeConsensus`. The recv-loop integration and metric-increment test belongs with #1067.

cool thanks for that @shane-moore, it has taken me a while to get my head around this.

I made a couple of changes in response to this, but need some guidance as I don't fully understand this yet.
- Removed `RoundChangeQuorum` from `RoundAdvanceReason`.
- Added 4 `Qbft` state-machine tests more realistically driving each round-advance path and feeding the observed `(state_kind, round)` into `classify_round_advance`.

**Discoveries**

  1. `RoundChangeQuorum` is unreachable under planned per-message snapshot (in the recv loop, next issue #1067) granularity. This snapshots the state once per `receive()` call. The f+1 branch will always fire before a quorum is reached and the round will advance initially on the f+1 message. The quorum-completing message never produces a round change.
  2. The 4 new tests prove that the classifier's assumptions match the state machine's actual behaviour. It constructs a fresh mock `Qbft` instance and drives it through a specific path. It then asserts the result produced by the classifier function given the observed state. Confirms 1.
  3. The classifier function's Message match arm is now exhaustive. `AwaitingProposal`, `RoundChangeConsensus`, `Commit`, and `Complete` now return `None` and are not behind a catch all `_`.

Does that sound correct?

shane-moore:
Thanks for digging in on this, and the real-instance tests are exactly the right move, keep them. On "does that sound correct?" though, Discovery #1 doesn't hold, so I'd keep `RoundChangeQuorum` rather than remove it.

**The quorum branch can advance the round.** The `f+1 < 2f+1` argument assumes the f+1 catch-up counts per round and jumps to that round. It doesn't: `lowest_partial_quorum_above_round` counts the union of operators across all higher rounds and jumps to the *lowest* one. So a stale low-round round-change parks `current_round` below a higher round that independently reaches quorum, and `set_round` (which sets the round unconditionally) then advances on the quorum-completing message.

Running the existing test helpers on this sequence (n=4, node = op1, starting round 1):

```text
receive RC(round=2, op2)   // low anchor
receive RC(round=3, op2)   // same op also ahead -> union still {op2}, no f+1 jump yet
receive RC(round=3, op3)   // union {op2,op3} >= f+1 -> jump to LOWEST future round = 2
// before: round = 2, state = SentRoundChange
receive RC(round=3, op4)   // round 3 hits 2f+1 -> quorum branch -> set_round(3): 2 -> 3
// after:  round = 3, state = AwaitingProposal
```

This advances `2 -> 3` and `classify_round_advance(...)` returns `None`, so `PROPOSER_ROUND_ADVANCE_TOTAL` silently undercounts. It's reachable whenever a node catches up while peers are spread across rounds, which is exactly the round-churn the metric is meant to explain. The same-round test passes because with no lower anchor the f+1 jump and the quorum land on the same round, so `set_round` is a no-op.

The fix keeps this PR's theme (no QBFT control-flow change, no consensus change, no getter beyond `state_kind()`): add `RoundChangeQuorum` back to `RoundAdvanceReason` (with `as_str` -> `"rc_quorum"`) and map `AwaitingProposal` to it:

```rust
RecvArmTag::Message => match after_state {
    InstanceStateKind::SentRoundChange => RoundAdvanceReason::FPlusOneRoundChange,
    InstanceStateKind::Prepare => RoundAdvanceReason::FutureRoundProposal,
    // A Message-arm round advance into AwaitingProposal is the round-change quorum path:
    // received_round_change sets RoundChangeConsensus, then set_round() calls start_round(),
    // which leaves both leader and non-leader instances in AwaitingProposal.
    InstanceStateKind::AwaitingProposal
    | InstanceStateKind::RoundChangeConsensus => RoundAdvanceReason::RoundChangeQuorum,
    InstanceStateKind::Commit | InstanceStateKind::Complete => return None,
},
```

This is correct and complete: `set_round` is only ever called from the quorum branch, and `start_round` always lands in `AwaitingProposal` (the leader's `send_proposal` doesn't change state), so `AwaitingProposal` on the Message arm uniquely fingerprints the quorum path.

Then turn the sequence above into a regression test asserting `Some(RoundAdvanceReason::RoundChangeQuorum)` (the same-round test stays as-is, asserting `None`).

shane-moore:
Follow-up on the test side: the match change above flips two of the existing hand-fed unit tests, which currently assert `None` for a Message-arm advance:

- `none_when_awaiting_proposal_on_message_arm`
- `none_when_round_change_consensus_on_message_arm`

Both feed `after_round > before_round`, so they now hit the restored arm and should become `Some(RoundAdvanceReason::RoundChangeQuorum)`. The real-instance `round_change_quorum_does_not_advance_round_beyond_f_plus_one` correctly stays `None` (its `after_round == before_round`).

jnhsigmap:
thanks @shane-moore, super helpful

Changed in https://github.com/sigp/anchor/pull/1068/commits/8a7e047b4a338d8f68014d2e77ffc6887b646c88:
  - Restored `RoundChangeQuorum` in `RoundAdvanceReason` and `as_str` `"rc_quorum"`.
  - `AwaitingProposal` and `RoundChangeConsensus` mapped by `classify_round_advance` to `RoundChangeQuorum` instead of returning `None`.
  - [Two unit tests](https://github.com/sigp/anchor/pull/1068#issuecomment-4606935240) from `none_when_*` to `round_change_quorum_when_*`, flipped assertions to `expect Some(RoundChangeQuorum)`.
  - Added `Qbft` regression test proving the scenario of a quorum completing _**at a round above the f+1-anchored round**_ advances `current_round` through the quorum branch into `AwaitingProposal`, classifies as `RoundChangeQuorum`.

shane-moore (approved):
LGTM. `rc_quorum` restored, match is explicit, and the real-instance regression test covers the cross-round case. tests + CI green 🎉

Issue #1015, opened 2026-05-14
feat(validator_store): fork-gate decide_abstract_block; skip blinded decode for Gloas


shane-moore:
## Goal

Update `decide_abstract_block` in `anchor/validator_store/src/lib.rs` to dispatch by `completed_data.version` so Gloas slots decode straight to a `BeaconBlock<E>` and skip the blinded-block path entirely. Pre-Gloas behavior preserved.

## Context

Under Gloas the blinded-block path is removed (no relay-builder mechanism, no `BlindedBeaconBlock` decode). Post-#1014, the try-blinded-then-full dispatch in `decide_abstract_block` still works for Gloas, but only accidentally: `decode_blinded_block` fails fast on a version check (`fork >= Gloas` returns `NoMatchingVariant`) and the `or_else` fallback lands in `decode_block_contents`, which the LH pin decodes for Gloas as `FullBlockContents::Block(BeaconBlock)` (`common/eth2/src/types.rs:1882-1894`). This issue makes the dispatch explicit instead of fallback-driven. After this change, the Gloas arm returns a block-only `UnsignedBlock` and `sign_abstract_block` signs it under `Domain::BeaconProposer`. No envelope is decoded, fetched, or carried through the decided value (envelope signing is out of scope per SIP-94 commit `cfb672e`).

Builds on the Gloas decode path added by #1014, which landed as `decode_block` on `ProposerConsensusData` (consensus.rs:267), with `decode_blinded_block` rejecting Gloas via a version check; `validate_block_proposal`'s fork branch shows the dispatch shape to mirror.

## Suggested approach

In `decide_abstract_block`, replace the try-blinded-then-full decode with an explicit fork branch on `ForkName::from(completed_data.version)` (the conversion `#1014` uses):
- Gloas and later (`fork >= ForkName::Gloas`, mirroring the landed `validate_block_proposal` branch): call `decode_block` on `ProposerConsensusData`.
- Pre-Gloas: keep the existing two-step blinded-then-full fallback.

`completed_data.version` is authoritative because it is set from the producing BN's block (`block.fork_name_unchecked()`), reflecting what the network actually signed. Do not gate on `ForkSchedule::active_fork_at_slot` at this boundary.

[...]

PR #1070, opened 2026-06-01
feat(ssv_types): Gloas decode branch on ProposerConsensusData

### LINKED ISSUE #1014: feat(ssv_types): Gloas decode branch on ProposerConsensusData (block-only)
## Goal

Add a Gloas decode path to `ProposerConsensusData` in `anchor/common/ssv_types/src/consensus.rs` that returns `BeaconBlock<E>` only. Under Gloas, `DataSSZ` carries an SSZ-encoded `Gloas.BeaconBlock`; no inline blobs, no KZG proofs, no envelope.

## Context

This issue assumes SIP-94 §4 lands as currently drafted (`ssvlabs/SIPs`, commit `cfb672e`): the SSV cluster signs only `BeaconBlock` under `Domain::BeaconProposer`, distributed signing of `SignedExecutionPayloadEnvelope` is out of scope, and PTC may attest `payload_present = FALSE` on self-build slots as a known consequence. SIP-94 still has an open question on self-build envelope signing; if it lands the other way, this issue together with #1015 and #1017 must carry, sign, and publish the envelope rather than drop it.

Upstream alignment:
- Lighthouse's `FullBlockContents::from_ssz_bytes_for_fork` (`common/eth2/src/types.rs`) special-cases Gloas to return `FullBlockContents::Block(BeaconBlock)` only.
- Beacon-APIs `produceBlockV4` (ethereum/beacon-APIs#580) returns `Gloas.BeaconBlock` when `include_payload=false`; LH's VC takes this path under Gloas.
- Gloas does ship `SignedExecutionPayloadEnvelope` as a first-class object with its own gossip path; the narrower claim here is that Anchor/SSV is choosing not to support distributed envelope signing in this iteration.
- Blinded-block path is gone under Gloas.

## Suggested approach

Add `decode_gloas_block(&self) -> Result<BeaconBlock<E>, DecodeError>` on `ProposerConsensusData` using `BeaconBlock::from_ssz_bytes_for_fork`. Branch `validate_block_proposal` on `value.version`:
- Gloas: skip the blinded-first try; decode straight to a Gloas block.
- Pre-Gloas: keep existing two-variant logic.

Slashing pre-check under `Domain::BeaconProposer` runs unchanged for both branches. `ProposerConsensusData`'s SSZ shape is unchanged; `version` is the existing `DataVersion` field, which already lists Gloas.

## Acceptance criteria

- `decode_gloas_block` accepts Gloas SSZ bytes and returns a Gloas `BeaconBlock<E>`.
- `decode_blinded_block` returns `DecodeError` on Gloas bytes (negative test).
- `validate_block_proposal` Gloas branch: skips blinded try; runs slashing pre-check on the Gloas block header.
- `DataVersion(Gloas)` SSZ round-trip works.
- Pre-Gloas behavior unchanged.

## Risks

- Gloas `BeaconBlockBody` SSZ shape diverges from Electra (adds `signed_execution_payload_bid`, `payload_attestations`, `parent_execution_requests`; removes `execution_payload`, `blob_kzg_commitments`, `execution_requests`). Trust LH's `from_ssz_bytes_for_fork` helper rather than hand-rolling the decode.

jnhsigmap:

- Under Gloas (EIP-7732), the execution payload is decoupled from the beacon block body. The SSV cluster signs only `BeaconBlock` under `Domain::BeaconProposer` — no envelope, blobs, or KZG proofs enter consensus (SIP-94).
- `ProposerConsensusData` needs a dedicated decode path that makes this contract explicit rather than relying on the incidental success of the blinded-block decoder.
- Resolves #1014.

## Change Overview

- Modifies `consensus.rs` in the `ssv_types` crate.
- Adds `ProposerConsensusData::decode_block<E>()` — decodes `DataSSZ` as a plain `BeaconBlock` via `from_ssz_bytes_for_fork(.., self.version)`.
- Fork-gates `decode_blinded_block<E>()` to reject Gloas-and-later with `NoMatchingVariant`. The removed blinded proposer path cannot be reached under EIP-7732.
- Fork-branches `validate_block_proposal`: `fork >= Gloas` uses the new decoder directly; pre-Gloas preserves the existing blinded-then-full fallback unchanged.
- Adds 5 and adjusts 1 unit test covering the decode round-trip, the fork-branch happy/error paths, a negative test that blinded decoding rejects Gloas, and a pre-Gloas regression guard.
- `tempfile` added as a dev-dependency for test fixture construction.

- **Not behaviourally load-bearing for valid Gloas input today.** EIP-7732 removes the payload-parameterized field from `BeaconBlockBodyGloas`, making full and blinded SSZ byte-identical. The old fallback would also succeed. The branch is intentional — it documents the SIP-94 contract, avoids redundant decode attempts, and anchors the path for downstream consumer changes.
- **Blinded decoder now rejects Gloas.** Decoding is gated on `self.version` so that the SIP-94 contract (blinded proposer path removed under Gloas) is enforced at the decoder rather than relying on incidental fallback success.
- **Blast radius:** confined to `ssv_types` crate; no downstream crate changes.

## Validation

- `cargo test -p ssv_types`
- `make cargo-fmt && make cargo-fmt-check`
- `make lint`
- `cargo build --release`

## Rollback

- Revert single commit; no data, config, or runtime impact.

### DIFF
--- a/anchor/common/ssv_types/src/consensus.rs
+++ b/anchor/common/ssv_types/src/consensus.rs
@@ -24,8 +24,8 @@ use typenum::{
 };
 use types::{
     AggregateAndProofBase, AggregateAndProofElectra, AttestationBase, AttestationData,
-    AttestationElectra, BlindedBeaconBlock, ChainSpec, Checkpoint, CommitteeIndex, Domain, EthSpec,
-    ForkName, Hash256, Slot, SyncCommitteeContribution,
+    AttestationElectra, BeaconBlock, BlindedBeaconBlock, ChainSpec, Checkpoint, CommitteeIndex,
+    Domain, EthSpec, ForkName, Hash256, Slot, SyncCommitteeContribution,
 };
 
 use crate::{CommitteeId, ValidatorIndex, message::*, partial_sig::PartialSignatureKind};
@@ -259,6 +259,11 @@ impl ProposerConsensusData {
         let fork = ForkName::from(self.version);
         FullBlockContents::from_ssz_bytes_for_fork(&self.data_ssz, fork)
     }
+
+    /// Decode the block data as a Gloas beacon block (block-only, no envelope/blobs/KZG).
+    pub fn decode_gloas_block<E: EthSpec>(&self) -> Result<BeaconBlock<E>, DecodeError> {
+        BeaconBlock::from_ssz_bytes_for_fork(&self.data_ssz, ForkName::Gloas)
+    }
 }
 
 impl QbftData for ProposerConsensusData {
@@ -379,17 +384,30 @@ impl<E: EthSpec> ProposerConsensusDataValidator<E> {
         &self,
         value: &ProposerConsensusData,
     ) -> Result<(), DataValidationError> {
-        // Always do this check, even if we're not validating slashing. This is to ensure that we
-        // have a decodable value.
-        let header = value
-            .decode_blinded_block::<E>()
-            .map(|block| block.block_header())
-            .or_else(|_| {
-                value
-                    .decode_block_contents::<E>()
-                    .map(|block| block.block().block_header())
-            })
-            .map_err(DataValidationError::DecodeError)?;
+        // Decode the block header to ensure the value is decodable (even when slashing
+        // protection is disabled). Under Gloas (EIP-7732), DataSSZ is decoded directly as a plain
+        // BeaconBlock. This behaviour is not behaviourally load-bearing as the execution
+        // payload is decoupled from the block body. The outcome of `decode_blinded_block`
+        // and `decode_gloas_block` are identical for this variant. The Pre-Gloas branch
+        // preserves the existing try-blinded-then-full fallback.
+        let fork = ForkName::from(value.version);
+
+        let header = if fork >= ForkName::Gloas {
+            value
+                .decode_gloas_block::<E>()
+                .map(|block| block.block_header())
+                .map_err(DataValidationError::DecodeError)?
+        } else {
+            value
+                .decode_blinded_block::<E>()
+                .map(|block| block.block_header())
+                .or_else(|_| {
+                    value
+                        .decode_block_contents::<E>()
+                        .map(|block| block.block().block_header())
+                })
+                .map_err(DataValidationError::DecodeError)?
+        };
 
         if !self.disable_slashing_protection {
             let epoch = header.slot.epoch(E::slots_per_epoch());
@@ -1270,8 +1288,12 @@ mod tests {
     use std::collections::HashMap;
 
     use bls::{AggregateSignature, FixedBytesExtended};
+    use eth2::types::FullBlockContents;
     use ssz_types::{BitList, BitVector};
-    use types::{Checkpoint, Epoch, MainnetEthSpec, SyncCommitteeContribution};
+    use types::{
+        BeaconBlockDeneb, BeaconBlockGloas, Checkpoint, EmptyBlock, Epoch, MainnetEthSpec,
+        SyncCommitteeContribution,
+    };
 
     use super::*;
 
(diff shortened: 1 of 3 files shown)


jnhsigmap:
@claude, review this pull request. validate the solution against #1014 and the [ePBS milestone](https://github.com/sigp/anchor/milestone/2).

jnhsigmap:
@claude, in response to intial feedback

1. decode_gloas_block ignores self.version — API asymmetry / future-fork foot-gun
I would not say this is a foot-gun, the method is clearly named and the contract is clear in the API fork gate, however for alignment to the other methods implemented in `ProposerConsensusData`, I have changed `decode_gloas_block` to depend on the instance's `fork` value and removed the hard coded value.

2. validator_store consumer is untouched
I believe this will be addressed in #1015 when we add fork gating to the upstream caller.

I have corrected points 3 and 5 with comments.

Correct me if I am misinterpreting the implication of changing point 1 on point 4, but now `decode_gloas_block` should pass on any non-blinded structure?

Re-review given these changes in https://github.com/sigp/anchor/pull/1070/commits/500f74a8d79f1f041740e47d9ce132a1ce9bee1c

jnhsigmap:
> **Claude finished @jnhsigmap's task in 3m 6s** —— [View job](https://github.com/sigp/anchor/actions/runs/26798186731)
> 
> ### Claude PR Re-review (commit `500f74a`)
> * [x]  Read the latest changes in commit `500f74a`
> * [x]  Verify fix to point 1 (decode_gloas_block uses self.version)
> * [x]  Verify fixes to points 3 and 5 (comments)
> * [x]  Address question about point 4 (test gap implications)
> 
> ### Verification of fixes
> Point	Status	Note
> 1. `decode_gloas_block` uses `self.version`	✅ Fixed	Now matches sibling decoder pattern at `consensus.rs:264-266`
> 2. `validator_store` consumer	📋 Deferred to #1015	Acknowledged
> 3. Truncated doc comment	✅ Fixed	`consensus.rs:2397`
> 4. Decoder/version mismatch test	See below	Question answered
> 5. EIP-7732 invariant comment on canary test	✅ Fixed	`consensus.rs:2321`
> The fix to point 1 is the right call. The branch policy in `validate_block_proposal` is now the _only_ place fork policy is encoded, and the decoder methods are uniformly per-version. That's a cleaner separation of concerns.
> 
> ### Re: your question on point 4
> > Correct me if I am misinterpreting the implication of changing point 1 on point 4, but now `decode_gloas_block` should pass on any non-blinded structure?
> 
> **Mostly yes, with one nuance worth being explicit about.**
> 
> With the fix, `decode_gloas_block::<E>()` now decodes `data_ssz` as a `BeaconBlock<E>` for whatever fork `self.version` indicates. So:
> 
> * ✅ **Any-version full `BeaconBlock` SSZ** → succeeds (the bytes are a plain `BeaconBlock` for that fork).
> * ❌ **Any-version `BlindedBeaconBlock` SSZ** (pre-Gloas) → fails, because pre-Gloas `BeaconBlock` and `BlindedBeaconBlock` differ in SSZ layout (`ExecutionPayload` vs `ExecutionPayloadHeader`).
> * ❌ **Any-version `FullBlockContents` SSZ** → fails, because `FullBlockContents` wraps `BeaconBlock + (blobs, proofs)` and the trailing bytes aren't a valid `BeaconBlock` tail.
> * ✅ **Gloas-versioned bytes** → succeed regardless of which decoder you pick (`decode_blinded_block` / `decode_gloas_block`), because EIP-7732 makes blinded ≡ full.
> 
> So the foot-gun that motivated point 4 — "calling `decode_gloas_block` on a Fulu-versioned value silently produces a Gloas-shaped result" — **is closed**. The function will now correctly decode against the value's declared version.
> 
> Point 4 is therefore moot in its original form: there's no longer a static fork-mismatch hazard inside the decoder. The remaining check ("does `validate_block_proposal` reject a value whose declared `version` disagrees with the SSZ shape?") is already covered by `validate_block_proposal_gloas_rejects_invalid_bytes` for the Gloas branch and by the existing pre-Gloas tests for that branch.
> 
> ### One small follow-on (non-blocking)
> The function name `decode_gloas_block` is now slightly misleading — after the fix, it's really `decode_block` (per-version, full BeaconBlock shape). It's only "Gloas" in the sense that Gloas is the first fork where this decoder is _used_. Two equally-fine options for a future cleanup pass:
> 
> 1. Rename to `decode_block` and let the caller's fork gate carry the policy (most consistent with `decode_blinded_block` / `decode_block_contents`).
> 2. Leave the name and tighten the doc to `"Decode as a full BeaconBlock — used by the Gloas+ branch which signs BeaconBlock under Domain::BeaconProposer per SIP-94."`
> 
> Either is fine; not worth blocking this PR.
> 
> ### Recap
> LGTM. All actionable feedback addressed; the remaining naming nit is cosmetic and can ride a later commit. Pre-Gloas behavior is unchanged, the SIP-94 contract is documented at the branch site, and the test suite covers the round-trip + happy/error paths in both branches.
> 
> > **Note on validation against #1014 / [ePBS milestone](https://github.com/sigp/anchor/milestone/2):** still unable to fetch issue/milestone bodies in this run (no `gh`/WebFetch permissions for `api.github.com`). Code-level checks against the PR body's stated SIP-94 contract pass; milestone acceptance criteria not independently verified.
> > · branch `feature/epbs-proposer-workflow`

`decode_gloas_block` renamed to `decode_block`.

anchor/common/ssv_types/src/consensus.rs
shane-moore, line 257: Can we guard this decoder against Gloas here?

Even though Lighthouse’s current Gloas full/blinded block shapes are byte-identical, `decode_blinded_block` is Anchor’s semantic decoder. Under the #1014/#1015 milestone contract, Gloas proposer consensus data is block-only and the blinded proposer path is gone. Returning `Ok(BlindedBeaconBlock)` for `fork >= ForkName::Gloas` leaves a footgun for any caller that forgets to fork-gate first.

  I think this should return `DecodeError` for Gloas-and-later, and the added blinded/header-match test should become the negative test #1014 asks for. #1015 should still fork-gate `decide_abstract_block`, but putting the invariant here makes it much harder to bypass accidentally

shane-moore:
small PR-body cleanup: the description still refers to `decode_gloas_block` and the “hardcoded-Gloas decoder” risk. The current code now uses `decode_block` and derives the fork from `self.version`, so those sections are stale

anchor/common/ssv_types/src/consensus.rs
jnhsigmap: yep!
 In https://github.com/sigp/anchor/pull/1070/commits/2b2f582d101f71617e015ac8c519a6dd29842f1a:
  - Fork-gated `decode_blinded_block` to return `DecodeError::NoMatchingVariant` for _Gloas-and-later_. Removed blinded proposer path can't be reached after EIP-7732.
  - Flipped the associated test to negative `decode_block_header_matches_blinded_decode` to `decode_blinded_block_rejects_gloas`, asserting Gloas input is **rejected**.

shane-moore (approved):
well done breh, lgtm!

Issue #1027, opened 2026-05-15
feat(validator_store): apply decided AttestationDataIndex in metadata-service and committee signing


shane-moore:
## Goal

Thread the BN-supplied / cluster-decided `attestation_data.index` through Anchor's attestation pipeline so all operators sign over the same root under Gloas. Three changes in the `validator_store` crate:

1. `metadata_service::update_voting_context`: carry `attestation_data.index` from the BN's `getAttestationData` response into the new `GloasBeaconVote.attestation_data_index` field that Phase 2 publishes. Since #1041 (synced into the feature branch 2026-06-11), Phase 2 sources `AttestationData` from one of two paths: an eager path on the slot's head event (queries the event-firing BN directly, verifies the returned `block_root` matches the event, bypasses WAD) and the WAD/first_success fallback at the fork-aware attestation deadline. The index rides along either path unchanged. This surface also owns switching `sign_committee_attestations`' Gloas seed from `first_att_data` (lib.rs:1513-1518) to the voting_context vote, matching the sync path, which already seeds from `voting_context.beacon_vote` (lib.rs:1395). The two call sites seed from independently-fetched sources (LH's attestation data, itself head-event-triggered on current LH unstable, vs Anchor's voting_context vote), so the seeds can diverge on timing or BN choice; #1061's shared-instance hash-equality only holds if both read the same voting_context field.
2. `metadata_service::fetch_aggregated_attestations`: compute the `attestation_data_root` (passed to `get_validator_aggregate_attestation_v2`) using the Gloas index carried in `voting_context` (surface 1's field) on Gloas slots. Note this is the QBFT *seed* value, not the decided value: `fetch_aggregated_attestations` is called with `&voting_context.beacon_vote` (metadata_service.rs:664-666) and metadata_service has no path to QBFT-decided values (watch channels flow one way). Divergence between seed and decided is absorbed downstream by the AggregatorCommittee QBFT, exactly as it is for `block_root`/`source`/`target` today.
3. `validator_store::sign_committee_attestations`: after QBFT decide, apply `decided.attestation_data_index` to each validator's `attestation.data.index` before signing.

Pre-Gloas behavior on each surface is preserved.

## Context

Builds on #1025 (`GloasBeaconVote` type) and #1026 (`GloasBeaconVoteValidator`). Per SIP-94 §2 (`ssvlabs/SIPs`, commit `e849eeb`; quote re-verified at head `4e02e6a`), the aggregation path note states:

[...]

Issue #1025, opened 2026-05-15
feat(ssv_types): add GloasBeaconVote with QbftData for Gloas slots


shane-moore:
## Goal

Add a new SSZ struct `GloasBeaconVote { block_root, source, target, attestation_data_index: u64 }` in `anchor/common/ssv_types/src/consensus.rs` mirroring today's `BeaconVote` plus the Gloas `AttestationData.index` field. Implement `QbftData` on it and add a `QbftDecidable` dispatch in `anchor/qbft_manager/src/lib.rs` so that QBFT instances at Gloas-and-later slots decide on `GloasBeaconVote`, while pre-Gloas slots continue on the existing `BeaconVote`.

## Context

Per SIP-94 §2 (`ssvlabs/SIPs`, commit `e849eeb`), under Gloas the SSV cluster must thread the BN-supplied `AttestationData.index` value through QBFT so that all operators sign over identical attestation roots. The wire-format change is a new SSZ type rather than an in-place extension of `BeaconVote`; the SIP names it `GloasBeaconVote` and motivates the two-type design as the cleanest way to make pre-Gloas and Gloas wire bytes mutually-rejecting on SSZ length mismatch (SSZ derives do not support fork-conditional fields).

Both `BeaconVote` (3 fields, unchanged) and `GloasBeaconVote` (4 fields) coexist on `feat/epbs` for the duration of the fork window. A future SIP retires `BeaconVote` and renames `GloasBeaconVote` back to `BeaconVote` once Gloas has activated on all networks.

Field correspondence: the SIP types `attestation_data_index` as `phase0.CommitteeIndex` (a `uint64` alias), so the Rust field is `u64`. Range and same-slot constraints are enforced by the value check in a follow-up issue, not at the struct level.

## Suggested approach

- Add `GloasBeaconVote` to `consensus.rs` next to the existing `BeaconVote`, with SSZ derive plus a `QbftData` impl that mirrors `BeaconVote::hash()` (SHA-256 over SSZ bytes for cross-operator determinism).
- Add a `QbftDecidable` impl for `GloasBeaconVote` in `qbft_manager` and dispatch on `active_fork(epoch) >= Fork::CStar` at the consumption boundary.
- Public re-exports as needed.

At consumption points (`metadata_service.rs`, `sign_committee_attestations`), dispatch checks `active_fork` at the relevant slot and uses `BeaconVote` or `GloasBeaconVote`.

## Acceptance criteria

[...]

PR #1071, opened 2026-06-02
feat(ssv_types): fork-aware GloasBeaconVoteValidator with slashability

### LINKED ISSUE #1026: feat(ssv_types): fork-aware GloasBeaconVoteValidator with slashability
## Goal

Add a new `GloasBeaconVoteValidator` in `anchor/common/ssv_types/src/consensus.rs` next to today's `BeaconVoteValidator`. The new validator runs on Gloas-and-later slots, range-checks `attestation_data_index < 2`, and threads the decided index into the slashability check so double-vote detection works correctly under Gloas. The dispatch in `anchor/validator_store/src/lib.rs` already landed with #1025/PR #1059: `sign_committee_attestations`' fork branch calls `create_gloas_beacon_vote_validator` (lib.rs:891), currently a `todo!()` stub, so this issue is a body-only change (implement the validator, fill the stub). Open PR #1071 implements it.

## Context

Builds on #1025 (`GloasBeaconVote` type + `QbftDecidable` dispatch). Per SIP-94 §2 (`ssvlabs/SIPs`, commit `e849eeb`; quotes re-verified at head `4e02e6a`), the Gloas value check must:

1. Range-check `attestation_data_index` in `{0, 1}` (SIP §2: "rejects `AttestationDataIndex` values other than `0` or `1`").
2. Thread the decided `attestation_data_index` into the `AttestationData` passed to slashability, so the double-vote predicate trips when an operator is asked to sign both `index=0` and `index=1` for the same `(source, target, slot)`. Today's flow uses `committee_index` from the per-validator map; under Gloas the single decided index must be used (SC-1 invariant in SIP §2 Security Considerations).
3. NOT compare the proposed `attestation_data_index` against each operator's local BN view (SC-2 in SIP §2 Security Considerations). Operators' fork-choice `execution_payload_availability` views may legitimately disagree: envelopes are published around the 75% mark of the producing slot, and by the standard attestation deadline at 25% of the next slot some BNs have processed the update while others haven't. Requiring local agreement at QBFT value-check time would fail rounds whenever the cluster spans this kind of transient.

Pre-Gloas slots continue running today's `BeaconVoteValidator` unchanged.

## Suggested approach

Mirror `BeaconVoteValidator::do_validation` for `GloasBeaconVote`. All five existing checks carry over unchanged:

- Far-future target rejection (`target.epoch > current_epoch + 1`).
- `source.epoch < target.epoch` (with genesis exception).
- Majority-fork-protection: `strict_majority_fork_protection` vs `epoch_majority_fork_protection` per the `strict_mfp` flag, comparing `source`/`target` only (no `attestation_data_index` comparison).
- `check_attestation_slashing` per validator pubkey via `preliminary_check_attestation`.

Two Gloas-specific additions:

- After the source/target checks: range-check `attestation_data_index < 2`.
- Inside `check_attestation_slashing`, set `attestation_data.index = value.attestation_data_index` for every validator in the loop, replacing today's `attestation_data.index = *committee_index`. The per-validator `committee_index` map (from `MetadataService`) is still used to identify which of the operator's validators this attestation is for, but stops controlling `index`.

In `validator_store/src/lib.rs`, the `active_fork(epoch) >= Fork::CStar` dispatch already exists (landed with #1059); fill the `create_gloas_beacon_vote_validator` `todo!()` body (lib.rs:891-897) to construct the new validator.

## Acceptance criteria

- Gloas slot: `attestation_data_index` in `{0, 1}` accepted; `>= 2` rejected at value check.
- Gloas slot: `check_attestation_slashing` reconstructs each validator's `AttestationData` with `index = decided.attestation_data_index`.
- Gloas slot: all five carried-over checks (far-future target, source < target, MFP, slashing-DB call, genesis exception) behave identically to today's `BeaconVoteValidator`.
- Pre-Gloas slot: existing `BeaconVoteValidator` behavior unchanged.
- Negative test (slashing-protection reconstruction): same `(slot, source, target)` decided with two different indices produces two different signing roots; the local slashing DB must trip on the second signature. The acceptance is about index reconstruction, not on-chain validity. Under SIP-94 §2, same-slot attestations require `data.index = 0`, so an `index=1` same-slot attestation is missed/rejected for normal inclusion rather than "valid on-chain."
- Value check does NOT call any "does this match my BN view?" guard for `attestation_data_index` (no extension of MFP to include the new field).

## Risks

- **Slashability false negative if missed.** If `check_attestation_slashing` keeps using the per-validator `committee_index` for the `AttestationData.index` reconstruction, a Gloas operator can be tricked into signing both `(index=0)` and `(index=1)` for the same `(source, target, slot)` without local DB tripping. Both signed roots verify on-chain; a constructible `AttesterSlashing` is the result.
- **Same-slot mandate is BN-enforced, not SSV-enforced.** The spec rule `same-slot → index = 0` is enforced at BN production time (LH's `produce_unaggregated_attestation` hard-codes `index = 0` when `is_same_slot_attestation`) and at the gossip REJECT layer. Detecting same-slot at the SSV value-check layer would require a BN lookup (block-at-slot or root-to-slot), which is the "compare local BN view" pattern forbidden by SC-2. Worst case if a malicious leader proposes `index = 1` for a same-slot attestation is a single missed attestation; no slashing risk.
- **MFP does not extend to `attestation_data_index`.** Adding the new field to `strict_majority_fork_protection`/`epoch_majority_fork_protection` is intentionally NOT done. The same liveness reasoning that led to `epoch_majority_fork_protection` ignoring `target.root` applies here (BN divergence around envelope arrival), and `attestation_data_index` has no structural sub-component to partial-match on. SC-2 picks "no comparison" as the only useful lenient option.

petarjuki7:

Under Gloas (`Fork::CStar`), `AttestationData.index` is BN-supplied and part of the signed attestation root and the double-vote slashing predicate. #1025 added the `GloasBeaconVote` type and the fork-gated QBFT dispatch but left the value check as a placeholder. Without a Gloas-aware validator, the cluster cannot range-check the index, and slashing protection would reconstruct the wrong root — an operator could be walked into signing `index=0` and `index=1` for the same `(source, target, slot)` without local protection tripping.

Resolves #1026. Spec source: SIP-94 §2.

## Change Overview

- New `GloasBeaconVoteValidator` next to `BeaconVoteValidator`. Every pre-Gloas check carries over (far-future target, source-before-target with genesis exception, majority-fork-protection, per-validator slashing-DB call). Two Gloas rules are added: `attestation_data_index` is range-checked to `{0, 1}`, and the slashing-DB check reconstructs `AttestationData` with the QBFT-decided index instead of the per-validator committee index.
- The decided index is trusted from the leader: never compared against the local BN view, and not folded into majority-fork-protection. BNs legitimately diverge on payload status around the envelope reveal, so requiring local agreement would fail rounds during a normal transient. The same-slot `index = 0` rule is BN/gossip-enforced, not checked here.
- Wired into `sign_committee_attestations` via the fork branch from #1025, replacing the placeholder.
- `IndexOutOfRange` is added to the shared `BeaconVoteValidationError` rather than a cloned enum — the two validators are a temporary fork-window pair that collapses to one when `BeaconVote` is retired, so the shared error and deliberate mirror-not-abstract duplication keep the eventual deletion a one-step change.

- **Slashing-adjacent code.** The index reconstruction is the load-bearing safety property. Covered by a negative test that seeds a real slashing DB at `index=0` and asserts `index=1` over the same `(source, target, slot)` trips, with a control proving the reconstruction uses the decided index (not the committee index). The seed domain is computed byte-identically to the validator so the trip is index-driven, not a domain artifact.
- **Interim completeness.** The value check now reconstructs slashing with the decided index, but threading that index into the actual signed attestation is #1027 (marked with a `TODO`). The chain `#1025 → #1026 → #1027` lands together on `epbs` before CStar activation; `#1026` in isolation makes the value check correct.
- **Deliberate duplication** of the pure source/target checks between the two validators, rather than a shared generic. The pair is collapsed at fork retirement, so the method-on-validator shape is the post-fork target; extracting a shared abstraction now would move away from it and drag stable code into this PR.

## Validation

- New unit tests in `ssv_types`: index `{0,1}` accepted / `>= 2` rejected (variant + payload); the carried-over checks behave identically to `BeaconVoteValidator`; majority-fork-protection ignores the index under both strict and non-strict modes; and the slashing reconstruction negative test described above.
- `cargo test -p ssv_types` (103) and `-p anchor_validator_store` (33) pass; `make cargo-fmt-check` and clippy clean.

## Rollback

Additive at the type and validator layer. Reverting restores the placeholder; pre-Gloas behavior is unchanged throughout.

## Blockers / Dependencies

Builds on #1025 (merged). #1027 threads the decided index into the signing path.

### DIFF
--- a/anchor/common/ssv_types/src/consensus.rs
+++ b/anchor/common/ssv_types/src/consensus.rs
@@ -24,8 +24,8 @@ use typenum::{
 };
 use types::{
     AggregateAndProofBase, AggregateAndProofElectra, AttestationBase, AttestationData,
-    AttestationElectra, BlindedBeaconBlock, ChainSpec, Checkpoint, CommitteeIndex, Domain, EthSpec,
-    ForkName, Hash256, Slot, SyncCommitteeContribution,
+    AttestationElectra, BeaconBlock, BlindedBeaconBlock, ChainSpec, Checkpoint, CommitteeIndex,
+    Domain, EthSpec, ForkName, Hash256, Slot, SyncCommitteeContribution,
 };
 
 use crate::{CommitteeId, ValidatorIndex, message::*, partial_sig::PartialSignatureKind};
@@ -251,6 +251,9 @@ impl ProposerConsensusData {
     /// Decode the block data as a blinded beacon block.
     pub fn decode_blinded_block<E: EthSpec>(&self) -> Result<BlindedBeaconBlock<E>, DecodeError> {
         let fork = ForkName::from(self.version);
+        if fork >= ForkName::Gloas {
+            return Err(DecodeError::NoMatchingVariant);
+        }
         BlindedBeaconBlock::from_ssz_bytes_for_fork(&self.data_ssz, fork)
     }
 
@@ -258,6 +261,12 @@ impl ProposerConsensusData {
     pub fn decode_block_contents<E: EthSpec>(&self) -> Result<FullBlockContents<E>, DecodeError> {
         let fork = ForkName::from(self.version);
         FullBlockContents::from_ssz_bytes_for_fork(&self.data_ssz, fork)
+    }
+
+    /// Decode as a full beacon block shape.
+    pub fn decode_block<E: EthSpec>(&self) -> Result<BeaconBlock<E>, DecodeError> {
+        let fork = ForkName::from(self.version);
+        BeaconBlock::from_ssz_bytes_for_fork(&self.data_ssz, fork)
     }
 }
 
@@ -379,17 +388,30 @@ impl<E: EthSpec> ProposerConsensusDataValidator<E> {
         &self,
         value: &ProposerConsensusData,
     ) -> Result<(), DataValidationError> {
-        // Always do this check, even if we're not validating slashing. This is to ensure that we
-        // have a decodable value.
-        let header = value
-            .decode_blinded_block::<E>()
-            .map(|block| block.block_header())
-            .or_else(|_| {
-                value
-                    .decode_block_contents::<E>()
-                    .map(|block| block.block().block_header())
-            })
-            .map_err(DataValidationError::DecodeError)?;
+        // Decode the block header to ensure the value is decodable (even when slashing
+        // protection is disabled). Under Gloas (EIP-7732), DataSSZ is decoded directly as a plain
+        // BeaconBlock. This behaviour is not behaviourally load-bearing as the execution
+        // payload is decoupled from the block body. The outcome of `decode_blinded_block`
+        // and `decode_block` are identical for this variant. The Pre-Gloas branch
+        // preserves the existing try-blinded-then-full fallback.
+        let fork = ForkName::from(value.version);
+
+        let header = if fork >= ForkName::Gloas {
+            value
+                .decode_block::<E>()
+                .map(|block| block.block_header())
+                .map_err(DataValidationError::DecodeError)?
+        } else {
+            value
+                .decode_blinded_block::<E>()
+                .map(|block| block.block_header())
+                .or_else(|_| {
+                    value
+                        .decode_block_contents::<E>()
+                        .map(|block| block.block().block_header())
+                })
+                .map_err(DataValidationError::DecodeError)?
+        };
 
         if !self.disable_slashing_protection {
             let epoch = header.slot.epoch(E::slots_per_epoch());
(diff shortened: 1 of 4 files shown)


anchor/common/ssv_types/src/consensus.rs
```diff
@@ -2100,6 +2283,515 @@ mod tests {
...
+                assert_eq!(mismatch.proposed_source_epoch, our_source.epoch);
+                assert_eq!(mismatch.our_target_epoch, our_target.epoch);
+                assert_eq!(mismatch.proposed_target_epoch, proposed_target.epoch);
+            }
+            err => panic!("Expected EpochMismatch error, got: {:?}", err),
+        }
+    }
+
+    #[test]
+    fn test_gloas_strict_mismatched_target_different_roots() {
```
shane-moore, line 2517: Section B ports `test_strict_mismatched_target_different_roots` but not its non-strict sibling `test_valid_matching_epochs_different_roots` (same source/target *epochs*, different *roots* → accepted under `epoch_majority_fork_protection`). Cheapest fix: extend `test_gloas_mfp_ignores_index` to also vary the source/target roots, so it proves different-roots *and* different-index both pass under non-strict MFP.

shane-moore:
@claude, please take a look, and keep my pr comments in mind as well

anchor/common/ssv_types/src/consensus.rs
shane-moore: Resolved in 2788297, which adds `test_gloas_valid_matching_epochs_different_roots`: a direct port of the non-strict sibling, with `attestation_data_index` held at 0 on both votes to isolate the root difference.

shane-moore (approved):
lgtm!

Issue #1017, opened 2026-05-14
feat(validator_store): adapt sign_block destructure for Gloas FullBlockContents


shane-moore:
## Goal

Adapt Anchor's `sign_block` in `anchor/validator_store/src/lib.rs` to Lighthouse's updated `FullBlockContents<E>` shape once LH implements the beacon-APIs PR [580](https://github.com/ethereum/beacon-APIs/pull/580) stateless path on Gloas. The work is mechanical destructure-adjustment; QBFT inputs and SSV cluster signing surfaces are unchanged.

## Context

Today, LH's V4 helper decodes Gloas block responses as `BeaconBlock<E>` directly and wraps them as `FullBlockContents::Block(...)`. LH carries a known TODO in `common/eth2/src/types.rs:1894` (re-verified at the current Anchor pin `81d576943`; inside `FullBlockContents::from_ssz_bytes_for_fork`: "revisit when produceBlockV4 PR is finalised") flagging that the current SSZ encoding diverges from PR ethereum/beacon-APIs#580's `anyOf: [BeaconBlock, BlockContents]` body shape on self-build slots. When LH lands the update, the V4 helper will return a Gloas variant of `FullBlockContents` carrying the inline `ExecutionPayloadEnvelope` (and likely `kzg_proofs`/`blobs`) on stateless self-build slots.

Anchor's destructure in `sign_block` must grow a new match arm that extracts `.block` into `blinded_block` and explicitly drops the envelope from the QBFT input per SIP-94 §4/§6 (the block path does not sign or publish the envelope; envelope signing is the separate §6 EnvelopeProposer duty, tracked in milestone 8: #1120, #1121, #1122, #1124, #1126). Everything downstream of the destructure (`decide_abstract_block`, `decode_gloas_block`, `sign_abstract_block`) is unaffected because `blinded_block` is already a `BeaconBlock`-shaped value before it enters QBFT; `data_ssz` continues to be SSZ-encoded `Gloas.BeaconBlock` bytes, byte-identical to today.

## Suggested approach

Extend the `block_type` match and add a match arm for the new Gloas variant of `FullBlockContents`. Extract `.block` into `blinded_block`; explicitly drop the envelope. Decide whether `kzg_proofs`/`blobs` belong in `local_full_block` for publish, contingent on LH's blob-handling choice under Gloas (LH may move blob publication to the envelope-side flow, in which case `local_full_block` should not carry blobs on self-build slots).

[...]

PR #1072, opened 2026-06-04
feat(validator_store): fork-gate decide_abstract_block for Gloas block decoding

### LINKED ISSUE #1015: feat(validator_store): fork-gate decide_abstract_block; skip blinded decode for Gloas

jnhsigmap:

- Under Gloas (EIP-7732), `DataSSZ` carries a plain `BeaconBlock` — the existing blinded-then-full fallback wastes a guaranteed-to-fail `decode_blinded_block` attempt and could mask decode errors by silently falling through.
- Per SIP-94, the SSV cluster signs only `BeaconBlock` under `Domain::BeaconProposer`; envelope, blobs, and KZG proofs do not enter consensus.
- Closes #1015. Depends on #1070 (merged).

## Change Overview (Required)

- Fork-gates the decode+wrap logic in `decide_abstract_block`: Gloas decodes directly via `decode_block()` → `UnsignedBlock::Full(FullBlockContents::Block(_))`; pre-Gloas preserves the existing blinded-then-full fallback unchanged.
- Adds an integration test that exercises `decide_abstract_block` end-to-end with a Gloas `BeaconBlock` through the mock consensus harness.
- Adds unit tests for the Gloas decode primitive (happy path + garbage rejection).

- **Risk:** Fork boundary off-by-one. Mitigated: boundary (`>= ForkName::Gloas`) is consistent with `decode_blinded_block`'s fork-guard and `validate_block_proposal` in `ssv_types`.
- **Risk:** Regression in pre-Gloas path. Mitigated: `else` branch is the original expression verbatim; full test suite passes.
- **Trade-off:** Tests the dispatch boundary via integration test rather than extracting a pure function. Extraction is a candidate for follow-up refactor (see below).

## Validation (Required)

- `make cargo-fmt-check`
- `make lint`
- `make test`
- `cargo test -p anchor_validator_store`

- Revert returns to the unconditional blinded-then-full fallback — no data or config impact. Pre-Gloas behavior is unchanged so revert is safe at any time before Gloas activation.

- Depends on #1070 (merged) which added `decode_block` and the `decode_blinded_block` fork-guard.
- #1017 (adapt `sign_block` destructure for future LH `FullBlockContents` Gloas variant) is blocked on Lighthouse and builds on this PR.

- **Potential refactor (separate issue):** The decode+wrap expression in `decide_abstract_block` could be extracted into a small pure function (e.g. `decode_decided_block(completed_data) -> Result<UnsignedBlock<E>, DecodeError>`). This would allow direct unit testing of the fork-gating dispatch without the async QBFT infrastructure, making the boundary testable in isolation and reducing test coupling. Not blocking for this PR — the integration test catches branch-inversion bugs — but worth tracking as tech debt cleanup.


jnhsigmap:
@claude, review this PR

jnhsigmap:
Apparently - Revised verdict: it's a Bedrock-side inference hang — not a file-access problem :shrug:

jnhsigmap:
> **Claude finished @jnhsigmap's task in 3m 24s** —— [View job](https://github.com/sigp/anchor/actions/runs/26950595905)
> 
> ### PR Review: fork-gate `decide_abstract_block` for Gloas
> * [x]  Read changed files
> * [x]  Analyze the fork-gate logic
> * [x]  Check tests for correctness
> * [x]  Post review feedback
> 
> ### Summary
> The fork-gate change is small, surgical, and correct. The pre-Gloas branch is preserved verbatim, and the Gloas branch correctly avoids the now-guaranteed-to-fail `decode_blinded_block()` and the silent-fallthrough hazard documented in the PR description. The boundary `>= ForkName::Gloas` is consistent with the fork-guard in [`ProposerConsensusData::decode_blinded_block`](anchor/common/ssv_types/src/consensus.rs#L252-L257) and [`validate_block_proposal`](anchor/common/ssv_types/src/consensus.rs#L399-L414), so a future SIP-style boundary nudge has a single, narrow set of call sites to revisit.

...

> **Unit tests** — `anchor/validator_store/src/lib.rs:3484–3564` ⚠️ Redundant
> 
> Both `gloas_decided_data_decodes_to_block_via_full_path` and `gloas_decided_data_rejects_invalid_bytes` call `ProposerConsensusData::decode_block` directly. They do **not** exercise the new fork-gate dispatch in `decide_abstract_block` — the `if ForkName::from(...) >= ForkName::Gloas` branch is never traversed by these tests. They are functionally equivalent to existing tests in `ssv_types`:
> 
> * `decode_block_round_trip` ([`anchor/common/ssv_types/src/consensus.rs:2478`](anchor/common/ssv_types/src/consensus.rs#L2478))
> * The garbage-rejection assertion is essentially what `BeaconBlock::from_ssz_bytes_for_fork` already covers
> 
> The first test even constructs `UnsignedBlock::Full(FullBlockContents::Block(b))` manually — i.e., it's testing tuple-struct construction, not anything in this PR.
> 
> Suggestion: drop both unit tests, or — if you want fork-gate test coverage at the unit level — extract the decode-and-wrap expression into a small pure function (the PR description already flags this as a follow-up). Either way, the integration test is the load-bearing one. Keeping the redundant unit tests adds maintenance cost without adding signal.
> 
> [Fix this → drop the redundant `gloas_decide_tests` module](https://claude.ai/code?q=On%20branch%20feat%2F1015%2Fdecide-abstract-block%2C%20remove%20the%20%60gloas_decide_tests%60%20module%20at%20anchor%2Fvalidator_store%2Fsrc%2Flib.rs%3A3484-3564.%20Reason%3A%20both%20tests%20call%20%60ProposerConsensusData%3A%3Adecode_block%60%20directly%20and%20do%20not%20exercise%20the%20fork-gate%20dispatch%20in%20%60decide_abstract_block%60.%20They%20duplicate%20%60decode_block_round_trip%60%20in%20anchor%2Fcommon%2Fssv_types%2Fsrc%2Fconsensus.rs%3A2478.%20The%20integration%20test%20at%20anchor%2Fvalidator_store%2Fsrc%2Ftesting%2Fproposer_block.rs%20provides%20the%20real%20coverage.%20After%20deleting%20the%20module%2C%20run%20%60make%20cargo-fmt%20%26%26%20make%20lint%20%26%26%20cargo%20test%20-p%20anchor_validator_store%60%20to%20confirm.&repo=sigp/anchor)
> 
> ### Minor observations (non-blocking)
> * The integration test only covers the Gloas branch. A symmetric pre-Gloas test (e.g. Electra `BeaconBlockContents`) would catch a branch-inversion bug in the _other_ direction. Not required for this PR — pre-Gloas behavior is unchanged — but worth considering as the test harness already supports it.
> * The doc comment at `lib.rs:597–598` is good and explains the _why_. 👍
...

Correct, I overlooked that. Removed in https://github.com/sigp/anchor/pull/1072/commits/99bbefbcb93cfb9beda13e7b4471c7767bf5e350. Perhaps we test the current pre-gloas path in a refactor. Will open an issue post team feedback.

shane-moore:
The current test exercises the new `decide_abstract_block` Gloas branch, but not the production caller shape. `sign_block` first projects full inputs with `block.to_ref().into()` and calls `decide_abstract_block` with that blinded-shaped block, so for Gloas `data_ssz` comes from the blinded projection and is decoded via `decode_block()` as a full `BeaconBlock`. That works only because the pinned LH Gloas full/blinded SSZ is byte-identical, and neither this test (full input) nor #1070 covers the blinded-projection input. Confirmed locally against pin `1a6863118`: `full.as_ssz_bytes() == blinded.as_ssz_bytes()`, and `decode_block` yields `Full(Block(_))` for both.

Since this shapes the branch the PR adds (behavior-neutral, improves coverage), worth doing here rather than as follow-up: extract the decode+wrap into a pure helper and unit-test it, covering Gloas full, Gloas blinded-projection, invalid Gloas bytes, and the pre-Gloas fallback. The helper also collapses the duplicated `map_err`:

```rust
fn decode_decided_block<E: EthSpec>(
    data: &ProposerConsensusData,
) -> Result<UnsignedBlock<E>, DecodeError> {
    if ForkName::from(data.version) >= ForkName::Gloas {
        data.decode_block().map(|b| UnsignedBlock::Full(FullBlockContents::Block(b)))
    } else {
        data.decode_blinded_block()
            .map(UnsignedBlock::Blinded)
            .or_else(|_| data.decode_block_contents().map(UnsignedBlock::Full))
    }
}
```

Worth asserting `full.as_ssz_bytes() == blinded.as_ssz_bytes()` in the blinded-projection test to pin that invariant directly.

shane-moore:
To be explicit on the upside: with the pure helper these become plain `#[test]`s that build a `ProposerConsensusData`, call `decode_decided_block`, and assert, same shape as `decode_block_round_trip` in #1070. That drops `ValidatorStoreTestHarness`, `create_committee_setup`, `MockConsensusDecider`, and `#[tokio::test]` from this test, so it's a net reduction rather than added coverage cost. (The harness stays for the aggregate/attestation tests; only `proposer_block.rs` stops depending on it.)

shane-moore:
Re: "do the refactor now and test from `sign_block`?" - do it here, but aim a level *below* `decide_abstract_block`, not above at `sign_block`. Your Additional Info already calls the target: extract `decode_decided_block` so the dispatch is testable "without the async QBFT infrastructure." `sign_block` goes the other way and adds infra (sync gate, slashing protection, threshold-sig collection), not less.

What you're after from `sign_block` is the production caller shape, the blinded projection. You get that at the helper level without the signing path: build the full block, project with `.to_ref().into()` in arrange, call `decode_decided_block`, and assert `full.as_ssz_bytes() == blinded.as_ssz_bytes()` to pin the invariant directly. That's the `as_ssz_bytes` assert from my earlier comment, doing exactly the job `sign_block` would, minus the harness. Net reduction too: `proposer_block.rs` drops `ValidatorStoreTestHarness`, `MockConsensusDecider`, and `#[tokio::test]`. So worth promoting from follow-up into this PR.

jnhsigmap:
> The current test exercises the new `decide_abstract_block` Gloas branch, but not the production caller shape. `sign_block` first projects full inputs with `block.to_ref().into()` and calls `decide_abstract_block` with that blinded-shaped block, so for Gloas `data_ssz` comes from the blinded projection and is decoded via `decode_block()` as a full `BeaconBlock`. That works only because the pinned LH Gloas full/blinded SSZ is byte-identical, and neither this test (full input) nor #1070 covers the blinded-projection input. Confirmed locally against pin `1a6863118`: `full.as_ssz_bytes() == blinded.as_ssz_bytes()`, and `decode_block` yields `Full(Block(_))` for both.
> 
> Since this shapes the branch the PR adds (behavior-neutral, improves coverage), worth doing here rather than as follow-up: extract the decode+wrap into a pure helper and unit-test it, covering Gloas full, Gloas blinded-projection, invalid Gloas bytes, and the pre-Gloas fallback. The helper also collapses the duplicated `map_err`:
> 
> ```rust
> fn decode_decided_block<E: EthSpec>(
>     data: &ProposerConsensusData,
> ) -> Result<UnsignedBlock<E>, DecodeError> {
>     if ForkName::from(data.version) >= ForkName::Gloas {
>         data.decode_block().map(|b| UnsignedBlock::Full(FullBlockContents::Block(b)))
>     } else {
>         data.decode_blinded_block()
>             .map(UnsignedBlock::Blinded)
>             .or_else(|_| data.decode_block_contents().map(UnsignedBlock::Full))
>     }
> }
> ```
> 
> Worth asserting `full.as_ssz_bytes() == blinded.as_ssz_bytes()` in the blinded-projection test to pin that invariant directly.

In response in https://github.com/sigp/anchor/pull/1072/commits/c2dc3ff86c1182e036679389984fc7204d087144:
- Removed integration test and `proposer_block.rs`.
- `ValidatorStoreTestHarness` and async runtime no longer necessary for testing.

  | Test | Branch |
  |------|--------|
  | `gloas_full_block_decodes_via_full_path` | Gloas: full block SSZ → `Full(Block(_))` |
  | `gloas_blinded_projection_decodes_via_full_path` | Gloas: blinded projection SSZ (production caller shape) + `assert_eq!(full.as_ssz_bytes(), blinded.as_ssz_bytes())` to pin byte-identity |
  | `gloas_invalid_bytes_rejected` | Gloas: garbage → `Err` |
  | `pre_gloas_blinded_block_decodes_directly` | Pre-Gloas path 1: blinded SSZ → `decode_blinded_block` succeeds → `Blinded(_)` |
  | `pre_gloas_full_block_contents_decodes_via_fallback` | Pre-Gloas path 2: `FullBlockContents` SSZ → `decode_blinded_block` fails → `or_else` → `decode_block_contents` succeeds → `Full(_)` |
  | `pre_gloas_invalid_bytes_rejected` | Pre-Gloas path 3: garbage → both decoders fail → `Err` |

shane-moore (approved):
lgtm!

PR #1076, opened 2026-06-05
refactor(ssv_types): remove PayloadAttestationVote QBFT value object

shane-moore:
## Problem, Evidence, and Context

- `PayloadAttestationVote` and its QBFT machinery (the value object, its `QbftData` impl, the value validator, and their tests) are dead code in `ssv_types`. This reverts #1047.
- SIP-94 §3 was rewritten to a validator-scoped, leaderless PTC design with no per-slot QBFT instance, so there is no consensus value to negotiate for PTC. The only intended consumer, the `qbft_manager` PTC routing (#1057), was never merged.
- Evidence: `git grep PayloadAttestationVote anchor/` matches only the defining module (definition + tests), with no cross-crate users, so removal cannot break another crate.
- Links: closes #1074; reverts #1047; part of the no-QBFT PTC [milestone 6](https://github.com/sigp/anchor/milestone/6), which supersedes the QBFT-design [milestone 4](https://github.com/sigp/anchor/milestone/4) (retained as a fallback). Spec: ssvlabs/SIPs#94 §3.

## Change Overview

- Pure deletion confined to `ssv_types` (-203 lines, no other crates): the `PayloadAttestationVote` value object, the `PayloadAttestationVoteValidator` and its error enum, and the corresponding unit tests.
- For the reviewer: this is a straight revert of #1047 with no replacement code. Confirm the removed symbols are the QBFT value object and its validator, and that nothing outside this crate referenced them.
- Intentionally unchanged: `BeaconVote` and `GloasBeaconVote`, their validators, and their tests. No new types, no behavior change.

## Risks, Trade-offs, and Mitigations

- Blast radius: limited to `ssv_types`; the removed symbols had no callers elsewhere (verified by `git grep`).
- Trade-off: drops the QBFT-design PTC path in favor of the no-QBFT redesign; the QBFT version remains recoverable via #1047 / milestone #4 if that decision is reversed.
- Mitigation: no behavior change; the compiler and clippy confirm no orphaned imports or dead code remain.

## Validation

- `cargo check -p ssv_types` and `cargo clippy -p ssv_types --all-targets`: clean, no `unused import` / `dead_code` warnings.
- `cargo test -p ssv_types`: 92 passed, 0 failed (+ 2 doc-tests); the `BeaconVote` / `GloasBeaconVote` suites are untouched and green.
- `cargo fmt -p ssv_types -- --check`: clean.
- `git grep -n PayloadAttestationVote anchor/`: no matches.

## Rollback

[...]

Issue #1074, opened 2026-06-05
refactor(ssv_types): remove PayloadAttestationVote QBFT value object


shane-moore:
## Goal

Remove the `PayloadAttestationVote` QBFT consensus value (struct + `QbftData` impl + `PayloadAttestationVoteValidator` + error enum + tests) from `anchor/common/ssv_types/src/consensus.rs`. This reverts #1034 / PR #1047. Under the no-QBFT PTC redesign there is no per-slot QBFT instance over PTC data, so these are dead code.

## Context

SIP-94 §3 ([ssvlabs/SIPs#94](https://github.com/ssvlabs/SIPs/pull/94)) was rewritten to a validator-scoped, leaderless PTC design: no QBFT instance, no leader-decided value. The committee/QBFT artifacts for PTC are therefore removed. `PayloadAttestationVote` (added in #1034 / PR #1047) was only ever a QBFT consensus value, and its intended consumer, the `qbft_manager` PTC routing (#1035 / PR #1057), never merged. `git grep PayloadAttestationVote anchor/` on `epbs` returns only `consensus.rs` (definition + tests), so there are no cross-crate users and deletion cannot break another crate. Part of milestone #6; the QBFT-design version is #1034 (milestone #4), retained as a fallback.

## Suggested approach

All in `anchor/common/ssv_types/src/consensus.rs` (lead with the symbol; line hints drift, re-grep at file time):

- delete `struct PayloadAttestationVote` (~979) and its doc block
- delete `impl QbftData for PayloadAttestationVote` (~989)
- delete `enum PayloadAttestationVoteValidationError` (~1277)
- delete `struct PayloadAttestationVoteValidator` + `impl QbftDataValidator<PayloadAttestationVote>` + `impl PayloadAttestationVoteValidator { do_validation }` (~1286-1329)
- delete the `// PayloadAttestationVote Tests` region (the `create_payload_attestation_vote` helper + the 6 `test_payload_attestation_vote_*` tests, ~2201-2320), up to but not including the `// GloasBeaconVote Tests` banner
- let `cargo check` + clippy flag any now-orphaned `use` (confirm `Sha256` is still used by `BeaconVote` / `GloasBeaconVote`; likely retained)

## Acceptance criteria

- `git grep -n PayloadAttestationVote anchor/` returns nothing.
- `cargo check -p ssv_types` compiles with no `unused import` / `dead_code` warnings.
[...]

PR #1080, opened 2026-06-06
feat(ssv_types): retarget PTC role to validator-scoped PTCAttester

### LINKED ISSUE #1075: feat(ssv_types): retarget PTC role to validator-scoped PTCAttester + PartialSignatureKind::PTCAttester
## Goal

Retarget the PTC role from committee-scoped to validator-scoped: rename `Role::PTCCommittee` -> `Role::PTCAttester` (`is_committee_role = false`, `duty_executor = Validator`, `max_round = None`) and add a dedicated `PartialSignatureKind::PTCAttester = 7` (no longer reusing `PostConsensus`). Reworks #1032 / PR #1033.

## Context

SIP-94 §3 ([ssvlabs/SIPs#94](https://github.com/ssvlabs/SIPs/pull/94)) was rewritten to a validator-scoped, leaderless PTC design, modeled on `ProposerPreferences` (§5). With no consensus round, a PTC partial signature is a standalone single-validator signature, so it gets its own `PartialSignatureKind` rather than reusing `PostConsensus`, and the role uses `DutyExecutor::Validator` with no `qbft_manager` routing. `PTCAttester = 7` keeps value-parity with the SIP's `RunnerRole` / `PartialSigMsgType` / `BeaconRole` (all `7`); the wire byte stays `[7,0,0,0]` (unchanged from `PTCCommittee`), and `Domain::PTCAttester` is `0x0C`. `ProposerPreferences` (§5) renumbers to `8`. Part of milestone #6.

This is one logical change, but it breaks exhaustive matches across several crates, so it lands as one PR: the `Role` rename hits matches in `msgid.rs`, `message_validator`, and `qbft_manager`; the new `PartialSignatureKind` hits the two matches in `message_counts.rs`. `cargo check --workspace` is the gate that catches the cross-crate breaks (`-p ssv_types` alone will not).

## Suggested approach

Lead with the symbol; line hints drift, re-grep at file time.

`anchor/common/ssv_types/src/partial_sig.rs`:

| Symbol (line hint) | current -> target |
|---|---|
| `PartialSignatureKind` enum (~22-40) | add `PTCAttester = 7,` after `AggregatorCommitteePartialSig = 6,` |
| `TryFrom<u64>` (~42-57) | add `7 => Ok(PartialSignatureKind::PTCAttester),` before the `_ =>` arm |
| invalid-variant sentinel tests (**two**: `partial_signature_kind_ssz_decode_invalid_variant` ~239 and `partial_signature_kind_try_from_u64_invalid_values` ~307) | bump `7u64 -> 8u64` in **both** (7 is now valid; `ProposerPreferences = 8` is not yet in Anchor, so 8 is the next free invalid value) |

`anchor/common/ssv_types/src/msgid.rs` (keep wire byte `[7,0,0,0]`; rename + reclassify):

| Symbol (line hint) | current -> target |
|---|---|
| `Role` enum variant (~24) | `PTCCommittee` -> `PTCAttester` |
| `From<Role> for [u8;4]` (~37) | `Role::PTCAttester => [7, 0, 0, 0]` |
| `TryFrom<&[u8]>` (~54) | `[7, 0, 0, 0] => Ok(Role::PTCAttester)` |
| `is_committee_role` (~69-73) | drop PTC from the committee `matches!` arm |
| `max_round` (~80-88) | delete the `PTCCommittee => Some(4)` arm; add PTC to the `ValidatorRegistration \| VoluntaryExit => None` arm |
| `duty_executor` (~158-168) | remove PTC from the `Committee` arm; add `Role::PTCAttester` to the `DutyExecutor::Validator` (pubkey) arm |
| `is_qbft_role` (new helper) | add `pub fn is_qbft_role(self) -> bool { self.max_round().is_some() }` (positively named, mirroring `is_committee_role`, and negated at call sites; renamed from `is_non_qbft_role` per review); consumed by the `consensus_message` guard below to reject consensus messages for validator-scoped non-QBFT roles |

`anchor/message_validator/src/message_counts.rs` (easy-to-miss break site: two `match PartialSignatureKind` blocks, no wildcard). PTC joins the **pre-consensus** arm (not `PostConsensus`): `post_consensus` is specifically the partial sig over the QBFT-decided value; PTC is a standalone, non-QBFT single-validator signature like `ValidatorRegistration` / `VoluntaryExit`, which both sit in the pre-consensus arm. (Functionally moot for PTC in isolation, since its MessageId is distinct and both buckets cap at 1, but the correct/consistent bucket is pre-consensus.)

| Symbol (line hint) | current -> target |
|---|---|
| `validate_partial_signature_message`, pre-consensus arm (~64-69) | add `\| PartialSignatureKind::PTCAttester` to the pre-consensus arm |
| `record_partial_signature`, pre-consensus arm (~107-112) | add `\| PartialSignatureKind::PTCAttester` to the `=> self.pre_consensus += 1` arm |

`anchor/message_validator/src/partial_signature.rs`:

| Symbol (line hint) | current -> target |
|---|---|
| `partial_signature_type_matches_role` (~150) | split `Role::Committee \| Role::PTCCommittee => kind == PostConsensus` into `Role::Committee => kind == PostConsensus` + new arm `Role::PTCAttester => kind == PartialSignatureKind::PTCAttester` |
| per-validator packet-bound arm (~327-333) | **move `Role::PTCAttester` into the per-validator arm** (with `Aggregator \| Proposer \| ValidatorRegistration \| VoluntaryExit`), which enforces `message_count > 1 -> reject`. The `> 1` bound subsumes the old committee occurrence cap, so **both** now-dead PTC helpers are deleted: `validate_ptc_committee_message_count` (rejected only `message_count > validator_count`, i.e. committee batching) and `validate_validator_index_occurrence_limit` (the per-index cap, also no longer used). No signing-root dedup set (PTC has one root per `(validator, slot)`). |

`anchor/message_validator/src/lib.rs`:

| Symbol (line hint) | current -> target |
|---|---|
| `committee_info` dispatch (~453) | remove PTC from the committee arm; add `Role::PTCAttester` to the **validator** arm (with ValidatorRegistration/VoluntaryExit) |
| `validate_role_for_fork` (~862) | rename arm; keep the `< Fork::CStar` reject |
| `message_lateness` / TTL (~932) | move `Role::PTCAttester` into the short slot-bound arm with `Proposer \| SyncCommittee` (`1 + LATE_SLOT_ALLOWANCE` = 3 slots). Payload attestations are gossip-valid only for their own slot and block-includable only at slot + 1, so stale partials are useless; the epoch-long bucket was inherited from the committee-scoped design. (Landed in #1080 as a follow-up commit.) |
| `duty_limit` (~1052) | fold `Role::PTCAttester` into the `Role::Aggregator \| Role::ValidatorRegistration => Ok(Some(2))` arm. **Not** the old `min(slots_per_epoch, V)` formula: PTC is validator-scoped, so `committee_info.validator_indices` is always length 1 and the duty counter is keyed per-validator, making that formula evaluate to `Some(1)` in production (and `Some(0)` on missing metadata). `Some(2)` = ~1 PTC duty/epoch/validator + a one-duty boundary margin, matching the sibling validator-scoped roles. |
| test helper `create_message_id_for_test` (~1301) | move PTC from committee arm to `DutyExecutor::Validator(...)` |

`anchor/message_validator/src/consensus_message.rs`: the "duty role has consensus" guard in `validate_consensus_message_semantics` switches from a hardcoded `ValidatorRegistration | VoluntaryExit` list to `!role.is_qbft_role()`, so a consensus message addressed to `PTCAttester` is now rejected as `UnexpectedConsensusMessage` (PTC is non-QBFT). The previously unreachable `FailedToGetMaxRound` fallback below the guard is deleted: the defensive branch now also returns `UnexpectedConsensusMessage` and the error variant is removed from `ValidationFailure`. Tests: `test_consensus_message_for_non_consensus_role` is parameterized over all non-QBFT roles including `PTCAttester` (as landed in #1080).

`anchor/qbft_manager/src/lib.rs` (**both `match msg_id.role()` arms are exhaustive by explicit enumeration with no `_` wildcard, so a stray `PTCAttester` is E0004**):

| Symbol (line hint) | current -> target |
|---|---|
| validator-executor error arm (~277-283) | rename `Role::PTCCommittee -> Role::PTCAttester` and **move it into the non-QBFT group** with `ValidatorRegistration \| VoluntaryExit`. Do not drop it. PTC is validator-scoped + non-QBFT, so a QBFT message for it returns `InconsistentMessageId`. |
| stub arm `Some(Role::PTCCommittee) => { warn!("...routing not wired"); Err(RoleNotActive) }` (~341-347) | delete the **stub body/warn only**, and add `Role::PTCAttester` to the committee-executor error arm (~348-352), grouped with the non-QBFT roles (`ValidatorRegistration \| VoluntaryExit`) as in the validator-executor branch (it is a non-QBFT role, not a wrong-executor QBFT role). No catch-all arm exists here; omitting `PTCAttester` is E0004. |

## Acceptance criteria

- `git grep PTCCommittee anchor/` empty; `Role::PTCAttester` exists.
- `Role::PTCAttester.is_committee_role() == false`; `.max_round() == None`; `.is_qbft_role() == false`; `MessageId::new(.., Role::PTCAttester, &DutyExecutor::Validator(pk)).duty_executor() == Some(DutyExecutor::Validator(_))`.
- `PartialSignatureKind::PTCAttester as u64 == 7`; `TryFrom::<u64>::try_from(7) == Ok(PTCAttester)`; SSZ round-trips to `[7,0,0,0,0,0,0,0]`.
- `partial_signature_type_matches_role(PartialSignatureKind::PTCAttester, Role::PTCAttester) == true`; `(PostConsensus, PTCAttester) == false`.
- `validate_role_for_fork`: pre-CStar rejects with `RoleNotActiveBeforeFork { minimum_fork: Fork::CStar }`; CStar `Ok(())`.
- `message_lateness`: `Role::PTCAttester` uses the short slot-bound TTL (`1 + LATE_SLOT_ALLOWANCE` = 3 slots, alongside Proposer/SyncCommittee); a PTC partial 20 slots late is rejected as `LateSlotMessage`.
- `validate_consensus_message_semantics` rejects a consensus message for any non-QBFT role (`UnexpectedConsensusMessage`), now including `PTCAttester`, via the negated `Role::is_qbft_role()`.
- `duty_limit(Role::PTCAttester) == Some(2)` (folded into the `Aggregator \| ValidatorRegistration` arm). The earlier `Some(min(slots_per_epoch, V))` was wrong: validator-scoped dispatch supplies a length-1 `validator_indices` and the duty counter is per-validator, so the cap is a flat per-validator value, not a cluster-wide one.
- per-validator packet bound: > 1 `PayloadAttestationMessage` in one packet rejected; `validator_index` occurring > 1 time rejected (cap 1).
- `qbft_manager` has no dedicated PTC routing arm (`PTCAttester` sits in the `InconsistentMessageId` error arms of both branches); no `//todo(epbs)` stub.
- **`cargo check --workspace` green.**

## Tests (rename + retarget the existing PTC tests; as landed in #1080)

- `msgid.rs`: retarget the PTC duty-executor test -> `ptc_attester_uses_validator_duty_executor` (mirror `aggregator_committee_*`), plus the combined classification test `ptc_attester_is_validator_scoped_non_qbft` pinning `!is_committee_role()`, `max_round() == None`, and `!is_qbft_role()` (added in #1080 after review traced that no existing test fails if the classifications silently revert), and `role_qbft_classification_is_pinned`, which asserts every role's QBFT/non-QBFT classification on both sides of the partition (review feedback: `max_round`'s exhaustive match forces a decision for a new role, but nothing checked the decision was right).
- `partial_sig.rs`: bump `7 -> 8` in **both** invalid-variant tests; **extend the hand-written enumerated-variant arrays** (`..._ssz_roundtrip_all_variants`, `..._ssz_byte_layout`, `..._try_from_u64_all_valid_values`, TreeHash array) to include `PTCAttester` (no `EnumIter`, so they don't pick it up automatically). Those array rows already cover the variant value and SSZ encoding, so no dedicated `ptc_attester_*` value/encoding tests are added.
- `partial_signature.rs`: retarget the PTC tests to `test_ptc_attester_*` (`_within_ttl_accepted`, `_beyond_short_ttl_rejected`, `_rejected_before_cstar`, `_binds_ptc_attester_kind`, `_rejects_multiple_messages_per_packet`); fix stale "bucketed as committee" comments. `_within_ttl_accepted` asserts acceptance 2 slots late inside the short TTL; `_beyond_short_ttl_rejected` pins that 20 slots late (accepted under the old committee TTL) is rejected.
- `consensus_message.rs`: rename `test_duty_limit_ptc_committee -> test_duty_limit_ptc_attester`, switch to `DutyExecutor::Validator`, and assert `Some(2)` (and that the cap does not scale with `validator_indices` length).

## Notes

Implemented by #1080. #1074 (the PayloadAttestationVote removal) not required as a dep (disjoint symbols). Builds on `Fork::CStar` (merged). The sign path and client spawn build on this PR's `Role::PTCAttester` + `PartialSignatureKind::PTCAttester`.

shane-moore:

SIP-94 §3 was rewritten to a validator-scoped, leaderless PTC design (modeled on
the ProposerPreferences flow in §5): a payload-timeliness attestation is a
standalone single-validator signature with no QBFT consensus round. The role
that merged earlier (#1033) modeled PTC as a committee-scoped, QBFT-bearing role,
which no longer matches the spec.

This PR retargets that role to a validator-scoped `PTCAttester` and gives PTC its
own partial-signature kind instead of overloading `PostConsensus`.

- Closes #1075 (issue 2 of the no-QBFT PTC milestone).
- Independent of #1074 / #1076 (disjoint symbols). Builds only on `Fork::CStar`.
  The later PTC sign path and client spawn build on this PR's new symbols.

## Change Overview

PTC moves out of the committee role-class into the validator-scoped, non-QBFT
class at every dispatch site: validator-scoped duty executor (not a committee
id), no max QBFT round, the pre-consensus message bucket, and a per-validator
one-signature-per-packet bound. It also gets a dedicated `PartialSignatureKind`,
so a PTC attestation is no longer indistinguishable on the wire from a
post-consensus signature.

Suggested reading order:
1. `ssv_types` (`msgid`, `partial_sig`): the role rename + reclassification and
   the new partial-signature kind. A small `is_qbft_role` predicate is
   introduced (positively named, mirroring `is_committee_role`) and negated
   at the consensus-message guard.
2. `message_validator`: the validation-path consequences (kind/role binding,
   packet bound, duty limit, and rejection of consensus messages for non-QBFT
   roles). Two committee-only helpers are deleted, subsumed by the shared
   per-validator bound.
3. `qbft_manager`: the PTC routing stub is removed; PTC now rejects QBFT
   messages as an inconsistent message id in both executor branches.

What intentionally did not change:
- Wire identifiers: the role byte stays `7`; the dedicated kind takes value `7`.
- The `Fork::CStar` activation gate.

- Contained blast radius: every `Role` / `PartialSignatureKind` match in the
  touched crates is exhaustive (no `_` wildcard), so the compiler flagged each
  site that needed updating. `cargo check --workspace` is the gate.
- Behavior change: a PTC consensus (QBFT) message is now rejected rather than
  routed (matching the leaderless design), and PTC partial-sig packets are
  capped at one signature per validator per packet (the old committee occurrence
  cap is subsumed by this bound). The rejection is a `Reject` gossip verdict
  (peer penalty on synced nodes), which is intentional: post-`CStar` a PTC
  consensus message is a protocol violation under the leaderless design.
  Previously such a message was accepted and propagated, then dropped
  internally by the `qbft_manager` stub.
- Behavior change (bucketing flip): because `PTCAttester` leaves the committee
  role-class, two checks previously skipped for `PTCCommittee` now apply to PTC
  packets: the validator-index-in-committee check (`ValidatorIndexMismatch`) and
  the per-signer slot-advancement check (`SlotAlreadyAdvanced`). Both follow from
  validator-scoping (one validator per packet, monotonic per-validator slot).
- Behavior change (TTL tightened): `PTCAttester` moves from the epoch-long
  lateness/TTL bucket (`slots_per_epoch + 2` = 34 slots, inherited from the
  committee-scoped design) to the slot-bound Proposer/SyncCommittee bucket
  (3 slots). Per the Gloas spec, a payload attestation is gossip-valid only
  for its own slot and block-includable only at slot + 1, so older partials
  are useless and accepting them only widens the replay/relay surface.
- Wire compatibility: the role byte is unchanged, but a PTC packet's
  partial-signature kind flips from `PostConsensus` (`0`) to `PTCAttester`
  (`7`), and the byte-7 `MessageId` layout changes from `CommitteeId`
  (`[24..56]`) to validator pubkey (`[8..]`). No deployed impact: no producer
  emits PTC partial signatures yet and `CStar` has never been fork-active, so
  only an epbs devnet mixing pre- and post-retarget nodes would need a
  lockstep upgrade.

## Validation

- `cargo test -p ssv_types -p message_validator -p qbft_manager`: 98 / 63 / 19
  pass, 0 failures (plus doc-tests). Existing PTC tests were retargeted to the
  validator-scoped role; coverage asserts the validator duty executor, the
  non-QBFT classification (not committee-scoped, no max round), the dedicated
  kind's SSZ round-trip, the kind-to-role binding, the per-validator packet
  bound, the consensus-message rejection for `PTCAttester`, the pre-`CStar`
  fork reject, the flat duty limit, and the slot-bound TTL (accepted 2 slots
  late, rejected 20 slots late).
- `cargo +nightly fmt --all -- --check`: clean.
- `git grep PTCCommittee anchor/`: empty.

## Rollback

Revert the branch's four commits. No config, data, or migration impact; wire encodings
are unchanged, so there is no cross-version compatibility concern.

## Blockers / Dependencies

None for merge. Follow-ups that build on this PR's new symbols: the PTC sign path
and the client spawn wiring.

### DIFF
--- a/anchor/common/ssv_types/src/msgid.rs
+++ b/anchor/common/ssv_types/src/msgid.rs
@@ -79,11 +79,11 @@ impl Role {
         }
     }
 
-    /// Returns true if this role is validator-scoped and does not run a QBFT
-    /// consensus round (ValidatorRegistration, VoluntaryExit, PTCAttester),
-    /// i.e. it has no max QBFT round.
-    pub fn is_non_qbft_role(self) -> bool {
-        self.max_round().is_none()
+    /// Returns true if this role runs a QBFT consensus round, i.e. it has a
+    /// max QBFT round. The validator-scoped roles that do not
+    /// (ValidatorRegistration, VoluntaryExit, PTCAttester) return false.
+    pub fn is_qbft_role(self) -> bool {
+        self.max_round().is_some()
     }
 }
 
@@ -348,6 +348,35 @@ mod tests {
         // giving it a max round would compile silently; this pins the values.
         assert!(!Role::PTCAttester.is_committee_role());
         assert_eq!(Role::PTCAttester.max_round(), None);
-        assert!(Role::PTCAttester.is_non_qbft_role());
+        assert!(!Role::PTCAttester.is_qbft_role());
+    }
+
+    #[test]
+    fn role_qbft_classification_is_pinned() {
+        // Adding a new Role forces a decision in max_round()'s exhaustive
+        // match, but nothing checks the decision is right: a role landing in
+        // the wrong arm silently flips its consensus-message and routing
+        // behavior. Pin every existing role on both sides of the partition.
+        for role in [
+            Role::Committee,
+            Role::Aggregator,
+            Role::AggregatorCommittee,
+            Role::Proposer,
+            Role::SyncCommittee,
+        ] {
+            assert!(role.is_qbft_role(), "{role:?} runs QBFT");
+            assert!(role.max_round().is_some(), "{role:?} must have a max round");
+        }
+        for role in [
+            Role::ValidatorRegistration,
+            Role::VoluntaryExit,
+            Role::PTCAttester,
+        ] {
+            assert!(!role.is_qbft_role(), "{role:?} must not run QBFT");
+            assert!(
+                role.max_round().is_none(),
+                "{role:?} must not have a max round"
+            );
+        }
     }
 }
--- a/anchor/common/ssv_types/src/msgid.rs
+++ b/anchor/common/ssv_types/src/msgid.rs
@@ -21,7 +21,7 @@ pub enum Role {
     ValidatorRegistration,
     VoluntaryExit,
     AggregatorCommittee,
-    PTCCommittee,
+    PTCAttester,
 }
 
 impl From<Role> for [u8; 4] {
@@ -34,7 +34,7 @@ impl From<Role> for [u8; 4] {
             Role::ValidatorRegistration => [4, 0, 0, 0],
             Role::VoluntaryExit => [5, 0, 0, 0],
             Role::AggregatorCommittee => [6, 0, 0, 0],
-            Role::PTCCommittee => [7, 0, 0, 0],
+            Role::PTCAttester => [7, 0, 0, 0],
         }
     }
 }
@@ -51,15 +51,14 @@ impl TryFrom<&[u8]> for Role {
             [4, 0, 0, 0] => Ok(Role::ValidatorRegistration),
             [5, 0, 0, 0] => Ok(Role::VoluntaryExit),
             [6, 0, 0, 0] => Ok(Role::AggregatorCommittee),
-            [7, 0, 0, 0] => Ok(Role::PTCCommittee),
+            [7, 0, 0, 0] => Ok(Role::PTCAttester),
             _ => Err(DecodeError::NoMatchingVariant),
         }
     }
 }
 
 impl Role {
-    /// Returns true if this role is a committee-based role (Committee, AggregatorCommittee, or
-    /// PTCCommittee).
+    /// Returns true if this role is a committee-based role (Committee or AggregatorCommittee).
     ///
     /// Committee roles handle multiple validators in batched operations and have relaxed
     /// validation rules compared to per-validator roles:
@@ -67,10 +66,7 @@ impl Role {
     /// - Skip slot advancement checks (allow processing "older" slots within 34-slot window)
     /// - Have different message count limits and validator index occurrence limits
     pub fn is_committee_role(self) -> bool {
-        matches!(
-            self,
-            Role::Committee | Role::AggregatorCommittee | Role::PTCCommittee
-        )
+        matches!(self, Role::Committee | Role::AggregatorCommittee)
     }
 
     pub fn max_round(self) -> Option<u64> {
@@ -78,15 +74,16 @@ impl Role {
         match self {
             Role::Committee | Role::Aggregator | Role::AggregatorCommittee => Some(12),
             Role::Proposer | Role::SyncCommittee => Some(6),
-            // PTC is expected near 75% of a 12s slot, leaving about 3s.
-            // With 2s quick round timeouts, only round 1 is likely to finish
-            // before slot end. Allow up to 4 rounds as a small local grace
-            // window for delayed starts or message loss; this is not a
-            // consensus-spec requirement.
-            Role::PTCCommittee => Some(4),
             // These roles don't use QBFT consensus
-            Role::ValidatorRegistration | Role::VoluntaryExit => None,
+            Role::ValidatorRegistration | Role::VoluntaryExit | Role::PTCAttester => None,
         }
+    }
+
+    /// Returns true if this role is validator-scoped and does not run a QBFT
+    /// consensus round (ValidatorRegistration, VoluntaryExit, PTCAttester),
+    /// i.e. it has no max QBFT round.
+    pub fn is_non_qbft_role(self) -> bool {
+        self.max_round().is_none()
     }
 }
 
@@ -156,14 +153,15 @@ impl MessageId {
     pub fn duty_executor(&self) -> Option<DutyExecutor> {
         // which kind of executor we need to get depends on the role
         match self.role()? {
-            Role::Committee | Role::AggregatorCommittee | Role::PTCCommittee => {
+            Role::Committee | Role::AggregatorCommittee => {
                 self.0[24..].try_into().ok().map(DutyExecutor::Committee)
             }
             Role::Aggregator
             | Role::Proposer
             | Role::SyncCommittee
             | Role::ValidatorRegistration
-            | Role::VoluntaryExit => PublicKeyBytes::deserialize(&self.0[8..])
+            | Role::VoluntaryExit
+            | Role::PTCAttester => PublicKeyBytes::deserialize(&self.0[8..])
                 .ok()
                 .map(DutyExecutor::Validator),
         }
@@ -322,38 +320,23 @@ mod tests {
     }
 
     #[test]
-    fn ptc_uses_committee_duty_executor() {
-        // PTCCommittee must use Committee-style duty executor (CommitteeId),
-        // not Validator-style (PublicKeyBytes)
+    fn ptc_attester_uses_validator_duty_executor() {
+        // PTCAttester is validator-scoped: it must use Validator-style duty
+        // executor (PublicKeyBytes), not Committee-style (CommitteeId).
         let domain = DomainType([0, 0, 0, 0]);
-        let committee_id = CommitteeId::from(vec![OperatorId(100), OperatorId(200)]);
-        let duty_executor = DutyExecutor::Committee(committee_id);
+        let public_key = PublicKeyBytes::empty();
+        let duty_executor = DutyExecutor::Validator(public_key);
 
-        let msg_id = MessageId::new(&domain, Role::PTCCommittee, &duty_executor);
+        let msg_id = MessageId::new(&domain, Role::PTCAttester, &duty_executor);
 
-        assert_eq!(msg_id.role(), Some(Role::PTCCommittee));
+        assert_eq!(msg_id.role(), Some(Role::PTCAttester));
 
         match msg_id.duty_executor() {
-            Some(DutyExecutor::Committee(id)) => assert_eq!(id, committee_id),
-            Some(DutyExecutor::Validator(_)) => {
-                panic!("PTCCommittee should use Committee duty executor, not Validator")
+            Some(DutyExecutor::Validator(pk)) => assert_eq!(pk, public_key),
+            Some(DutyExecutor::Committee(_)) => {
+                panic!("PTCAttester should use Validator duty executor, not Committee")
             }
             None => panic!("Failed to extract duty executor"),
         }
-    }
-
-    #[test]
-    fn ptc_max_round_is_four() {
-        // PTC duty starts at 75% slot; Some(4) is the deliberate cap distinct
-        // from Committee's Some(12). Regressions copying the Committee value
-        // would compile silently — this guards against that.
-        assert_eq!(Role::PTCCommittee.max_round(), Some(4));
-    }
-
-    #[test]
-    fn ptc_is_committee_role() {
-        // is_committee_role drives the validator-index-mismatch skip and
-        // slot-advancement skip in message_validator. PTC must qualify.
-        assert!(Role::PTCCommittee.is_committee_role());
     }
 }
--- a/anchor/common/ssv_types/src/msgid.rs
+++ b/anchor/common/ssv_types/src/msgid.rs
@@ -339,4 +339,15 @@ mod tests {
             None => panic!("Failed to extract duty executor"),
         }
     }
+
+    #[test]
+    fn ptc_attester_is_validator_scoped_non_qbft() {
+        // Every behavior flip in the PTCAttester retarget rides on these three
+        // classifications (validation bucketing, consensus-message rejection,
+        // qbft_manager routing). Re-adding PTCAttester to the committee arm or
+        // giving it a max round would compile silently; this pins the values.
+        assert!(!Role::PTCAttester.is_committee_role());
+        assert_eq!(Role::PTCAttester.max_round(), None);
+        assert!(Role::PTCAttester.is_non_qbft_role());
+    }
 }
--- a/anchor/common/ssv_types/src/partial_sig.rs
+++ b/anchor/common/ssv_types/src/partial_sig.rs
@@ -37,6 +37,9 @@ pub enum PartialSignatureKind {
     // AggregatorCommitteePartialSig is a partial signature for combined aggregator and sync
     // committee selection proofs (committee-based batching)
     AggregatorCommitteePartialSig = 6,
+    // PTCAttester is a standalone single-validator partial signature over PayloadAttestationData
+    // (validator-scoped, leaderless; not a QBFT consensus value)
+    PTCAttester = 7,
 }
 
 impl TryFrom<u64> for PartialSignatureKind {
@@ -51,6 +54,7 @@ impl TryFrom<u64> for PartialSignatureKind {
             4 => Ok(PartialSignatureKind::ValidatorRegistration),
             5 => Ok(PartialSignatureKind::VoluntaryExit),
             6 => Ok(PartialSignatureKind::AggregatorCommitteePartialSig),
+            7 => Ok(PartialSignatureKind::PTCAttester),
             _ => Err(()),
         }
     }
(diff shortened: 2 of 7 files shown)


shane-moore:
@claude, please take a look

shane-moore:
@claude, thanks for the careful pass. Dispositions:

**1. Two newly-enforced checks.** Agreed, surfaced in the description: added a "bucketing flip" bullet noting that `PTCAttester` leaving the committee class now applies the validator-index-in-committee check (`ValidatorIndexMismatch`) and the `SlotAlreadyAdvanced` check to PTC packets.

**2. Orphaned QBFT-PTC types.** Already handled by the open #1076 (issue #1074), which removes exactly these and touches only `consensus.rs`. They were orphaned before this PR too (no callers on `epbs`); #1080 deliberately leaves `consensus.rs` untouched (disjoint symbols), so no change here.

**3. `is_non_qbft_role` via explicit `matches!`.** Keeping `max_round().is_none()`. `matches!(self, A | B | C)` expands with an implicit `_ => false`, so it would not force a compile-time decision when a new `Role` is added (the variant silently classifies as QBFT). `max_round()` is the one genuinely exhaustive, no-wildcard match that does force that decision, so deriving the predicate from it keeps a single source of truth that can't drift from it. If a role ever needs `max_round = None` while still running consensus, we'd split the two concepts then.

**4. Log wording at `qbft_manager`.** That `error!(...)` is an unchanged context line: it already covered `ValidatorRegistration | VoluntaryExit` before this PR, which only added `PTCAttester` to the same arm. Leaving the wording as-is to stay within this PR's diff; it can be refined as a separate cleanup.

**5.** Confirmed, thanks for tracing it.

jnhsigmap:
> @claude, thanks for the careful pass. Dispositions:
> 
> **1. Two newly-enforced checks.** Agreed, surfaced in the description: added a "bucketing flip" bullet noting that `PTCAttester` leaving the committee class now applies the validator-index-in-committee check (`ValidatorIndexMismatch`) and the `SlotAlreadyAdvanced` check to PTC packets.
> 
> **2. Orphaned QBFT-PTC types.** Already handled by the open #1076 (issue #1074), which removes exactly these and touches only `consensus.rs`. They were orphaned before this PR too (no callers on `epbs`); #1080 deliberately leaves `consensus.rs` untouched (disjoint symbols), so no change here.
> 
> **3. `is_non_qbft_role` via explicit `matches!`.** Keeping `max_round().is_none()`. `matches!(self, A | B | C)` expands with an implicit `_ => false`, so it would not force a compile-time decision when a new `Role` is added (the variant silently classifies as QBFT). `max_round()` is the one genuinely exhaustive, no-wildcard match that does force that decision, so deriving the predicate from it keeps a single source of truth that can't drift from it. If a role ever needs `max_round = None` while still running consensus, we'd split the two concepts then.
> 
> **4. Log wording at `qbft_manager`.** That `error!(...)` is an unchanged context line: it already covered `ValidatorRegistration | VoluntaryExit` before this PR, which only added `PTCAttester` to the same arm. Leaving the wording as-is to stay within this PR's diff; it can be refined as a separate cleanup.
> 
> **5.** Confirmed, thanks for tracing it.

For **# 3:** This would trip me up without the help of a coding agent in the distant future if making mods to this domain and still learning my way around here. Valid to use the max round if its tied semantically, but would suggest a test to ensure all roles follow the rules. Something like
```
  for role in [Role::Committee, Role::Aggregator, Role::AggregatorCommittee,
               Role::Proposer, Role::SyncCommittee] {
      assert!(!role.is_non_qbft_role(), "{role:?} runs QBFT");
      assert!(role.max_round().is_some(), "{role:?} must have a max round");
  }
```
technically non-blocking but would probably help maintainers later on. I see `ptc_attester_is_validator_scoped_non_qbft` checks `PTCAttester` adheres, but no test coverage for the others ^.

Less important - could be followed up later or ignored but I see [`Some(role) if role.is_non_qbft_role()`](https://github.com/sigp/anchor/pull/1080/changes#diff-bc158384f68c2a3a456524d9cb730fcda7a276b8db5f084311f9cc4310a7d275R124) seems to invalidate [`FailedToGetMaxRound`](https://github.com/sigp/anchor/pull/1080/changes#diff-bc158384f68c2a3a456524d9cb730fcda7a276b8db5f084311f9cc4310a7d275L137) if it's now tied to `max_round`. That branch won't be reached if `is_non_qbft_role()` returns `None` right?

otherwise lgtm :100:

jnhsigmap (approved):

jnhsigmap:
ha ha, cancelled that run ^.
don't need claude to answer that one :laughing:

petarjuki7:
Great, looks good! 
I have one idea, not a blocker.
Could we maybe change `is_non_qbft_role()` to `is_qbft_role()` and then use `!is_qbft_role()` where needed?
I think it reads a bit easier, so if sometime later in the code we have to use something like `!is_non_qbft_role()` it's easier to reason about.
It also mirrors how we use `is_committee_role`; it's named positively and used negated at every call site: `!role.is_committee_role()` appears multiple times.

shane-moore:
@jnhsigmap Both addressed in [b50aaf63](https://github.com/sigp/anchor/pull/1080/commits/b50aaf638e2d87e88e60a43dc763bf28e154f30c).

**Classification test:** added `role_qbft_classification_is_pinned`, extended from your sketch to pin both sides of the partition (the five QBFT roles plus `ValidatorRegistration` / `VoluntaryExit` / `PTCAttester` as non-QBFT), so a new role landing in the wrong `max_round` arm fails loudly either way.

**`FailedToGetMaxRound`:** your read is right, the branch is unreachable. One nuance: it was already dead before this PR, since the old `ValidatorRegistration | VoluntaryExit` guard exactly covered the `None`-returning roles (`PTCCommittee` had `Some(4)`). This PR just made the deadness definitional, so I cleaned it up here: the fetch is now a let-else whose fallback returns `UnexpectedConsensusMessage` (semantically what that branch would mean if it ever fired, and no panic on network input), and the variant is deleted. Nothing else referenced it.

shane-moore:
@petarjuki7 Good call, done in [b50aaf63](https://github.com/sigp/anchor/pull/1080/commits/b50aaf638e2d87e88e60a43dc763bf28e154f30c). `is_qbft_role()` returning `max_round().is_some()`, with the guard now `Some(role) if !role.is_qbft_role()`. Agreed on the `is_committee_role` symmetry, and renaming now is cheap since the follow-up PRs haven't started consuming the symbol yet.

Issue #1078, opened 2026-06-05
feat(client): spawn LH PayloadAttestationService with AnchorValidatorStore backend


shane-moore:
## Goal

Spawn Lighthouse's `PayloadAttestationService` with `AnchorValidatorStore` as the backend in `anchor/client/src/lib.rs` (the service is absent from `client/` today). This is the final wiring step for the PTC duty. Reworks the not-yet-implemented #1038.

## Context

The LH service drives the per-slot PTC loop: at the 75% cutoff it fetches `PayloadAttestationData`, abstains on no-block, and calls `sign_payload_attestation` per validator (implemented in #1077). PTC duties are populated by LH's `poll_beacon_ptc_attesters` under the existing `duties_service::start_update_service` (Gloas-gated), so no extra duty wiring is needed. Dep: #1077. Part of milestone #6.

**Construction shape (verified against the pin):** there is **no** `PayloadAttestationServiceBuilder`. Construct with `PayloadAttestationService::new(...)` (6 positional args, the 6th being `chain_spec`, which `SyncCommitteeService::new` does NOT take), and start with `start_update_service()` which **takes no args** (consumes `self`), NOT `start_update_service(&spec)`. Mirror `SyncCommitteeService::new` (`client/src/lib.rs:682-687`) for the *constructor* only. **Gate the start** on `spec.is_gloas_scheduled()`, matching LH's own VC (`validator_client/src/lib.rs:657`), which starts `payload_attestation_service` (and `proposer_preferences_service`) only when Gloas is scheduled. If #1064's ProposerPreferences spawn has already landed its gated block, join that block rather than adding a second `if`. Do NOT mirror the *ungated* `sync_committee_service` start: the per-slot `gloas_enabled()` self-gate makes an ungated start harmless, but it spawns a perpetual idle task pre-Gloas, which diverges from LH.

## Suggested approach

`anchor/client/src/lib.rs` (lead with the symbol; line hints drift):

| Symbol (line hint) | change |
|---|---|
| `validator_services::{...}` use block (~54-63) | add `payload_attestation_service::PayloadAttestationService` |
| after the `sync_committee_service` construction (~687) | construct (below) |
| after `metadata_service.start_update_service()` (~730) | start (below) |

```rust
let payload_attestation_service = PayloadAttestationService::new(
    duties_service.clone(),
    validator_store.clone(),
    slot_clock.clone(),
    beacon_nodes.clone(),
    executor.clone(),
    spec.clone(),
);

[...]

Issue #1077, opened 2026-06-05
feat(validator_store): implement sign_payload_attestation (SingleValidator partial-sig)


shane-moore:
## Goal

Implement `sign_payload_attestation`, replacing the `Unsupported` stub at `anchor/validator_store/src/lib.rs:3102`: a `SingleValidator` partial-signature collection over the `PayloadAttestationData` Lighthouse passes in, signed under `Domain::PTCAttester`. Reworks the not-yet-implemented #1037.

## Context

LH's `PayloadAttestationService` fetches `data` once at the 75% cutoff and abstains (early-returns) on no-block, then calls `sign_payload_attestation(pubkey, data)` per validator (`payload_attestation_service.rs:116-189`). **This method does NOT fetch from the beacon node and does NOT touch the slashing-protection DB**; it only signs the `data` it is handed. Mirror `collect_voluntary_exit_partial_signatures` (`validator_store/src/lib.rs:~824-855`), the closest merged precedent (SingleValidator, single signature, no fetch, no slashing DB). `collect_signature` is the 8-arg fn at `validator_store/src/lib.rs:447-457`. Dep: #1075 (implemented by #1080). Part of milestone #6.

## Suggested approach

Replace the stub body at `sign_payload_attestation` (~3102). Lead with the symbol; line hints drift.

```rust
async fn sign_payload_attestation(
    &self,
    validator_pubkey: PublicKeyBytes,
    data: PayloadAttestationData,
) -> Result<PayloadAttestationMessage, Error> {
    // LH fetched `data` at 75% and abstained on no-block already. No BN fetch, no slashing DB.
    let (validator, cluster) = self.get_validator_and_cluster(validator_pubkey)?;
    let validator_index = validator.index.ok_or(SpecificError::MissingIndex)?;

    let epoch = data.slot.epoch(E::slots_per_epoch());
    let domain = self.get_domain(epoch, Domain::PTCAttester);
    let signing_root = data.signing_root(domain);

    // Do NOT use `?` here: the no-threshold/divergence path returns an Err we must
    // classify for the metric before returning it (see telemetry note).
    let signature = match self
        .collect_signature(
            PartialSignatureKind::PTCAttester,
            Role::PTCAttester,
            CollectionMode::SingleValidator,
            &validator,
            &cluster,
            signing_root,
            data.slot,
        )
        .await
    {
        Ok(sig) => sig,
        Err(e) => {
[...]

Issue #1079, opened 2026-06-05
feat(signature_collector): distinguishable no-threshold error for the PTC divergence metric


shane-moore:
## Goal

Make `signature_collector` resolve a no-threshold collection (a still-pending collector evicted at cleanup) to a **distinguishable** error instead of the ambiguous `QueueClosedError`, so the PTC reconstruction-failure metric added in #1077 can isolate a clean `observation_divergence` rate instead of the coarse `{reason="no_signature"}` upper bound. **Observability improvement, not a protocol requirement.**

## Context / motivation

Under the no-QBFT PTC design (#1077, milestone #6), a minority operator whose beacon-node observation never reaches threshold records a non-slashable, penalty-free silent miss. How often that happens is useful telemetry to watch (the question was raised in the SIP-94 thread), but it is **not a normative SIP-94 §3 requirement**, and nothing functional depends on it: #1077 ships a coarse `{reason="no_signature"}` counter that works today.

The blocker to a clean rate is in the collector: `CollectionError::CollectionTimeout` is dead code (never constructed; `signature_collector/src/lib.rs:514`), and `sign_and_collect` ends at `Ok(result_rx.await?)` (`:294`) with no timeout. So a minority/no-threshold operator blocks until the `cleaner` (`:411-429`, retains `for_slot >= slot - SIGNATURE_COLLECTOR_RETAIN_SLOTS`) evicts its collector, dropping the notifier -> `RecvError` -> `QueueClosedError` (`:529-532`) -- indistinguishable from a genuine channel close.

**Blocked by:** #1077 (the reconstruction-failure metric this would refine).

## Suggested approach

Make eviction of a still-pending collector resolve its registered waiters to a distinguishable no-threshold error instead of `QueueClosedError`. Two shapes:

- **(a) clean protocol fix:** change the notifier oneshot to carry `Result<Arc<Signature>, CollectionError>` and have the cleanup / State-drop path send `Err(CollectionTimeout)` (or a new `NoThreshold` variant) before dropping. Touches every `collect_signature` caller's success unwrap (`Ok((*collector.await…)?).clone()` -> handle the inner `Result`).
- **(b) lower blast radius:** at eviction, detect that the collector still had registered notifiers and signal those waiters distinctly without changing the success payload type (a dedicated close-reason).

Either way, a no-threshold expiry must reach `collect_signature`'s caller as a variant distinct from a genuine queue close. Pick at implementation time.

[...]

PR #1082, opened 2026-06-11
feat(validator_store): implement sign_payload_attestation

shane-moore:

Under Gloas/ePBS, PTC validators must sign payload attestations (SSV-side section 3 of SIP-94). Lighthouse's `PayloadAttestationService` calls `sign_payload_attestation(pubkey, data)` per validator, but Anchor's implementation is an `Unsupported` stub, so the duty cannot run. Both dependencies are merged to `epbs`: #1076 (PTC value object removal) and #1080 (`Role::PTCAttester` + `PartialSignatureKind::PTCAttester`).

Addresses #1077.

## Change Overview

The method runs a SingleValidator partial-signature collection over the `PayloadAttestationData` LH passes in, signed under `Domain::PTCAttester`, mirroring the voluntary-exit flow (the one merged SingleValidator single-signature precedent). LH fetches the data once at the 75% slot cutoff and abstains on no-block, so the method only signs what it is handed.

Collection failures are classified for telemetry before the error is surfaced, because LH crit-logs whatever we return and the trait has no abstain path: a no-threshold miss (surfacing as `QueueClosedError` via collector eviction) warns and increments `anchor_ptc_reconstruction_failures_total{reason="no_signature"}`; local faults count as `reason="infra"`. Errors from the method's own validator/cluster/index resolution propagate untouched; non-collection errors raised inside `collect_signature` (share lookup, key decryption, threshold arithmetic) are error-logged for context but emit no metric.

Suggested reading order:
- `lib.rs`: `sign_payload_attestation` (the duty) and `report_ptc_collection_failure` (the effects)
- `instrumentation.rs`: pure failure classifier, next to the existing `failure_reason` mapper
- `metrics.rs`: the new counter
- `testing/`: harness extensions (failure injection, slashing flag, richer call capture) and the 5 new tests

Intentionally unchanged:
- No `PayloadAttestationService` spawning (#1078)
- No `signature_collector` changes; at current collector granularity a no-threshold miss and a genuine channel close are both `QueueClosedError`, so `reason="no_signature"` is an upper bound on true observation divergence. Clean measurement needs a collector change (separate follow-up issue).
- No `is_synced` gating: LH only schedules duties when synced. Flagging for reviewer confirmation.

- Blast radius is small: nothing calls this method until #1078 spawns the service.
- On a minority split, the call blocks until the collector cleaner evicts (about 2 slots) and LH awaits validators sequentially. This is a pre-existing property of `collect_signature` shared by all duties, not introduced here.
- Recoverable no-threshold misses still surface through LH at `crit!`; unavoidable Anchor-side given the trait shape. The metric, not the log, is the measurement channel.

## Validation

`cargo test -p anchor_validator_store`: 44 passed (5 new). The new tests lock:
- message construction, and the signing root recomputed independently in the test (locks the `Domain::PTCAttester` choice, epoch derivation, and signed object), plus kind/role/slot wiring into the collector
- `MissingIndex` fails before any collection attempt
- `QueueClosedError` returns the original error and increments `{reason="no_signature"}`
- `EmptySignature` increments `{reason="infra"}` and leaves `{reason="no_signature"}` at zero delta, pinning that infra faults cannot leak into the divergence bucket
- slashing-DB tripwire: protection enabled against an empty slashing DB, so any future slashing check added to this path fails the test (payload attestations are not slashable)

`make cargo-fmt` and `make lint` clean.

## Rollback

Revert the commit; the method returns to the `Unsupported` stub. No config, data, or operational impact (no caller yet; the metric simply disappears).

## Blockers / Dependencies

None for merge. Follow-ups: #1078 (spawn the LH service) and a research issue for clean divergence measurement in `signature_collector`.

### DIFF
--- a/anchor/validator_store/src/instrumentation.rs
+++ b/anchor/validator_store/src/instrumentation.rs
@@ -1,3 +1,5 @@
+use signature_collector::CollectionError;
+
 use crate::{Error, SpecificError};
 
 /// Outcome of a block signing attempt, used for metrics labeling.
@@ -19,6 +21,43 @@ pub mod checkpoints {
     pub const PUBLISH_BLOCK: &str = "publish_block";
     pub const DUTY_COMPLETED: &str = "duty_completed";
     pub const DUTY_FAILED: &str = "duty_failed";
+}
+
+/// Telemetry classification of `collect_signature` failures during PTC duties.
+///
+/// Returned as an enum rather than a label string because the class drives two independent
+/// effects in the caller: the log level and whether a reconstruction-failure metric is
+/// incremented at all.
+pub enum PtcFailureClass {
+    /// The committee never reached the partial signature threshold. This surfaces as
+    /// `QueueClosedError` because the collector is evicted after
+    /// `SIGNATURE_COLLECTOR_RETAIN_SLOTS`, dropping the result channel while we await it, so it
+    /// cannot be distinguished from a genuine channel close.
+    NoSignature,
+    /// Local infrastructure failed while collecting or reconstructing the signature.
+    Infra,
+    /// The failure happened before signature collection started (unknown pubkey, threshold
+    /// arithmetic, key share decryption, missing index).
+    NonCollection,
+}
+
+pub fn classify_ptc_collection_failure(error: &Error) -> PtcFailureClass {
+    match error {
+        // The inner match is deliberately wildcard-free so a future `CollectionError` variant
+        // forces a conscious classification decision here at compile time.
+        Error::SpecificError(SpecificError::SignatureCollectionFailed(collection_error)) => {
+            match collection_error {
+                CollectionError::QueueClosedError | CollectionError::CollectionTimeout => {
+                    PtcFailureClass::NoSignature
+                }
+                CollectionError::QueueFullError
+                | CollectionError::OwnOperatorIdUnknown
+                | CollectionError::EmptySignature
+                | CollectionError::RecoverError(_) => PtcFailureClass::Infra,
+            }
+        }
+        _ => PtcFailureClass::NonCollection,
+    }
 }
 
 /// Common reasons for block signing failures.
--- a/anchor/validator_store/src/lib.rs
+++ b/anchor/validator_store/src/lib.rs
@@ -80,6 +80,8 @@ use validator_store::{
     ValidatorStore,
 };
 
+use crate::instrumentation::PtcFailureClass;
+
 /// Number of epochs of slashing protection history to keep.
 ///
 /// This acts as a maximum safe-guard against clock drift.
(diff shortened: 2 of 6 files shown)


shane-moore:
@claude, please take a look

shane-moore:
@claude, went through each review item:

**1 (Infra arm coverage):** Added `sign_payload_attestation_infra_failure_increments_infra_metric` in 2cad7fb1: injects `EmptySignature`, asserts `{reason="infra"}` delta == 1 and `{reason="no_signature"}` delta == 0. The zero-delta half is the valuable direction: an infra variant drifting into the no_signature bucket would silently inflate the SIP-94 divergence upper bound.

**2 (double index resolution):** Leaving as is. Resolving `validator.index` at the duty site is the crate-wide pattern (six other call sites do the same outside `collect_signature`), and the early resolution is load-bearing: `MissingIndex` must surface before any collection attempt and bypass the failure reporter, which `sign_payload_attestation_missing_index_errors` pins.

**3 (`inspect_err`):** Keeping the explicit match. `inspect_err` has zero uses in the workspace, so the combinator would be a novel idiom for an equivalent form.

**4 (metric delta race):** Resolved alongside item 1. The new zero-delta read crosses a label the existing test writes, so the sole-writer argument no longer held; both metric tests now serialize via a shared lock, which also closes the symmetric race on the existing delta assertion.

**5 (description vs behavior):** Description updated. Behavior is per #1077's spec (non-collection errors from `collect_signature` are error-logged without the metric); the overview now distinguishes errors that bypass the reporter from those logged without a metric.

jnhsigmap (approved):
The observation divergence property looks like it's going to occur more frequently than what could be considered 'exceptional' as it looks more like 1. an artefact of Lighthouse's architecture and 2. an artefact of how the network behaves. The scope of this PR is just to ensure metrics attempt to capture? is this something we can try to resolve in Lighthouse?

shane-moore:
agreed it won't be exceptional, but that's the expected design, not a defect:
the sip deliberately runs PTC as a pre-consensus partial-sig round with no QBFT, so each operator signs what its own BN observed and near-boundary divergence is inherent. There's no convergence phase to add because the spec chose not to have one. This PR just measures it: `{reason="no_signature"}` is an upper bound on the rate, no behavior change.

On "resolve in LH": no. It isn't an LH bug (any client observes independently), and the only LH-touchable part is the `crit!` on a recoverable miss (no `Ok(None)` abstain path), which is pure ergonomics. We push to LH pretty much only for ssv-spec compliance, not ergonomics, and even those aren't guaranteed: they're gated on how much risk the change adds to LH.

The one real follow-up is Anchor-side and measurement-only, tracked in #1079 (research): make `signature_collector` resolve a no-threshold eviction to a distinguishable error, so the metric can split true `observation_divergence` from a
genuine channel-close instead of today's coarse `{reason="no_signature"}` upper bound

Issue #254, opened 2025-04-14
Processor: Split into multiple queues


dknopik:
For initial development, we only had two queues: permitless (for tasks that are not compute heavy but routed through the processor for consistency) and urgent_consensus (for everything else).

Obviously, this is not the idea behind the processor, and now that we are near-feature complete and have a good overview where and when we invoke the Processor, we should add more queues. We should also evaluate whether this improves client performance in resource constrained cases, and set reasonable queue sizes.

shane-moore:
**A production case to inform the split, sizing, and priority decisions here.**

We hit the first real-world `urgent_consensus` saturation on mainnet (relayed via SSV Labs from a large operator). It turns out to be a useful design input for this issue, because the root cause is the sizing/priority model rather than CPU.

### The case
- Large mainnet operator: ~75 committees, which under the current subnet algorithm is **57 of 128 subnets (~45% of all mainnet SSV gossip)** validated by one node. v1.2.3.
- Symptom: bursts of `Processor queue full` / `Unable to pass message to message receiver` on `urgent_consensus`, hundreds within single seconds at slot boundaries, quiet in between.
- **Not host CPU starvation**: the node uses ~2.6 of 8 cores during the drops, and co-located single-cluster Anchor nodes stay under 25%. There are idle cores while we drop messages.

### Root cause: the inbound buffer is ~8x shallower than go-ssv's
- Anchor funnels every inbound gossip message through the single `urgent_consensus` queue (default **1000**, `processor/src/lib.rs:78`), drained by `num_cpus` workers that each run the full validate-and-route to completion while holding the permit (`processor/src/lib.rs:180-188`, `message_receiver/src/manager.rs:78`). So the 1000 queue is the *only* burst buffer, and it backs up at `num_cpus`-wide throughput.
- go-ssv runs its validator asynchronously inside go-libp2p's pubsub pipeline, which allows up to **ValidateThrottle = 8192** concurrent validations plus a 512 feed queue (`network/topics/pubsub.go:34,38`). Validation is dispatched off the workers, so a slot-boundary burst is absorbed as parked goroutines and worked off over the following moment, dropping only past ~8192.
- Both clients are CPU-bound to ~`num_cpus` for the actual signature verification, so this is **buffering before drop, not parallelism**. go-ssv tolerates ~8x larger bursts; we drop everything past 1000.

### Implications for this issue
[...]

PR #1086, opened 2026-06-12
fix(subnet_service): don't subscribe to subnets of liquidated clusters

### LINKED ISSUE #1084: Do not subscribe to subnets for liquidated clusters
## Description

Subnet subscriptions are computed from every cluster the operator is a member of, with no filter on liquidation status:

- `SubnetGenerator::committees_from_network_state` (`subnet_service/src/subscriptions.rs`) maps `get_own_clusters()` directly to subnets. `ClusterLiquidated` only flips `cluster.liquidated` (`database/src/pending_updates.rs`), and a cluster only leaves the own-clusters set when its last validator is removed, so liquidated clusters keep their subnets subscribed indefinitely. The same logic exists in all releases (v1.2.3: `subnet_service/src/lib.rs`).
- Expected message rates for gossipsub topic scoring (`get_committee_info_for_subnet` in `subnet_service/src/scoring.rs`) also count validators from liquidated clusters, network-wide, inflating expected rates even though liquidated clusters generate no traffic.

Everywhere else, Anchor already treats liquidation as "do no work" (#558, #264): duties are filtered in the validator store and metadata service, index sync skips liquidated clusters, and exit processing skips them. The gossip layer is the remaining gap.

go-ssv never subscribes for liquidated clusters: subscriptions happen per started validator (`Validator.Start()`), and validators are only initialized/started from shares filtered with `ByNotLiquidated`. One asymmetry worth knowing: on mid-run liquidation go-ssv stops the validator without unsubscribing the topic (`onShareStop` does not call `p2pNetwork.Unsubscribe`), so it sheds the stale subscription only on restart. The fix proposed here unsubscribes live, which is slightly stronger than go-ssv but consistent with Anchor's reactive subscription architecture.

## Impact

Every inbound message on a subscribed subnet is fully validated on the `urgent_consensus` processor queue before "not interested" filtering, so unnecessary subnet subscriptions translate directly into wasted validation CPU and queue pressure (#254).

This is not hypothetical: while investigating a mainnet operator's sustained `Processor queue full` / `Unable to pass message to message receiver` errors, we found roughly a third of their registered validators sit in liquidated clusters. Their committees map to 57 of 128 subnets today; filtering liquidated clusters would drop that to 38, i.e. about a third less mainnet gossip to validate. Operators in many public clusters accumulate liquidated clusters over time and never shed the subscriptions.

## Proposed solution

Filter liquidated clusters at the two read sites in `subnet_service`:

1. `SubnetGenerator::committees_from_network_state`: skip clusters with `cluster.liquidated` when building the committee set.
2. `get_committee_info_for_subnet` (scoring): skip liquidated clusters when computing expected message rates, so topic score parameters reflect actual traffic.

No event plumbing is needed: `ClusterLiquidated`/`ClusterReactivated` already publish a `NetworkState` change through the db watch channel, which triggers `handle_subnet_changes`; the existing subscribe/unsubscribe diffing then handles gossipsub topics, peer-manager join/leave, and the ENR subnet bitfield. This unsubscribe path is already exercised in production when a cluster's last validator is removed. Set semantics keep a subnet if a non-liquidated cluster shares it.

The filter belongs at the read sites, not inside `get_own_clusters()`: other consumers of cluster state (e.g. exit-event verification) intentionally still need to see liquidated clusters.

Testing: the `subnet_service` test harness already exercises the full `run()` loop against an in-memory database and captures emitted `TopicEvent`s; add cases asserting an `Unsubscribe` after liquidating a cluster (when no other own cluster shares the subnet), a `Subscribe` after reactivation, and subnet retention when a non-liquidated cluster shares the subnet.

## Edge cases considered

- **Zero expected rates**: filtering can drive a topic's expected message rate to 0; the score-param generation already guards this (`expected_messages_per_decay_interval > 0.0` checks in `topic_score_config.rs`, mesh-delivery weight disabled), so no degenerate params result.
- **Message validation and publishing are unaffected**: both `message_validator` and `message_sender` use the pure committee-to-subnet computation, never the subscription state.
- **Reactivation race**: resubscription is immediate on `ClusterReactivated`, but mesh grafting takes a few heartbeats, so the first duty after reactivation may ride on lazy gossip rather than eager mesh push. go-ssv has the same exposure, and the event already lags by the EL follow distance.
- **In-flight QBFT instances** for a committee liquidated mid-slot stop receiving messages; their output is already unusable since duty gating and signing refusal apply from the same event.
- **Historical replay**: an own cluster liquidated and later reactivated in history can produce one transient unsubscribe/resubscribe during initial sync (state updates are published per block batch and watch notifications coalesce, so same-batch transitions are no-ops). Bounded by own liquidation history.
- **Mesh density**: shedding these subnets stops relaying for other committees sharing them; the go-ssv fleet already sheds them on restart, so this does not push the network below the dominant client's baseline.
- Filtering on validator activity (exited/pending validators), as go-ssv's started-validator gating effectively also does, is deliberately out of scope: it is metadata-driven rather than contract-event-driven and has riskier edge cases (e.g. pending-activation validators must keep their subnet warm).

shane-moore:

- Subnet subscriptions are computed from every own cluster with no liquidation filter, and a cluster only leaves the own-clusters set when its last validator is removed, so liquidated clusters keep their subnets subscribed indefinitely.
- Every inbound message on a subscribed subnet is fully validated on the `urgent_consensus` queue before "not interested" filtering, so stale subscriptions translate directly into wasted validation CPU and queue pressure (#254).
- Evidence: a mainnet operator with sustained `Processor queue full` errors has roughly a third of registered validators in liquidated clusters; filtering them drops their subscriptions from 57 to 38 of 128 subnets, about a third less gossip to validate.
- Anchor already treats liquidation as "do no work" everywhere else (#558, #264); the gossip layer was the remaining gap. go-ssv never subscribes for liquidated clusters (validators are only started from `ByNotLiquidated`-filtered shares).
- Closes #1084.

## Change Overview (Required)

- Filter liquidated clusters at the two read sites in `subnet_service`: subnet generation (drives subscribe/unsubscribe) and expected message rates for gossipsub topic scoring.
- No event plumbing needed: `ClusterLiquidated`/`ClusterReactivated` already publish through the database watch channel, and the existing diffing handles gossipsub topics, peer-manager join/leave, and the ENR subnet bitfield. The unsubscribe path is already exercised in production when a cluster's last validator is removed. Set semantics keep a subnet if a non-liquidated cluster shares it.
- Reading order: the two one-line filters first, then the tests, which extend the existing harness to drive committee-based subscriptions from a real database instead of `subscribe_all_subnets`.
- Intentionally unchanged: `get_own_clusters()` semantics (other consumers must still see liquidated clusters), message validation/sending (they compute subnets purely from committees), and no filtering on validator beacon status (deliberately out of scope, see #1084).

- Shedding subnets stops relaying for other committees sharing them; the go-ssv fleet already sheds the same subscriptions on restart, so this stays at the dominant client's baseline.
- Reactivation resubscribes immediately, but mesh grafting takes a few heartbeats, so the first duty after reactivation may ride lazy gossip. go-ssv has the same exposure, and the event already lags by the EL follow distance.
- Scoring now diverges from go-ssv, which counts liquidated committees in expected rates. The rate only shapes the positive first-delivery score curve; the mesh-delivery penalty has zero weight in both clients, so there is no path to penalizing honest peers. Zero expected rates are already guarded in score-param generation.

## Validation (Required)

- Four new tests exercise the full `run()` loop against a real database: unsubscribe on liquidation, resubscribe on reactivation, subnet retention while a non-liquidated cluster shares it (with an ordering probe so the negative assertion cannot false-pass), and exclusion of liquidated clusters from the expected message rate.
- All four fail with the filters removed and pass with them in place.
- fmt, clippy (including `--all-targets` for the crate), and dependency sort are clean; release tests pass for the affected crates.

- Plain revert; no config, schema, or operational impact. Subscriptions recompute from database state on the next change or restart.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7483,6 +7483,7 @@ dependencies = [
  "ethereum_serde_utils 0.7.0",
  "fork",
  "libp2p",
+ "rusqlite",
  "serde",
  "sha2",
  "slot_clock",
--- a/anchor/subnet_service/Cargo.toml
+++ b/anchor/subnet_service/Cargo.toml
@@ -23,4 +23,6 @@ typenum = { workspace = true }
 types = { workspace = true }
 
 [dev-dependencies]
+database = { workspace = true, features = ["test-utils"] }
+rusqlite = { workspace = true }
 tempfile = "3.14.0"
--- a/anchor/subnet_service/src/scoring.rs
+++ b/anchor/subnet_service/src/scoring.rs
@@ -92,6 +92,8 @@ impl<S: SlotClock> SubnetService<S> {
         network_state
             .clusters()
             .values()
+            // Liquidated clusters generate no traffic, so they must not inflate expected rates.
+            .filter(|cluster| !cluster.liquidated)
             .filter(|cluster| {
                 let operator_ids: Vec<OperatorId> =
                     cluster.cluster_members.iter().copied().collect();
--- a/anchor/subnet_service/src/subscriptions.rs
+++ b/anchor/subnet_service/src/subscriptions.rs
@@ -61,6 +61,8 @@ impl SubnetGenerator {
             .get_own_clusters()
             .iter()
             .flat_map(|cluster| state.clusters().get_by(cluster))
+            // Liquidated clusters perform no duties, so their subnets need no subscription.
+            .filter(|cluster| !cluster.liquidated)
             .map(|cluster| cluster.cluster_members.iter().copied().collect::<Vec<_>>())
             .collect();
         SubnetGenerator::Committees(all_committee_members)
@@ -302,19 +304,25 @@ mod tests {
         time::Duration,
     };
 
-    use database::NetworkDatabase;
+    use database::{
+        NetworkDatabase, PendingStateUpdates,
+        test_utils::{DEFAULT_NUM_OPERATORS, commit_and_publish, generators},
+    };
     use fork::{ALAN_TOPIC_PREFIX, Fork, ForkConfig, ForkLifecycle, ForkSchedule};
+    use rusqlite::Transaction;
     use slot_clock::{ManualSlotClock, SlotClock};
-    use ssv_types::{OperatorId, domain_type::DomainType};
+    use ssv_types::{Cluster, ClusterId, Operator, OperatorId, domain_type::DomainType};
     use task_executor::test_utils::TestRuntime;
     use tempfile::TempDir;
     use tokio::{
         sync::{mpsc, watch},
         time::timeout,
     };
-    use types::{ChainSpec, Epoch, MinimalEthSpec, Slot};
+    use types::{
+        ChainSpec, Epoch, MinimalEthSpec, Slot, test_utils::generate_deterministic_keypair,
+    };
 
-    use crate::{SUBNET_COUNT, TopicEvent, start_subnet_service};
+    use crate::{SUBNET_COUNT, SubnetId, TopicEvent, start_subnet_service};
 
     const TEST_NETWORK: &str = "test";
     const ALAN_DOMAIN: DomainType = DomainType([0, 0, 0, 1]);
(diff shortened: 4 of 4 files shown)


shane-moore:
@claude, please take a look

jnhsigmap (approved):
makes sense 🙂 looks good!

PR #1090, opened 2026-06-15
refactor!: gate ePBS on Ethereum Gloas fork, remove SSV Fork::CStar

shane-moore:

ePBS (Gloas) behavior on the `epbs` branch is gated behind a placeholder SSV protocol fork, `Fork::CStar`, that has never had a real schedule entry (only commented-out examples in the built-in `ssv_fork_schedule.yaml` files). That couples an Ethereum consensus-layer transition (Gloas / EIP-7732) to an SSV-invented fork ordinal, so the fork boundary effectively lives in two places (the SSV fork schedule *and* the eth2 `config.yaml`) and can drift, risking ePBS activating at the wrong slot.

Our pinned Lighthouse `ChainSpec` already exposes the Gloas boundary directly (`gloas_fork_epoch`, `fork_name_at_slot` / `fork_name_at_epoch`, `gloas_enabled()`), so the canonical source of truth is already in hand. This aligns with SIP-94, which frames ePBS as an Ethereum-fork-driven change rather than a new SSV network fork.

## Change Overview

Gate the Gloas-specific paths on the Ethereum Gloas fork from the `ChainSpec` instead of the SSV fork:

```
active_fork(epoch) >= Fork::CStar
  ->  spec.fork_name_at_slot::<E>(slot).gloas_enabled()
```

- Remove the `Fork::CStar` variant; `Fork` is now `{Alan, Boole}`.
- Thread `spec: Arc<ChainSpec>` into `QbftManager` and the message `Validator` (`validator_store` already held one).
- Add a `RoleNotActiveBeforeEthFork` validation failure for `PTCAttester` messages received before the Gloas fork.
- Drop the `cstar` entries from the built-in `ssv_fork_schedule.yaml` files.

**Reading order:** start with `common/fork/src/fork.rs` + `schedule.rs` (the variant removal), then `qbft_manager/src/lib.rs` and `validator_store/src/lib.rs` (gate swap + dispatch), then `message_validator/src/lib.rs` (role gate + new failure variant), then the tests.

**What did NOT change:** `BeaconVote` <-> `GloasBeaconVote` dispatch behavior, domain types, and gossip topic prefixes. The SSV `Fork` schedule still governs SSV-only concerns (Alan/Boole gates, deprecated roles). Only the *source* of the Gloas boundary moved, from the SSV fork ordinal to the eth `ChainSpec`.

- **Blast radius:** 7 crates, but the change is an atomic enum-variant removal plus signature threading; it cannot be split without breaking compilation at intermediate steps. Net diff is small (+145 / -181).
- **Operational:** ePBS now activates from `GLOAS_FORK_EPOCH` in the eth2 network config. This is the same config that already drives every other Ethereum fork transition (Electra, Fulu), so there is no new operational surface, and it removes the prior risk of the SSV schedule and eth2 config disagreeing.
- **Trade-off:** the `spec.fork_name_at_*(...).gloas_enabled()` gate is open-coded at 3 sites rather than wrapped in a helper. This deliberately matches the existing house idiom (`metadata_service` gates Electra the same way); a wrapper was not added for 3 callers.

## Validation

- `cargo +nightly fmt --all -- --check`: clean.
- `cargo check --workspace --all-targets`: clean.
- Tests pass: `anchor_validator_store` 46, `message_validator` 63, `qbft_manager` 31. The `qbft_manager` Gloas-dispatch and `message_validator` `PTCAttester` tests were rewritten to drive the boundary via a `spec_with_gloas(Option<epoch>)` helper that sets `ChainSpec::gloas_fork_epoch`, covering pre-Gloas, at-activation-boundary, and post-Gloas.
- Rebased on the latest `epbs` (including #1082 `sign_payload_attestation`) before pushing.

## Rollback

Pure `git revert` restores the `Fork::CStar` gate. No migrations, persisted state, or config changes are involved, and `CStar` had no real schedule entry, so there is no runtime/operational impact.

Follow-on ePBS attestation/sync work can gate on the same `gloas_enabled()` predicate rather than CStar, which also removes the need to keep Gloas on the SSV `Fork` ordinal axis.

### DIFF
--- a/anchor/client/src/lib.rs
+++ b/anchor/client/src/lib.rs
@@ -549,6 +549,7 @@ impl Client {
             slot_clock.clone(),
             subnet_service.clone(),
             fork_schedule.clone(),
+            spec.clone(),
             &executor,
         );
 
@@ -590,6 +591,7 @@ impl Client {
             message_sender,
             NonZeroU64::new(E::slots_per_epoch()).expect("slots_per_epoch is non-zero"),
             fork_schedule.clone(),
+            spec.clone(),
         )
         .map_err(|e| format!("Unable to initialize qbft manager: {e:?}"))?;
 
--- a/anchor/common/fork/src/fork.rs
+++ b/anchor/common/fork/src/fork.rs
@@ -33,19 +33,12 @@ pub enum Fork {
     /// - Subnet topology: `min(SHA256(operator_id)) % 128`
     /// - Topic format: `/ssv/<network>/boole/<subnet>`
     Boole,
-
-    /// The CStar fork, the SSV-side rollout of Ethereum's Gloas (ePBS) features.
-    ///
-    /// Characteristics:
-    /// - Subnet topology: inherits Boole behavior (no subnet change at this fork).
-    /// - Topic format: `/ssv/<network>/cstar/<subnet>`
-    CStar,
 }
 
 impl Fork {
     /// Returns all known forks in chronological order.
     pub const fn all() -> &'static [Fork] {
-        &[Fork::Alan, Fork::Boole, Fork::CStar]
+        &[Fork::Alan, Fork::Boole]
     }
 
     /// Returns the name of this fork as a string.
@@ -53,7 +46,6 @@ impl Fork {
         match self {
             Fork::Alan => "alan",
             Fork::Boole => "boole",
-            Fork::CStar => "cstar",
         }
     }
 
@@ -82,7 +74,6 @@ impl std::str::FromStr for Fork {
         match s.to_lowercase().as_str() {
             "alan" => Ok(Fork::Alan),
             "boole" => Ok(Fork::Boole),
-            "cstar" => Ok(Fork::CStar),
             _ => Err(format!("Unknown fork: {s}")),
         }
     }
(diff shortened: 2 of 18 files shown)


shane-moore:
@claude, please take a look

shane-moore:
@claude, thanks for the review. Addressed in 942c551c:

**1. `current_fork` on `RoleNotActiveBeforeEthFork` — done.** Worth noting this was a field the refactor *dropped*: the pre-refactor PTCAttester gate used `RoleNotActiveBeforeFork`, which already carried `current_fork`. Restored it, now carrying the Ethereum `ForkName` active at the offending epoch (`spec.fork_name_at_epoch(epoch)`), computed once and reused for the `gloas_enabled()` check — the correct axis for an eth-fork gate, and back in parity with the sibling variants.

**2. `spec_with_gloas` duplication — declining, as convention.** Sharing it cross-crate needs a dev-dependency test-utils crate (or a feature-gated `fork::testing` module); the repo currently keeps per-file test helpers (e.g. `generate_fork_schedule` is duplicated the same way). Happy to extract into a shared test util in a follow-up if/when a fourth copy lands.

**3. `validate_role_for_fork` doc comment — done.** Added the `PTCAttester before the Ethereum Gloas (ePBS) fork` bullet. (Minor note: the PTCAttester rejection itself predates this PR — it was previously CStar-gated — so the doc was already missing it; corrected now.)

**4. qbft_manager imports — confirmed, no change needed.** `Fork` stays live via `Fork::Boole` (lib.rs:223, :333) and `Epoch` via `Epoch::new` (lib.rs:222).

**Verification:** `cargo clippy --workspace --tests -- -D warnings -D clippy::allow_attributes` is clean, `cargo +nightly fmt --all -- --check` is clean, and `message_validator` 63/63 pass (the `RoleNotActiveBeforeEthFork` test uses `..`, so the new field needed no test change).

petarjuki7:
Makes sense, lgtm!

petarjuki7 (approved):

PR #1091, opened 2026-06-15
fix(processor): raise urgent_consensus queue default

### LINKED ISSUE #1088: processor: undersized urgent_consensus queue drops gossip bursts (missed attestations on high-footprint operators)
## Summary

A large mainnet operator (~75 committees, 57 of 128 subnets, v1.2.3) is missing attestations because the processor's single `urgent_consensus` queue drops inbound gossip during sub-second slot-boundary bursts, sometimes the node's own committee messages. The default depth (1000) is an underived placeholder. **This issue is scoped to one focused change: raise that default to a measured, sane value (4096).** Work-item expiry and the queue split are tracked in #254. (The affected operator is already unblocked via the existing `--work-queue-size` override; this fixes the default so the next high-footprint operator does not silently hit the same wall.)

## Problem / impact

- Bursts of `Processor queue full` / `Unable to pass message to message receiver` on `urgent_consensus`, hundreds within single seconds at slot boundaries, quiet between.
- Not CPU-starved (~2.6 of 8 cores during the drops, idle cores available); load is bursty, not sustained.
- Most dropped messages are other operators' relay traffic (harmless), but a fraction are the node's own consensus/partial-signature messages, which causes the misses.

## Root cause

Every inbound gossip message is one work item on the single `urgent_consensus` queue (default 1000, `processor/src/lib.rs:78`), drained by `num_cpus` workers each running the full validate-and-route (`message_receiver/src/manager.rs:78`). Past 1000 queued, the newest is dropped. RSA verify (`message_validator/src/lib.rs:698`) is lock-free and parallel, so throughput is bounded by the worker pool and the 1000-slot queue is the only burst buffer.

go-ssv validates inside go-libp2p's pubsub pipeline, which buffers up to `ValidateThrottle = 8192` in-flight validations before dropping, so it absorbs ~8x larger slot-boundary bursts on comparable hardware. Anchor's 1000 was never derived from traffic (placeholder from #57).

## Measured on mainnet (2026-06-14, read-only)

From the mainnet Anchor canaries' `/metrics` (4 active nodes, up to 80 days uptime):

- **Per-item validation service time ~165-430 us** (`anchor_processor_worker_time{type="message_receiver"}`, exact mean via `sum/count`). Drain ~2.5k msg/s/core, ~20k/s at 8 cores.
- **Average arrival ~50-125 msg/s/node**; `message_receiver` is **99.7%** of all `urgent_consensus` traffic.
- **Zero drops and zero queue depth on every canary** over full uptime, so 1000 is adequate at single-cluster footprint and only the high-footprint operator's bursts overflow it. Average CPU ~2% of a core; this is burst-vs-buffer, not CPU.
- The operator overshoots 1000 by "hundreds" per burst (true peak >= ~1500, clipped by the drop), and runs clean on go-ssv's ~8192 ceiling.

## Proposed fix

Raise the `urgent_consensus` default from **1000 to 4096** (`processor/src/lib.rs:78`), with a comment recording the derivation, plus tests (default size; burst-absorption past the old cap). Keep the queue name (it is the `--work-queue-size` CLI contract); value-only change, operator overrides preserved.

**Why 4096 (not 8192):** it covers the observed overshoot with ~2.7x margin and is modest enough to be safe *without* work-item expiry. A deeper default (toward go-ssv's 8192) buffers ~8x more before dropping, which on a sustained-overloaded node means grinding through proportionally more stale messages, so it would need expiry as a guard. 4096 sidesteps that, keeps the change to a single value, and is a strict improvement; if production shows it is short, the escalation below raises it with the proper guard. Peak held memory ~= depth x real msg size (~4 MB at 4096); tokio mpsc allocates lazily so idle cost is ~0.

## Going deeper (if 4096 is not enough)

Recorded so a future maintainer can escalate without re-deriving; every trigger is a processor metric already exported. None needed now.

1. **Deeper default (toward 8192).** More burst headroom if `send_error{type="message_receiver"}` keeps climbing after 4096. A deeper buffer alone only *delays* drops, so under sustained overload it needs a shedding policy (next), not just a bigger backlog.
2. **Priority / queue split (#254), the real overload lever.** Shed *other operators' relay traffic* before our own duty-critical/outbound work; ownership answers "still useful to us?" far better than age. Trigger: `send_error{type="signer"|"sender"}` fires (own outbound dropped) or multi-second `urgent_consensus` backlog.
   - *Why not time-based expiry:* the dominant traffic is attestation QBFT / partial-signature messages, which stay useful for ~the attestation inclusion window (up to 32 slots / 1 epoch; a late-but-within-epoch attestation still earns the target reward, the largest flag component). A correct duty-aware TTL is therefore epoch-scale, so expiry would essentially never fire (a node 30+ slots behind is already dead) for near-zero benefit, while a naive ~1-slot TTL would wrongly drop still-rewardable work. Ownership-based shedding is the better mechanism; expiry might help only for genuinely short-window types (e.g. sync-committee, included the next slot).
3. **Decouple accept-from-validate.** Dispatch validation off the worker pool (go-libp2p style) so queue depth stops being the only burst buffer.
4. **Per-subnet-scaled depth.** Size the queue from the node's subnet count so benefit and memory cost land on the same high-footprint nodes.

Considered and not doing: un-hiding `--work-queue-size`. Its values are queue names slated to change in #254 (`lib.rs:58` TODO) and an unknown name is a hard startup failure (`config.rs:308-310`), so documenting it now is a rename footgun.

## Scope

- **In:** raise the `urgent_consensus` default (value) + tests.
- **Deferred to #254:** work-item expiry, the queue split, and the deeper escalation above.

## Related

- #254 (processor: split into multiple queues), the structural follow-up and home for the escalation path.
- #1084 (do not subscribe to liquidated-cluster subnets), complementary load reduction (~57 -> ~38 subnets for the affected operator), not a substitute.

shane-moore:

A large mainnet operator (~75 committees, ~45% of all SSV subnets carried by a single node) is missing attestations because the processor's `urgent_consensus` queue drops inbound gossip during sub-second slot-boundary bursts, sometimes the node's own committee messages. The queue depth was a hardcoded `1000`, an underived placeholder from the original processor PR (#57); go-ssv tolerates the same load on comparable hardware because its validation pipeline buffers far more before dropping.

Measured on the mainnet Anchor canaries (read-only; full data in #1088): per-message validation is ~0.2-0.4 ms (so drain capacity is ~20k/s at 8 cores), average arrival is ~50-125 msg/s/node, and inbound gossip validation is the overwhelming majority of this queue's traffic. Zero drops occur at single-cluster footprint, confirming this is burst-vs-buffer, not a CPU shortage.

- Closes #1088
- Follow-up: #254 (deeper sizing, work-item expiry, queue split)

## Change Overview

Raise the `urgent_consensus` queue default from 1000 to **4096** so high-footprint nodes absorb slot-boundary bursts instead of dropping them. It is a value-only change: the queue name (the `--work-queue-size` CLI contract) and any operator overrides are unchanged.

Intentionally **not** changed: no work-item expiry, no queue split, no new config surface, and `permitless` stays at 1000. Those are deliberately deferred to #254, where the escalation path and its metric triggers are recorded.

- **Memory:** the queue allocates lazily (tokio mpsc), so nodes that never burst pay ~0; worst-case held memory is roughly depth x real message size (~4 MB at 4096), bounded further by continuous draining.
- **Why 4096 and not go-ssv's 8192:** a deeper buffer would hold ~8x more before dropping, which on a *sustained*-overloaded node means grinding through more stale messages, so it would need an expiry guard to stay safe. 4096 sidesteps that while still covering the observed overshoot with margin. The deeper-buffer-plus-expiry path is documented in #254 if production shows 4096 is short.
- **Blast radius:** one queue default in one crate; no behavior change for nodes that never hit the old cap.

## Validation

- `cargo test -p processor`: existing tests plus a new deterministic burst-absorption test that saturates all worker permits, then confirms a burst above the old 1000 cap (and below 4096) all enqueues without `TrySendError::Full`. Verified non-vacuous (a 1000-capacity queue rejects at exactly item 1000).
- `make cargo-fmt-check` and `cargo clippy -p processor --tests`: clean.
- The affected operator is already running with `--work-queue-size urgent_consensus=...` as an interim workaround; this change makes a sane depth the default so other high-footprint operators are not silently affected.

## Rollback

Revert the single commit; the default returns to 1000. No config, data, or migration impact. Operators using `--work-queue-size` are unaffected either way.

### DIFF
--- a/anchor/processor/tests/processor_tests.rs
+++ b/anchor/processor/tests/processor_tests.rs
@@ -1,19 +1,23 @@
-use std::{error::Error, sync::Arc, time::Duration};
+use std::{collections::HashMap, error::Error, sync::Arc, time::Duration};
 
 use task_executor::TaskExecutor;
 use tokio::{
     select,
-    sync::{Barrier, Notify, oneshot},
+    sync::{Barrier, Notify, mpsc::error::TrySendError, oneshot},
     time::sleep,
 };
 
-/// Workers for the burst test; kept small so a few blocking tasks saturate every
+/// Workers for the capacity tests; kept small so a few blocking tasks saturate every
 /// permit and stop the processor draining `urgent_consensus`.
 const BURST_TEST_MAX_WORKERS: usize = 3;
 
-/// Burst size for the capacity test: above the old 1000 cap and below the new 4096
-/// default, so every send succeeds now but would have overflowed before (#1088).
+/// Burst size for the positive capacity test: above the old 1000 cap and below the new
+/// 4096 default, so every send succeeds now but would have overflowed before (#1088).
 const BURST_TEST_ITEM_COUNT: usize = 1500;
+
+/// The previous `urgent_consensus` default. The negative test pins the queue here to prove
+/// the burst test is non-vacuous: at this cap the same burst really is rejected.
+const OLD_QUEUE_CAP: usize = 1000;
 
 #[tokio::test]
 async fn test_max_workers() -> Result<(), Box<dyn Error>> {
@@ -104,12 +108,58 @@ async fn test_max_workers() -> Result<(), Box<dyn Error>> {
     Ok(())
 }
 
-/// Verifies that with the default config the `urgent_consensus` queue absorbs a burst
-/// larger than its old 1000 cap without returning `TrySendError::Full` (#1088).
-///
-/// Determinism comes from saturating every worker permit before the burst, so the
-/// processor cannot drain `urgent_consensus` and the channel only fills; the inline
-/// comments justify each step.
+/// Handle returned by [`saturate_workers`]. Releasing it lets the parked blockers finish so
+/// the runtime can shut down cleanly.
+struct BlockerRelease {
+    release_blockers: Arc<Barrier>,
+}
+
+impl BlockerRelease {
+    async fn release(self) {
+        self.release_blockers.wait().await;
+    }
+}
+
+/// Saturates every worker permit with a parked task so the processor cannot drain
+/// `urgent_consensus`, isolating the channel's capacity from the worker pool. Returns once
+/// all permits are provably held; the returned handle frees the blockers afterwards.
+async fn saturate_workers(senders: &processor::Senders, max_workers: usize) -> BlockerRelease {
+    // Opens only once all blockers plus this task have arrived, proving the processor has
+    // spawned every blocker and therefore acquired every permit.
+    let permits_saturated = Arc::new(Barrier::new(max_workers + 1));
+    // A second barrier releases them. Unlike `Notify::notify_waiters`, a barrier cannot lose
+    // the wakeup if a blocker has not yet parked when the caller releases.
+    let release_blockers = Arc::new(Barrier::new(max_workers + 1));
+
+    // Occupy every worker permit with a task that parks without releasing its permit.
+    for _ in 0..max_workers {
+        let permits_saturated = permits_saturated.clone();
+        let release_blockers = release_blockers.clone();
+        senders
+            .urgent_consensus
+            .send_async(
+                async move {
+                    permits_saturated.wait().await;
+                    release_blockers.wait().await;
+                },
+                "saturate_blocker",
+            )
+            .expect("blocker should enqueue while the queue still has capacity");
+    }
+
+    // Wait until all permits are confirmed held. After this, the processor can no longer
+    // drain `urgent_consensus`, so the channel only accumulates work.
+    select! {
+        _ = sleep(Duration::from_secs(5)) => panic!("blockers should saturate all worker permits"),
+        _ = permits_saturated.wait() => {},
+    }
+
+    BlockerRelease { release_blockers }
+}
+
+/// Verifies that with the default config the `urgent_consensus` queue absorbs a burst larger
+/// than its old 1000 cap without returning `TrySendError::Full` (#1088). Determinism comes
+/// from saturating every worker permit first, so the processor cannot drain the channel.
 #[tokio::test]
 async fn test_urgent_consensus_absorbs_burst_above_old_cap() -> Result<(), Box<dyn Error>> {
     let handle = tokio::runtime::Handle::current();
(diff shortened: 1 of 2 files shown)


shane-moore:
@claude, please take a look

shane-moore:
@claude, thanks for the review. Addressed the two test observations in 55f3f719:

**2 (non-vacuity).** Added `test_urgent_consensus_rejects_burst_past_old_cap`: it pins the queue to the old 1000 cap via `queue_size`, runs the same saturated-worker harness, fills exactly 1000, and asserts the next send returns `Error::Queue(TrySendError::Full)`. So "rejects past the cap" is now enforced in code rather than only claimed in the PR body, and the burst test can no longer pass vacuously (e.g. if the channel became unbounded or `Full` were ever swallowed).

**3 (duplication).** Extracted the saturate-then-release setup into a `saturate_workers` fixture now that two capacity tests share it. Left `test_max_workers` alone since it exercises priority/ordering with mixed queues, a different shape.

Also hardened the release while in there: blockers are now freed via a second `Barrier` instead of `Notify::notify_waiters()`. The saturation barrier proves the permits are held but not that every blocker has parked on `.notified()` before release, so the barrier closes that lost-wakeup window. Harmless to the assertions (they complete first), but cleaner for a shared fixture.

**1 (test name).** Kept `test_urgent_consensus_absorbs_burst_above_old_cap`. The `BURST_TEST_ITEM_COUNT` doc already carries the "above old cap / below new default" context, and the negative test now makes the "old cap" reference concrete.

anchor/processor/tests/processor_tests.rs
```diff
@@ -95,3 +107,134 @@ async fn test_max_workers() -> Result<(), Box<dyn Error>> {
...
+impl BlockerRelease {
+    async fn release(self) {
+        self.release_blockers.wait().await;
+    }
+}
+
+/// Saturates every worker permit with a parked task so the processor cannot drain
+/// `urgent_consensus`, isolating the channel's capacity from the worker pool. Returns once
+/// all permits are provably held; the returned handle frees the blockers afterwards.
+async fn saturate_workers(senders: &processor::Senders, max_workers: usize) -> BlockerRelease {
```
petarjuki7, line 126: Cool stuff!

petarjuki7 (approved):
Lgtm!

PR #1102, opened 2026-06-22
chore: Backmerge from stable for v1.2.4

dknopik:
DO NOT MERGE VIA GITHUB, for review only.

contains the stuff that was in stable, plus the contents from the v1.2.4 release.


jnhsigmap (approved):
appears to be as expected

petarjuki7 (approved):

Issue #1061, opened 2026-05-27
feat(validator_store): migrate sign_committee_sync_committee_signatures to GloasBeaconVote at CStar


shane-moore:
## Goal

Migrate `sign_committee_sync_committee_signatures` to call `decide_instance::<GloasBeaconVote>` at `Fork::CStar+`, passing the same `GloasBeaconVote` value that `sign_committee_attestations` passes. This restores the pre-#1025 shared-instance behavior at CStar: attestation and sync-committee messages for the same `(committee, slot)` share a single QBFT task via `Initialized.on_completed: Vec<oneshot::Sender>`.

Pre-CStar code path unchanged.

## Context

#1025 introduced fork-gated wire dispatch in `QbftManager::receive_data` that routes ALL `Role::Committee` messages at `epoch >= CStar` to `gloas_beacon_vote_instances`. The sync-committee messages path (`anchor/validator_store/src/lib.rs:1362-1452`) was not migrated and still calls `decide_instance::<BeaconVote>`, spawning a local `BeaconVote` instance in `beacon_vote_instances`. At CStar, that instance receives no wire input from peers (their dispatch sends matching messages to `gloas_beacon_vote_instances`) and times out every slot.

Pre-#1025, attestation and sync-committee messages shared ONE QBFT task per `(committee, slot)` via `Initialized.on_completed: Vec<oneshot::Sender>` (`anchor/qbft_manager/src/instance.rs:51`). The second `decide_instance` call pushes its `on_completed` sender into the existing instance's Vec rather than spawning a new task (`instance.rs:75`). Post-#1025 they bifurcate at CStar because the two `D` types (`BeaconVote` vs `GloasBeaconVote`) resolve to different maps in `D::get_map`. This issue restores the shared-instance behavior by aligning both call sites on `GloasBeaconVote` with byte-identical initial values.

Why `attestation_data_index` cannot be a placeholder (e.g., `0`): the second caller's `initial_value` is discarded — only the first caller's value seeds the QBFT instance (`instance.rs:71-77`). If sync-committee races attestation and calls first with `attestation_data_index = 0`, attestation's later call attaches to `on_completed` and the QBFT instance seeds with `0`; if the local operator is the QBFT leader, the cluster decides `index = 0` regardless of any operator's local BN view. That defeats SIP-94 §2's purpose (operators reflecting their local payload-status view through the signed root). Both callers must pass the same `attestation_data_index`, sourced from the same BN-supplied `getAttestationData` response.

[...]

PR #1103, opened 2026-06-29
feat(validator_store): apply decided Gloas attestation index in committee signing

### LINKED ISSUE #1027: feat(validator_store): apply decided AttestationDataIndex in metadata-service and committee signing

petarjuki7:

Under Gloas (SIP-94 §2), `AttestationData.index` is BN-supplied and part of both the signed attestation root and the double-vote slashing predicate. #1025/#1026 added `GloasBeaconVote` and a fork-aware validator (range-check + slashing reconstruction with the index), but nothing threaded the decided index into the signed attestation — so divergent BN views would produce divergent signing roots.

Resolves #1027. Carries the production change for #1061 (this lands the sync-path migration).

## Change Overview

- `SlotVote { Base(BeaconVote), Gloas(GloasBeaconVote) }` fork-tagged enum on `VotingContext`, so the BN index rides through the slot's shared vote.
- Producer carries `attestation_data.index`; the aggregate-fetch root is computed with it.
- `sign_committee_attestations` applies the QBFT-decided index to each validator's `attestation.data.index` before signing (Gloas only).
- `sign_committee_sync_committee_signatures` decides over the fork-matching type from the same vote and aligns its QBFT start to `get_attestation_due`, restoring the shared committee instance at Gloas.
- Pre-Gloas attestation index and signing behavior are unchanged; committee QBFT seeding now uses the shared `VotingContext` source for both attestation and sync paths.

- The decided index now lands in the signed root, matching #1026's slashing reconstruction.
- Pre-Gloas sync-QBFT start shifts 4000ms → 3999ms (immaterial) and removes a call-order-dependent divergence on the shared instance.
- Aggregation keys the fetch off the locally-cached index per SIP-94 §2; convergence is handled downstream by the AggregatorCommittee consensus — tracked as a follow-up.
- Deliberate `Base`/`Gloas` duplication collapses when `BeaconVote` retires.

## Validation

9 new tests (5 integration + 4 unit): decided-index application, cluster-wide signing-root agreement, pre-Gloas pre-Electra/Electra+ invariance, sync/attestation seed identity; non-vacuity confirmed. metadata-service fork and aggregate-root construction are unit-tested through production helpers (the crate avoids BN mocks). Workspace compiles; clippy + fmt clean.

## Rollback

Additive at the type/validator layer; reverting restores pre-Gloas behavior throughout.

### DIFF
--- a/anchor/validator_store/src/lib.rs
+++ b/anchor/validator_store/src/lib.rs
@@ -64,12 +64,12 @@ use tracing::{Instrument, Span, debug, error, field, info, info_span, trace, war
 use types::{
     AbstractExecPayload, Address, AggregateAndProof, AggregateAndProofBase,
     AggregateAndProofElectra, Attestation, AttestationBase, AttestationElectra, BeaconBlock,
-    BeaconBlockRef, BlindedPayload, ChainSpec, ContributionAndProof, Domain, Epoch, EthSpec,
-    ExecutionPayloadEnvelope, ForkName, FullPayload, Graffiti, Hash256, PayloadAttestationData,
-    PayloadAttestationMessage, ProposerPreferences, SelectionProof, SignedAggregateAndProof,
-    SignedBeaconBlock, SignedBlindedBeaconBlock, SignedContributionAndProof,
-    SignedExecutionPayloadEnvelope, SignedProposerPreferences, SignedRoot,
-    SignedValidatorRegistrationData, SignedVoluntaryExit, Slot, SlotData,
+    BeaconBlockRef, BlindedPayload, ChainSpec, Checkpoint, ContributionAndProof, Domain, Epoch,
+    EthSpec, ExecutionPayloadEnvelope, ForkName, FullPayload, Graffiti, Hash256,
+    PayloadAttestationData, PayloadAttestationMessage, ProposerPreferences, SelectionProof,
+    SignedAggregateAndProof, SignedBeaconBlock, SignedBlindedBeaconBlock,
+    SignedContributionAndProof, SignedExecutionPayloadEnvelope, SignedProposerPreferences,
+    SignedRoot, SignedValidatorRegistrationData, SignedVoluntaryExit, Slot, SlotData,
     SyncAggregatorSelectionData, SyncCommitteeContribution, SyncCommitteeMessage,
     SyncSelectionProof, SyncSubnetId, ValidatorRegistrationData, VoluntaryExit,
 };
@@ -1424,37 +1424,77 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> AnchorValidator
         let validator_attestation_committees =
             self.get_attesting_validators_in_committee(&voting_context, committee_id);
 
-        // Run QBFT consensus once for the entire committee
+        // Run QBFT consensus once for the entire committee. The start time mirrors the
+        // attestation path's fork-aware deadline: attestation and sync-committee duties for the
+        // same `(committee, slot)` share one QBFT instance, and the first caller's timing wins,
+        // so both must compute `instance_start_time` identically or round timing becomes
+        // call-order-dependent at Gloas.
         let timer = metrics::start_timer_vec(&metrics::CONSENSUS_TIMES, &[metrics::BEACON_VOTE]);
         let timeout_mode = TimeoutMode::SlotTime {
             instance_start_time: self
-                .get_instant_in_slot(slot, self.spec.get_slot_duration() / 3)?,
+                .get_instant_in_slot(slot, self.spec.get_attestation_due::<E>(slot))?,
         };
 
-        let completed = self
-            .consensus
-            .decide_instance(
-                CommitteeInstanceId {
-                    committee: committee_id,
-                    instance_height: slot.as_usize().into(),
-                },
-                voting_context.beacon_vote.clone(),
-                self.create_beacon_vote_validator(slot, validator_attestation_committees),
-                timeout_mode,
-                &cluster.cluster_members,
-            )
-            .await
-            .map_err(SpecificError::from)?;
-        drop(timer);
+        let instance_id = CommitteeInstanceId {
+            committee: committee_id,
+            instance_height: slot.as_usize().into(),
+        };
 
-        let data = match completed {
-            Completed::TimedOut => return Err(Error::SpecificError(SpecificError::Timeout)),
-            Completed::Success(data) => data,
+        // Decide over the fork-matching QBFT type so this call attaches to the SAME shared
+        // instance the attestation path creates (instances are keyed by `D` via `D::get_map`).
+        // Both paths build the seed from the same `voting_context.vote`, so the initial-value
+        // hashes match by construction and the sync caller joins the attestation instance via
+        // `on_completed` instead of spawning a sibling. Only `block_root` (the sync signing
+        // root) and the decided hash (the partial-signature base) are consumed downstream.
+        let vote = voting_context.vote.clone();
+        let (block_root, decided_hash) = if self.spec.fork_name_at_slot::<E>(slot).gloas_enabled() {
+            let completed = self
+                .consensus
+                .decide_instance(
+                    instance_id,
+                    GloasBeaconVote {
+                        block_root: vote.block_root(),
+                        source: vote.source(),
+                        target: vote.target(),
+                        attestation_data_index: vote.index(),
+                    },
+                    self.create_gloas_beacon_vote_validator(slot, validator_attestation_committees),
+                    timeout_mode,
+                    &cluster.cluster_members,
+                )
+                .await
+                .map_err(SpecificError::from)?;
+            drop(timer);
+            match completed {
+                Completed::TimedOut => return Err(Error::SpecificError(SpecificError::Timeout)),
+                Completed::Success(decided) => (decided.block_root, decided.hash()),
+            }
+        } else {
+            let completed = self
+                .consensus
+                .decide_instance(
+                    instance_id,
+                    BeaconVote {
+                        block_root: vote.block_root(),
+                        source: vote.source(),
+                        target: vote.target(),
+                    },
+                    self.create_beacon_vote_validator(slot, validator_attestation_committees),
+                    timeout_mode,
+                    &cluster.cluster_members,
+                )
+                .await
+                .map_err(SpecificError::from)?;
+            drop(timer);
+            match completed {
+                Completed::TimedOut => return Err(Error::SpecificError(SpecificError::Timeout)),
+                Completed::Success(decided) => (decided.block_root, decided.hash()),
+            }
         };
 
         // Prepare all validators (metadata already resolved by `group_by_committee`)
         let domain = self.get_domain(epoch, Domain::SyncCommittee);
-        let signing_root = data.block_root.signing_root(domain);
+        let signing_root = block_root.signing_root(domain);
 
         let prepared: Vec<SigningRequest<u64>> = messages
             .into_iter()
@@ -1477,7 +1517,7 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> AnchorValidator
                 slot,
                 &cluster,
                 validator_partial_signature_batch_size,
-                data.hash(),
+                decided_hash,
                 &prepared,
             )
             .await?;
@@ -1505,7 +1545,7 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> AnchorValidator
             );
             results.push(SyncCommitteeMessage {
                 slot,
-                beacon_block_root: data.block_root,
+                beacon_block_root: block_root,
                 validator_index,
                 signature,
             });
@@ -1529,7 +1569,6 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> AnchorValidator
             return Ok(vec![]);
         };
         let slot = first_attestation.attestation.data().slot;
-        let first_att_data = first_attestation.attestation.data();
 
         let voting_context_tx = self.get_voting_context(slot).await?;
         let validator_attestation_committees =
@@ -1547,7 +1586,9 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> AnchorValidator
             instance_height: slot.as_usize().into(),
         };
 
-        let (block_root, source, target, decided_hash) = if self
+        let vote = voting_context_tx.vote.clone();
+
+        let (block_root, source, target, decided_hash, decided_index) = if self
             .spec
             .fork_name_at_slot::<E>(slot)
             .gloas_enabled()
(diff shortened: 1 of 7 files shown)


anchor/validator_store/src/lib.rs
```diff
@@ -1424,37 +1424,77 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> AnchorValidator
...
-            Completed::TimedOut => return Err(Error::SpecificError(SpecificError::Timeout)),
-            Completed::Success(data) => data,
+        // Decide over the fork-matching QBFT type so this call attaches to the SAME shared
+        // instance the attestation path creates (instances are keyed by `D` via `D::get_map`).
+        // Both paths build the seed from the same `voting_context.vote`, so the initial-value
+        // hashes match by construction and the sync caller joins the attestation instance via
+        // `on_completed` instead of spawning a sibling. Only `block_root` (the sync signing
+        // root) and the decided hash (the partial-signature base) are consumed downstream.
+        let vote = voting_context.vote.clone();
+        let (block_root, decided_hash) = if self.spec.fork_name_at_slot::<E>(slot).gloas_enabled() {
```
jnhsigmap: The fork branching works nicely, this is just a refactor and doesn't modify the signing flow:

`self.spec` appears to own the `SlotVote`, however we are checking based on `fork_name_at_slot(...).gloas_enabled()` and referencing `vote` for the `BeaconVote` fields. There's an opportunity to match directly on the `SlotVote` variant and wrap the match / `decide_instance` into a helper method.

Something like
```
  /// Decided committee beacon-vote fields for a slot.
  struct DecidedVote {
      block_root: Hash256,
      source: Checkpoint,
      target: Checkpoint,
      decided_hash: Hash256,
      /// `Some` at Gloas; `None` pre-Gloas (`BeaconVote` carries no index).
      decided_index: Option<u64>,
  }

  impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> AnchorValidatorStore<T, E, C> {
      ...
      /// Runs the committee QBFT for `slot`, dispatching on the `SlotVote` variant.
      async fn decide_committee_vote(
          &self,
          instance_id: CommitteeInstanceId,
          vote: SlotVote,
          slot: Slot,
          validator_attestation_committees: HashMap<PublicKeyBytes, u64>,
          timeout_mode: TimeoutMode,
          cluster_members: &IndexSet<OperatorId>,
      ) -> Result<DecidedVote, Error> {
          match vote {
              SlotVote::Gloas(seed) => {
                  let completed = self
                      .consensus 
                      .decide_instance(
                          instance_id,
                          seed,
                          self.create_gloas_beacon_vote_validator(slot, validator_attestation_committees),
                          timeout_mode,
                          cluster_members,
                      )
                      .await
                  match completed {
                      Completed::TimedOut => Err(Error::SpecificError(SpecificError::Timeout)),
                      Completed::Success(d) => Ok(DecidedVote {
                          block_root: d.block_root,
                          source: d.source,
                          target: d.target,
                          decided_hash: d.hash(),
                          decided_index: Some(d.attestation_data_index),
                      }),
                  }
              }
              SlotVote::Base(seed) => {
                  let completed = self
                      .consensus
                      .decide_instance(
                          instance_id,
                          seed,
                          self.create_beacon_vote_validator(slot, validator_attestation_committees),
                          timeout_mode,
                          cluster_members,
                      )
                      .await
                  match completed {
                      Completed::TimedOut => Err(Error::SpecificError(SpecificError::Timeout)),
                      Completed::Success(d) => Ok(DecidedVote {
                          block_root: d.block_root,
                          source: d.source,
                          target: d.target, 
                          decided_hash: d.hash(),
                          decided_index: None,
                      }),
                  }
              }
          }
      }
  }
```

For example here in `sign_committee_sync_committee_signatures`:
```
          ...
          let instance_id = CommitteeInstanceId {
              committee: committee_id,
          };

          let decided = self
              .decide_committee_vote(
                  instance_id,
                  voting_context.vote.clone(),
                  slot,
                  validator_attestation_committees,
                  timeout_mode,
                  &cluster.cluster_members,
              )
              .await?;
          ...
```

Given we're duplicating the existing check in [`sign_committee_attestation`](https://github.com/sigp/anchor/pull/1103/changes#diff-25a340f5184292e9132e4b012313bfd9e056682eb85555b96830b1909801b2f1R1591).

jnhsigmap (commented):
lets also create a follow-up issue for the aggregation seed-vs-decided component :)

jnhsigmap:
@claude, provide a review

anchor/validator_store/src/lib.rs
```diff
@@ -1588,9 +1628,9 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> AnchorValidator
                 .decide_instance(
                     instance_id,
                     BeaconVote {
-                        block_root: first_att_data.beacon_block_root,
-                        source: first_att_data.source,
-                        target: first_att_data.target,
+                        block_root: vote.block_root(),
```
shane-moore: Could you tweak the “Pre-Gloas paths unchanged” wording in PR description?

  One nuance I missed at first: this PR also changes the pre-Gloas attestation QBFT seed from the incoming
  `AttestationToSign` data (`first_att_data`) to the shared metadata-service `VotingContext.vote`. I think
  that is the right direction, since it makes attestation and sync seed the same committee QBFT instance
  from the same source, and it means fallback WAD selection is actually honored by the attestation path too.

  But it is still a pre-Gloas behavior change when `VotingContext` and Lighthouse’s attestation service
  would have selected different attestation data. Maybe describe it as something like:

  “Pre-Gloas attestation index/signing behavior is unchanged; committee QBFT seeding now uses the shared
  `VotingContext` source for both attestation and sync paths.”

shane-moore: also, could we add one non-vacuity check for the seed-source switch?

The current Gloas tests seed both `VotingContext.vote` and the incoming `AttestationToSign` with zero roots/checkpoints, so they prove the decided index is applied but do not prove the rest of the QBFT seed comes from `VotingContext.vote`. A regression that took `block_root/source/target` back from `first_att_data` while still using `vote.index()` would pass.

Suggested shape: make the voting-context vote and the duty attestation disagree on `beacon_block_root` (or checkpoint root), force the mock decider to preserve the seed's non-index fields while deciding `attestation_data_index = 1`, then assert the signed attestation carries the voting-context root plus the decided index. That pins the intended contract that metadata-service/WAD selection feeds the attestation path too.

anchor/validator_store/src/lib.rs
shane-moore: Agree with this approach. It should apply to both `sign_committee_attestations` and `sign_committee_sync_committee_signatures`, both blocks are structurally identical after the timing alignment.

One extension: build `instance_id`, `timeout_mode`, and the consensus timer inside the helper rather than passing them in. Both are pure functions of `(committee_id, slot)`, and since the shared instance takes the first caller's timing, computing `instance_start_time` in one place makes the alignment structural instead of by-convention.

With the call sites matching on the variant directly, `SlotVote::source()`/`target()` go dead and can be removed.

shane-moore:
@petarjuki7, lgtm man, great work! made the small tweaks mentioned above. feel free to merge if you're good with it

shane-moore:
gonna merge this so that I can build my aggregations PR on top of `SlotVote`. if have any follow-ups for me, feel free to file another gh issue in the attestation milestone ⚡

shane-moore (approved):

Issue #1104, opened 2026-06-29
Re-align proposer QBFT instrumentation taxonomy with the original observability model


jnhsigmap:
## Summary

The proposer-duty observability model was defined and agreed upon (#919) as the [milestone's](https://github.com/sigp/anchor/milestone/1) consistent taxonomy of span names, checkpoint vocabulary, outcome taxonomy, and metric naming. The instrumentation shipped in #1066 (PR #1068, merged) was the primary source of drift. Issue #1067 consolidates it through wiring.

Realignment here is justified to ensure consistency and a single reference point for instrumentation structure. This issue tracks reconciling the two so the remaining milestone work (#923) builds on one consistent taxonomy.

This is **observability-naming tech debt only** with no intention of changing consensus behavior and it is filed separately as a follow-up to #1067.

## Source of drift

- **#1066 / PR #1068 (merged)** introduced: all 7 checkpoint constants, the 4 `RoundAdvanceReason` labels, and all 5 metric names. This is the origin of most of the drift.
- **#1067** added only the `ProposerOutcome` vocabulary (`decided`/`max_round_timeout`/`channel_closed`) at the qbft-instance layer.

## Divergences (implemented → #919 canonical)

| Area | Implemented (#1066/#1067) | #919 canonical | Note |
|------|---------------------------|----------------|------|
| Child span | `proposer_qbft_instance` | `qbft { decided_round }` (child of `proposer_sign_block`) | Implemented span provides more detail, keep. |
| Decided-round metric | `anchor_proposer_qbft_decided_round` | `anchor_proposer_consensus_decided_round` | `qbft_` aligns to span naming, keep |
| Round metric | `anchor_proposer_round_advance_total{reason}` | `anchor_proposer_round_timeout_total{round}` | adopt #919's `round_timeout_total{round}` as the friction counter; drop the `reason` counter (keep `reason` as a span field, see below) |
| Checkpoints | flat `qbft_*`; `round_timeout` + `round_advanced` **merged** into `qbft_round_advance{reason}`; no distinct `commit_quorum`; `proposal_accepted` vs `proposal_received` | dotted `block.*` and `randao.*`; separate `round_timeout`/`round_advanced`; distinct `commit_quorum` | namespace + structure; split the merged event back into `round_timeout{round}` + `round_advanced{old,new,reason}` |
[...]

PR #1105, opened 2026-06-29
feat: instruments proposer QBFT lifecycle

jnhsigmap:

- This implements boundary-layer components of proposer observability [milestone](https://github.com/sigp/anchor/milestone/1) in the main `qbft_instance` loop.
- #1066 (merged in #1068) added taxonomy, metrics, and classifier as un-used scaffold. This change wires them into the live QBFT instance loop.
- _**Evidence:**_ mirrors the visibility SSV shipped in `ssvlabs/ssv#2609` (QBFT stage-duration logging); satisfies the four acceptance criteria of #921.
- _**Links:**_ issue #1067 (this PR), parent #921, taxonomy PR (issue #1066) #1068, observability model #919.

**Note:** I understand that #1066 drifted materially from the original agreed observability model in #919 and I think that we should explicitly aim to realign. This would obviously touch already-merged components and should sensibly land in a separate change that follows this one. See #1104 for detail.

## Change Overview (Required)

- Observes a proposer QBFT instance's state at the boundary by calling into it.
- _**Observability added:**_
    - `tracing` span with lifecycle checkpoint events (start, proposal accepted, prepare quorum, round advance, decided/timed-out/channel-closed).
    - 5 Prometheus metrics (decided round, round advances by reason, handoff budget, outcome, total duration).
- Handoff budget (slot time remaining at consensus start) computed inside `qbft_manager::decide_instance` from the duty slot _for proposer instances only_.
- QBFT state is read at recv-loop boundaries via existing accessors. Emission is encapsulated in an observer instance owned by the `instrumentation.rs` module. Hot loop controls instrumentation semantically through methods offered by observer.
- The decided-round metric/span now records the round from the decided certificate via a new read-only `Qbft::decided_round()` and not the local round cursor. This is a better choice for cross-round decided commits.
- Instance span/start event now carries `start_round`/`start_state` so round advances during buffered-message replay (if applicable) before the observer exists are at least traceable through recording the starting state.
- _**Reviewer guide**_: Start with the `instrumentation.rs` module (observer + taxonomy), then `handoff_budget` calculation in `qbft_manager/src/lib.rs` method `decide_instance`, and the `common/qbft` determination of `decided_round`, then the instance loop call sites in `instance.rs`.
- Intentionally unchanged: `common/qbft` consensus logic (gains only a read-only state accessor from #1068), timeout formulas, leader selection, and all non-proposer duties.

- _**Blast radius:**_ `qbft_manager` (instance loop, instrumentation, decide_instance internals) and a read-only behavior-neutral addition to `common/qbft` (a `decided_round` field and getter. `current_round` is never mutated).
- _**Trade-off:**_ Observation is implemented for proposer instances only. Non-proposer duties skip the state machine snapshot on each loop iteration. Other duties bypass instrumentation gated by observer and snapshot being created.
- _**Mitigation**_: Round-advance classifier is pure and unit-tested. Proposer path covered end-to-end for both terminal outcomes (instance decided and max-round timeout). metric/label cardinality is bounded (high-cardinality values stay in spans, never in Prometheus labels).

## Validation (Required)

- _**Existing:**_ `qbft_manager` tests and real-`Qbft` correspondence tests (from #1068).
- _**New:**_
  - 3 proposer e2e tests (decided / max-round-timeout / channel-closed).
  - max-timeout test now asserts the handoff-budget histogram (count + ms→s sum).
  - `decided_round` coverage in `common/qbft` (a discriminating certificate-round != cursor case and a commit-quorum case).
  - unit tests for the relocated `compute_handoff_budget_ms` helper.
- `make cargo-fmt-check`, `make lint` (`-D warnings`), and `make test` all pass.

- Revert this commit. This will remove instrumentation and metric wiring that has no behavioural impact beyond adding instrumentation gates for proposer duty flow.
- Operational impact is additive. New metric series and span/log events and the tracing span is a no-op without a OTLP collector. No existing metric is removed or renamed in this PR.

- New field `handoff_budget` computed in `qbft_manager` from the duty-slot boundary. Is `None` when the slot clock cannot resolve start_of(duty_slot)/now.
- the handoff-budget value now anchors on the duty-slot boundary rather than the previous current-slot modulo, so the emitted `handoff_budget_ms` changes in late-start scenarios
- _**Noteworthy:**_ Taxonomy has drifted from the original #919 definition (see #1104). Reconciliation is warranted and intentionally deferred to a separate issue for further determination, since realignment affects already-merged taxonomy and the sibling milestone spans.


jnhsigmap:
@claude review this PR based on the criteria of #1066/#1067 and #921. Then provide separate analysis and evaluation of recommendation in #1104 regarding drift from original taxonomy.

jnhsigmap:
@claude, https://github.com/sigp/anchor/pull/1105/commits/be1aefda007d1b4fb1de0406a8e07dad0c1df2cb addresses the following:

**`observe_state_transition` doc comment clarified**
- The dual-emit is intentional as the state change and round change events are independent facts and only overlap in this particular case.

**PROPOSER_QBFT_DECIDED_ROUND description is incomplete on the ChannelClosed outcome**
- This was the correct interpretation. Decided round and duration are not meaningful pieces of information when a still-initialized instance is terminated and these metrics may be polluted by recordings from this scenario (though, it is unlikely).
- `finish()` now only records decided-round and duration histograms for consensus-terminal outcomes and as such gates on `!=ChannelClosed`.
- outcome counter remains intact.

**u64::try_from(*get_instance_height()).unwrap_or(u64::MAX) saturates to a misleading sentinel**
- Replaced dead u64::try_from().unwrap_or(u64::MAX) with lossless as u64 cast
 
**Additionally for test coverage**
- `compute_handoff_budget_ms()` extracted as a testable pure helper function with unit tests. `decide_abstract_block` calls it directly.

Items acknowledged without action:

**5× None handoff-budget propagation is noisy**
- Agreed, however is a design artefact, noted.

**OTLP span "built-in duration ≈ event moments, not lifetime" **
- The span is built in `ProposerObserver::start` and lives for the whole instance until `ProposerObserver::finish` when a terminal outcome is recorded and then is dropped at function end. `duration_ms` is a field relied upon by metrics and log consumers.

**Buffered-replay window during initialize() is unobserved**
- Agreed and accepted for now. Mentioned in PR body.

**Part 2 — Recommendation in https://github.com/sigp/anchor/issues/1104: taxonomy drift from https://github.com/sigp/anchor/issues/919**
- As mentioned this PR is additive (wiring of #1068). Including transformational change blurs its objectives.
- The `round_advance_total{reason} counter` → `round_timeout_total{round}` migration and the checkpoint split.
  - Doing it here expands the blast radius into already-merged #1068 territory and mixes previously noted taxonomy change into a wiring PR.
  - To make it clear that this is set to change and the counter is _new operator surface_, an acceptable strategy is a one-line PR-body note freezing its status ("this counter is provisional pending #1104") to be clear that this is going to change. We can also accept that this change is going into `unstable` where users are aware of this fact.

Review again given this angle, also provide a critique of my assessment here.

jnhsigmap:
fair enough, however, if #1104 is agreed upon then I disagree on the perceived severity of splitting the change between this PR and leaving renamed metrics and taxonomy changes to a future change.

If this is direction is approved and agreed upon, it would be a priority to ensure #1104 follows promptly and that the general nature of `unstable` is honoured here. We would never merge to `stable` an incomplete or fragile surface knowing that it is still volatile due to an ongoing initiative.

Marking automated review as 'done' pending team consensus on actual direction and whether my beliefs outlined in #1104 are valid and worth pursuing or whether we remain on the current course with taxonomy shape :+1:

jnhsigmap:
regarding the `run-local-testnet` error: `gh api repos/Zacholme7/ssv-network` returns `404`.

`Zacholme7/ssv-network` is not accessible, either deleted, renamed, or made private. The Kurtosis image build appears to hard-code a clone of it. This not being found is causing the run to fail.

anchor/validator_store/src/lib.rs
```diff
@@ -576,6 +587,8 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> AnchorValidator
 
         let data_validator = self.create_proposer_consensus_data_validator(validator.public_key);
 
+        let handoff_budget_ms = compute_handoff_budget_ms(&self.slot_clock);
```
shane-moore: This should be anchored to the duty slot, not the current wall-clock slot. `compute_handoff_budget_ms()` uses `millis_from_current_slot_start()`, which is modulo `slot_clock.now()`. That happens to match while the wall-clock slot equals the duty slot, but the semantic quantity [#1067/](https://github.com/sigp/anchor/pull/1105/changes#1067/)#921 need is remaining budget for `block.slot()`. If QBFT ever starts after that duty-slot boundary, the histogram reports next-slot budget instead of 0. Since `slot = block.slot()` is already in scope here and the QBFT timer is anchored to that slot, can we compute `(start_of(slot) + slot_duration).saturating_sub(now_duration())` instead?

shane-moore: Actually, thinking more about it, I think the cleaner fix is to move the handoff-budget calculation out of `validator_store` and into `QbftManager::decide_instance`, rather than threading a caller-supplied `Option<u64>` through the consensus API.

Today `anchor/validator_store/src/lib.rs` computes `handoff_budget_ms` at line 590 and then every non-proposer caller has to pass `None` with a comment. But `anchor/qbft_manager/src/lib.rs` already has all the information needed at the actual handoff boundary: after lines 212-214 it has `instance_height`, `message_id.role()`, and `self.slot_clock`. For `Role::Proposer`, `instance_height` is the duty slot because `decide_abstract_block` builds `ProposerInstanceId { duty: ValidatorDutyKind::Proposal, instance_height: slot.as_usize().into() }`.

Suggested shape in `QbftManager::decide_instance`, before constructing `QbftInitialization`:

```rust
let handoff_budget_ms = if message_id.role() == Some(Role::Proposer) {
    let duty_slot = Slot::new(*instance_height as u64);
    self.slot_clock
        .start_of(duty_slot)
        .and_then(|start| {
            self.slot_clock.now_duration().map(|now| {
                (start + self.slot_clock.slot_duration()).saturating_sub(now)
            })
        })
        .map(|budget| budget.as_millis() as u64)
} else {
    None
};
```

Then delete the caller plumbing:

- remove the `handoff_budget_ms` parameter from `QbftManager::decide_instance` and the `ConsensusDecider` trait/impl in `anchor/qbft_manager/src/lib.rs`;
- remove `let handoff_budget_ms = compute_handoff_budget_ms(&self.slot_clock);` from `anchor/validator_store/src/lib.rs`;
- remove the final `handoff_budget_ms` / `None` argument from all `decide_instance` call sites in `validator_store`;
- remove the now-unused `_handoff_budget_ms` parameter from `anchor/validator_store/src/testing/common.rs`;
- delete or move the `compute_handoff_budget_ms` helper/tests from `validator_store`.

`QbftInitialization` can still carry `handoff_budget_ms`, since that is the value the instance loop needs when it starts the proposer observer. The important part is that the value is derived once by the manager from the duty slot, not supplied by callers. Not a separate blocker if you prefer the smaller `compute_handoff_budget_ms(slot, slot_clock)` fix, but this version removes the current-slot bug and deletes the extra API surface at the same time.

anchor/qbft_manager/src/instance.rs
```diff
@@ -284,20 +291,51 @@ pub async fn qbft_instance<D: QbftData<Hash = Hash256>>(
...
                         debug!(msg_id = ?initialization.message_id, "Received initialization message");
                         instance = instance.initialize(initialization, &message_sender).await;
+
+                        // If Qbft instance is running a proposer duty and initialized, start
+                        // proposer observability. NOTE: the observer is
+                        // created only after `initialize()` returns, so any round advances that
+                        // occur while replaying buffered messages during initialization are not
+                        // observed. The terminal outcome is still captured
+                        // by the post-loop `completed()` check below.
+                        if let QbftInstance::Initialized(initialized) = &instance
```
shane-moore, line 331: There is one lifecycle blind spot here that seems cheap to at least flag. If peer QBFT messages arrive before our local `Initialize`, `Uninitialized::receive` buffers them, and `Uninitialized::initialize` replays them with direct `qbft.receive(...)` calls before this observer is created. Any proposal acceptance, prepare quorum, or round advance that happens during that replay will not emit the new checkpoint events or increment `anchor_proposer_round_advance_total`.

That matters most for late-start / catch-up proposer instances, but it can also happen for a normal non-leader if the leader's round-1 proposal is buffered before our local initialize. In that case the observer starts with the instance already in `Prepare` or later, but the `qbft_proposal_accepted` checkpoint never fired.

Could we record the post-replay starting state on the start event/span? For example, at this call site:

```rust
let start_round = initialized.qbft.get_round();
let start_state = initialized.qbft.state_kind();
observer = Some(ProposerObserver::start(
    instance_height,
    handoff_budget_ms,
    start_round,
    start_state,
));
```

Then `ProposerObserver::start` can add `start_round` and `start_state` to the `proposer_qbft_instance` span and `qbft_instance_started` event. That does not recover every replayed transition, but it makes the blind spot visible: `start_round > 1` means replay already advanced rounds, and `start_state != AwaitingProposal` means replay already accepted a proposal or reached a later state before instrumentation started.

anchor/qbft_manager/src/instance.rs
```diff
@@ -340,21 +405,46 @@ pub async fn qbft_instance<D: QbftData<Hash = Hash256>>(
...
+                    }
                 };
                 None
             }
             RecvResult::Closed => {
                 // If the instance can receive no more messages, we no longer need it. Signal
                 // time out to listeners, as this instance was likely cleaned up.
                 if let QbftInstance::Initialized(initialized) = instance {
+                    // If proposer duty: record the `channel_closed` outcome.
+                    if let Some(observer) = &observer {
```
shane-moore, line 449: Can we add one regression test for this ChannelClosed branch and the histogram gate it relies on?

This is now the only path that records `outcome="channel_closed"`, and `ProposerObserver::finish` intentionally keeps ChannelClosed out of `anchor_proposer_qbft_decided_round` and `anchor_proposer_qbft_duration_seconds`. The current proposer e2e tests only cover `decided` and `max_round_timeout`, and they only assert the outcome counter, so deleting this `finish(ChannelClosed, ...)` call or removing the `outcome != ChannelClosed` histogram gate would not be caught.

Suggested shape in `anchor/qbft_manager/src/tests/timeout_tests.rs`:

1. Add a helper using `Histogram::get_sample_count()`:

```rust
fn histogram_sample_count(metric: &metrics::Result<metrics::Histogram>) -> u64 {
    metric
        .as_ref()
        .map(|histogram| histogram.get_sample_count())
        .unwrap_or(0)
}
```

2. Add a third proposer e2e test:

- initialize a proposer instance with `spawn_proposer_instance`;
- record `proposer_outcome_count("channel_closed")`;
- record sample counts for `PROPOSER_QBFT_DECIDED_ROUND` and `PROPOSER_QBFT_DURATION_SECONDS`;
- drop `message_tx`;
- assert `result_rx.await` is `Ok(Completed::TimedOut)`;
- assert the channel-closed outcome counter increments;
- assert both histogram sample counts are unchanged.

One caveat: these Prometheus metrics are process-global, so the no-sample assertion should not race with the existing decided/timeout proposer tests under plain `cargo test`. A small module-level `tokio::sync::Mutex` around the three proposer-metric e2e tests would keep the before/after metric assertions deterministic.

anchor/qbft_manager/src/tests/timeout_tests.rs
```diff
@@ -8,6 +11,221 @@ async fn test_timeouts() {
...
+}
+
+/// Objective: Drive proposer instance to max round timeout with a non-`None` `handoff_budget_ms`.
+/// Proposer instance exhausts its rounds and the observer records the `max_round_timeout` outcome.
+#[tokio::test(start_paused = true)]
+async fn test_proposer_instance_max_round_timeout_runs_observer() {
+    // Mirrors `ProposerOutcome::MaxRoundTimeout.as_str()`.
+    const MAX_ROUND_TIMEOUT_OUTCOME: &str = "max_round_timeout";
+    const MAX_ROUNDS: usize = 2;
+    const HANDOFF_BUDGET_MS: u64 = 4_000;
```
shane-moore, line 150: Can we add a small assertion here for `anchor_proposer_qbft_handoff_budget_seconds`?

This test already passes `Some(4_000)` into `spawn_proposer_instance`, but it currently only checks `outcome_total{outcome="max_round_timeout"}`. That means the live qbft-manager wiring from `QbftInitialization.handoff_budget_ms` to `ProposerObserver::start` to `PROPOSER_QBFT_HANDOFF_BUDGET_SECONDS.observe(...)` can be removed while tests still pass. The pure `compute_handoff_budget_ms` tests cover the helper only, not this observer path.

Suggested shape:

1. Before spawning, record `PROPOSER_QBFT_HANDOFF_BUDGET_SECONDS.get_sample_count()` and `get_sample_sum()`.
2. After `result_rx.await` returns `Ok(Completed::TimedOut)`, assert the sample count increased by 1.
3. Also assert the sample sum increased by `HANDOFF_BUDGET_MS as f64 / 1000.0`, so the test catches ms-to-seconds regressions rather than only proving that some sample was observed.

One caveat: these Prometheus metrics are process-global. If we add before/after assertions to this test and the ChannelClosed test, the proposer-metric e2e tests should be serialized with a small module-level `tokio::sync::Mutex` so they do not race under plain `cargo test`.

This intentionally verifies the qbft-manager observer wiring. It does not by itself prove the `validator_store` computation site; that is covered by the helper tests today, or would be covered more directly if the handoff-budget computation moves into `QbftManager::decide_instance`.

anchor/qbft_manager/src/instance.rs
```diff
@@ -363,6 +453,18 @@ pub async fn qbft_instance<D: QbftData<Hash = Hash256>>(
...
+            // If proposer duty and there is finished data: record the outcome before instance is
+            // overwritten.
+            if let Some(observer) = &observer
+                && let Some(completed) = initialized.qbft.completed()
+            {
+                let outcome = match completed {
+                    Completed::Success(_) => ProposerOutcome::Decided,
+                    Completed::TimedOut => ProposerOutcome::MaxRoundTimeout,
+                };
+                observer.finish(outcome, u64::from(initialized.qbft.get_round()));
```
shane-moore: One semantic caveat on `anchor_proposer_qbft_decided_round`: the value recorded here is the local QBFT round cursor at completion, not necessarily the round in the decided certificate.

`common/qbft` deliberately accepts quorum-signed multi-signer `Commit` messages as decided certificates even when their `qbft_message.round` is behind or ahead of the local `current_round`. `received_decided()` then marks the instance complete and stores the aggregated commit without changing `current_round`. So if this node is still at round 1 and learns a quorum-signed round-3 commit, the histogram records 1. Conversely, if the local timer moved to round 2 and it learns a round-1 decided commit, it records 2.

Could we record the certificate round in `common/qbft` at the success sites instead of passing `get_round()` here? Both paths already have the right value in scope: the local commit-quorum path has `round`, and `received_decided()` has `wrapped_msg.qbft_message.round`. The shape I mean is roughly:

    // common/qbft/src/lib.rs, next to `completed`
    decided_round: Option<Round>,

    // Qbft::new
    decided_round: None,

    // local commit-quorum success path
    self.decided_round = Some(round);
    self.completed = Some(Completed::Success(hash));

    // received_decided success path
    self.decided_round = Some(wrapped_msg.qbft_message.round.into());
    self.completed = Some(Completed::Success(wrapped_msg.qbft_message.root));

    pub fn decided_round(&self) -> Option<Round> {
        self.decided_round
    }

Then the `instance.rs` finish call can use:

    let terminal_round = initialized
        .qbft
        .decided_round()
        .unwrap_or_else(|| initialized.qbft.get_round());
    observer.finish(outcome, u64::from(terminal_round));

The fallback keeps timeout semantics unchanged: `MaxRoundTimeout` still records the local round where the instance timed out. Only the `Decided` value changes when completion came from a cross-round decided certificate. This also fixes the span field and completion event, since both are populated by `ProposerObserver::finish`, and avoids cloning or decoding the aggregated commit. This should not mutate `current_round`; it is only terminal outcome metadata.

anchor/validator_store/src/lib.rs
jnhsigmap: this was a good call - handled in https://github.com/sigp/anchor/pull/1105/commits/13d11cbe3f6c1329e2eedcb7c44acf667d6ea329.

- Relocated `compute_handoff_budget_ms` into `qbft_manager::decide_instance`.
- Duty-slot anchored which fixes the stated late-start bug where budget from following slot is reported.
- Deleted the threaded `handoff_budget_ms: Option<u64>` from `decide_instance`/`ConsensusDecider`/`impl`/mock and 5 call sites (includes non-proposer paths).

anchor/qbft_manager/src/instance.rs
jnhsigmap: Implemented in https://github.com/sigp/anchor/commit/13d11cbe3f6c1329e2eedcb7c44acf667d6ea329
- Helper functions retrieving count and sum of metric values.
- `tokio::sync::Mutex` added to tests that measure metric values.
- `test_proposer_instance_max_round_timeout_runs_observer` checks that budget count is incremented and correct magnitude of budget increase occurs.
- New test `test_proposer_instance_channel_closed_runs_observer` for `ChannelClosed` path where `message_tx` is dropped:
  - Outcome counter incremented for `channel_closed`.
  - Decided round does not record.
  - Duration histogram does not record.

anchor/qbft_manager/src/instance.rs
jnhsigmap: Cool, in https://github.com/sigp/anchor/commit/13d11cbe3f6c1329e2eedcb7c44acf667d6ea329 - we are not tracing how starting state is reached, however `start_round` and `start_state` are recorded in `qbft_instance` loop and provided to `ProposerObserver` span and start event. `start_state` is a typed `InstanceStateKind`.

anchor/qbft_manager/src/tests/timeout_tests.rs
jnhsigmap: as part of above - https://github.com/sigp/anchor/pull/1105#discussion_r3556254193.
Grouped in with proposer metric tests in https://github.com/sigp/anchor/commit/13d11cbe3f6c1329e2eedcb7c44acf667d6ea329.

anchor/qbft_manager/src/instance.rs
jnhsigmap: In https://github.com/sigp/anchor/commit/13d11cbe3f6c1329e2eedcb7c44acf667d6ea329. Agreed that this widens the scope of this change to slightly alter state machine behaviour by adding `decided_round` wiring to `Qbft` and records this on success paths. Implemented as stated and used instead of `get_round()` (which is now a fallback) on the `finish` call.

shane-moore (approved):
LGTM man!  `unstable` branch is unfrozen now, so good to merge this whenver you feel good about it

jnhsigmap:
thanks @shane-moore, I'm not really happy with it in its current form. I think it needs more thought before we go ahead.

Once we're done sprinting on `epbs` I'll come back to this and pull it apart again with those new changes in mind. Going to put it back to Draft until then :+1:

PR #1106, opened 2026-07-02
feat: add Role::ProposerPreferences with message-validator fork-gate

### LINKED ISSUE #1062: feat: add Role::ProposerPreferences with message-validator fork-gate
## Goal

Add `Role::ProposerPreferences` (wire byte `[8, 0, 0, 0]`, matches SIP-94 `BNRoleProposerPreferences = 8`) in `anchor/common/ssv_types/src/msgid.rs`, and `PartialSignatureKind::ProposerPreferences = 8` in `anchor/common/ssv_types/src/partial_sig.rs` (next free after `PTCAttester = 7`, added in #1080; matches SIP-94 `ProposerPreferencesPartialSig = 8`). Wire the message-validator rules for the role per SIP-94 §7: fork gate, proposal-slot semantics (earliness allowance, tight lateness, slot-advance exemption), proposer-assignment check, distinct-signing-root dedup, and duty limit, so the workspace compiles in one shot. ProposerPreferences is non-QBFT, so no `qbft_manager` stub or routing is required, though `receive_data`'s two exhaustive non-QBFT match arms must gain the new variant to compile.

## Context

Per SIP-94 §5 ([ssvlabs/SIPs#94](https://github.com/ssvlabs/SIPs/pull/94)), ProposerPreferences is a validator-scoped, non-QBFT duty added at the Gloas / EIP-7732 / ePBS fork. Each operator signs `ProposerPreferences { dependent_root, proposal_slot, validator_index, fee_recipient, target_gas_limit }` under `DOMAIN_PROPOSER_PREFERENCES` with its BLS share; partial signatures go through one collection round, and the reconstructed `SignedProposerPreferences` is published over the new `proposer_preferences` gossip topic.

**Message slot = `proposal_slot` (SIP-94 §5/§7; supersedes the earlier emission-slot design in this issue's history).** The `PartialSignatureMessages` envelope carries the duty's `proposal_slot` itself: one runner per proposal slot, matching ssv-spec's `validatePartialSigMsgForSlot` (`msg.Slot == duty.DutySlot()`). go-ssv parity: the reference runner keys one sub-runner per proposal slot and stamps it on the wire (`protocol/v2/ssv/runner/proposer_preferences.go:40,80,199`: "duty.Slot is the proposal slot the preference targets, which is also the slot carried on the [wire]"). Because that slot is legitimately up to the proposer lookahead in the future at emission time, and because a signer holds its whole lookahead of proposal slots concurrently, four validation rules deviate from the non-committee defaults (each with its go-ssv parity site on `epbs-gloas`):

1. **Earliness allowance.** Role-8 messages may arrive up to `(1 + MIN_SEED_LOOKAHEAD) * slots_per_epoch` slots before their slot (the proposer lookahead: current epoch plus `MIN_SEED_LOOKAHEAD = 1`); every other role keeps the default no-future rule (50ms clock tolerance). Parity: `message/validation/common_checks.go:52-58` (`earlySlotAllowance`), `const.go:32` (`proposerPreferencesEarlyEpochs = 2`).
2. **Tight lateness.** TTL = `LATE_SLOT_ALLOWANCE` (2) slots past `proposal_slot` (plus the 3s margin), NOT the epoch-class arm: a preference for a slot already behind us is a replay. Parity: `common_checks.go:68-72`.
3. **Slot-advance exemption.** The "signer already advanced to a later slot" rule does not apply to role 8: a lower proposal slot is a concurrent lookahead duty, not a stale one; its past bound is rule 2. Parity: `common_checks.go:23-25` (`monotonicSlotRole`), call site `partial_validation.go:177-185`.
4. **Proposer-assignment check.** A role-8 message requires a proposer duty for the validator at `proposal_slot`, checked only once the slot-epoch's proposer duties are known locally (a not-yet-fetched epoch is tolerated); absence is `NoDuty` (Ignore). Parity: `common_checks.go:195-200` (`IsEpochSet`-guarded `ValidatorDuty`). Anchor's `DutiesProvider` already exposes both needed methods (`is_epoch_known_for_proposers`, `is_validator_proposer_at_slot`), used today by the `Role::Proposer` arm of `validate_beacon_duty` (`message_validator/src/lib.rs:787-820`).

**Dedup: up to 4 distinct signing roots per (signer, `proposal_slot`).** Base rule stays 1 pre-consensus partial sig per (signer, slot); role 8 gains the bounded exception agreed on the SIP thread: a NEW `signing_root` is accepted up to 4 per (MessageId, operator, slot); an exact repeat of a seen root is a true duplicate. Distinct roots at one proposal slot arise when the preference inputs change between emissions for the same slot, especially a `dependent_root` shift under reorg (the SIP-94 §5 re-emission trigger); `fee_recipient` / `target_gas_limit` config changes across operator restarts produce new roots the same way. The epoch-boundary batch never shares this key since each proposal slot is its own (signer, slot). Not yet implemented in go-ssv (its `seen_msg_types.go:83-84` still enforces 1 per slot, which drops honest re-emission; the fix is agreed on the SIP thread and lands on both sides).

**Fork gate with no special case.** Role 8 messages with `epoch(msg.slot) < GLOAS_FORK_EPOCH` are rejected, mirroring the `PTCAttester` gate (`lib.rs:876-885`, `RoleNotActiveBeforeEthFork { minimum_fork: ForkName::Gloas }`, evaluated via the `ChainSpec` per #1090). SIP-94 §5's required pre-fork emission passes this gate with no exception: those messages carry post-fork `proposal_slot` values by construction. (This resolves the earlier note in this issue about pre-fork emission needing a fork-gating revisit: under proposal-slot stamping it does not.) `validate_role_for_fork` is `if`-based, not an exhaustive match, so the gate must be added by hand.

The new role is wire-incompatible with pre-Gloas operators: they decode the byte as `Err(NoMatchingVariant)`. ProposerPreferences is validator-scoped, not committee-scoped: routing is by `(Role, ValidatorPubKey)`, mirroring `Role::ValidatorRegistration` / `Role::VoluntaryExit` / `Role::PTCAttester`. It is non-QBFT (single-round partial-sig collection), so the partial-sig kind is the new `ProposerPreferences` variant, not `PostConsensus`, and any consensus message carrying the role is rejected (`UnexpectedConsensusMessage`, falls out of `is_qbft_role()`; go parity: `consensus_validation.go:164-169`).

Adding the enum variants breaks exhaustive matches in `ssv_types` (`max_round`, `duty_executor`), `message_validator` (committee_info dispatch, `message_lateness`, `duty_limit`, `partial_signature_type_matches_role`, message counts), and `qbft_manager` (`receive_data`'s two non-QBFT error arms), so those all live in this issue to keep the workspace green.

## Suggested approach

`ssv_types/src/msgid.rs`:
- Add `Role::ProposerPreferences` with wire byte `[8, 0, 0, 0]`, `is_committee_role() == false`, `max_round() == None` (non-QBFT; `is_qbft_role()` must return `false`).
- Extend `From<Role> for [u8;4]` and `TryFrom<&[u8]>`, and the validator-scoped arm of `duty_executor()` (`msgid.rs:159-166`).

`ssv_types/src/partial_sig.rs`:
- Add `PartialSignatureKind::ProposerPreferences = 8`; extend `TryFrom<u64>`; bump the two invalid-variant test sentinels to 9; add the variant to the four enumerated test arrays the compiler does not force (`ssz_roundtrip_all_variants`, `ssz_byte_layout`, `try_from_u64_all_valid_values`, `tree_hash_differs_per_variant`).

`message_validator`:
- `validate_role_for_fork` (`lib.rs:849-888`): add the role-8 Gloas gate mirroring the `PTCAttester` block (`lib.rs:876-885`). Evaluated on the message slot (= `proposal_slot`); no pre-fork exception (see Context).
- `validate_slot_time` / `message_earliness` (`lib.rs:904-938`): add a role-scoped earliness allowance, mirroring go's shape (`earliness > CLOCK_ERROR_TOLERANCE + early_slot_allowance(role)`): `(1 + spec.min_seed_lookahead) * slots_per_epoch` slots' worth of time for `Role::ProposerPreferences`, zero for every other role. `validation_context.spec` is already threaded in (#1090).
- `message_lateness` (`lib.rs:943-967`): give `ProposerPreferences` its own arm with `ttl = LATE_SLOT_ALLOWANCE` (neither the `1 + LATE_SLOT_ALLOWANCE` proposer arm nor the epoch arm).
- Slot-advance check (`partial_signature.rs:192-200`): exempt the role, e.g. extend the guard to `!role.is_committee_role() && role != Role::ProposerPreferences` (or introduce a `monotonic_slot_role()` helper on `Role` mirroring go's `monotonicSlotRole`).
- `validate_beacon_duty` (`lib.rs:778-841`): add a `Role::ProposerPreferences` arm: if `duty_provider.is_epoch_known_for_proposers(epoch)` and `!duty_provider.is_validator_proposer_at_slot(slot, validator_index)`, return `NoDuty` (Ignore). Same provider calls as the `Role::Proposer` arm, minus the RANDAO first-slot tolerance (there is no RANDAO here); keep the epoch-known guard so messages arriving before the duties fetch are tolerated.
- `duty_limit` (`lib.rs:1008-1064`): `Role::ProposerPreferences => Ok(Some(slots_per_epoch))` (a validator proposes at most once per slot). Parity: `common_checks.go:149-152`. NOT the `Some(2)` registration arm. Note the counter counts distinct message slots per (signer, epoch-of-message-slot), so under proposal-slot stamping the count lands in the proposal slot's epoch, which is the bound's natural epoch.
- Per-role ring sizing (`get_duty_state`, `lib.rs:527-539`): size the `DutyState` circular buffer by role: default stays `slots_per_epoch * 2`; for `Role::ProposerPreferences` add the lookahead span on top, `(1 + spec.min_seed_lookahead) * slots_per_epoch + slots_per_epoch * 2`. The role is already extracted before the call (`lib.rs:449-452`), and the map is keyed by MessageId (role-scoped), so only role-8 states grow. This is REQUIRED, not hygiene: the role-8 live window is lateness (2) + earliness (`2 * slots_per_epoch`) = 67 slots on mainnet, wider than the 64-slot default ring, so a validly accepted message for slot `S + 64` lands in bucket `S % 64` and `set_signer_state_for_first_round` (`duty_state.rs:236-237`) evicts live slot `S`'s state, resetting its dedup caps; the lateness rule cannot prevent this because the eviction comes from an accepted future slot, not an old one. Parity: `validation.go:383-395` (`storedSlotCount`, role-scoped ring = lookahead + default), collision test `proposer_preferences_test.go:83-100`.
- Distinct-root dedup (`message_counts.rs` + `duty_state.rs`): do NOT route `ProposerPreferencesPartialSig` through the shared `pre_consensus` counter (`message_counts.rs:63-76`, cap 1, which would drop every honest re-emission). Instead track seen `signing_root`s per (MessageId, operator, slot) on `SignerState` (`duty_state.rs:269-280`, next to `seen_signers` / `proposal_hash`): an exact-duplicate root is rejected like today's duplicates (`InvalidPartialSignatureTypeCount`-class, Reject); a new distinct root is accepted until the set holds 4, after which further new roots are dropped with a NEW dedicated failure variant (e.g. `TooManyDistinctSigningRoots`) added to the Ignore arm of `From<&ValidationFailure> for MessageAcceptance` (`lib.rs:233-260`); unlisted variants default to Reject there, and Ignore is deliberate: a 5th distinct root is not provably malicious under extreme reorg churn. The `signing_root` is read from the packet's single `PartialSignatureMessage` entry (packet cap 1 below). Cap 4 per the SIP-94 §7 rule agreed with the go side.
- `partial_signature_type_matches_role` (`partial_signature.rs:145-168`): `Role::ProposerPreferences => kind == PartialSignatureKind::ProposerPreferences`. Parity: `partial_validation.go:391-392`.
- Per-validator packet arm (`partial_signature.rs:318-330`): include `Role::ProposerPreferences` (max 1 message entry per packet, `TooManyPartialSignatureMessages`).
- `committee_info` dispatch (`lib.rs:462-485`) and the test `duty_executor` helper: include the role in the validator-scoped arm.
- `validate_beacon_duty`'s `Role::Proposer` and `Role::SyncCommittee` arms, and everything else about existing roles: unchanged.

`qbft_manager`: `receive_data`'s two exhaustive non-QBFT error arms (`Role::ValidatorRegistration | Role::VoluntaryExit | Role::PTCAttester`, both executor branches) gain `Role::ProposerPreferences`, still mapping to `QbftError::InconsistentMessageId`.

## Acceptance criteria

- `Role::ProposerPreferences` byte `[8, 0, 0, 0]`; roundtrip covered by `role_roundtrip_all_variants` via `strum::EnumIter`.
- `is_committee_role() == false`, `max_round() == None`, `is_qbft_role() == false`; role added to the non-QBFT array in `role_qbft_classification_is_pinned`.
- `PartialSignatureKind::ProposerPreferences as u64 == 8`; SSZ `[8, 0, 0, 0, 0, 0, 0, 0]` roundtrip.
- Kind-role: `ProposerPreferences` kind accepted only for the role; any other kind → `PartialSignatureTypeRoleMismatch`.
- Consensus message with the role → `UnexpectedConsensusMessage` (add the role to `test_consensus_message_for_non_consensus_role`).
- Fork gate: pre-Gloas message slot → `RoleNotActiveBeforeEthFork { minimum_fork: ForkName::Gloas }`; at/post-activation accepted. Driven via `ChainSpec` (`spec_with_gloas` helper from #1090).
- Earliness: a message whose slot is up to `(1 + min_seed_lookahead) * slots_per_epoch` slots in the future is accepted; one further out → `EarlySlotMessage`. Every other role keeps the no-future rule (regression-pin one).
- Lateness: accepted through `slot_start(proposal_slot + 2) + margin`; later → `LateSlotMessage`.
- Slot-advance: after a role-8 message for a later slot, one for an earlier (still-timely) proposal slot is accepted; the same sequence for a monotonic role still fails (`SlotAlreadyAdvanced`).
- Duty check: assigned proposer → accepted; known epoch + not assigned → `NoDuty`; unknown epoch → accepted.
- Dedup: 4 packets with distinct `signing_root`s at one (signer, slot) all accepted; an exact-duplicate root → Reject; a 5th distinct root → Ignore (the new `TooManyDistinctSigningRoots`-style variant, present in the Ignore arm). `ProposerPreferences` does NOT touch the shared `pre_consensus` counter (registration-at-same-slot unaffected). Peer-aware classification of the duplicate case (same-peer Reject vs relayed Ignore, the SIP-94 §7 two-tier handling) is handled in #1131.
- Ring sizing: two accepted role-8 messages for the current slot and for `current + 2 * slots_per_epoch` (the earliness bound) retain distinct per-slot signer states, and the earlier slot's dedup state survives the later write (mirrors go's `TestProposerPreferencesRingAvoidsLookaheadCollision`); other roles keep the `slots_per_epoch * 2` ring.
- `duty_limit(Role::ProposerPreferences) == Ok(Some(slots_per_epoch))`.
- Packet cap: 2 message entries → `TooManyPartialSignatureMessages { limit: 1 }`.
- `cargo check --workspace` green; pre-existing tests unaffected.

## Risks

- **Wire-byte / discriminant collisions.** Role 8 and kind 8 must match ssv-spec at the SIP's cited pin and go-ssv `epbs-gloas`; verify both before commit (an earlier SIP revision had ProposerPreferences kind at 7, which would collide with `PTCAttester = 7`).
- **Earliness allowance scope.** The allowance must be strictly role-scoped; widening the default earliness window for other roles would reopen the future-slot replay surface the rule exists to close.
- **Ring eviction if the per-role sizing is skipped.** With the default `2 * slots_per_epoch` ring, the role-8 live window (67 slots on mainnet) exceeds the buffer and a validly accepted future-slot message evicts a live slot's state, resetting its dedup caps (duplicate amplification; the 4-root cap defeated by replay). Lateness cannot prevent it: the eviction comes from an accepted future slot. The per-role sizing bullet above closes this; the root set itself stays bounded (4 x 32-byte roots per signer-slot).
- **Duty-limit classification.** Anchor classifies duty-limit breaches Reject (`ExcessiveDutyCount` is not in the Ignore arm, `lib.rs:233-260`) while go-ssv and SIP-94 §7 classify them Ignore; tracked as a separate alignment issue in #1118, out of scope here.

petarjuki7:
Closes #1062 (ePBS Proposer Preferences duty milestone, SIP-94 §5).

Wire surface + inbound message validation for the new ProposerPreferences duty:

- `Role::ProposerPreferences` (wire byte `[8,0,0,0]`) and `PartialSignatureKind::ProposerPreferences = 8`; validator-scoped, non-QBFT.
- message-validator: Gloas fork-gate, TTL, duty limit, partial-sig kind binding, per-packet bound, and a per-slot signing-root dedup (accepts a validator's multiple lookahead preference roots, rejects exact-duplicate resends and anti-spam overflow).
- `qbft_manager` non-QBFT arms.

Sign path and service wiring land in follow-up milestone issues.

### DIFF
--- a/anchor/message_validator/src/duty_state.rs
+++ b/anchor/message_validator/src/duty_state.rs
@@ -7,8 +7,9 @@ use ssv_types::{
     CommitteeId, Epoch, OperatorId, Slot,
     consensus::{QbftMessage, QbftMessageType},
     message::SignedSSVMessage,
-    partial_sig::PartialSignatureMessages,
+    partial_sig::{PartialSignatureKind, PartialSignatureMessages},
 };
+use types::Hash256;
 
 use crate::{FIRST_ROUND, ValidationFailure, message_counts::MessageCounts};
 // duty_state.rs
@@ -96,6 +97,30 @@ impl DutyState {
                 )
             }
         };
+
+        // ProposerPreferences-specific per-slot signing-root dedup (not the shared pre_consensus
+        // counter): a validator legitimately signs multiple distinct preference roots at one send
+        // slot (the current + next-epoch lookahead batch), so we track the distinct roots.
+        if partial_signature_messages.kind == PartialSignatureKind::ProposerPreferences {
+            let root = partial_signature_messages
+                .messages
+                .first()
+                .ok_or(ValidationFailure::NoPartialSignatureMessages)?
+                .signing_root;
+            if signer_state.seen_preferences.len() as u64 >= 2 * slots_per_epoch {
+                return Err(ValidationFailure::InvalidPartialSignatureTypeCount {
+                    got: format!(
+                        "proposer-preferences distinct roots exceed cap {}",
+                        2 * slots_per_epoch
+                    ),
+                });
+            }
+            if !signer_state.seen_preferences.insert(root) {
+                return Err(ValidationFailure::DuplicatedMessage {
+                    got: format!("proposer-preferences root {root:?}"),
+                });
+            }
+        }
 
         // Record the partial signature (only once)
         signer_state
@@ -277,6 +302,9 @@ pub(crate) struct SignerState {
     pub(crate) proposal_hash: Option<[u8; 32]>,
     /// A set of CommitteeIds indicating which committees have already been seen.
     seen_signers: HashSet<CommitteeId>,
+    /// Distinct ProposerPreferences signing roots already seen for this
+    /// (MessageId, operator, slot), used to reject exact-duplicate resends.
+    seen_preferences: HashSet<Hash256>,
 }
 
 impl SignerState {
@@ -288,6 +316,7 @@ impl SignerState {
             message_counts: MessageCounts::default(),
             proposal_hash: None,
             seen_signers: HashSet::new(),
+            seen_preferences: HashSet::new(),
         }
     }
 
--- a/anchor/message_validator/src/duty_state.rs
+++ b/anchor/message_validator/src/duty_state.rs
@@ -132,6 +132,15 @@ impl DutyState {
                 .first()
                 .ok_or(ValidationFailure::NoPartialSignatureMessages)?
                 .signing_root;
+            // Duplicate identity takes precedence over capacity: an already-seen root is a
+            // Reject-class DuplicatedMessage even once the distinct-root set is full, whereas only
+            // a NEW root beyond the cap is the Ignore-class TooManyDistinctSigningRoots (SIP-94
+            // §7).
+            if signer_state.seen_preferences.contains(&root) {
+                return Err(ValidationFailure::DuplicatedMessage {
+                    got: format!("proposer-preferences root {root:?}"),
+                });
+            }
             if signer_state.seen_preferences.len() >= MAX_PROPOSER_PREFERENCES_DISTINCT_ROOTS {
                 return Err(ValidationFailure::TooManyDistinctSigningRoots {
                     got: format!(
@@ -140,11 +149,7 @@ impl DutyState {
                     ),
                 });
             }
-            if !signer_state.seen_preferences.insert(root) {
-                return Err(ValidationFailure::DuplicatedMessage {
-                    got: format!("proposer-preferences root {root:?}"),
-                });
-            }
+            signer_state.seen_preferences.insert(root);
         }
 
         // Record the partial signature (only once)
--- a/anchor/message_validator/src/duty_state.rs
+++ b/anchor/message_validator/src/duty_state.rs
@@ -21,6 +21,29 @@ use crate::{FIRST_ROUND, ValidationFailure, message_counts::MessageCounts};
 //  - SignerState: The state of a signer at a particular slot, including message counts and proposal
 //    data.
 
+/// Maximum distinct `ProposerPreferences` signing roots accepted per
+/// (`MessageId`, operator, `proposal_slot`).
+///
+/// A validator can legitimately sign several distinct roots for one `proposal_slot` when its
+/// preference inputs change between emissions — chiefly a `dependent_root` shift under reorg (the
+/// SIP-94 §5 re-emission trigger), and also `target_gas_limit` / `fee_recipient` config changes
+/// across operator restarts. Per SIP-94 §7 the cap is policy headroom for a few realistic
+/// reorg-driven corrections while bounding spam — not a safety/consensus bound. That is why
+/// exceeding it is an Ignore (`TooManyDistinctSigningRoots`), whereas a repeat of an already-seen
+/// root stays a Reject-class duplicate.
+const MAX_PROPOSER_PREFERENCES_DISTINCT_ROOTS: usize = 4;
+
+/// Test-only, crate-visible mirror of the private cap so pipeline tests in sibling modules
+/// (e.g. `partial_signature`) can reference the real value instead of hardcoding `4`. The
+/// compile-time assertion below makes the mirror impossible to drift from production.
+#[cfg(test)]
+pub(crate) const MAX_PROPOSER_PREFERENCES_DISTINCT_ROOTS_FOR_TEST: usize =
+    MAX_PROPOSER_PREFERENCES_DISTINCT_ROOTS;
+#[cfg(test)]
+const _: () = assert!(
+    MAX_PROPOSER_PREFERENCES_DISTINCT_ROOTS_FOR_TEST == MAX_PROPOSER_PREFERENCES_DISTINCT_ROOTS
+);
+
 /// DutyState manages the state for duty validation across operators and slots
 pub(crate) struct DutyState {
     /// Tracks the duty state for an operator
@@ -99,19 +122,21 @@ impl DutyState {
         };
 
         // ProposerPreferences-specific per-slot signing-root dedup (not the shared pre_consensus
-        // counter): a validator legitimately signs multiple distinct preference roots at one send
-        // slot (the current + next-epoch lookahead batch), so we track the distinct roots.
+        // counter): the envelope slot is the duty's proposal_slot, and a validator may sign several
+        // distinct roots for one proposal_slot as its preference inputs change between emissions
+        // (chiefly a dependent_root shift under reorg). Track the distinct roots per
+        // (MessageId, operator, proposal_slot), capped at MAX_PROPOSER_PREFERENCES_DISTINCT_ROOTS.
         if partial_signature_messages.kind == PartialSignatureKind::ProposerPreferences {
             let root = partial_signature_messages
                 .messages
                 .first()
                 .ok_or(ValidationFailure::NoPartialSignatureMessages)?
                 .signing_root;
-            if signer_state.seen_preferences.len() as u64 >= 2 * slots_per_epoch {
-                return Err(ValidationFailure::InvalidPartialSignatureTypeCount {
+            if signer_state.seen_preferences.len() >= MAX_PROPOSER_PREFERENCES_DISTINCT_ROOTS {
+                return Err(ValidationFailure::TooManyDistinctSigningRoots {
                     got: format!(
-                        "proposer-preferences distinct roots exceed cap {}",
-                        2 * slots_per_epoch
+                        "proposer-preferences distinct roots exceed cap \
+                         {MAX_PROPOSER_PREFERENCES_DISTINCT_ROOTS}"
                     ),
                 });
             }
--- a/anchor/message_validator/src/lib.rs
+++ b/anchor/message_validator/src/lib.rs
@@ -191,6 +191,9 @@ pub enum ValidationFailure {
     InvalidPartialSignatureTypeCount {
         got: String,
     },
+    TooManyDistinctSigningRoots {
+        got: String,
+    },
     TooManyPartialSignatureMessages {
         got: usize,
         limit: usize,
@@ -251,9 +254,8 @@ impl From<&ValidationFailure> for MessageAcceptance {
             | ValidationFailure::ValidatorIndexMismatch
             | ValidationFailure::TooManyDutiesPerEpoch
             | ValidationFailure::NoDuty
-            | ValidationFailure::EstimatedRoundNotInAllowedSpread { .. } => {
-                MessageAcceptance::Ignore
-            }
+            | ValidationFailure::EstimatedRoundNotInAllowedSpread { .. }
+            | ValidationFailure::TooManyDistinctSigningRoots { .. } => MessageAcceptance::Ignore,
             _ => MessageAcceptance::Reject,
         }
     }
@@ -500,7 +502,7 @@ impl<S: SlotClock + 'static, D: DutiesProvider> Validator<S, D> {
 
         drop(network_state);
 
-        let mut duty_state = self.get_duty_state(ssv_message.msg_id(), self.slots_per_epoch);
+        let mut duty_state = self.get_duty_state(ssv_message.msg_id(), role, self.slots_per_epoch);
 
         let validation_context = ValidationContext {
             signed_ssv_message,
@@ -528,12 +530,23 @@ impl<S: SlotClock + 'static, D: DutiesProvider> Validator<S, D> {
     fn get_duty_state(
         &self,
         message_id: &MessageId,
+        role: Role,
         slots_per_epoch: u64,
     ) -> RefMut<'_, MessageId, DutyState> {
         self.duty_state_map
             .entry(message_id.clone())
             .or_insert_with(|| {
-                let stored_slot_count = slots_per_epoch * 2; // Store last two epochs
+                // Default: keep the last two epochs. ProposerPreferences also spans the proposer
+                // lookahead into the future (its envelope slot is a future proposal_slot), so its
+                // ring must cover lookahead + default; otherwise a validly accepted future-slot
+                // write would evict a live slot's dedup state (the ring is indexed by slot % len).
+                let stored_slot_count = match role {
+                    Role::ProposerPreferences => {
+                        (1 + self.spec.min_seed_lookahead.as_u64()) * slots_per_epoch
+                            + slots_per_epoch * 2
+                    }
+                    _ => slots_per_epoch * 2,
+                };
 
                 DutyState::new(stored_slot_count as usize)
             })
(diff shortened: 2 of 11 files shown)


shane-moore:
Heads-up: #1062, which this PR implements, was reworked after the ProposerPreferences slot semantics settled on the SIP side (the envelope slot is now the duty's `proposal_slot`, not the send slot). Summary of what changes for this PR: https://github.com/sigp/anchor/issues/1062#issuecomment-4899949725.

anchor/message_validator/src/lib.rs
```diff
@@ -937,6 +987,27 @@ fn message_earliness(
         .unwrap_or_default())
 }
 
+/// Extra future-slot tolerance (on top of `CLOCK_ERROR_TOLERANCE`) allowed for a role's message
+/// slot. Only `ProposerPreferences` is non-zero: its envelope slot is the duty's future
+/// `proposal_slot`, so the whole proposer lookahead (current epoch + `min_seed_lookahead`) ahead of
+/// that slot must be accepted. Every other role keeps the strict no-future rule.
+fn early_slot_allowance(validation_context: &ValidationContext<impl SlotClock>) -> Duration {
+    match validation_context.role {
+        Role::ProposerPreferences => {
```
shane-moore, line 1002: **[P2] Keep this future duty slot out of doppelganger freshness checks**

This allowance is correct for SIP-94, but it breaks an existing downstream invariant. `NetworkMessageReceiver` passes every validated message to `OperatorDoppelgangerService`, whose `extract_message_slot` treats every partial-signature envelope slot as message-creation time. It then reports a twin whenever a single-signer packet from our operator has `msg_slot > startup_slot`.

For this role, `msg_slot` is the future `proposal_slot`. For example, an old process can sign a valid packet for slot 120, stop at slot 101, and be replaced at slot 102. Replaying the original packet during startup still passes this future-slot allowance, but doppelganger detection evaluates `120 > 102` and aborts the replacement even though the processes never overlapped.

Please make `OperatorDoppelgangerService::is_doppelganger` role-aware and exclude `Role::ProposerPreferences` from slot-based freshness evidence. Add a regression test using `ValidatedSSVMessage::PartialSignatureMessages` with our operator ID and `proposal_slot > startup_slot`, asserting that it does not detect a twin. Keep the existing current-slot QBFT positive test unchanged. There is no authenticated emission time on this wire object, so adjusting the slot threshold cannot distinguish an old packet from a live twin.

anchor/message_validator/src/duty_state.rs
```diff
@@ -97,6 +121,32 @@ impl DutyState {
...
+        // distinct roots for one proposal_slot as its preference inputs change between emissions
+        // (chiefly a dependent_root shift under reorg). Track the distinct roots per
+        // (MessageId, operator, proposal_slot), capped at MAX_PROPOSER_PREFERENCES_DISTINCT_ROOTS.
+        if partial_signature_messages.kind == PartialSignatureKind::ProposerPreferences {
+            let root = partial_signature_messages
+                .messages
+                .first()
+                .ok_or(ValidationFailure::NoPartialSignatureMessages)?
+                .signing_root;
+            if signer_state.seen_preferences.len() >= MAX_PROPOSER_PREFERENCES_DISTINCT_ROOTS {
```
shane-moore, line 144: **[P3] Check duplicate membership before the distinct-root cap**

Once `seen_preferences` contains four roots, this branch runs before membership is tested. For example, after accepting roots A, B, C, and D for the same operator and `proposal_slot`, another packet whose `signing_root` is A sees `len() == 4` and returns `TooManyDistinctSigningRoots`/Ignore. It never reaches `insert(A)`, which would identify the already-seen root. #1062 and this function's documented contract require an already-seen root to be `DuplicatedMessage`/Reject; only a new root beyond the cap should be Ignore.

Please make duplicate identity take precedence over capacity, for example:

```rust
if signer_state.seen_preferences.contains(&root) {
    return Err(ValidationFailure::DuplicatedMessage {
        got: format!("proposer-preferences root {root:?}"),
    });
}

if signer_state.seen_preferences.len() >= MAX_PROPOSER_PREFERENCES_DISTINCT_ROOTS {
    return Err(ValidationFailure::TooManyDistinctSigningRoots {
        got: format!(
            "proposer-preferences distinct roots exceed cap \
             {MAX_PROPOSER_PREFERENCES_DISTINCT_ROOTS}"
        ),
    });
}

signer_state.seen_preferences.insert(root);
```

Please also extend `test_proposer_preferences_root_cap_enforced`: after filling the set with four distinct roots, send another packet using the first root and assert `DuplicatedMessage` plus `MessageAcceptance::Reject`. Keep the fifth-new-root assertion as `TooManyDistinctSigningRoots` plus Ignore.

anchor/message_validator/src/partial_signature.rs
```diff
@@ -1876,78 +1905,1217 @@ mod tests {
...
+        // survived), rather than being reset by a ring-index collision.
+        let committee_info = create_committee_info(FOUR_NODE_COMMITTEE);
+        let (private_key, public_key) = generate_test_key_pair();
+        let map =
+            create_operator_pub_keys(committee_info.committee_members.clone(), vec![public_key]);
+        let signer_id = OperatorId(1);
+
+        // Production ProposerPreferences ring size:
+        // (1 + min_seed_lookahead) * spe + spe * 2, with min_seed_lookahead = 1.
+        let ring = ((1 + 1) * SLOTS_PER_EPOCH_TEST + 2 * SLOTS_PER_EPOCH_TEST) as usize;
```
shane-moore: **[P3] Exercise the production ring-size selector in this test**

The production mapping is correct, but this test recreates the expected role-8 formula and passes it directly to `DutyState::new`. It proves that a 128-slot ring avoids the collision, not that `Validator::get_duty_state` selects that ring. I verified the gap by changing only the production role-8 arm to the 64-slot default in a temporary copy: this test still passed.

Please extract the production calculation into a small helper used by both `get_duty_state` and this test, for example:

```rust
fn stored_slot_count(role: Role, slots_per_epoch: u64, spec: &ChainSpec) -> usize {
    let default = slots_per_epoch * 2;
    match role {
        Role::ProposerPreferences => {
            ((1 + spec.min_seed_lookahead.as_u64()) * slots_per_epoch + default) as usize
        }
        _ => default as usize,
    }
}
```

Have this collision test call the helper, and add a direct assertion that a representative non-role-8 role still gets `2 * slots_per_epoch`. This keeps the existing collision scenario while pinning the production selection required by #1062.

This is a non-blocking regression-coverage recommendation, not a current runtime defect.

shane-moore (approved):
Looks good, thanks for addressing the feedback!

Issue #37, opened 2024-10-23
Testing 


jking-aus:
# Testing 
This is a major issue that tracks completion of testing issues for anchor

## Issues
- [ ] #39 
- [ ] #40 
- [x] #12 
- [x] #36 
- [x] #1

Issue #30, opened 2024-10-23
Logging


AgeManning:
We need to set up a custom logging formatter for our tracing library and decide how our logs should look.

Issue #29, opened 2024-10-23
Release CI


AgeManning:
We need to create a release CI. 

There is an example in lighthouse. This needs to create multiple binaries for our supported architectures, sign them and generate release notes.

PR #1112, opened 2026-07-06
ci: bump ssv-mini pin to fix local-testnet

shane-moore:

Two **required** checks (`.github/mergify.yml`) were red repo-wide and **circularly blocked** each other:

1. **`run-local-testnet`** — red on every branch since ~2026-07-02. It builds a contract-deployer image from the pinned `ssvlabs/ssv-mini`, and that pin (`ae910a2d`) cloned a personal fork since deleted:
   ```
   git clone --branch foundry https://github.com/Zacholme7/ssv-network .
   fatal: could not read Username for 'https://github.com'  (exit 128)
   ```
2. **`check-code`** — red on every PR since 2026-07-06 via `cargo audit`: [RUSTSEC-2026-0204](https://rustsec.org/advisories/RUSTSEC-2026-0204) flags `crossbeam-epoch` 0.9.18.

Each PR that fixed one failed the other, so neither could merge without a maintainer override. This PR carries **both** fixes so it goes green on its own (supersedes #1116).

**The local-testnet fix is more than a pin bump.** Upstream ssv-mini's v2.0.0 migration (#30) fixed the deleted-fork clone *and* removed on-chain validator pre-registration, delegating it to a private "aetheria" executor we don't run. So bumping the pin alone left the anchor operators with **zero validators** — no consensus, no proposals. Registration is restored by a new default-off `pre_register_validators` flag added upstream in [ssvlabs/ssv-mini#37](https://github.com/ssvlabs/ssv-mini/pull/37).

## Change Overview

CI config plus one transitive lockfile bump:

- **`local-testnet.yml`** — pin ssv-mini to the fork branch behind #37 (the flag + the v2.0.0 payable-registration fix). `TEMPORARY`: swap to the merged `ssvlabs/ssv-mini` SHA once #37 lands.
- **`params.yaml`** — reconcile to the v2.0.0 schema and the validator layout: `images.deployer` (was `foundry`), `preregistered_validator_count: 74`, `use_static_keys: true`, `pre_register_validators: true`, and `validator_count: 32` so 64 lighthouse VCs + 10 pre-registered anchor validators = 74/74 participation (needed for the finality + 95% attestation checks).
- **`test_anchor_proposal.yaml`** — `blockCount` 50 → 3 (anchor runs 10/74 validators ≈ 13.5% proposer share, so 50 `Anchor.*`-graffiti blocks was unreachable), and startup `sleep` 5m → 8m so anchor syncs the registry before the checks sample.
- **`Cargo.lock`** — bump `crossbeam-epoch` 0.9.18 → 0.9.20 (RUSTSEC-2026-0204 solution; transitive, no manifest change).

- **CI-only + one transitive lockfile bump** — no runtime or production code.
- The ssv-mini pin is a personal fork for now (pending #37 review); the `TEMPORARY` marker flags the swap back to upstream.
- Self-validating: `local-testnet.yml` is `pull_request`-triggered, so this PR runs the fixed workflow on itself — a green `run-local-testnet` here is the proof, not a post-merge surprise.

## Validation

- **`run-local-testnet`: green** on this PR — the deployer clones `ssvlabs/ssv-network@v2.0.0`, anchor pre-registers and runs its validators, and assertoor passes block proposals (`Anchor.*` graffiti), 95% attestation, and finality over 2 epochs.
- **`check-code`: green** with the crossbeam bump (remaining audit output is the same allowed warnings as before).

## Rollback

Revert the commits; CI-only, no config, data, or operational impact.

- Lands on `unstable`; propagates to `epbs` and other branches via the normal syncs. Does not need `stable` (`local-testnet.yml` is `pull_request`-triggered; `stable` only matters for the workflow's `push: [stable]` release trigger).
- After [ssvlabs/ssv-mini#37](https://github.com/ssvlabs/ssv-mini/pull/37) merges, repin `local-testnet.yml` from the fork to the upstream SHA (marked `TEMPORARY`).
- Supersedes #1116 — the crossbeam bump is folded in here to break the circular check deadlock; #1116 can be closed.

### DIFF
--- a/.github/custom/local-testnet/params.yaml
+++ b/.github/custom/local-testnet/params.yaml
@@ -42,6 +42,10 @@ network:
 # Static (deterministic) validator keys, required by the newer ssv-mini key-generation flow.
 use_static_keys: true
 
+# Pre-register the static keyshares on-chain (ssv-mini main.star Step 4). Anchor CI runs ssv-mini
+# without the aetheria executor, so operators only run validators if we register them ourselves.
+pre_register_validators: true
+
 monitor:
   enabled: false
 
--- a/.github/custom/local-testnet/params.yaml
+++ b/.github/custom/local-testnet/params.yaml
@@ -19,9 +19,10 @@ network:
     fulu_fork_epoch: 0
     perfect_peerdas_enabled: false
 
-    # 40 = 10 validators * 2(number of el/cl nodes) + 20 (running on SSV/Anchor nodes)
-    # aligns with validator_count configuration under participants section
-    preregistered_validator_count: 40
+    # ssv-mini main.star asserts preregistered_validator_count >= 74: VCs run
+    # [0, validator_count*count) = [0, 20); genesis deposits [0, 74). Indices 64-73 are
+    # reserved for the ssv-mini/aetheria seed (deposited but VC-idle), so keep this >= 74.
+    preregistered_validator_count: 74
     # NOTE: changes the number of slots in the epoch and potentially some other network settings.
     # docs: https://github.com/ethpandaops/ethereum-package/blob/main/README.md#configuration
     # Supported by Anchor.
@@ -37,6 +38,9 @@ network:
     tests:
       - https://raw.githubusercontent.com/sigp/anchor/<<COMMIT>>/.github/custom/local-testnet/test_anchor_proposal.yaml
 
+
+# Static (deterministic) validator keys, required by the newer ssv-mini key-generation flow.
+use_static_keys: true
 
 monitor:
   enabled: false
@@ -56,5 +60,7 @@ images:
   monitor: "monitor"
   redis: "redis:7.4.2"
   postgres: "postgres:15"
-  foundry: "localssv/ssv-network"
+  # ssv-mini renamed this image key from `foundry` to `deployer` (contract deployer now builds
+  # the official ssvlabs/ssv-network@v2.0.0 via hardhat).
+  deployer: "localssv/ssv-network"
 
--- a/.github/custom/local-testnet/params.yaml
+++ b/.github/custom/local-testnet/params.yaml
@@ -19,12 +19,11 @@ network:
     fulu_fork_epoch: 0
     perfect_peerdas_enabled: false
 
-    # ssv-mini main.star asserts preregistered_validator_count >= 74: lighthouse VCs run
-    # [0, validator_count*count) = [0, 64); genesis deposits [0, 74). Indices 64-73 are the
-    # anchor validators (deposited but lighthouse-VC-idle; anchor runs them via the
-    # pre-registered static keyshares), so keep this >= 74. Running 64 VC + 10 anchor = 74/74
-    # gives the >2/3 participation the assertoor finality + 95% attestation checks require.
-    preregistered_validator_count: 74
+    # Dynamic keygen layout: lighthouse VCs run [0, validator_count*count) = [0, 64); anchor runs
+    # [64, preregistered) = [64, 128) via freshly generated keyshares. 64 VC + 64 anchor = 128/128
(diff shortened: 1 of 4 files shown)


petarjuki7:
I wanted to verify the outputs here so I went through the logs from the failed run myself. Good news is the pin fix works, the testnet comes up, contracts deploy. But all 4 anchor nodes end with `Operator ready, no validators assigned`, so they never do any duties and only the 20 lighthouse validators attest.

What I found is that before, ssv-mini did the validator pre-registration for us as part of `main.star`. But the same commit that fixed the deleted-fork clone (6d310a4b) also removed that step: it now prints `Step 4/5: Skipping validator pre-registration (executor registers its own; see #29)`. Their "executor" is ssvlabs' internal harness which we don't run. So now nobody registers the validators, and anchor never learns it should operate them. That also means there's no pin that works, every commit either has the broken clone or has registration removed.

Also, it's not just a timing problem: all 74 deposited validators are active from genesis, but we only run 20 of them. Even if we register the 10 static keyshares, that's 30/74 ≈ 40% participation (under 2/3), so the chain can never finalize and `check_consensus_finality` can't pass no matter how long we sleep. The new ssv-mini layout assumes 64 VC validators + 10 SSV ones (indices 64–73), while our params are still on the old 20 VC layout.

So I think on top of this PR we need to:
- bump validator_count 10 → 32 in params.yaml (this should also make the contract deploy much faster)
- register the keyshares ourselves after kurtosis run. `register-validators.cjs` is already sitting in the deployer container at /app/registration, ssv-mini just never calls it
- bump the 5m sleep in test_anchor_proposal.yaml to ~8m so the checks start after registration takes effect

The test still checks anchor / the graffiti check only anchor can satisfy, and 95% attestation is only reachable if anchor's 10 validators attest. The CI run will show us if this was the right fix.

petarjuki7:
Went through the new run, the registration fix works. `Registered 10 validator(s)`, anchor picks them up, and epochs 3–4 show 100% target votes (74/74), attestation check passed. Anchor proposes blocks fine too.

The remaining failure is just `blockCount: 50` in the proposal check. That was sized for the old layout where anchor ran half the validators. Now anchor proposes ~13.5% of blocks, so 50 matches need ~37 min and the test times out at 20m (it collected 14). Dropping `blockCount` to 10 should do it, maybe with the test timeout at 25m for slack.

shane-moore:
Update, since you nailed the participation issue:

- **Registration**: handled upstream via a default-off `pre_register_validators` flag in ssv-mini ([ssvlabs/ssv-mini#37](https://github.com/ssvlabs/ssv-mini/pull/37)) that registers the keyshares in `main.star` Step 4. The deployer container is torn down after operator registration, so an in-flow flag beat a post-run `register-validators.cjs` call. CI pins the fork branch until #37 merges.
- **`validator_count` 10 -> 32** (your fix): 64 VC + 10 anchor = 74/74, so attestation now passes at **100%** and the chain finalizes.

The last red check is `check_consensus_block_proposals`, set to `blockCount: 50` = 50 **Anchor-graffiti** blocks. Anchor runs 10 of 74 validators (~13.5% proposer share), so 50 anchor blocks needs ~37 min of chain time; CI produced 4 before the concurrent-task deadline and failed. So I'm dropping it to **3**, which still proves anchor proposes valid blocks (a run produced 4 in the window) without requiring an unreachable count.

I also tried a dynamic-keygen layout (`use_static_keys: false`, `preregistered 128`) to make anchor a ~50% proposer so a high blockCount would be meaningful. It works once mature (100% attestation, ~62% of blocks anchor-proposed), but the 64 validators ramp up too slowly for the assertoor sleep on CI, so attestation sampled mid-ramp at 81% (needs 95%). Not worth the timing fragility for a merge-queue gate, so going with static + `blockCount: 3`.

petarjuki7 (approved):
Nice, thanks!

PR #1125, opened 2026-07-09
feat(validator_store): implement sign_proposer_preferences (SingleValidator partial-sig)

### LINKED ISSUE #1063: feat(validator_store): implement sign_proposer_preferences (SingleValidator partial-sig)
## Goal

Implement the `sign_proposer_preferences` LH trait method on `AnchorValidatorStore`, filling the existing `TODO(gloas)` stub (`anchor/validator_store/src/lib.rs:3367-3374`), as a `SingleValidator`-mode partial-signature collection (no QBFT): sign the supplied `ProposerPreferences` under `DOMAIN_PROPOSER_PREFERENCES` with the validator's BLS share and reconstruct via the signature collector. The trait signature is `(validator_pubkey: PublicKeyBytes, preferences: ProposerPreferences) -> SignedProposerPreferences` (`validator_client/validator_store/src/lib.rs:217` at LH pin `81d576943`, PR #1125's pin; re-verify the live pin in `Cargo.toml` before trusting line refs); the preferences arrive fully built from the LH `ProposerPreferencesService`, so no duty or config derivation happens in this method. Mirrors `sign_validator_registration_data` / `sign_voluntary_exit` in shape, and LH's own implementation (`lighthouse_validator_store/src/lib.rs:1490` at the pin) for the domain and signing flow.

## Context

Per SIP-94 §5, each operator independently derives `ProposerPreferences { dependent_root, proposal_slot, validator_index, fee_recipient, target_gas_limit }` and signs it; reconstruction succeeds only when `2f + 1` operators converge on one signing root. No QBFT: `target_gas_limit` (operator config) and `dependent_root` (per-operator BN observation) determine each operator's root locally.

**Collection slot = `proposal_slot` (SIP-94 §5/§7; supersedes the earlier `slot_clock.now()` design in this issue's history).** The slot passed to `collect_signature` becomes the `PartialSignatureMessages.slot` stamped on the wire (`SignatureMetadata` built at `lib.rs:461` with its `slot` field at `:469`; wire stamp `slot: metadata.slot` at `signature_collector/src/lib.rs:312-316`), and SIP-94 §5 pins that slot to `proposal_slot` itself: one runner per proposal slot, matching ssv-spec's `validatePartialSigMsgForSlot` (`msg.Slot == duty.DutySlot()`). go-ssv parity: `protocol/v2/ssv/runner/proposer_preferences.go:40,80,199` (one sub-runner per proposal slot; "duty.Slot is the proposal slot the preference targets, which is also the slot carried on the [wire]"). Message validation accepts the future slot via the role's earliness allowance and validates proposer assignment against it (#1062).

**The collector retention story improves under this design.** Collectors are reaped once `for_slot < current_slot - SIGNATURE_COLLECTOR_RETAIN_SLOTS(1)` (`signature_collector/src/lib.rs:50,423-427`), so a collector keyed to a future `proposal_slot` lives from creation until the proposal slot passes, and operators no longer race a 1-slot arrival window for lookahead emissions (the failure mode the earlier send-slot design had to work around). Collectors key on `(signing_root, validator_index)` (`signature_collector/src/lib.rs:375-381`), so a `dependent_root` re-emission for the same proposal slot creates a fresh collector under its new root; operators need only sign on the same trigger (epoch boundary or the shared re-emission trigger, §5).

## Suggested approach

1. Look up the validator metadata and cluster for `validator_pubkey` (the same lookup `sign_validator_registration_data` / `sign_voluntary_exit` use). The `preferences` argument arrives fully built from the LH `ProposerPreferencesService` (`proposer_preferences_service.rs:141-190` at the pin: `dependent_root` and `proposal_slot` from the duties map, `fee_recipient` and `target_gas_limit` from `proposal_data(pubkey)`); duty and config derivation is that service's layer, tracked under #1064, and SIP-94 §5's byte-for-byte agreement requirement applies there. (The container field is now `target_gas_limit`, matching consensus-specs and SIP-94; older pins named it `gas_limit`.)
2. Compute the signing domain via the 2-arg wrapper `self.get_domain(epoch, Domain::ProposerPreferences)` (`lib.rs:277-284`) with `epoch = preferences.proposal_slot.epoch(E::slots_per_epoch())`: the domain epoch is keyed on `proposal_slot` so the signature verifies against the fork of the slot being proposed, including the §5 pre-fork emission case (Gloas domain while the chain is pre-fork). Same derivation as LH's reference implementation.
3. `signing_root = preferences.signing_root(domain)`.
4. Collect with the duty's slot: `collect_signature(PartialSignatureKind::ProposerPreferences, Role::ProposerPreferences, CollectionMode::SingleValidator, &validator, &cluster, signing_root, preferences.proposal_slot)`, following the RANDAO call shape at `lib.rs:2550-2558`. Do NOT stamp `slot_clock.now()` and do NOT stamp epoch-start; the wire slot is the duty's `proposal_slot` (see Context).
5. Return the reconstructed `SignedProposerPreferences` to the caller (the LH `ProposerPreferencesService` spawned in #1064 handles submission).

## Acceptance criteria

- The broadcast `PartialSignatureMessages.slot` equals `preferences.proposal_slot` (assert via the collector/manager path, not just the call site).
- Domain epoch equals `epoch(proposal_slot)`; a preference for a first-Gloas-epoch slot signed pre-fork uses the Gloas fork domain.
- Kind is `PartialSignatureKind::ProposerPreferences`, role `Role::ProposerPreferences`, `SingleValidator` mode.
- No slashing-protection database read or write (registration-style non-slashable duty).
- Re-emission for the same `proposal_slot` with a changed `dependent_root` produces a different signing root and a fresh collector; both reconstruct independently when quorums exist.
- A lookahead emission (proposal slot up to `(1 + min_seed_lookahead)` epochs ahead) reconstructs even when partial sigs arrive several slots after emission (collector alive until `proposal_slot + 1`).
- A missing quorum fails per-validator with a bounded collection timeout; it must never hang the caller indefinitely, so one failed validator cannot stall sibling validators' signing or publication (see Risks: batch liveness).
- Collection failures are surfaced cause-neutrally (e.g. threshold not reached / timeout). The wire carries only the signing root, so telemetry must not claim attribution to `target_gas_limit` vs `dependent_root` divergence; either can only be listed as a possible cause.
- No duty or config derivation inside `validator_store`: the signed message equals the supplied `preferences` verbatim (the method only looks up validator/cluster, computes domain and root, and collects).
- No change to `sign_validator_registration_data` (registration keeps its send-slot semantics pre-Gloas).

## Risks

- **Slot source of truth.** `proposal_slot` must come from the `preferences` argument end-to-end; deriving it from the clock anywhere reintroduces the send-slot design and breaks #1062's proposer-assignment check.
- **Trigger alignment (caller-side, #1064).** Operators must emit on the shared triggers (epoch boundary; preference-input change) for quorums to form; a straggler operator observing a different `dependent_root` splits roots (accepted §5 behavior, surfaces as a reconstruction timeout). This method signs whatever it is handed; cadence lives in the LH `ProposerPreferencesService` spawned in #1064. At pin `81d576943` the service polls every slot, covers the current and next epoch (proposer lookahead), re-emits when `dependent_root` changes, and emits pre-fork for the first Gloas epoch (the fork gate is per preference epoch; duties race fixed in sigp/lighthouse#9309).
- **Batch liveness (caller-side, upstream).** At pin `81d576943` the service awaits each signer sequentially (`proposer_preferences_service.rs:190`), so one pending collection (e.g. a missing quorum running to its timeout) delays every later validator's publication in that batch. Fixed upstream by sigp/lighthouse#9617 (concurrent signing and publication, per-pair retry), open at the time of this edit and picked up on a future pin bump. The per-validator timeout criterion above is what makes the concurrent service effective for Anchor.
- **Cross-checking #1062 (caller-side).** The dedup cap (4 distinct roots per signer-slot) bounds how many same-slot re-emissions can reconstruct; unbounded re-emission requests would burn that budget. Enforcement sits with the caller (#1064), not in this method.
- **Doppelganger freshness (existing review finding, not duplicated here).** Future proposal slots incorrectly affect doppelganger freshness; tracked in the PR #1106 review thread: https://github.com/sigp/anchor/pull/1106#discussion_r3555669823

petarjuki7:
Closes #1063 (ePBS Proposer Preferences duty milestone, SIP-94 §5/§7).

Implements `sign_proposer_preferences` on `AnchorValidatorStore`: a single-validator partial-signature collection round over `ProposerPreferences`, domain keyed on `proposal_slot`'s epoch, envelope slot stamped with the duty's `proposal_slot` (not the send slot), and no slashing-DB interaction. Adds a reconstruction-failure reporter + metric and generalizes the shared `collect_signature` failure classifier.

Stacked on #1106 — only the final commit (`implement sign_proposer_preferences`) is in scope here; the ssv_types / message-validator commits belong to #1106 and drop out of this diff once it merges.

### DIFF
--- a/anchor/validator_store/src/lib.rs
+++ b/anchor/validator_store/src/lib.rs
@@ -81,7 +81,7 @@ use validator_store::{
     ValidatorStore,
 };
 
-use crate::instrumentation::PtcFailureClass;
+use crate::instrumentation::CollectionFailureClass;
 
 /// Number of epochs of slashing protection history to keep.
 ///
@@ -98,6 +98,7 @@ const AGGREGATE_LOG_NAME: &str = "aggregate";
 const SELECTION_PROOF_LOG_NAME: &str = "selection proof";
 const SYNC_SELECTION_PROOF_LOG_NAME: &str = "sync selection proof";
 const SYNC_COMMITTEE_CONTRIBUTION_LOG_NAME: &str = "sync committee contribution";
+const PROPOSER_PREFERENCES_LOG_NAME: &str = "proposer preferences";
 
 /// A request to collect a committee signature for a single validator.
 ///
@@ -871,8 +872,8 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> AnchorValidator
         validator_pubkey: &PublicKeyBytes,
         slot: Slot,
     ) {
-        match instrumentation::classify_ptc_collection_failure(error) {
-            PtcFailureClass::NoSignature => {
+        match instrumentation::classify_collection_failure(error) {
+            CollectionFailureClass::NoSignature => {
                 warn!(
                     ?validator_pubkey,
                     %slot,
@@ -884,7 +885,7 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> AnchorValidator
                     &[metrics::PTC_FAILURE_NO_SIGNATURE],
                 );
             }
-            PtcFailureClass::Infra => {
+            CollectionFailureClass::Infra => {
                 error!(
                     ?validator_pubkey,
                     %slot,
@@ -896,12 +897,59 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> AnchorValidator
                     &[metrics::PTC_FAILURE_INFRA],
                 );
             }
-            PtcFailureClass::NonCollection => {
+            CollectionFailureClass::NonCollection => {
                 error!(
                     ?validator_pubkey,
                     %slot,
                     ?error,
                     "Failed to sign payload attestation"
+                );
+            }
+        }
+    }
+
+    /// Classify and report a ProposerPreferences signature-collection failure.
+    fn report_proposer_preferences_collection_failure(
+        &self,
+        error: &Error,
+        preferences: &ProposerPreferences,
+        signing_root: Hash256,
+    ) {
+        match instrumentation::classify_collection_failure(error) {
+            CollectionFailureClass::NoSignature => {
+                warn!(
+                    validator_index = preferences.validator_index,
+                    proposal_slot = %preferences.proposal_slot,
+                    target_gas_limit = preferences.target_gas_limit,
+                    dependent_root = ?preferences.dependent_root,
+                    ?signing_root,
+                    ?error,
+                    "ProposerPreferences reconstruction failed; operators likely diverged on the \
+                     signing root (target_gas_limit or dependent_root)"
+                );
+                metrics::inc_counter_vec(
+                    &metrics::PROPOSER_PREFERENCES_RECONSTRUCTION_FAILURES,
+                    &[metrics::PROPOSER_PREFERENCES_FAILURE_SIGNING_ROOT_DIVERGENCE],
+                );
+            }
+            CollectionFailureClass::Infra => {
+                error!(
+                    validator_index = preferences.validator_index,
+                    proposal_slot = %preferences.proposal_slot,
+                    ?error,
+                    "ProposerPreferences signature collection infrastructure failure"
+                );
+                metrics::inc_counter_vec(
+                    &metrics::PROPOSER_PREFERENCES_RECONSTRUCTION_FAILURES,
+                    &[metrics::PROPOSER_PREFERENCES_FAILURE_INFRA],
+                );
+            }
+            CollectionFailureClass::NonCollection => {
+                error!(
+                    validator_index = preferences.validator_index,
+                    proposal_slot = %preferences.proposal_slot,
+                    ?error,
+                    "Failed to sign ProposerPreferences"
                 );
             }
         }
@@ -3487,11 +3535,57 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> ValidatorStore
 
     async fn sign_proposer_preferences(
         &self,
-        _validator_pubkey: PublicKeyBytes,
-        _preferences: ProposerPreferences,
+        validator_pubkey: PublicKeyBytes,
+        preferences: ProposerPreferences,
     ) -> Result<SignedProposerPreferences, Error> {
-        // TODO(gloas)
-        Err(Error::SpecificError(SpecificError::Unsupported))
+        let (validator, cluster) = self.get_validator_and_cluster(validator_pubkey)?;
+
+        let epoch = preferences.proposal_slot.epoch(E::slots_per_epoch());
+        let domain = self.get_domain(epoch, Domain::ProposerPreferences);
+        let signing_root = preferences.signing_root(domain);
+
+        // Envelope slot = the duty's proposal_slot (SIP-94 §5/§7), NOT slot_clock.now() and NOT
+        // epoch-start. It becomes PartialSignatureMessages.slot on the wire, which peers accept via
+        // the ProposerPreferences earliness allowance and validate proposer-assignment against
+        // (#1062). A future proposal_slot also keeps the collector alive until that slot passes.
+        let proposal_slot = preferences.proposal_slot;
+
+        let future = async {
+            let signature = match self
+                .collect_signature(
+                    PartialSignatureKind::ProposerPreferences,
+                    Role::ProposerPreferences,
+                    CollectionMode::SingleValidator,
+                    &validator,
+                    &cluster,
+                    signing_root,
+                    proposal_slot,
+                )
+                .await
+            {
+                Ok(signature) => signature,
+                Err(err) => {
+                    self.report_proposer_preferences_collection_failure(
+                        &err,
+                        &preferences,
+                        signing_root,
+                    );
+                    return Err(err);
+                }
+            };
+
+            Ok(SignedProposerPreferences {
+                message: preferences,
+                signature,
+            })
+        };
+
+        run_and_update_metrics(
+            PROPOSER_PREFERENCES_LOG_NAME,
+            &metrics::SIGNED_PROPOSER_PREFERENCES_TOTAL,
+            future,
+        )
+        .await
     }
 }
 
--- a/anchor/validator_store/src/lib.rs
+++ b/anchor/validator_store/src/lib.rs
@@ -98,7 +98,10 @@ const AGGREGATE_LOG_NAME: &str = "aggregate";
 const SELECTION_PROOF_LOG_NAME: &str = "selection proof";
 const SYNC_SELECTION_PROOF_LOG_NAME: &str = "sync selection proof";
 const SYNC_COMMITTEE_CONTRIBUTION_LOG_NAME: &str = "sync committee contribution";
-const PROPOSER_PREFERENCES_LOG_NAME: &str = "proposer preferences";
+
+/// Upper bound, in slots, on how long `sign_proposer_preferences` waits for a validator's signature
+/// to be reconstructed.
+const PROPOSER_PREFERENCES_COLLECTION_TIMEOUT_SLOTS: u32 = 2;
 
 /// A request to collect a committee signature for a single validator.
 ///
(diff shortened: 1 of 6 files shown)


anchor/validator_store/src/lib.rs
```diff
@@ -894,6 +895,53 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> AnchorValidator
...
+        match instrumentation::classify_collection_failure(error) {
+            CollectionFailureClass::NoSignature => {
+                warn!(
+                    validator_index = preferences.validator_index,
+                    proposal_slot = %preferences.proposal_slot,
+                    target_gas_limit = preferences.target_gas_limit,
+                    dependent_root = ?preferences.dependent_root,
+                    ?signing_root,
+                    ?error,
+                    "ProposerPreferences reconstruction failed; operators likely diverged on the \
```
shane-moore: **Telemetry: `signing_root_divergence` overclaims the cause of a failed ProposerPreferences reconstruction**

`report_proposer_preferences_collection_failure` labels every `NoSignature` collection failure `signing_root_divergence` and warns "operators likely diverged" (`anchor/validator_store/src/lib.rs:906-921`, `metrics.rs:163`). But that class is reached only via `QueueClosedError` (`instrumentation.rs:54-56`), which covers threshold-not-reached **and** shutdown (`signature_collector/src/lib.rs:520-533`), and the wire carries no preference fields (`ssv_types/src/partial_sig.rs:137-143`), so the collector cannot know a split occurred. The metric doc three lines up already says as much (`metrics.rs:160-163`), and the sibling PTC path labels the identical class honestly as `no_signature` / "Insufficient partial signatures" (`lib.rs:862-873`). Peers agree: go-ssv returns `nil` on no-quorum with no divergence claim (`protocol/v2/ssv/runner/proposer_preferences.go:258-260`), and SIP-94 frames this as a silent miss "same shape as ValidatorRegistration today."

Suggest: rename the label to `insufficient_partial_signatures`, reword the warn to list divergence as one of several possible causes (also too-few-operators and delivery loss), keep the local tuple fields, and rename the two tests at `validator_store/src/testing/proposer_preferences.rs:238,303`. `CollectionTimeout` (`signature_collector/src/lib.rs:514`) is never produced; consider removing it or reserving it explicitly rather than testing it.

shane-moore:
think CI will be fixed if rebase this branch from the `epbs` branch

anchor/validator_store/src/metrics.rs
```diff
@@ -27,6 +27,15 @@ pub static SIGNED_RANDAO_REVEALS_TOTAL: LazyLock<Result<IntCounterVec>> = LazyLo
     )
 });
 
+pub static SIGNED_PROPOSER_PREFERENCES_TOTAL: LazyLock<Result<IntCounterVec>> =
+    LazyLock::new(|| {
+        try_create_int_counter_vec(
+            "signed_proposer_preferences_total",
```
shane-moore: This is the only unprefixed metric in the file; the paired failure counter uses `anchor_` and the closest top-line analog uses `vc_signed_randao_reveals_total`. Worth picking a prefix before ship (renaming later breaks dashboards); `anchor_` to match `anchor_proposer_preferences_reconstruction_failures_total` seems most consistent.

anchor/validator_store/src/testing/proposer_preferences.rs
```diff
@@ -0,0 +1,377 @@
...
+    assert_eq!(
+        call.metadata.role,
+        Role::ProposerPreferences,
+        "the network message should be routed under the ProposerPreferences role"
+    );
+    // The broadcast `PartialSignatureMessages.slot` (captured off the collector call) must equal
+    // `proposal_slot`: SIP-94 §5/§7 pins the wire slot to the slot being proposed, and #1062
+    // validates proposer assignment against it.
+    assert_eq!(
+        call.metadata.slot, preferences.proposal_slot,
```
shane-moore, line 176: These tests assert `call.metadata.slot` at the trait boundary (the mock captures the metadata; the real `create_message` never runs), but the comments here and in the module doc say the broadcast `PartialSignatureMessages.slot` is asserted. Low risk since the metadata-to-wire copy is covered by signature_collector's own tests, but suggest either softening the comments or adding one integration case through a real `SignatureCollectorManager`, which would also satisfy #1063's first AC as written ("assert via the collector/manager path").

anchor/validator_store/src/lib.rs
```diff
@@ -3487,11 +3537,57 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> ValidatorStore
...
+                }
+            };
+
+            Ok(SignedProposerPreferences {
+                message: preferences,
+                signature,
+            })
+        };
+
+        run_and_update_metrics(
```
shane-moore: Collection failures get double-reported here at contradictory levels: a no-quorum failure (`SignatureCollectionFailed(QueueClosedError)`) is correctly emitted by the reporter as `warn!` + `insufficient_partial_signatures`, then propagates into this wrapper, matches none of the `SameData`/`Slashable`/`Timeout` arms, and hits the catch-all: `error!("Unexpected error while signing ...")` + `signed_proposer_preferences_total{status="other_error"}`. LH error-logs the returned error too, so one SIP-expected divergence event (per-operator `target_gas_limit` / `dependent_root`) yields one correct warn plus two errors, and the top-line metric miscounts it as `other_error`.

The wrapper's `SameData`/`Slashable`/`Timeout` arms are all unreachable on this path (no slashing protection; nothing in the collection chain produces `SpecificError::Timeout`), so its only live effects are the success counter and this catch-all. Suggest dropping the wrapper to match `sign_payload_attestation` and incrementing `SIGNED_PROPOSER_PREFERENCES_TOTAL{status="success"}` explicitly on the Ok path if the top-line counter is wanted.

anchor/validator_store/src/testing/proposer_preferences.rs
```diff
@@ -0,0 +1,377 @@
...
+        "the broadcast partial-signature slot should equal the proposal slot"
+    );
+
+    // Recompute the root independently to lock the ProposerPreferences-specific signing decisions
+    // that no other test covers: the `Domain::ProposerPreferences` choice, the epoch derived from
+    // `proposal_slot` (here a future slot, so this pins epoch keying to the proposal, not the
+    // clock), and the signed object being the `ProposerPreferences` itself. A regression to a
+    // different domain or a different epoch key fails here.
```
shane-moore: The epoch half of this recompute is unfalsifiable under the harness spec. The harness uses the default `ChainSpec::mainnet()`, where the send epoch (`TEST_SLOT = 1`, epoch 0) and the proposal epoch (slot 65, epoch 2) are both pre-Altair (Altair at 74240), so `fork_at_epoch`/`get_fork_version` return the genesis fork version `[0,0,0,0]` for both and `get_domain` produces byte-identical domains. A regression to keying the domain on the send slot's epoch would therefore produce the same signing root and this assert would still pass, so the "different epoch key fails here" claim doesn't hold. No test schedules Gloas either, so the #1063 criterion "a first-Gloas-epoch slot signed pre-fork uses the Gloas fork domain" is unexercised.

Suggest one success case on a mainnet-based spec with `gloas_fork_epoch = Some(Epoch::new(LOOKAHEAD_EPOCHS))`, wired in via `HarnessOptions::spec`. Note `gloas_at_genesis_spec()` doesn't help here: with Gloas at epoch 0 both epochs share the Gloas fork version, so the boundary must fall strictly inside the lookahead window. Then assert the captured signing root equals a Gloas-fork-version recompute AND differs from a genesis-fork-version recompute (the inequality is what makes the epoch keying falsifiable); the same case covers the pre-fork Gloas-domain criterion, since the harness clock sits in epoch 0. While touching it, `expected_signing_root` could take the spec and `genesis_validators_root` from the harness instead of hardcoding `ChainSpec::mainnet()` / `Hash256::zero()`.

anchor/validator_store/src/testing/proposer_preferences.rs
```diff
@@ -0,0 +1,377 @@
...
+/// The code classifies collection failures only into the coarse `insufficient_partial_signatures` /
+/// `infra` buckets; it never attributes a divergence to a specific input field. A
+/// `target_gas_limit` or `dependent_root` mismatch is only observable as a signing-root split
+/// (threshold-not-reached), which surfaces here as the same `QueueClosedError`.
+///
+/// This asserts the `insufficient_partial_signatures` label increments while the per-input
+/// attribution labels (`"target_gas_limit_divergence"`, `"dependent_root_divergence"`) stay at 0,
+/// proving the code emits no per-input attribution reason.
+#[tokio::test(flavor = "multi_thread")]
+async fn proposer_preferences_does_not_classify_remote_input_without_metadata() {
```
shane-moore: The two attribution labels this test asserts at zero (`target_gas_limit_divergence`, `dependent_root_divergence`) are strings the production code never writes (only `insufficient_partial_signatures` and `infra` exist), so those asserts pass even if the whole `NoSignature` reporting arm is deleted, and a future per-input attribution under any other label name would not trip them either; the remaining assertions repeat the test above. If the cause-neutrality criterion is worth pinning, the falsifiable boundary is between the two labels that do exist: I'd drop this test, add a zero-delta check on the `infra` label to the `QueueClosedError` test, and add an infra-class counterpart (e.g. `EmptySignature`, mirroring `sign_payload_attestation_infra_failure_increments_infra_metric`), which would also give the new Infra arm of `report_proposer_preferences_collection_failure` its first coverage.

Separately, the "no slashing-protection read or write" criterion is currently unpinned: all four tests run with protection disabled (it's the harness default). A copy of `sign_payload_attestation_does_not_touch_slashing_db` with protection enabled would cover it as a tripwire. And since the two success tests share their setup and slot assertion, folding them into one (or converting the second into the fork-boundary domain case suggested in the thread above) keeps the file at three well-differentiated tests.

anchor/validator_store/src/lib.rs
```diff
@@ -3487,11 +3537,57 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> ValidatorStore
...
+                .collect_signature(
+                    PartialSignatureKind::ProposerPreferences,
+                    Role::ProposerPreferences,
+                    CollectionMode::SingleValidator,
+                    &validator,
+                    &cluster,
+                    signing_root,
+                    proposal_slot,
+                )
+                .await
```
shane-moore: #1063's AC asks that a missing quorum "fails per-validator with a bounded collection timeout". As implemented, a no-quorum await resolves only when the cleaner reaps the collector at `proposal_slot + 2` (`signature_collector/src/lib.rs:423-427`), which for a next-epoch lookahead emission is up to ~65 slots (~13 min on mainnet). At the current LH pin the `ProposerPreferencesService` awaits each validator sequentially in a single loop task (`proposer_preferences_service.rs:190-205`), so one no-quorum validator (expected recurring behavior per SIP-94 §5, e.g. a `target_gas_limit` config mismatch or a `dependent_root` observation split) parks the entire service, both epochs' emissions included, for that duration. Suggest wrapping this await in a short `tokio::time::timeout` (1-2 slots) mapped to `CollectionError::CollectionTimeout`: `classify_collection_failure` already handles that variant, and when a whole batch fails the LH service's per-slot retry then acts as a poll that picks up the cached signature from the still-alive collector if quorum forms later, so the lookahead-reconstruction AC is unaffected.

One side effect for completeness: on retry the service re-signs and `sign_and_collect` re-broadcasts the same root; that interaction with gossip dedup is tracked separately in #1148 and does not block this change. If the timeout is deferred instead, the bounded-timeout AC should be explicitly descoped into a follow-up, since this PR closes #1063.

shane-moore:
looks good! just got small `check-fmt` CI fixes then should be all good

shane-moore:
pushed the small `check-fmt` fix && approved, so we can merge if you feel good about it  🙌

shane-moore (approved):

PR #1128, opened 2026-07-10
feat(validator_store): use decided votes for aggregation

### LINKED ISSUE #1113: feat(validator_store): use committee-decided vote for aggregation inputs
## Goal

After committee QBFT decides, use the committee-decided `SlotVote` as the input to the slot's aggregation phase, with `VotingContext.vote` as fallback when no decision is available yet.

This covers both aggregation inputs:

- attestation aggregate fetch: compute `attestation_data_root` from the decided vote, including the Gloas `attestation_data_index`
- sync contribution fetch: use the decided vote's `block_root`

## Context / motivation

Follow-up requested in #1103 review: https://github.com/sigp/anchor/pull/1103#pullrequestreview-4608710592.

At #1103 head `54e7230`, committee signing applies the QBFT-decided vote before signing, but aggregator data is still fetched from the local slot seed:

- `metadata_service::update_aggregation_assignments` groups duties by SSV committee, but also builds global `attestation_committee_indexes` and `all_subnet_ids`.
- `build_consensus_data_for_all_committees` fetches one global `committee_index -> Attestation` map from `voting_context.vote`, and one global sync contribution map from `voting_context.vote.block_root()`.
- `fetch_aggregated_attestations` reconstructs `AttestationData` from the supplied seed vote.
- `AggregatorCommitteeConsensusData` is per SSV committee, so it can carry different aggregate bytes per SSV committee object, but the current global fetch map collapses before that point.

Under Gloas, the aggregate-fetch root must include the full attestation data the committee actually signed, including `attestation_data_index`. More generally, if committee QBFT decides a vote different from the local seed, aggregation should fetch data for what the cluster attested, not what this operator seeded locally. Otherwise the cluster can fail to fetch an aggregate, or aggregate data it did not sign. This is non-slashable, but it is a liveness/correctness gap.

## Implementation approach

Store committee decisions in the slot-scoped `VotingContext`:

```rust
struct VotingContext {
    voting_assignments: Arc<VotingAssignments>,
    vote: SlotVote,
    decided_votes: Mutex<HashMap<CommitteeId, SlotVote>>,
}
```

The `VotingContext` is replaced with the slot's voting assignments, so the cache is naturally bounded to that slot and does not require a separate multi-slot map, retention constant, or pruning loop. The mutex is never held across `.await`.

Centralize decision recording in `decide_committee_vote`, which is shared by both committee signing paths:

- `sign_committee_attestations`
- `sign_committee_sync_committee_signatures`

Record only after `Completed::Success`. The first decision for a committee is authoritative, an identical repeat is idempotent, and a conflicting repeat preserves the first value and returns `CommitteeDecisionConflict` before the second caller can sign the conflicting vote.

When constructing aggregation data, select one vote per SSV committee:

```rust
voting_context.vote_for_committee(ssv_committee_id)
```

This returns the committee decision when present and falls back to the slot seed otherwise.

Represent Beacon API request identity with complete composite keys:

```rust
AggregateFetchKey {
    attestation_data_root,
    committee_index,
}

SyncContributionFetchKey {
    block_root,
    subnet_id,
}
```

Build the union of requests across SSV committees and deduplicate only identical composite keys. Committees with the same request share one Beacon API call, while committees with divergent decided votes remain isolated even when they reference the same beacon committee index or sync subnet.

Fetch aggregate attestations and sync contributions concurrently using the existing partial-result deadlines. Deterministically assemble each committee's `AggregatorCommitteeConsensusData` from its selected vote and the shared composite-key result maps, preserving go-ssv-compatible ordering for the same candidates and successfully fetched request identities:

- aggregators by validator index
- contributors by selection-proof signing root, then validator index
- committee indexes and subnets in first-seen order from those sorted candidates
- aggregate attestations and sync contributions in the corresponding index/subnet order

## Acceptance criteria

- When a decided vote exists for the slot's SSV committee, aggregate-fetch `attestation_data_root` is computed from that decided vote.
- Under Gloas, the decided `attestation_data_index` is included in the aggregate-fetch root.
- Sync contribution fetch uses the decided vote's `block_root` when available.
- If the committee decision is not available when aggregation data is built, behavior falls back to `VotingContext.vote` and does not block the duty.
- The decided-vote cache is scoped to the slot's `VotingContext`, keyed by `CommitteeId`, and cannot grow across slots.
- The first decision is preserved. Identical repeats are accepted, while conflicting repeats fail closed before signing.
- Two SSV committees in the same slot can use different decided votes even if they reference the same beacon committee index or sync subnet.
- Identical Beacon API requests are deduplicated globally by their complete request identity.
- Aggregator committee consensus data preserves deterministic go-ssv-compatible ordering and object association for the same available inputs.
- `AggregatorCommitteeConsensusData.version` remains stamped from the duty slot's fork through the assembly refactor (currently `DataVersion::from(fork_name_at_epoch)` in `build_consensus_data_for_committee`), so Gloas slots carry `DataVersionGloas`.
- Pre-Gloas behavior is preserved, except that aggregation may now converge on decided `block_root/source/target` when available.

## Tests

- `VotingContext` returns a cached committee decision or falls back to the slot seed.
- Repeating an identical committee decision is idempotent.
- A conflicting committee decision returns contextual `CommitteeDecisionConflict` data and preserves the first value.
- Attestation-first and sync-first signing both populate the shared decided-vote cache, and both paths attach to the same committee QBFT instance.
- Fork-specific aggregate-fetch data uses the pre-Electra committee index, Electra zero index, and Gloas decided `attestation_data_index`.
- Consensus-data assembly at a Gloas slot stamps `version` as `DataVersionGloas`; a pre-Gloas slot preserves the existing version.
- Composite aggregate and sync keys deduplicate only identical request identities.
- Production consensus-data assembly excludes foreign results with the same bare committee index or subnet and preserves the exact ordering of aggregators, committee indexes, aggregate bytes, contributors, and sync contributions.
- Empty fetch results return no consensus data, while aggregator-only and contributor-only results remain valid.

## Notes

This cache only bridges a committee decision into the same slot's aggregation construction. It is not an attestation pool and stores only `SlotVote`, not full attestations or aggregate bytes.

The implementation intentionally defines parity over the same selected candidates and successfully fetched request identities. Beacon API retry scheduling and node-specific network outcomes are not part of the encoded consensus-data contract.

If implementation ever caches per-committee aggregate roots instead of `SlotVote`, be careful not to use Lighthouse `validator_committee_index` as the beacon committee index. That field is the validator's position within the committee.

shane-moore:

Committee signing from #1103 uses the QBFT-decided vote, but aggregation still derived aggregate-attestation and sync-contribution requests from the local slot seed. Under Gloas, this can omit the decided `AttestationData.index` from the requested root or use a block root the committee did not sign, causing missed or mismatched aggregation.

Closes #1113.

Related context:

- #1103
- https://github.com/ssvlabs/SIPs/pull/94

## Change Overview (Required)

- Cache each committee's successful decision in the slot-scoped voting context. The first decision wins, equal repeats are idempotent, and conflicting repeats fail closed before signature collection.
- Select one vote per SSV committee during aggregation, falling back to the local slot seed when no decision is available.
- Deduplicate Beacon API requests using complete identities: aggregate root plus beacon committee index, and block root plus sync subnet.
- Assemble consensus data deterministically using the existing Go-compatible ordering.
- Keep wire types, SSZ layout, P2P role and message formats, fetch deadlines, and partial-result behavior unchanged.

The commits are ordered for review: the first establishes the decision handoff and conflict safety, and the second consumes those decisions in aggregation.

### Design rationale and data flow

Before this change, aggregation could use the slot's locally selected seed vote for every SSV committee. The fetched aggregate-attestation results were effectively associated by beacon committee index, and sync contributions by subnet. That was sufficient while every SSV committee used the same block and attestation-data inputs.

After #1103, each SSV committee can complete QBFT with its own `SlotVote`. In Gloas, that vote also contains the decided `AttestationData.index`. Two committees can therefore request the same beacon committee index or sync subnet against different decided roots. Reusing a result based only on committee index or subnet can attach data fetched for vote A to the consensus data for vote B.

The implementation separates decision retention from fetch-result deduplication:

1. `VotingContext` retains the first successful decision for each SSV committee. It is already slot-scoped, so decisions expire with the existing voting context and require no independent eviction policy. If a committee has not decided before aggregation runs, lookup falls back to the slot seed to preserve the previous partial-progress behavior.
2. `MetadataService` resolves one vote for each SSV committee, then derives the exact Beacon API requests required by that committee.
3. Requests are deduplicated globally by their complete API identity:
   - aggregate attestation: `(attestation_data_root, beacon_committee_index)`
   - sync contribution: `(block_root, sync_subnet_id)`
4. Fetched results are held only for the duration of that aggregation build. They are not a long-lived service cache. Each committee reconstructs the same composite keys from its selected vote, so it can consume only matching results.
5. Matching candidates are passed through the existing deterministic ordering rules before the wire value is built.

The composite keys are intentional. Keying results only by SSV committee would prevent useful sharing when committees agree and would still need another level for multiple beacon committees or sync subnets. Keying only by beacon committee index or subnet is unsafe once committee votes diverge. The complete request identity provides isolation while retaining cross-committee deduplication.

- Divergent committee decisions can create more requests than the previous global maps. Exact composite-key deduplication still shares identical requests.
- Missing decisions deliberately fall back to the local slot seed instead of blocking the duty.
- Conflicting completed decisions reject that signing attempt to preserve consistency. The cached first value remains unchanged and the conflict is logged with context.
- Production assembly tests cover foreign-result isolation and exact list association and ordering.

## Validation (Required)

- [x] `cargo test -p anchor_validator_store` (57 passed)
- [x] `cargo clippy -p anchor_validator_store --tests -- -D warnings -D clippy::allow_attributes`
- [x] `cargo fmt --all -- --check`
- [x] `git diff --check upstream/epbs...HEAD`
- [x] The first commit independently passed the validator-store test suite and Clippy before the aggregation changes were applied.

New tests cover attestation-first and sync-first decisions, fallback, idempotence, conflicting decisions, Gloas index binding, composite-key deduplication and isolation, partial results, and deterministic ordering.

Revert the two commits. The cache is in-memory and slot-scoped, with no database, configuration, or network migration.

#1103 is merged into `epbs`. This branch is based on its merge commit, `3b174388`.

### DIFF
--- a/anchor/validator_store/src/lib.rs
+++ b/anchor/validator_store/src/lib.rs
@@ -2229,6 +2229,15 @@ impl VotingContext {
             }
         }
     }
+
+    /// Return the committee decision when available, otherwise use the slot's local seed vote.
+    fn vote_for_committee(&self, committee_id: &CommitteeId) -> SlotVote {
+        self.decided_votes
+            .lock()
+            .get(committee_id)
+            .cloned()
+            .unwrap_or_else(|| self.vote.clone())
+    }
 }
 
 /// The slot's agreed attestation vote, tagged by fork.
@@ -3913,6 +3922,23 @@ mod tests {
             target: distinct_checkpoint(root_byte as u64 + 1, root_byte.wrapping_add(2)),
             attestation_data_index: index,
         })
+    }
+
+    #[test]
+    fn voting_context_returns_committee_decision_or_seed() {
+        let seed = test_gloas_vote(0x10, 0);
+        let committee_a = CommitteeId([0xA1; 32]);
+        let committee_b = CommitteeId([0xB2; 32]);
+        let decided_a = test_gloas_vote(0x21, 1);
+        let context = test_voting_context(seed.clone());
+
+        assert_eq!(context.vote_for_committee(&committee_a), seed);
+        context
+            .remember_decided_vote(committee_a, decided_a.clone())
+            .expect("first committee decision should be stored");
+
+        assert_eq!(context.vote_for_committee(&committee_a), decided_a);
+        assert_eq!(context.vote_for_committee(&committee_b), seed);
     }
 
     #[test]
--- a/anchor/validator_store/src/lib.rs
+++ b/anchor/validator_store/src/lib.rs
@@ -3924,6 +3924,8 @@ mod tests {
         })
     }
 
+    /// Verifies committee-local lookup: a committee with a recorded decision gets that vote, while
+    /// a different undecided committee still falls back to the slot seed.
     #[test]
     fn voting_context_returns_committee_decision_or_seed() {
         let seed = test_gloas_vote(0x10, 0);
--- a/anchor/validator_store/src/lib.rs
+++ b/anchor/validator_store/src/lib.rs
@@ -1426,7 +1426,7 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> AnchorValidator
         &self,
         committee_id: CommitteeId,
         slot: Slot,
-        vote: SlotVote,
+        voting_context: &VotingContext,
         validator_attestation_committees: HashMap<PublicKeyBytes, u64>,
         cluster_members: &IndexSet<OperatorId>,
     ) -> Result<DecidedVote, Error> {
@@ -1440,7 +1440,7 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> AnchorValidator
             instance_height: slot.as_usize().into(),
         };
 
-        let result = match vote {
+        let result = match voting_context.vote.clone() {
             SlotVote::Gloas(seed) => {
                 let completed = self
                     .consensus
@@ -1459,13 +1459,7 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> AnchorValidator
 
                 match completed {
                     Completed::TimedOut => Err(Error::SpecificError(SpecificError::Timeout)),
-                    Completed::Success(decided) => Ok(DecidedVote {
-                        block_root: decided.block_root,
-                        source: decided.source,
-                        target: decided.target,
-                        decided_hash: decided.hash(),
-                        decided_index: Some(decided.attestation_data_index),
-                    }),
+                    Completed::Success(decided) => Ok(SlotVote::Gloas(decided)),
                 }
             }
             SlotVote::Base(seed) => {
@@ -1483,18 +1477,20 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> AnchorValidator
 
                 match completed {
                     Completed::TimedOut => Err(Error::SpecificError(SpecificError::Timeout)),
-                    Completed::Success(decided) => Ok(DecidedVote {
-                        block_root: decided.block_root,
-                        source: decided.source,
-                        target: decided.target,
-                        decided_hash: decided.hash(),
-                        decided_index: None,
-                    }),
+                    Completed::Success(decided) => Ok(SlotVote::Base(decided)),
                 }
             }
         };
         drop(timer);
-        result
+        let decided_vote = result?;
+        voting_context.remember_decided_vote(committee_id, decided_vote.clone())?;
+        Ok(DecidedVote {
+            block_root: decided_vote.block_root(),
+            source: decided_vote.source(),
+            target: decided_vote.target(),
+            decided_hash: decided_vote.qbft_hash(),
+            decided_index: decided_vote.decided_index(),
+        })
     }
 
     /// Sign sync committee messages for all validators in a single SSV committee.
@@ -1526,7 +1522,7 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> AnchorValidator
             .decide_committee_vote(
                 committee_id,
                 slot,
-                voting_context.vote.clone(),
+                &voting_context,
                 validator_attestation_committees,
                 &cluster.cluster_members,
             )
@@ -1618,7 +1614,7 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> AnchorValidator
             .decide_committee_vote(
                 committee_id,
                 slot,
-                voting_context_tx.vote.clone(),
+                &voting_context_tx,
                 validator_attestation_committees,
                 &cluster.cluster_members,
             )
@@ -2198,6 +2194,41 @@ struct VotingContext {
     /// The fork-tagged attestation vote for this slot (only available at 1/3 slot from the
     /// beacon node).
     vote: SlotVote,
+    /// Committee decisions completed during this slot, used by the later aggregation phase.
+    decided_votes: Mutex<HashMap<CommitteeId, SlotVote>>,
+}
+
+impl VotingContext {
+    /// Remember the first decided vote for a committee in this slot.
+    ///
+    /// Attestation and sync-message signing share one QBFT instance, so both callers may observe
+    /// the same completion. A different second value would violate that shared-instance invariant;
+    /// keep the first value so aggregation cannot be redirected by a late conflicting write.
+    fn remember_decided_vote(
+        &self,
+        committee_id: CommitteeId,
+        vote: SlotVote,
+    ) -> Result<(), SpecificError> {
+        let mut decided_votes = self.decided_votes.lock();
+        match decided_votes.entry(committee_id) {
+            std::collections::hash_map::Entry::Vacant(entry) => {
+                entry.insert(vote);
+                Ok(())
+            }
+            std::collections::hash_map::Entry::Occupied(entry) => {
+                if entry.get() == &vote {
+                    Ok(())
+                } else {
+                    Err(SpecificError::CommitteeDecisionConflict {
+                        slot: self.voting_assignments.slot,
+                        committee_id,
+                        existing_decided_hash: entry.get().qbft_hash(),
+                        new_decided_hash: vote.qbft_hash(),
+                    })
+                }
+            }
+        }
+    }
 }
 
 /// The slot's agreed attestation vote, tagged by fork.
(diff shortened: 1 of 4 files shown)


shane-moore:
@claude, please take a look

shane-moore:
@claude, please take a look

shane-moore:
@claude, for the three observations:

1. **Fail-closed on `remember_decided_vote` conflict is intentional.** The shared-instance invariant is structural: both signing paths hold the same slot-scoped `Arc<VotingContext>` (same seed vote, same instance key), and the QBFT instance sends the identical decided value to every `on_completed` waiter and replays the stored value to post-decision initializers. `TimedOut` returns before the cache write, so a conflict requires two divergent `Success` values, which would need the instance map entry cleaned up mid-context plus a different re-decision at the same height. If that ever happened, aborting is preferable to signing a vote that diverges from what aggregation will use.

2. Keeping the repeated `aggregate_fetch_key` as-is: a few extra hashes per aggregator once per slot, and recomputing keys from the committee's own vote is the isolation property the design leans on.

3. `CommitteeDecisionConflict` cannot reach either wildcard in practice: `failure_reason` is only called from the proposer spans and `classify_ptc_collection_failure` only from PTC collection failures, neither of which is on the committee path where this error originates. The arms are compile-time fallthrough only.

anchor/validator_store/src/metadata_service.rs
```diff
@@ -698,45 +730,68 @@ impl<E: EthSpec, T: SlotClock + 'static> MetadataService<E, T> {
...
+        let ssv_committees: HashSet<CommitteeId> = attesters_by_ssv_committee
+            .keys()
+            .chain(sync_by_ssv_committee.keys())
+            .copied()
+            .collect();
+        let mut committee_votes = HashMap::with_capacity(ssv_committees.len());
+        let mut aggregate_requests = HashSet::new();
+        let mut sync_requests = HashSet::new();
+
+        for ssv_committee_id in &ssv_committees {
```
jnhsigmap, line 745: This is a bit nit-picky but here `build_consensus_data_for_all_committees` creates and populates `committee_votes` in the first loop by iterating `ssv_committees`, then the second loop iterates over `ssv_committees` again and returns `Err` if no `vote` is found for a `ssv_committee_id`?
```
  let vote = committee_votes
      .get(&ssv_committee_id)
      .ok_or_else(|| format!("Missing vote for committee {ssv_committee_id:?}"))?;
```

`committee_votes` contains all of the `ssv_committee_id` values that we want, so `ok_or_else` technically should never be reached. Could we loop over `committee_votes` directly here?:
```
  let mut result = HashMap::with_capacity(committee_votes.len());
  for (ssv_committee_id, vote) in &committee_votes {
      let consensus_data = self.build_consensus_data_for_committee(
          slot,
          ssv_committee_id,
          attesters_by_ssv_committee.get(ssv_committee_id),
          sync_by_ssv_committee.get(ssv_committee_id),
          vote,
          &fetch_results,
      )?;
      if let Some(data) = consensus_data {
          result.insert(*ssv_committee_id, Arc::new(data));
      }
  }
```

anchor/validator_store/src/lib.rs
```diff
@@ -3845,4 +3906,93 @@ mod tests {
...
+        SlotVote::Gloas(GloasBeaconVote {
+            block_root: Hash256::repeat_byte(root_byte),
+            source: distinct_checkpoint(root_byte as u64, root_byte.wrapping_add(1)),
+            target: distinct_checkpoint(root_byte as u64 + 1, root_byte.wrapping_add(2)),
+            attestation_data_index: index,
+        })
+    }
+
+    #[test]
+    fn voting_context_returns_committee_decision_or_seed() {
```
jnhsigmap, line 3930: could be worth adding a doc comment like _"Tests `vote_for_committee` lookup rule. Does a committee get its own decided vote back once one is recorded, and does an undecided committee fall back to the slot seed."_ to make it clear what is being tested?

anchor/validator_store/src/metadata_service.rs
```diff
@@ -1638,6 +1721,182 @@ mod tests {
...
+        assert_ne!(sync_a, sync_divergent);
+        assert_ne!(sync_a, sync_other_subnet);
+        assert_eq!(
+            HashSet::from([sync_a, sync_identical, sync_divergent, sync_other_subnet,]).len(),
+            3
+        );
+    }
+
+    #[test]
+    fn build_consensus_data_uses_composite_results_and_preserves_wire_order() {
```
jnhsigmap, line 1946: fair bit going on here, test itself is sound, just could do with an explanation of objectives.

> Objectives:
>   1. Foreign-result isolation: Results fetched under a _different vote_ must never be consumed by a committee with `foreign_vote`.
>   2. Wire ordering: De-duplicated/surviving aggregators/contributors and their index lists come out in the exact go-ssv-compatible order as per _"The sorting order MUST match SSV-Go exactly for consensus compatibility."_

jnhsigmap (approved):
Approved b/c none of these are blocking but just some suggestions from claude and analysis 🙂 
another one that came up was an integration test for `build_consensus_data_for_all_committees` to ensure that it feeds each committee's decided vote and not the seed into the fetch keys when the two committees differ.

anchor/validator_store/src/metadata_service.rs
shane-moore: Resolved in [`e62322f7`](https://github.com/sigp/anchor/commit/e62322f7be1ec8b3ea0adec44eccdd66489b2340): iterate `committee_votes` directly and remove the unreachable missing-vote branch.

anchor/validator_store/src/lib.rs
shane-moore: Resolved in [`adc3858c`](https://github.com/sigp/anchor/commit/adc3858c7d94b0db8c126a3fd7627e95017fbf3d): document the committee decision lookup and seed-fallback behavior covered by the test.

anchor/validator_store/src/metadata_service.rs
shane-moore: Resolved in [`73159658`](https://github.com/sigp/anchor/commit/731596587c460829533425aea87d2690b5fc4bd0): document the composite-key isolation and go-ssv-compatible wire-ordering objectives.

PR #1130, opened 2026-07-10
ci: raise Claude review turn budget and bump to Opus 4.8

shane-moore:

Both Claude review workflows failed on PR #1128 with `Reached maximum number of turns (50)`, losing all review work roughly 13 minutes in and posting nothing:

- `claude-pr-review`: [run 29113701689](https://github.com/sigp/anchor/actions/runs/29113701689) - ~15 of 50 turns were wasted fighting the diff: `gh pr diff` output (~84KB) overflowed the tool-output limit, every attempt to save it to a file was blocked by the tool allowlist (no `Write`), and the agent fell back to re-paging the diff one `sed` slice per turn.
- `claude-mentions`: [run 29115026820](https://github.com/sigp/anchor/actions/runs/29115026820) - only 5 permission denials, so minimal friction, and it still exhausted 50 turns mid-review. The budget itself is too small for a ~1,400-line PR at Opus pace (~15s/turn, one tool call per turn).

Each failed attempt costs ~$3.30 in Bedrock spend with zero output.

## Change Overview

- Raise `--max-turns` 50 -> 80 and `timeout-minutes` 15 -> 25 in both workflows. These are the same limit at Opus pace (50 turns ~= 13.5 min), so they must move together.
- `claude-pr-review` only: pre-stage the PR diff into `.claude-pr-diff.txt` in a workflow step and point the prompt at it, so the agent reads it with the paginated `Read` tool instead of burning turns on `gh pr diff`. The file lives only on the ephemeral runner; nothing commits or uploads it.
- `review.md`: instruct the agent to post each inline comment as soon as it is confirmed, so a capped run still delivers partial findings instead of losing everything.
- Bump the model to `us.anthropic.claude-opus-4-8` in both workflows: same Bedrock pricing ($5/$25 per MTok), no API changes from 4.7, stronger at code review.

Intentionally unchanged: the tool allowlists and the silent-watchdog prompt semantics.

- Worst-case cost/latency per review rises (80 turns / 25 min ceiling). Typical runs finish well under the old cap, so expected cost is nearly flat; only pathological runs use the headroom.
- Opus 4.8 follows conservative-reporting instructions more literally than 4.7, so borderline findings may drop slightly. That matches the silent-watchdog intent.
- Requires Opus 4.8 model access in the AWS account (us-west-2). If missing, the run fails fast with an access error rather than burning the turn budget.

## Validation

- Both workflow files parse as valid YAML.
- Failure analysis is from the run logs of the two failed jobs (turn counts, permission denials, and per-turn pacing).
- The end-to-end check requires this to land on `stable` first (`pull_request_target`/`issue_comment` run workflow files from the default branch); after merge, applying the `claude-recheck` label to PR #1128 re-runs the review and doubles as verification of Bedrock Opus 4.8 access.

## Rollback

Revert the commit. No config, data, or operational impact beyond the workflows themselves.

After merge: apply `claude-recheck` to PR #1128 to confirm the review completes within budget.

### DIFF
--- a/.github/prompts/review.md
+++ b/.github/prompts/review.md
@@ -20,5 +20,7 @@ Do NOT comment on:
 
 When you DO find issues:
 - Use inline comments with concrete fix suggestions
+- Post each inline comment as soon as the issue is confirmed; do not save
+  them all up for the end of the review
 - Post a brief summary comment ONLY listing the issues found
 - No preamble, no praise, no filler
--- a/.github/workflows/claude-mentions.yml
+++ b/.github/workflows/claude-mentions.yml
@@ -83,14 +83,14 @@ jobs:
 
             - name: Run Claude Code Action
               if: steps.check.outputs.is_member == 'true'
-              timeout-minutes: 15
+              timeout-minutes: 25
               uses: anthropics/claude-code-action@v1.0.127
               with:
                   github_token: ${{ steps.app-token.outputs.token }}
                   use_bedrock: "true"
                   claude_args: |
-                      --max-turns 50
-                      --model us.anthropic.claude-opus-4-7
+                      --max-turns 80
+                      --model us.anthropic.claude-opus-4-8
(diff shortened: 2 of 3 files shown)


jnhsigmap (approved):
looks good - I wonder if it needs to be told explicitly not to duplicate commentary and to ensure that issues are logically segregated into mutually exclusive components. This would prevent it from wasting tokens examining something that it has already addressed previously - for example if an issue is confirmed at a surface level and a comment is posted, then further exploration results in a deeper root cause or wider blast radius being identified. An attempt to totally re-resolve may leave a previous comment defunct? I guess we can only find out by observing here.

jnhsigmap:
The `test-suite` checks should be fixed on `unstable`. Can ignore them I believe

Issue #1120, opened 2026-07-08
feat(ssv_types): add Role::EnvelopeProposer with message-validator fork gate


shane-moore:
## Goal

Add validator-scoped QBFT `Role::EnvelopeProposer` with wire byte `[9, 0, 0, 0]`, matching SIP-94 `RoleEnvelopeProposer = 9`, and bind it to `PartialSignatureKind::PostConsensus`.

Wire the compile-coupled `Role` match sites in `ssv_types`, `message_validator`, and `qbft_manager`, gate the role on the Ethereum Gloas fork, and leave a transient `qbft_manager` reject for the follow-up QBFT routing issue to replace.

## Context / motivation

SIP-94 §6 models self-build envelope signing as a second proposer-scoped QBFT duty after the §4 block is published. SIP-94 §7 pins the message-validation contract for this role:

- QBFT role, round cut-off `2`.
- `PostConsensusPartialSig` only, one partial signature per `(signer, slot)`.
- Message slot is the proposal slot.
- No earliness allowance, 3-slot lateness TTL.
- Proposer assignment check at the slot, but only once the epoch's proposer duties are known.
- Duty cap is `SLOTS_PER_EPOCH`.

Live SIP-94 and ssv-spec #632 name the role `RoleEnvelopeProposer`. The current go-ssv PR #2901 still names the same wire role `RoleEnvelopeBuilder`, but the relevant parity target is the shared wire byte `9` plus validation behavior.

Adding this role breaks exhaustive matches. This issue intentionally bundles the compile-coupled validation and temporary routing edits so the workspace remains buildable.

## Suggested approach

- `ssv_types/src/msgid.rs`: add `Role::EnvelopeProposer`, encode/decode it as `[9, 0, 0, 0]`, keep it out of `is_committee_role()`, set `max_round()` to `Some(2)`, and map its `MessageId` executor to `DutyExecutor::Validator`.
- `ssv_types/src/partial_sig.rs`: no new kind. Reuse `PartialSignatureKind::PostConsensus`.
- `message_validator/src/lib.rs`: add `EnvelopeProposer` to the validator-role committee lookup, the Gloas fork gate, the short TTL bucket, `create_message_id_for_test`, and `duty_limit`.
[...]

PR #1134, opened 2026-07-14
fix(message_validator): treat duty-limit breach as ignore, not reject

### LINKED ISSUE #1118: fix: duty-limit breach is classified REJECT, should be IGNORE
## Summary

A message-validation duty-limit breach constructs `ExcessiveDutyCount`, which is not in the IGNORE arm of the classification match, so it falls to the `_ => Reject` default. go-ssv and SIP-94 §7 classify duty-limit breaches as IGNORE. Anchor therefore gossip-penalizes a peer for relaying an over-limit message that should just be dropped.

## Current behavior

- `message_validator/src/lib.rs:996` constructs `ValidationFailure::ExcessiveDutyCount { .. }` on a duty-limit breach.
- `From<&ValidationFailure> for MessageAcceptance` (`lib.rs:233-260`) lists the IGNORE variants explicitly and defaults everything else to REJECT. `ExcessiveDutyCount` is not listed, so it maps to REJECT.
- Quirk: `TooManyDutiesPerEpoch` *is* in the IGNORE arm (`lib.rs:252`) but is never constructed anywhere, i.e. it is dead.

## Expected

A duty-limit breach is a rate condition, not a provable protocol violation: an honest relayer can forward a message that pushes a signer over its per-epoch count. go-ssv's `ErrTooManyDutiesPerEpoch` has no reject flag (IGNORE), and SIP-94 §7 pins duty limits as IGNORE for all three new roles.

## Fix

Map `ExcessiveDutyCount` to IGNORE (add it to the IGNORE arm), or switch the construction site to `TooManyDutiesPerEpoch` and delete the now-redundant variant. Add a regression test asserting IGNORE on a duty-limit breach.

Refs at anchor `upstream/epbs` `35191521`, go-ssv `epbs-gloas` `90df7fa0`; SIP-94 §7.

jnhsigmap:
Closes #1118

When a signer sends more messages than its allowed number of duties for an epoch, the message validator rejected the message. A reject also lowers the sending peer's gossip score, which can lead to an honest peer being disconnected.

Going over the duty limit is a rate condition and not proof that a peer is misbehaving. An honest relayer can forward a message that happens to push a signer over its limit. We should quietly drop (ignore) the message instead.

Evidence:
- The Go client (`go-ssv`) treats this case as ignore. Its error carries no reject flag.
- `SIP-94` section 7 lists the duty limit as an ignore case for the new roles.
- In our code the check builds a `ValidationFailure` value that was not listed in the ignore group, so it fell through to the default, which is reject.

## Change Overview (Required)

- A duty-limit breach now classifies as **ignore** rather than **reject**.
- Removed a second unused failure value (`TooManyDutiesPerEpoch`). It was already listed as ignore but was never used anywhere. The value produced is the one now correctly marked as ignore. It also carries additional details: the count seen, the limit, and the role.

What intentionally did not change: the point where the limit is checked, the limit values themselves, and the handling of every other failure value. This is a classification-only change.

- Blast radius is very small: one file, one crate, classification only.
- Trade-off: an ignored message produces no gossip penalty which is correct.
- Mitigation: the check itself is unchanged, so over-limit messages are still not accepted or acted on but instead dropped without penalising the sender.

## Validation (Required)

- New regression test asserting that a duty-limit breach maps to ignore.
- Full crate test run: `cargo test -p message_validator`: 64 passed, 0 failed
- `make cargo-fmt-check`

- Revert the single commit. No config, data, or migration impact.
- The change only affects how one failure case is scored for gossip, so reverting simply restores the previous reject behaviour.

- It is a prerequisite for the follow-up work in #1120, which relies on the duty limit mapping to ignore for the new roles.

- This was a pre-existing issue affecting every role that has a per-epoch duty limit, not just the new roles. Fixing it here makes the behaviour correct before the follow-up work depends on it.

### DIFF
--- a/anchor/message_validator/src/lib.rs
+++ b/anchor/message_validator/src/lib.rs
@@ -118,7 +118,6 @@ pub enum ValidationFailure {
     NonExistentCommitteeID,
     RoundTooHigh,
     ValidatorIndexMismatch,
-    TooManyDutiesPerEpoch,
     NoDuty,
     EstimatedRoundNotInAllowedSpread {
         got: String,
@@ -249,7 +248,7 @@ impl From<&ValidationFailure> for MessageAcceptance {
             | ValidationFailure::RoundTooHigh
             | ValidationFailure::RoundOverflow
             | ValidationFailure::ValidatorIndexMismatch
-            | ValidationFailure::TooManyDutiesPerEpoch
+            | ValidationFailure::ExcessiveDutyCount { .. }
             | ValidationFailure::NoDuty
             | ValidationFailure::EstimatedRoundNotInAllowedSpread { .. } => {
                 MessageAcceptance::Ignore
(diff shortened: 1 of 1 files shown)


shane-moore:
@claude, please take a look

shane-moore (approved):
Nice man, lgtm!

Issue #1092, opened 2026-06-15
research(qbft_manager): committee-duty QBFT hard-sleeps to attestation_due, negating the eager-attest head start


shane-moore:
## Issue Addressed

Research ticket, follow-up to #986. With the head monitor now live on `epbs`, `sign_attestations` fires on the head event (~+0.25s) and `VotingContext` is published early, but the committee QBFT instance then **hard-sleeps to the attestation deadline before starting round 1**, so consensus is still decided at ~+3.0s. This ticket is to confirm the gap, validate feasibility, and settle the design. The proposed direction below is **one candidate and needs more research / spec work**; the outcome should be a spun-out implementation issue/PR or a documented reason not to do it.

## The gap (measured on `epbs`, ssv-mini Gloas devnet, slot 192)

| Offset | Event |
|---|---|
| +0.253s | `Starting attestation production from_head_monitor=true` |
| +0.416s | first QBFT `Received initialization message` (instance seeded) |
| +0.4s to +3.0s | nothing — the instance is sleeping |
| +3.005s | `Starting new round round=1` (sleep wakes at `get_attestation_due` = 3.0s Gloas) |
| +3.020s | `Reached a COMMIT consensus` |

`Uninitialized::initialize` does `sleep_until(instance_start_time)` before round 1 (`qbft_manager/src/instance.rs:120-128`). `instance_start_time` is **dual-purpose**: the same value is the sleep target AND the cumulative round-timeout anchor (`qbft_manager/src/timeout.rs:21-30`; round N deadline = `instance_start_time + cumulative_timeout(N)`). So the sleep cannot simply be moved earlier without also moving every round-change deadline, which would desync operators. The data is in hand at ~+0.4s, but consensus cannot begin until the deadline.

## Why it is built this way

- #303 introduced the sleep to "start the round interval at precisely specified instants" (cross-operator round-change synchronization).
- #787 (closes #758) split block proposals to per-round `Relative` timing after a slot-anchored deadline caused a mainnet missed proposal, but **deliberately kept committee/attestation/sync on the absolute `SlotTime` anchor**.
- Head events were added later (#986) only at the trigger / voting-context layer. The QBFT start gate was never reconsidered for them. The sleep is a holdover, not a head-event-aware choice.

## How go-ssv differs

[...]

PR #1136, opened 2026-07-14
feat(qbft_manager): start SlotTime QBFT instances immediately

### LINKED ISSUE #1133: feat(qbft_manager): start SlotTime QBFT instances immediately, keep the anchor for round deadlines only
## Goal

Start `SlotTime` QBFT instances immediately at initialization instead of sleeping to the slot-relative start instant, and rename `instance_start_time` to `round_anchor` so the field serves only as the round-deadline anchor. `Relative` mode (block proposals) is untouched.

## Context / motivation

Spin-out of #1092, per the validation posted there: https://github.com/sigp/anchor/issues/1092#issuecomment-4952123445.

- `Uninitialized::initialize` hard-sleeps to `instance_start_time` before round 1 (`qbft_manager/src/instance.rs`). With the head monitor, `VotingContext` is published at ~+0.25s but committee consensus cannot begin before `attestation_due` (3.0s Gloas), so decisions land at ~+3.0s regardless of data availability.
- The field is dual-purpose today: sleep target and cumulative round-timeout anchor (`qbft_manager/src/timeout.rs`). Round-change deadlines must stay anchored to the absolute slot-relative instant so operators remain synchronized; only the start gate should move.
- Every `SlotTime` call site is already data-gated by its caller (committee: the VotingContext head-event/`attestation_due` race; aggregators: Phase-3 data published at 2/3 slot), so immediate start reproduces go-ssv's model (scheduler-level gate, immediate round 1, absolute round anchors), which is pre-existing, non-fork-gated mainnet behavior.
- The sleep also pins `urgent_consensus` processor permits: the Initialize work item holds its permit until `initialize()` returns, i.e. through the sleep. An operator with more committees than cores can have the whole pool pinned for ~2.5s per slot, delaying its own message sends in the critical window.
- Wire-safe: message validation checks earliness against slot start only and bounds rounds from above only, and go-ssv already emits early round-1 messages on mainnet.

## Implementation approach

- Delete the `SlotTime` sleep in `Uninitialized::initialize`; keep `Relative`'s sleep-then-reanchor behavior unchanged.
- Rename `SlotTime { instance_start_time }` to `SlotTime { round_anchor }`. Round N still times out at `round_anchor + cumulative_timeout(N)`; an instance initialized after the anchor starts immediately and expired deadlines cascade round changes until rounds catch up.
- No fork gate: cadence already matches go-ssv per fork, and the early start is not fork-dependent behavior.
- No propagation delay: not correctness-critical (lenient value-check, pre-init message buffer, LH's head arm root-checks fetched data). If devnet data shows round-1 failures, a delay belongs at the metadata-service head-event handler, not in `qbft_manager`.
- `--disable-beacon-head-monitor` remains the operational killswitch: without head events, fallback timing is identical to today.

## Acceptance criteria

- The pre-round-1 sleep is removed for `SlotTime`; `Relative` behavior is unchanged.
- Mock-clock unit tests assert round-N deadlines are invariant under early initialization, and that late initialization (after the anchor) starts immediately and cascades expired rounds to the correct anchor-based deadline.
- A multi-operator init-skew integration test converges (the `qbft_manager` harness supports per-operator init delay).
- The PR description notes the `CONSENSUS_TIMES` distribution shift: the histogram previously included the sleep, so committee-path values had a floor near `attestation_due`; post-change it measures true consensus latency and dashboards/alerts derived from it will move.

Closes #1092 when this lands (research parent; its text asks for this spin-out). Related: #1094's motivation explicitly anticipates this change ("measure the committee duty critical path end to end before changing committee QBFT start timing").

shane-moore:

Committee QBFT hard-sleeps to `attestation_due` (3.0s Gloas) before starting round 1, even when the head monitor has already delivered attestation data earlier in the slot. Consensus is therefore decided at ~+3.0s by construction, negating the eager-attest head start and leaving little margin under the tightened Gloas deadline. The sleeping task also pins an `urgent_consensus` processor permit for the full sleep.

Measured trace and full validation: #1092 and https://github.com/sigp/anchor/issues/1092#issuecomment-4952123445.

Closes #1133.

## Change Overview (Required)

**The real change is one hunk in `qbft_manager/src/instance.rs`**: `Uninitialized::initialize` no longer sleeps for `TimeoutMode::SlotTime`, so round 1 starts as soon as the instance is initialized. Every `SlotTime` call site is already data-gated by its caller, so the start now follows data availability instead of a fixed clock offset.

Everything else is the supporting rename: the field previously did two jobs (sleep target and round-deadline base) and now does one, so `instance_start_time` becomes `round_deadline_origin` (`lib.rs`, `timeout.rs`, four construction sites in `validator_store`, tests).

Suggested reading order: `instance.rs`, then `timeout.rs` (deadline math, unchanged), then the tests.

What did NOT change:

- **Round deadlines are byte-for-byte identical**: round N still times out at `round_deadline_origin + cumulative_timeout(N)` (slot-time based, 2s per round), so round-change cadence stays synchronized across operators regardless of when each one starts. The untouched expected values in the pre-existing timing tests demonstrate this.
- `Relative` mode (block proposals) is unchanged.
- The aggregator-path sleeps were already no-ops (callers fire at 2/3 slot).

This is go-ssv parity: it starts committee consensus immediately once duty data is ready and anchors round timeouts to absolute slot time, as pre-existing mainnet behavior.

- Round-1 messages now appear on the wire before `attestation_due`. Both Anchor's and go-ssv's message validators accept early round-1 messages (earliness is checked against slot start), and go-ssv operators already send them on mainnet, so mixed clusters exercise this path today. Older Anchor peers buffer and replay them at their own initialization.
- `--disable-beacon-head-monitor` remains a full operational killswitch: without head events, timing is identical to today.
- `CONSENSUS_TIMES` previously included the sleep; it now measures true consensus latency, so dashboards derived from it will shift left.

## Validation (Required)

- New mock-clock tests: round-N deadlines invariant under early initialization; late initialization cascades expired rounds immediately and still times out at the origin-based deadline.
- New multi-operator init-skew integration test (operators initializing 0-5s apart converge; the harness now shares one deadline origin across operators, matching production).
- `cargo test -p qbft_manager` (34) and `cargo test -p anchor_validator_store` (54) pass; fmt and clippy clean.

Revert the commit. No config, database, or network migration; `--disable-beacon-head-monitor` disables the early trigger operationally without a rollback.

### DIFF
--- a/anchor/qbft_manager/src/instance.rs
+++ b/anchor/qbft_manager/src/instance.rs
@@ -116,23 +116,16 @@ impl Uninitialized {
         init: QbftInitialization<D>,
         sender: &Arc<dyn MessageSender>,
     ) -> Initialized<D> {
-        // Sleep until the start time embedded in the timeout mode
-        let start_time = match init.timeout_mode {
-            TimeoutMode::SlotTime {
-                instance_start_time,
-            } => instance_start_time,
-            TimeoutMode::Relative {
-                current_round_start_time,
-            } => current_round_start_time,
-        };
-        tokio::time::sleep_until(start_time).await;
-
-        // For Relative mode, reset to Instant::now() after the sleep
+        // `SlotTime` instances start round 1 immediately: every call site is data-gated by
+        // its caller, and `round_deadline_origin` alone keeps round-change deadlines synchronized
+        // across operators. `Relative` instances (block proposals) still wait for the
+        // scheduled round start and then restart the round timer from now for per-round cadence.
         let mut timeout_mode = init.timeout_mode;
         if let TimeoutMode::Relative {
             current_round_start_time,
         } = &mut timeout_mode
         {
+            tokio::time::sleep_until(*current_round_start_time).await;
             *current_round_start_time = Instant::now();
         }
 
--- a/anchor/qbft_manager/src/lib.rs
+++ b/anchor/qbft_manager/src/lib.rs
@@ -52,9 +52,12 @@ const QBFT_RETAIN_SLOTS: u64 = 1;
 /// Determines how round timeouts are calculated.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum TimeoutMode {
-    /// Cumulative timeouts from instance start. Never resets.
+    /// Cumulative timeouts measured from a fixed slot-relative instant: round N times out at
+    /// `round_deadline_origin + cumulative_timeout(N)`. The instance starts round 1 immediately
+    /// upon initialization; the origin only governs round-change deadlines, keeping the round
+    /// cadence synchronized across operators regardless of when each one initializes.
     /// Used for: attestations, aggregations, sync committee.
-    SlotTime { instance_start_time: Instant },
+    SlotTime { round_deadline_origin: Instant },
     /// Per-round timeouts. Resets on round changes.
     /// Used for: block proposals.
     Relative { current_round_start_time: Instant },
--- a/anchor/qbft_manager/src/tests.rs
+++ b/anchor/qbft_manager/src/tests.rs
@@ -355,6 +355,12 @@ where
     ) -> UnboundedReceiver<(Hash256, Result<Completed<D>, QbftError>)> {
         let (result_tx, result_rx) = mpsc::unbounded_channel();
 
+        // All operators share the same round-deadline origin, mirroring production where the origin
+        // is derived from the slot clock rather than each operator's own initialization
+        // time. Operators initialized with a delay start with (partially) expired round
+        // deadlines and must catch up to the committee's round cadence.
+        let round_deadline_origin = Instant::now();
+
         for (data, data_id) in all_data {
             let height = *data.instance_height(&data_id) as u64;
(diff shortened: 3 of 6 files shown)


shane-moore:
@claude, please take a look

shane-moore:
@claude Re the non-blocking observations:

**1. Implicit coupling.** Intended design: callers own scheduling and data-gating, `qbft_manager` starts when told. `QbftInitialization` carries the consensus data itself, so no call site can start without data; the residual concern is only freshness, which is inherently caller-side. Round cadence is correct regardless of start time. A compile-time guard would move scheduling policy back into `qbft_manager`, which is the coupling this PR removes.

**2. Leader assumption in `run_slottime_scenario`.** Intentional and documented in the test; if `DefaultLeaderFunction` changes, the assertion fails loudly at the explaining comment.

Codecov: the two missed lines (1010, 1162) are renames inside pre-Boole paths that were uncovered before this PR and are marked for post-Boole removal.

jnhsigmap (approved):
couldn't find any issues, looks good to me 👍

Issue #1115, opened 2026-07-06
feat(message_validator): reject ValidatorRegistration partial-sig at/after Gloas


shane-moore:
## Goal

Reject inbound `ValidatorRegistration` partial-signature messages whose envelope
(send) slot is at or after the Ethereum Gloas fork. SIP-94 §5 deprecates the
`ValidatorRegistration` duty at Gloas: proposer preferences replace relay
registrations, and the relay path is gone with blinded blocks. This is the
receive-side half of that deprecation; the emit-side half (stop signing/broadcasting
post-Gloas) is #1065.

Add a fourth branch to `validate_role_for_fork` (`message_validator/src/lib.rs:849`)
and a new `RoleNotActiveAfterEthFork` error variant. Do NOT remove
`Role::ValidatorRegistration` (wire byte `[4, 0, 0, 0]`, `msgid.rs:34`),
`PartialSignatureKind::ValidatorRegistration = 4` (`partial_sig.rs:34`), or
`BEACON_ROLE_VALIDATOR_REGISTRATION = BeaconRole(5)` (`consensus.rs:490`, the
`BNRoleValidatorRegistration` the SIP names): SIP-94 §5 reserves those numeric values
for pre-Gloas operation and backward-compat decoding.

## Context / motivation

`validate_role_for_fork` already gates role-by-fork in both directions and on both
the SSV `Fork` axis and the Ethereum `ForkName` axis:
- `RoleNotActiveBeforeFork` (SSV): `AggregatorCommittee` before Boole.
- `RoleNotActiveAfterFork` (SSV): `Aggregator` / `SyncCommittee` deprecated after Boole.
- `RoleNotActiveBeforeEthFork` (eth): `PTCAttester` before Gloas.

The `ValidatorRegistration` deprecation is the missing symmetric case:
`RoleNotActiveAfterEthFork` on the eth axis. It reuses the exact `PTCAttester`
mechanism (`spec.fork_name_at_epoch(epoch).gloas_enabled()`, lib.rs:877) with inverted
polarity: reject when Gloas IS enabled.

`validate_role_for_fork` runs on the partial-sig path at `partial_signature.rs:36`
with `messages.slot` (the current send slot), so a single branch there covers every
inbound `ValidatorRegistration` message (the role is non-QBFT; partial-sig is its only
message type). The gate is defense-in-depth: honest post-Gloas operators emit nothing
(#1065); a Gloas-or-later-slot VR message is a buggy/stale/malicious sender.

[...]

Issue #1064, opened 2026-05-27
feat(client): spawn LH ProposerPreferencesService with AnchorValidatorStore backend


shane-moore:
## Goal

Wire `anchor/client/src/lib.rs` to construct Lighthouse's `ProposerPreferencesService` with `AnchorValidatorStore` as the backend, and start it via `start_update_service()`. The exact in-repo analog is `PayloadAttestationService` (`client/src/lib.rs:826` construction + `:834` start), the other Gloas-scheduled LH service: it uses the same 6-arg direct `::new(...)` constructor and the same no-arg `start_update_service()` that LH's `ProposerPreferencesService` exposes, and it already sits inside the `if spec.is_gloas_scheduled()` gate this service also needs. This completes the ProposerPreferences pipeline: LH's service drives `sign_proposer_preferences` (from #1063) per validator per epoch.

## Context

Per SIP-94 §5 ([ssvlabs/SIPs#94](https://github.com/ssvlabs/SIPs/pull/94)), Lighthouse provides `ProposerPreferencesService` as part of its validator-services crate. Anchor inherits duty fetching, slot timing, and per-validator dispatch from LH (per Anchor's LH-constraint operating principle); the only Anchor-side wiring is the client startup that constructs the service and spawns the update loop.

After this PR and its upstream emission dependency land, the pipeline is operational: the LH service schedules each required preference → #1063's `sign_proposer_preferences` runs single-validator partial-sig collection → reconstructed `SignedProposerPreferences` is submitted by LH to the BN via the `post_validator_proposer_preferences` endpoint. Wiring alone does not cover a validator added to an epoch whose `(epoch, dependent_root)` was already published; that remaining trigger gap is recorded below.

The current LH pin (`81d576943`) already provides `ProposerPreferencesService`, `Domain::ProposerPreferences`, and the `sign_proposer_preferences` trait method; this issue depends on #1063 for the trait impl. Re-verify the live pin in `Cargo.toml` before trusting the LH `file:line` refs below, which are pinned to `81d576943` and drift on bumps. The service at this pin covers lookahead, pre-fork emission, and `dependent_root` changes, but it does not re-emit for a newly added validator when the epoch's `dependent_root` is unchanged. #1064 therefore also depends on an upstream Lighthouse trigger fix before the milestone can claim SIP-94 §5 end-to-end behavior.

## LH service behavior verified at pin `81d576943` (2026-07-10)

[...]

PR #1137, opened 2026-07-14
feat(validator_store): fork-gate RegistrationService at GLOAS_FORK_EPOCH

### LINKED ISSUE #1065: feat(validator_store): fork-gate RegistrationService at GLOAS_FORK_EPOCH
## Goal

Insert a Gloas short-circuit in `RegistrationService::start_validator_registration_service`'s spawning loop (`anchor/validator_store/src/registration_service.rs:60-92`) so that the periodic `register_validators` call is skipped once the current slot's fork is Gloas (`fork_name_at_slot(slot).gloas_enabled()`). The service stays spawned; the periodic loop sleeps to the next epoch instead of issuing a wasted BN call. Pre-Gloas behavior unchanged.

Do NOT delete the service. Pre-Gloas testnets and the Gloas transition window still need it. SIP-94 §5 deprecates the `ValidatorRegistration` duty at the Gloas fork, its purpose replaced by `SignedProposerPreferences`; the pre-fork relay-builder mechanism it fed is gone along with blinded blocks (§4), so registrations are vestigial under Gloas. Note the pinned LH BN does NOT actually no-op `post_validator_register_validator` (verified: no Gloas gate in the handler at `beacon_node/http_api/src/validator/mod.rs:579`; it still updates the proposer-preparation cache and forwards to the builder regardless of fork, unlike `prepare_beacon_proposer`); the call still runs against a relay flow that no longer feeds block production, which is exactly why Anchor should stop issuing it. A Gloas-aware skip on the Anchor side avoids the wasted RPC call and operator-log noise.

## Context

Per SIP-94 §5 ([ssvlabs/SIPs#94](https://github.com/ssvlabs/SIPs/pull/94)), the pre-Gloas relay-builder registration flow is replaced under Gloas by the new `proposer_preferences` p2p mechanism (handled by #1062, #1063, #1064). The legacy `RegistrationService` continues to call `post_validator_register_validator` against the BN every slot; under Gloas those registrations no longer feed block production (the relay-builder flow is gone, though the pinned LH BN still accepts the call). Anchor should short-circuit the periodic loop so:

- Pre-Gloas testnets continue to publish registrations unchanged (Boole and earlier).
- The transition window (Anchor running post-LH-bump but pre-Gloas-activation on a given network) continues to publish.
- Once the slot clock crosses `GLOAS_FORK_EPOCH`, the periodic loop sleeps without invoking the registration path.

LH's `proposer_preferences_service.rs:77-87` and `payload_attestation_service.rs:88-100` already use this pattern (with inverted polarity, skip-when-not-Gloas). For `RegistrationService`, the polarity is reversed: skip when Gloas is enabled.

## Suggested approach

`anchor/validator_store/src/registration_service.rs`:

| Location | Change |
|---|---|
| `start_validator_registration_service` (around `:70-85`) | Inside the `if let Some(slot) = self.inner.slot_clock.now()` branch, add a Gloas fork-gate that sleeps to the next epoch and `continue`s if `spec.fork_name_at_slot::<S::E>(slot).gloas_enabled()`. |

Sketch (gate sits before the existing `inner.register_validators(slot).await` call):

```rust
if let Some(slot) = self.inner.slot_clock.now() {
    // SIP-94 §5: the ValidatorRegistration duty is deprecated at the Gloas
    // fork (the relay-builder mechanism it fed is gone with blinded blocks,
    // §4). The BN still accepts post_validator_register_validator, but the
    // registration no longer feeds block production, so skip the periodic
    // publish to avoid a vestigial RPC + operator-log noise. The service
    // stays spawned so pre-Gloas networks continue unchanged.
    if spec.fork_name_at_slot::<S::E>(slot).gloas_enabled() {
        let sleep_duration = self
            .inner
            .slot_clock
            .duration_to_next_epoch(S::E::slots_per_epoch())
            .unwrap_or_else(|| slot_duration * S::E::slots_per_epoch() as u32);
        sleep(sleep_duration).await;
        continue;
    }

    let inner = self.inner.clone();
    let executor = inner.executor.clone();
    let future = async move {
        if let Err(e) = inner.register_validators(slot).await {
            error!(error = ?e, "Error during validator registration");
        }
    };
    executor.spawn(future, "validator_registration");
} else {
    error!("Slot clock can not return current slot");
}
```

`spec` is already in scope (cloned at the top of `start_validator_registration_service`); no struct changes to `Inner` required. `S::E` is the generic already used elsewhere in the file (e.g., `S::E::slots_per_epoch()` inside `register_validators`).

### Testability

The `register_validators` function isn't called when the gate skips, so a unit test that exercises the gate directly is awkward (the spawning loop runs forever). Extract the gate decision into a small pure helper so it can be asserted without a runtime:

```rust
fn should_skip_for_gloas<E: EthSpec>(spec: &ChainSpec, slot: Slot) -> bool {
    spec.fork_name_at_slot::<E>(slot).gloas_enabled()
}
```

Tests assert this helper directly (no async, no slot clock, no BN mock).

## Acceptance criteria

- Pre-Gloas slot: `register_validators` is invoked from the spawning loop unchanged (existing behavior).
- Gloas slot: the spawning loop sleeps to the next epoch without invoking `register_validators` or making any BN call.
- Service stays spawned across the fork transition; the loop re-evaluates the gate each iteration.
- In-code comment references SIP-94 §5 (duty deprecation) so a future reader understands why the gate is there.
- `cargo check --workspace` green; existing `RegistrationService` tests unaffected.

Tests to deliver (against the extracted helper):

```rust
#[test]
fn should_skip_for_gloas_returns_true_after_fork() {
    let mut spec = ChainSpec::mainnet();
    spec.gloas_fork_epoch = Some(Epoch::new(0)); // active from genesis
    assert!(should_skip_for_gloas::<MainnetEthSpec>(&spec, Slot::new(0)));
}

#[test]
fn should_skip_for_gloas_returns_false_before_fork() {
    let mut spec = ChainSpec::mainnet();
    spec.gloas_fork_epoch = Some(Epoch::new(100_000)); // far future
    let pre_fork_slot = Slot::new(32); // epoch 1
    assert!(!should_skip_for_gloas::<MainnetEthSpec>(&spec, pre_fork_slot));
}

#[test]
fn should_skip_for_gloas_returns_false_when_unscheduled() {
    let mut spec = ChainSpec::mainnet();
    spec.gloas_fork_epoch = None; // never
    assert!(!should_skip_for_gloas::<MainnetEthSpec>(&spec, Slot::new(100_000)));
}
```

## Risks

- **Do not delete the service.** Pre-Gloas testnets and transition-window operators still need it. Fork-gating to a sleeping no-op preserves backward compatibility; deleting would break Holesky / Hoodi / any non-Gloas network Anchor is run against.
- **Gating choice (`gloas_enabled()` vs explicit epoch comparison).** Using LH's `gloas_enabled()` helper (`consensus/types/src/fork/fork_name.rs:205`) mirrors how LH gates its own Gloas-aware code in `proposer_preferences_service.rs`, `payload_attestation_service.rs`, `block_service.rs`, and elsewhere. If LH refactors the helper, Anchor inherits the change via the LH bump. Acceptable; no Anchor-side abstraction required.
- **Sleep duration unwrap.** `duration_to_next_epoch` returns `Option`; on `None` (slot-clock failure) the fallback is `slot_duration * slots_per_epoch`, matching LH's `proposer_preferences_service.rs:81-83` fallback shape.

## Dependencies

All required LH symbols are already in the current pin (`81d576943`; re-verify against `Cargo.toml`): `ChainSpec::fork_name_at_slot`, `ForkName::gloas_enabled`, and `gloas_fork_epoch: Option<Epoch>`. These predate the proposer-preferences work, so no LH bump is needed. Independent of #1062, #1063, #1064; can land in parallel with the rest of the milestone.

shane-moore:

- Under Gloas, SIP-94 §4 removes the pre-fork relay-builder registration mechanism along with blinded blocks, so `RegistrationService`'s periodic `post_validator_register_validator` calls no longer feed block production; they are wasted RPC and operator-log noise.
- The pinned LH BN still accepts the call (no Gloas gate in its handler), which is exactly why the skip belongs on the Anchor side.
- Closes #1065. SIP: ssvlabs/SIPs#94. Replacement flow: #1062, #1063, #1064. Receive-side twin: #1115.

## Change Overview (Required)

- Adds a Gloas gate at the top of the registration spawning loop: once `fork_name_at_slot(slot).gloas_enabled()`, the loop logs at debug, sleeps to the next epoch, and re-evaluates. No BN call, no signing.
- The gate decision is a pure `should_skip_for_gloas` helper so it is unit-testable without a runtime, mirroring `is_scheduled_for_slot` in the same file.
- Intentionally unchanged: the service is not deleted and its startup is not gated; pre-Gloas behavior is identical apart from the pure check. Inbound `ValidatorRegistration` partial-sig rejection is out of scope (#1115).

- Sleep-and-recheck was chosen over terminating the task: a `break` would rest on a single wall-clock reading, so a transient forward clock step (bad NTP) on a pre-Gloas network would permanently stop registrations until restart. The sleeping loop self-heals next epoch and matches how the sibling fork-aware services (proposer preferences, payload attestation, metadata) gate.
- Cost of keeping the task alive is one timer wakeup and one debug line per epoch.

## Validation (Required)

- Two runtime-free unit tests on the helper: the exact inclusive fork boundary (last pre-fork slot still registers, first fork-epoch slot skips) and the unscheduled-fork case.
- `cargo test -p anchor_validator_store`: 58 passed. Workspace fmt and clippy clean.

- Revert the single commit; no config, data, or wire impact.

N/A. Independent of #1062-#1064; all required LH symbols are already in the current pin.

### DIFF
--- a/anchor/validator_store/src/registration_service.rs
+++ b/anchor/validator_store/src/registration_service.rs
@@ -234,9 +234,8 @@ impl<S: ValidatorStore + 'static, T: SlotClock + 'static> Inner<S, T> {
     }
 }
 
-/// Whether the periodic registration publish should be skipped because the slot's fork is
-/// Gloas. SIP-94 §4: the relay-builder registration flow is gone under Gloas, replaced by
-/// the `proposer_preferences` p2p mechanism.
+/// Whether the periodic registration publish should be skipped because `slot`'s fork is
+/// Gloas or later. See the call site for the rationale.
 fn should_skip_for_gloas<E: EthSpec>(spec: &ChainSpec, slot: Slot) -> bool {
     spec.fork_name_at_slot::<E>(slot).gloas_enabled()
 }
@@ -261,32 +260,6 @@ mod tests {
     use types::{Epoch, MainnetEthSpec};
 
     use super::*;
-
-    #[test]
-    fn should_skip_for_gloas_returns_true_after_fork() {
-        // Arrange: Gloas is scheduled from genesis.
-        let mut spec = ChainSpec::mainnet();
-        spec.gloas_fork_epoch = Some(Epoch::new(0));
-
-        // Act
-        let skip = should_skip_for_gloas::<MainnetEthSpec>(&spec, Slot::new(0));
-
-        // Assert
-        assert!(skip, "registrations must be skipped once Gloas is active");
-    }
-
-    #[test]
-    fn should_skip_for_gloas_returns_false_before_fork() {
-        // Arrange: Gloas is scheduled far in the future; the queried slot is pre-fork.
-        let mut spec = ChainSpec::mainnet();
-        spec.gloas_fork_epoch = Some(Epoch::new(100_000));
-
-        // Act
-        let skip = should_skip_for_gloas::<MainnetEthSpec>(&spec, Slot::new(32));
-
-        // Assert
-        assert!(!skip, "registrations must continue before the Gloas fork");
-    }
 
     #[test]
     fn should_skip_for_gloas_flips_exactly_at_fork_boundary() {
--- a/anchor/validator_store/src/registration_service.rs
+++ b/anchor/validator_store/src/registration_service.rs
@@ -68,6 +68,22 @@ impl<S: ValidatorStore + 'static, T: SlotClock + 'static> RegistrationService<S,
         let validator_registration_fut = async move {
             loop {
                 if let Some(slot) = self.inner.slot_clock.now() {
+                    // SIP-94 §4: under Gloas, the relay-builder registration mechanism is
+                    // removed along with blinded blocks, so registrations no longer feed
+                    // block production. Skip the periodic publish to avoid wasted RPC and
+                    // operator-log noise. The service stays spawned and re-evaluates each
+                    // wakeup so pre-Gloas networks and the transition window are unaffected.
+                    if should_skip_for_gloas::<S::E>(&spec, slot) {
+                        debug!(%slot, "Skipping validator registration, deprecated since Gloas");
+                        let sleep_duration = self
+                            .inner
+                            .slot_clock
+                            .duration_to_next_epoch(S::E::slots_per_epoch())
+                            .unwrap_or_else(|| slot_duration * S::E::slots_per_epoch() as u32);
+                        sleep(sleep_duration).await;
+                        continue;
+                    }
+
                     let inner = self.inner.clone();
                     let executor = inner.executor.clone();
                     let future = async move {
@@ -218,6 +234,13 @@ impl<S: ValidatorStore + 'static, T: SlotClock + 'static> Inner<S, T> {
     }
 }
 
+/// Whether the periodic registration publish should be skipped because the slot's fork is
+/// Gloas. SIP-94 §4: the relay-builder registration flow is gone under Gloas, replaced by
+/// the `proposer_preferences` p2p mechanism.
+fn should_skip_for_gloas<E: EthSpec>(spec: &ChainSpec, slot: Slot) -> bool {
+    spec.fork_name_at_slot::<E>(slot).gloas_enabled()
+}
+
 /// To not sign for all validators at once, select based on the current slot.
 /// Note that it is important that this is the same across client implementations, as else
 /// the nodes broadcast partial signatures for their validators at varying times.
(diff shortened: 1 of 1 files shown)


shane-moore:
@claude, please take a look

shane-moore:
@claude,
1. Yes, deliberate: the spawn-gate polarity that works for PTC (`is_gloas_scheduled()`, post-fork service) has no safe analogue here, since a pre-fork service must spawn on every network today and the fork position is only knowable per-slot. The idle task costs one timer wakeup per epoch; the real cleanup is deleting the whole service once pre-Gloas networks are gone, which subsumes any startup gate.
2. Description updated to match the two consolidated tests.
3. Agreed, no change.

petarjuki7 (approved):
Looks good!

PR #1144, opened 2026-07-16
feat(client): spawn LH ProposerPreferencesService with AnchorValidatorStore backend

petarjuki7:
Part of #1064 (ePBS Proposer Preferences duty milestone, SIP-94 §5).

Wires Lighthouse's `ProposerPreferencesService` (with `AnchorValidatorStore` as the backend) into client startup, inside the existing `is_gloas_scheduled()` gate next to `PayloadAttestationService`. Completes the Anchor-side pipeline: LH's service drives `sign_proposer_preferences` (#1063) per validator per epoch, dispatching both `proposal_data` and the signing through Anchor's `ValidatorStore` impl.

No new tests: the client crate has none and no sibling spawned service is tested — the behavior is LH's; Anchor's role is wiring (compile + startup).

Stacked on #1125 (#1063) — those commits drop out of this diff once it merges.

End-to-end SIP-94 §5 behavior additionally needs the upstream LH fix that schedules newly-added `(dependent_root, proposal_slot, validator)` tuples (the current key is `(epoch, dependent_root)` only); tracked in the issue.

### DIFF
--- a/anchor/client/src/lib.rs
+++ b/anchor/client/src/lib.rs
@@ -63,6 +63,7 @@ use validator_services::{
     latency_service::start_latency_service,
     payload_attestation_service::PayloadAttestationService,
     preparation_service::PreparationServiceBuilder,
+    proposer_preferences_service::ProposerPreferencesService,
     sync_committee_service::SyncCommitteeService,
 };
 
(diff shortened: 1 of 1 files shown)


shane-moore (approved):
LGTM!

PR #1146, opened 2026-07-17
chore: add gossipsub subscription filters

shane-moore:
Configures a whitelist-based subscription filter for gossipsub, following the approach used in Lighthouse. Peers can now only subscribe us to topics that are valid on the network (any subnet of any scheduled fork), with subscription counts bounded by the size of that topic set.

### DIFF
--- a/anchor/network/src/behaviour.rs
+++ b/anchor/network/src/behaviour.rs
@@ -1,6 +1,6 @@
 use std::{hash::Hasher, time::Duration};
 
-use fork::ForkLifecycle;
+use fork::{ForkLifecycle, ForkSchedule};
 use libp2p::{
     gossipsub::{self, ConfigBuilderError, MessageAuthenticity, ValidationMode},
     identify, ping,
@@ -8,6 +8,7 @@ use libp2p::{
     upnp::tokio::Behaviour as Upnp,
 };
 use prometheus_client::registry::Registry;
+use subnet_service::topic;
 use thiserror::Error;
 use tokio::sync::watch;
 use twox_hash::XxHash64;
@@ -16,7 +17,6 @@ use version::version_with_platform;
 
 use crate::{
     Config,
-    behaviour::BehaviourError::Gossipsub,
     discovery::{Discovery, FIND_NODE_QUERY_CLOSEST_PEERS},
     handshake,
     peer_manager::PeerManager,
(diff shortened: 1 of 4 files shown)


anchor/network/src/behaviour.rs
```diff
@@ -25,6 +25,29 @@ use crate::{
...
+/// every subnet of every scheduled fork and nothing else. Note the filter also applies to our
+/// own `subscribe` calls.
+fn subscription_filter(fork_schedule: &ForkSchedule) -> SubscriptionFilter {
+    let possible_topics = topic::whitelist_topic_hashes(fork_schedule);
+    gossipsub::MaxCountSubscriptionFilter {
+        // A peer can at most subscribe to every valid topic.
+        max_subscribed_topics: possible_topics.len(),
+        max_subscriptions_per_request: possible_topics.len(),
+        filter: gossipsub::WhitelistSubscriptionFilter(possible_topics),
+    }
```
dknopik: I think the `MaxCountSubscriptionFilter` is redundant - it offers no benefit over just having the `WhitelistSubscriptionFilter` if it's parameters are just set to the number of the whitelisted topics.

dknopik (commented):
just a nit, otherwise LGTM

anchor/network/src/behaviour.rs
shane-moore: great catch man, makes sense! resolved in c23f5e26

shane-moore:
@claude, please take a look

anchor/network/src/behaviour.rs
```diff
@@ -25,6 +25,29 @@ use crate::{
...
+///
+/// The whitelist is built with the same topic construction used when subscribing, so it covers
+/// every subnet of every scheduled fork and nothing else. Note the filter also applies to our
+/// own `subscribe` calls.
+fn subscription_filter(fork_schedule: &ForkSchedule) -> SubscriptionFilter {
+    let possible_topics = topic::whitelist_topic_hashes(fork_schedule);
+    gossipsub::MaxCountSubscriptionFilter {
+        // A peer can at most subscribe to every valid topic.
+        max_subscribed_topics: possible_topics.len(),
+        max_subscriptions_per_request: possible_topics.len(),
```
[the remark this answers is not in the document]
shane-moore: keeping the bare whitelist: the raw-count check only runs after the RPC is already received and decoded, so it just saves a hash lookup per entry (request size is bounded by transport limits anyway), while adding a whole-RPC rejection path for legit bulk batches

shane-moore:
@claude, ack on the review points, all non-blocking as noted, no action needed

dknopik (approved):
lgtm!

Issue #1122, opened 2026-07-08
feat(qbft_manager): wire EnvelopeProposer per-proposer QBFT instances


shane-moore:
## Goal

Replace the transient `EnvelopeProposer` reject from #1120 with real per-proposer plumbing: `EnvelopeProposerInstanceId`, a fifth instance map, `receive_data` routing with a Gloas defense-in-depth gate, a fifth cleaner retain, and a `QbftDecidable` impl for `EnvelopeConsensusData`.

## Context / motivation

Modeled on the `ProposerConsensusData` path (`DutyExecutor::Validator`), not the committee path. `QbftManager` currently holds four instance maps (`qbft_manager/src/lib.rs:139-147` @ `a9f378a4`); the envelope value cannot reuse `ProposerInstanceId` because it has a different data type and no `ValidatorDutyKind`.

## Suggested approach

- Import `EnvelopeConsensusData` alongside the existing `ssv_types::consensus` values in `qbft_manager/src/lib.rs`.
- `EnvelopeProposerInstanceId { validator: PublicKeyBytes, instance_height: InstanceHeight }` after `ProposerInstanceId` (`qbft_manager/src/lib.rs:79`); no `duty` field (one envelope duty per slot), no new `ValidatorDutyKind` variant.
- Fifth map + `DashMap::new()` init (`qbft_manager/src/lib.rs:139-147`, `qbft_manager/src/lib.rs:174-177`).
- In `receive_data`'s `DutyExecutor::Validator` arm (`qbft_manager/src/lib.rs:277-305`): dedicated branch building the id and calling `pass_to_instance::<EnvelopeConsensusData>`, gated `if !self.spec.fork_name_at_slot::<E>(slot).gloas_enabled() { return Err(QbftError::RoleNotActive); }` (mirror the GloasBeaconVote gate at `qbft_manager/src/lib.rs:321`; `spec` field at `qbft_manager/src/lib.rs:155`). Remove the `TODO(gloas)` marker.
- Fifth `.retain(...)` in the cleaner (`qbft_manager/src/lib.rs:399-405`).
- `impl<E: EthSpec> QbftDecidable<E> for EnvelopeConsensusData` after the existing impls (`qbft_manager/src/lib.rs:472-545`): `type Id = EnvelopeProposerInstanceId`; `message_id` via `MessageId::new(domain, Role::EnvelopeProposer, &DutyExecutor::Validator(id.validator))`, mirroring the proposer impl.
[...]

PR #1149, opened 2026-07-20
feat: add validator-scoped QBFT Role::EnvelopeProposer (#1120)

### LINKED ISSUE #1120: feat(ssv_types): add Role::EnvelopeProposer with message-validator fork gate

jnhsigmap:

SIP-94 introduces ePBS self-build envelope signing (§6) as a second, proposer-scoped QBFT duty that runs after the §4 block is published. Anchor has no `Role` for it, so envelope messages cannot be classified, fork-gated, or validated. This PR adds that role and wires the message-validation contract SIP-94 §7 pins for it.

Because `Role` is a plain enum (not `#[non_exhaustive]`), adding a variant breaks every exhaustive match across `ssv_types`, `message_validator`, and `qbft_manager`. This change intentionally bundles those compile-coupled edits so the workspace stays buildable in one step, and leaves QBFT instance routing as a transient reject for the follow-up (#1122).

- Closes #1120
- Spec: SIP-94 §6 (self-build envelope signing) and §7 (message-validation contract)

## Change Overview (Required)

`Role::EnvelopeProposer` (wire byte `[9, 0, 0, 0]`, matching SIP-94 `RoleEnvelopeProposer = 9`) is a validator-scoped, QBFT, monotonic-slot role bound to `PartialSignatureKind::PostConsensus`. It reuses the existing post-consensus machinery rather than adding a new partial-signature kind; the runner role discriminates routing. Validation is gated on the Ethereum Gloas fork.

The production surface is small (~42 lines): the new variant plus the arms every compile-coupled match site needs. The bulk of the diff is tests (~950 lines) that pin each SIP-94 §7 rule against the real validation entry points.

**Issue Criteria Addressed:**

- **Wire round-trip** — `Role::EnvelopeProposer` encodes/decodes as `[9, 0, 0, 0]` via `From<Role>` / `TryFrom<&[u8]>`. *(msgid.rs)*
- **Classification** — `is_committee_role() == false`, `is_qbft_role() == true`, `max_round() == Some(2)` (its own arm — deliberately **not** grouped with `Role::Proposer`'s `Some(6)`, per §7's cut-off of 2). *(msgid.rs)*
- **Duty executor** — `MessageId::duty_executor()` resolves to `DutyExecutor::Validator`. *(msgid.rs)*
- **Partial-signature kind** — only `PostConsensus` is accepted; any other kind is rejected with `PartialSignatureTypeRoleMismatch`. No new kind added. *(partial_signature.rs)*
- **Packet cardinality** — same-packet `message_count > 1` is rejected via the per-validator bound. *(partial_signature.rs / lib.rs)*
- **Replay guard** — a repeated `PostConsensus` packet for the same signer/slot is rejected by the inherited post-consensus seen-message counter. *(message_counts.rs path; no new arm needed)*
- **Monotonic slot (shared state)** — consensus and post-consensus share one role-specific `MessageId` state; after accepting at slot N, a lower-slot envelope message from the same signer returns `SlotAlreadyAdvanced` (→ `Ignore`). Ordinary `Role::Proposer` state stays isolated by its different role byte. *(inherited non-committee checks; no envelope-specific stale branch)*
- **Fork gate** — pre-Gloas messages are rejected with `RoleNotActiveBeforeEthFork { minimum_fork: ForkName::Gloas }`. *(lib.rs `validate_role_for_fork`)*
- **TTL** — short bucket `1 + LATE_SLOT_ALLOWANCE` (3 slots), not the committee bucket. *(lib.rs)*
- **Beacon duty** — `validate_beacon_duty` rejects a known-epoch non-proposer with `NoDuty`, and tolerates a not-yet-fetched proposer epoch (accepts). *(lib.rs)*
- **Duty cap** — `duty_limit(EnvelopeProposer) == Ok(Some(slots_per_epoch))`. *(lib.rs)*
- **Gossip classification** — `NoDuty` and duty-limit overflow → `Ignore`; fork-gate, kind mismatch, packet-count overflow → `Reject`. *(lib.rs classification test)*
- **QBFT routing** — validator-executor `EnvelopeProposer` returns `RoleNotActive` with a `TODO(#1122)` marker (executor is correct, routing not yet wired — deliberately **not** `InconsistentMessageId`); committee-executor path handles it for exhaustiveness. *(qbft_manager/lib.rs)*

**Intentionally unchanged:** no new `PartialSignatureKind`; no envelope-specific stale-slot branch (the role inherits the existing monotonic checks); `Role::Proposer` behaviour and its `max_round() == Some(6)`; QBFT instance routing (deferred to #1122).

- **Blast radius:** touches shared `Role` match sites, so a missed arm would be a compile error, not a silent runtime bug — the type system is the safety net here. All new arms are additive.
- **Trade-off:** QBFT routing is a transient `RoleNotActive` reject rather than a working instance. This is deliberate and marked `TODO(#1122)`; envelope QBFT duties are inert until that lands, matching the issue's staged plan.
- **One deviation from the issue's suggested shape:** the committee-executor `EnvelopeProposer` case cannot produce `InconsistentMessageId` as literally worded, because `duty_executor()` always decodes bytes 8–55 as a `Validator` pubkey — a committee-executor `EnvelopeProposer` `MessageId` is unconstructable. The role is instead handled in the committee arm for exhaustiveness, and a test documents this encoding constraint. Called out for reviewer awareness.
- **Mitigation:** every §7 rule has a dedicated test exercising the real validation entry point (not a hand-rolled shortcut), so behaviour is pinned rather than assumed.

## Validation (Required)

- `cargo test -p ssv_types` — 111 passed
- `cargo test -p message_validator` — 100 passed (all new EnvelopeProposer coverage + sibling roles)
- `cargo test -p qbft_manager` — 36 passed
- `cargo fmt --check` — clean; `cargo clippy --tests -- -D warnings` — clean
- Coverage is new tests written for this change, exercising production entry points: role byte round-trip, duty executor, QBFT classification, kind binding, repeated-packet rejection, pre-Gloas fork gate, short TTL bucket, known/unknown proposer-epoch behaviour, duty cap, both monotonic-slot directions through the shared `MessageId` state, and the qbft_manager transient/permanent reject behaviour.

Revert the single commit. The role is fork-gated to Gloas (not yet active) and QBFT routing is a no-op reject, so there is no runtime or data impact on current networks. No config or migration involved.

**Reviewer guide — production vs test.** The diff is ~4% production, ~96% test. Read the production arms first (small, mechanical), then spot-check the tests.

*Production (~42 lines) — read these first:*
- `ssv_types/src/msgid.rs` (+6) — the variant, `[9,0,0,0]` codec, `max_round() = Some(2)`, `Validator` executor.
- `message_validator/src/lib.rs` (+22) — committee lookup, fork gate, short TTL, `duty_limit`, and the `validate_beacon_duty` proposer-assignment arm.
- `qbft_manager/src/lib.rs` (+11) — transient `RoleNotActive` reject + `TODO(#1122)`; exhaustiveness in the committee arm.
- `message_validator/src/partial_signature.rs` (+3 production) — the role added to the `PostConsensus` bind and the per-validator packet bound.

*Tests (~950 lines) — verify the contract:*
- `message_validator/src/partial_signature.rs` (+577) — the §7 partial-signature contract: kind bind, packet cap, short TTL (accept + reject), proposer-assignment (3 cases), monotonic equality boundary + both stale-slot directions, replay rejection. A shared `four_node_committee_and_keypair` fixture removes repeated setup across these.
- `message_validator/src/consensus_message.rs` (+111) — duty-cap and the pre-Gloas fork-gate rejection via the consensus entry point.
- `message_validator/src/lib.rs` (+46) — gossip `MessageAcceptance` classification.
- `qbft_manager/src/tests.rs` (+168) — transient/permanent reject behaviour and the committee-executor encoding constraint noted above.
- `ssv_types/src/msgid.rs` (+48) — role byte round-trip and classification pinning.

The single commit is `feat: add validator-scoped QBFT Role::EnvelopeProposer (#1120)`.

### DIFF
--- a/anchor/message_validator/src/lib.rs
+++ b/anchor/message_validator/src/lib.rs
@@ -480,7 +480,8 @@ impl<S: SlotClock + 'static, D: DutiesProvider> Validator<S, D> {
             | Role::ValidatorRegistration
             | Role::VoluntaryExit
             | Role::PTCAttester
-            | Role::ProposerPreferences => {
+            | Role::ProposerPreferences
+            | Role::EnvelopeProposer => {
                 let validator_pk = match ssv_message.msg_id().duty_executor() {
                     Some(DutyExecutor::Validator(pk)) => pk,
                     _ => return Err(ValidationFailure::UnknownValidator),
@@ -844,11 +845,11 @@ pub(crate) fn validate_beacon_duty(
         }
     }
 
-    // Rule: For a proposer-preferences message, the validator must be the assigned proposer at the
-    // preference's `proposal_slot` (= `slot` here). Checked only once the slot-epoch's proposer
-    // duties are known locally, so a not-yet-fetched epoch is tolerated. No RANDAO tolerance:
-    // ProposerPreferences carries no RANDAO signature.
-    if role == Role::ProposerPreferences {
+    // Rule: For a proposer-preferences or envelope-proposer message, the validator must be the
+    // assigned proposer at the slot. Checked only once the slot-epoch's proposer duties are known
+    // locally, so a not-yet-fetched epoch is tolerated. No RANDAO tolerance: neither
+    // ProposerPreferences nor EnvelopeProposer carry a RANDAO signature.
+    if matches!(role, Role::ProposerPreferences | Role::EnvelopeProposer) {
         // Non-committee roles always have one validator index
         let validator_index = validation_context
             .committee_info
@@ -895,6 +896,7 @@ pub(crate) fn validate_beacon_duty(
 /// - PTCAttester before the Ethereum Gloas (ePBS) fork (not yet active)
 /// - ValidatorRegistration at/after the Ethereum Gloas (ePBS) fork (deprecated by SIP-94)
 /// - ProposerPreferences before the Ethereum Gloas (ePBS) fork (not yet active)
+/// - EnvelopeProposer before the Ethereum Gloas (ePBS) fork (not yet active)
 pub(crate) fn validate_role_for_fork(
     slot: Slot,
     validation_context: &ValidationContext<impl SlotClock>,
@@ -921,18 +923,6 @@ pub(crate) fn validate_role_for_fork(
         });
     }
 
-    // Reject PTCAttester before the Ethereum Gloas (ePBS) fork, read from the consensus spec.
-    if role == Role::PTCAttester {
-        let current_fork = validation_context.spec.fork_name_at_epoch(epoch);
-        if !current_fork.gloas_enabled() {
-            return Err(ValidationFailure::RoleNotActiveBeforeEthFork {
-                role,
-                current_fork,
-                minimum_fork: ForkName::Gloas,
-            });
-        }
-    }
-
     // Reject ValidatorRegistration at/after the Ethereum Gloas (ePBS) fork; SIP-94
     // deprecates the duty (proposer preferences replace relay registrations). Gated
     // on the message's duty slot, not wall clock, so registrations for pre-fork
@@ -949,9 +939,12 @@ pub(crate) fn validate_role_for_fork(
         }
     }
 
-    // Reject ProposerPreferences before the Ethereum Gloas (ePBS) fork, read from the consensus
-    // spec.
-    if role == Role::ProposerPreferences {
+    // Reject post-Gloas roles (PTCAttester, ProposerPreferences, EnvelopeProposer) before the
+    // Ethereum Gloas (ePBS) fork, read from the consensus spec.
+    if matches!(
+        role,
+        Role::PTCAttester | Role::ProposerPreferences | Role::EnvelopeProposer
+    ) {
         let current_fork = validation_context.spec.fork_name_at_epoch(epoch);
         if !current_fork.gloas_enabled() {
             return Err(ValidationFailure::RoleNotActiveBeforeEthFork {
@@ -1044,7 +1037,9 @@ fn message_lateness(
     validation_context: &ValidationContext<impl SlotClock>,
 ) -> Result<Duration, ValidationFailure> {
     let ttl = match validation_context.role {
-        Role::Proposer | Role::SyncCommittee | Role::PTCAttester => 1 + LATE_SLOT_ALLOWANCE,
+        Role::Proposer | Role::SyncCommittee | Role::PTCAttester | Role::EnvelopeProposer => {
+            1 + LATE_SLOT_ALLOWANCE
+        }
         Role::Committee
         | Role::Aggregator
         | Role::ValidatorRegistration
@@ -1161,7 +1156,11 @@ fn duty_limit(
         }
         // Proposer and SyncCommittee have no duty limit
         Role::Proposer | Role::SyncCommittee => Ok(None),
-        Role::ProposerPreferences => Ok(Some(validation_context.slots_per_epoch)),
+        // Per-proposal-slot roles: max duties capped at SLOTS_PER_EPOCH (one preferences packet /
+        // one self-build envelope per proposal slot). Overflow is IGNORE-classified.
+        Role::ProposerPreferences | Role::EnvelopeProposer => {
+            Ok(Some(validation_context.slots_per_epoch))
+        }
     }
 }
 
@@ -1254,6 +1253,50 @@ mod tests {
             MessageAcceptance::from(&failure),
             MessageAcceptance::Ignore,
             "duty-limit breach (ExcessiveDutyCount) must classify as Ignore, not Reject."
+        );
+    }
+
+    /// Tests that the `MessageAcceptance` classification for EnvelopeProposer adheres to the SSV
+    /// spec.
+    #[test]
+    fn envelope_proposer_relevant_classifications_match_ssv_spec() {
+        // NoDuty and duty-limit overflow are Ignore; fork-gate, kind mismatch,
+        // packet-count overflow are Reject.
+        assert_eq!(
+            MessageAcceptance::from(&ValidationFailure::NoDuty),
+            MessageAcceptance::Ignore,
+            "NoDuty must be Ignore",
+        );
+        assert_eq!(
+            MessageAcceptance::from(&ValidationFailure::ExcessiveDutyCount {
+                got: 33,
+                limit: 32,
+                role: Role::EnvelopeProposer,
+            }),
+            MessageAcceptance::Ignore,
+            "duty-limit overflow must be Ignore",
+        );
+        assert_eq!(
+            MessageAcceptance::from(&ValidationFailure::RoleNotActiveBeforeEthFork {
+                role: Role::EnvelopeProposer,
+                current_fork: types::ForkName::Base,
+                minimum_fork: types::ForkName::Gloas,
+            }),
+            MessageAcceptance::Reject,
+            "fork-gate failure must be Reject",
+        );
+        assert_eq!(
+            MessageAcceptance::from(&ValidationFailure::PartialSignatureTypeRoleMismatch),
+            MessageAcceptance::Reject,
+            "kind mismatch must be Reject",
+        );
+        assert_eq!(
+            MessageAcceptance::from(&ValidationFailure::TooManyPartialSignatureMessages {
+                got: 2,
+                limit: 1
+            }),
+            MessageAcceptance::Reject,
+            "packet-count overflow must be Reject",
         );
     }
 
@@ -1436,7 +1479,8 @@ mod tests {
             | Role::ValidatorRegistration
             | Role::VoluntaryExit
             | Role::PTCAttester
-            | Role::ProposerPreferences => DutyExecutor::Validator(PublicKeyBytes::empty()),
+            | Role::ProposerPreferences
+            | Role::EnvelopeProposer => DutyExecutor::Validator(PublicKeyBytes::empty()),
         };
         MessageId::new(&domain, role, &duty_executor)
     }
--- a/anchor/message_validator/src/lib.rs
+++ b/anchor/message_validator/src/lib.rs
@@ -1256,50 +1256,6 @@ mod tests {
         );
     }
 
-    /// Tests that the `MessageAcceptance` classification for EnvelopeProposer adheres to the SSV
-    /// spec.
-    #[test]
-    fn envelope_proposer_relevant_classifications_match_ssv_spec() {
-        // NoDuty and duty-limit overflow are Ignore; fork-gate, kind mismatch,
-        // packet-count overflow are Reject.
-        assert_eq!(
-            MessageAcceptance::from(&ValidationFailure::NoDuty),
-            MessageAcceptance::Ignore,
-            "NoDuty must be Ignore",
-        );
-        assert_eq!(
-            MessageAcceptance::from(&ValidationFailure::ExcessiveDutyCount {
-                got: 33,
-                limit: 32,
-                role: Role::EnvelopeProposer,
-            }),
-            MessageAcceptance::Ignore,
-            "duty-limit overflow must be Ignore",
-        );
-        assert_eq!(
-            MessageAcceptance::from(&ValidationFailure::RoleNotActiveBeforeEthFork {
-                role: Role::EnvelopeProposer,
-                current_fork: types::ForkName::Base,
-                minimum_fork: types::ForkName::Gloas,
-            }),
-            MessageAcceptance::Reject,
-            "fork-gate failure must be Reject",
-        );
-        assert_eq!(
-            MessageAcceptance::from(&ValidationFailure::PartialSignatureTypeRoleMismatch),
-            MessageAcceptance::Reject,
-            "kind mismatch must be Reject",
-        );
-        assert_eq!(
-            MessageAcceptance::from(&ValidationFailure::TooManyPartialSignatureMessages {
-                got: 2,
-                limit: 1
-            }),
-            MessageAcceptance::Reject,
-            "packet-count overflow must be Reject",
-        );
-    }
-
     // Helper struct for directly creating consensus messages for tests
     pub(crate) struct QbftMessageBuilder {
         msg_type: QbftMessageType,
--- a/anchor/message_validator/src/partial_signature.rs
+++ b/anchor/message_validator/src/partial_signature.rs
@@ -147,6 +147,7 @@ fn partial_signature_type_matches_role(kind: PartialSignatureKind, role: Role) -
         Role::Committee => kind == PartialSignatureKind::PostConsensus,
         Role::PTCAttester => kind == PartialSignatureKind::PTCAttester,
         Role::ProposerPreferences => kind == PartialSignatureKind::ProposerPreferences,
+        Role::EnvelopeProposer => kind == PartialSignatureKind::PostConsensus,
         Role::Aggregator => {
             kind == PartialSignatureKind::PostConsensus
                 || kind == PartialSignatureKind::SelectionProofPartialSig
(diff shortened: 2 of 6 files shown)


jnhsigmap:
@claude, review this PR

anchor/qbft_manager/src/lib.rs
```diff
@@ -282,6 +282,11 @@ impl<E: EthSpec, S: SlotClock + Clone + 'static> QbftManager<E, S> {
                     Some(Role::Proposer) => ValidatorDutyKind::Proposal,
                     Some(Role::Aggregator) => ValidatorDutyKind::Aggregator,
                     Some(Role::SyncCommittee) => ValidatorDutyKind::SyncCommitteeAggregator,
+                    Some(Role::EnvelopeProposer) => {
+                        // TODO: wire EnvelopeProposer instance routing (#1122)
+                        warn!(?msg_id, "EnvelopeProposer routing not yet wired");
```
shane-moore: Suggest downgrading this `warn!` to `debug!`. Post-Gloas this arm is the expected path for every valid role-9 consensus message we accept and propagate, and message_receiver already logs an `error!` per failed `receive_data` (manager.rs:187), so each message double-logs until #1122 wires routing. Keeping the line at `debug!` is still useful: there are two production sources of `QbftError::RoleNotActive` (this stub and the AggregatorCommittee pre-Boole backstop below) and the caller renders the msg_id as raw hex, so this is the only readable breadcrumb distinguishing them.

anchor/qbft_manager/src/tests.rs
```diff
@@ -961,4 +961,172 @@ mod manager_tests {
...
+
+        // Assert: Must be transient RoleNotActive (routing not wired), NOT InconsistentMessageId
+        assert!(
+            matches!(result, Err(QbftError::RoleNotActive)),
+            "validator-executor EnvelopeProposer must be transient RoleNotActive, got: {result:?}"
+        );
+    }
+
+    #[tokio::test]
+    async fn envelope_proposer_committee_executor_decodes_as_validator_executor() {
```
shane-moore: These two tests end up feeding `receive_data` byte-identical input: for role 9, `duty_executor()` always reads bytes 8..55 as a validator pubkey, so the `Committee([0; 32])` MessageId built here encodes to the same 56 bytes as the `PublicKeyBytes::empty()` one above. Suggest collapsing them into one test that constructs both executor variants, asserts `RoleNotActive` for both, and states the decode-as-Validator invariant in a comment (the committee arm's `InconsistentMessageId` is unreachable for role 9, not just untested). While consolidating, the fork setup framing can go too: this arm returns unconditionally, and the "which mainnet ChainSpec has" comment is inverted since `ChainSpec::mainnet()` has `gloas_fork_epoch: None` at the current pin.

anchor/message_validator/src/partial_signature.rs
```diff
@@ -2006,6 +2008,171 @@ mod tests {
...
+                voluntary_exit_duty_count: 0,
+                ..Default::default()
+            }),
+        );
+
+        assert!(result.is_ok(), "Expected ok but got: {result:?}");
+    }
+
+    #[test]
+    fn envelope_proposer_beyond_short_ttl_rejected() {
```
shane-moore, line 2234: Could we add one earliness sibling to this TTL pair? SIP-94 §7 gives role 9 no future-slot allowance, but that cell is currently held only by `early_slot_allowance`'s `_ => Duration::ZERO` wildcard: adding a `Role::EnvelopeProposer` allowance arm later (the way ProposerPreferences has one) would fail no test, since the existing `test_non_proposer_preferences_future_slot_still_early` pin drives ValidatorRegistration. A future-slot EnvelopeProposer packet asserting `EarlySlotMessage` mirrors these two tests with the same helpers (clock at slot N, message slot N+1).

anchor/message_validator/src/lib.rs
```diff
@@ -1257,6 +1256,50 @@ mod tests {
         );
     }
 
+    /// Tests that the `MessageAcceptance` classification for EnvelopeProposer adheres to the SSV
+    /// spec.
+    #[test]
+    fn envelope_proposer_relevant_classifications_match_ssv_spec() {
+        // NoDuty and duty-limit overflow are Ignore; fork-gate, kind mismatch,
```
shane-moore: Suggest dissolving this test: no new tests needed, just move its three Reject asserts into the existing tests that already produce those failures, then delete it.

Concretely:
- the `PartialSignatureTypeRoleMismatch` assert_eq -> append to `envelope_proposer_binds_post_consensus_kind`
- the `TooManyPartialSignatureMessages` assert_eq -> append to `envelope_proposer_rejects_multiple_messages_per_packet`
- the `RoleNotActiveBeforeEthFork` assert_eq -> append to the pre-Gloas EnvelopeProposer test in consensus_message.rs
- the `NoDuty` and `ExcessiveDutyCount` asserts can just be dropped (already pinned in `test_proposer_preferences_known_epoch_not_assigned_no_duty` and `excessive_duty_count_maps_to_ignore` above)

Rationale: `From<&ValidationFailure>` never reads `role`, so nothing here is EnvelopeProposer-specific and the name overclaims. But the three Reject asserts are the only pins of those mappings in the crate, so they should survive, attached to the tests that produce the failure for real. Same pattern the ProposerPreferences tests already use ("Pin the Reject mapping so a future reclassification is caught here"). #1134 flipping `ExcessiveDutyCount` from Reject to Ignore is exactly the drift these catch.

anchor/message_validator/src/partial_signature.rs
```diff
@@ -3280,4 +3447,416 @@ mod tests {
...
+
+    // ==================== EnvelopeProposer proposer-assignment arm ====================
+    //
+    // `validate_beacon_duty`'s `Role::ProposerPreferences | Role::EnvelopeProposer` arm (keyed on
+    // the message's `slot`) rejects with `NoDuty` only when the slot's epoch is known AND the
+    // validator is NOT the assigned proposer; an unknown epoch is tolerated and no RANDAO
+    // tolerance applies. The three cases below drive the mock's two proposer knobs.
+
+    /// Standard single-signer four-node fixture (committee info + keypair + pubkey map).
+    fn four_node_committee_and_keypair() -> (
```
shane-moore: Nit: this fixture family is filed under the proposer-assignment banner but serves the whole EnvelopeProposer suite: `four_node_committee_and_keypair` is first called ~1400 lines up in the kind-binding test, and `create_envelope_proposer_context` in the packet-cap test. Suggest moving the three helpers up to the "EnvelopeProposer partial-signature tests" banner (or next to `generate_test_key_pair` in the shared helper region) so definitions precede uses; `create_signed_envelope_proposer_message` is only used below but travels with its siblings.

anchor/qbft_manager/src/lib.rs
jnhsigmap: handled in https://github.com/sigp/anchor/pull/1149/changes/5e397ac8be34b4b420c1502dc3148024b2482668

anchor/message_validator/src/partial_signature.rs
jnhsigmap: I initially read this as a direct `TTL` test - as in message received early for slot `N` - but _"A future-slot EnvelopeProposer packet asserting EarlySlotMessage mirrors these two tests with the same helpers (clock at slot N, message slot N+1)."_ suggests use `create_signed_envelope_proposer_message` / `create_signed_envelope_proposer_message` to create instance where message for slot `N+1` is received (time doesn't matter) during clock slot `N` (with `create_envelope_proposer_context`). 

Also mentioned [here](https://github.com/sigp/anchor/pull/1149#issuecomment-5022465894) as part of _"envelope slot == proposal slot (present), never future."_.

Handled in https://github.com/sigp/anchor/commit/5e397ac8be34b4b420c1502dc3148024b2482668 with `envelope_proposer_future_slot_still_early`. Hope this is the right path haha..

anchor/message_validator/src/partial_signature.rs
jnhsigmap: done too :saluting_face:

anchor/qbft_manager/src/tests.rs
jnhsigmap: https://github.com/sigp/anchor/commit/5e397ac8be34b4b420c1502dc3148024b2482668 includes `envelope_proposer_any_executor_decodes_as_validator_transient_role_not_active` - plan on renaming this after the `NotActive` is removed. Wiring this PR in was messy as there are several transient elements.

anchor/message_validator/src/lib.rs
jnhsigmap: `envelope_proposer_relevant_classifications_match_ssv_spec` removed :+1:

anchor/message_validator/src/lib.rs
```diff
@@ -949,9 +939,12 @@ pub(crate) fn validate_role_for_fork(
         }
     }
 
-    // Reject ProposerPreferences before the Ethereum Gloas (ePBS) fork, read from the consensus
-    // spec.
-    if role == Role::ProposerPreferences {
+    // Reject post-Gloas roles (PTCAttester, ProposerPreferences, EnvelopeProposer) before the
+    // Ethereum Gloas (ePBS) fork, read from the consensus spec.
+    if matches!(
```
shane-moore, line 944: nice consolidation of `Role::PTCAttester` into this condition 🎊

shane-moore (approved):
LGTM!

PR #1164, opened 2026-07-23
feat(ssv_types): reject decided ProposerConsensusData whose Version mismatches the duty-slot fork

### LINKED ISSUE #1129: feat(ssv_types): reject decided ProposerConsensusData whose Version mismatches the duty-slot fork
## Goal

Add a fork-schedule equality check to `ProposerConsensusDataValidator`: reject any `ProposerConsensusData` whose `version` does not equal the fork scheduled at `duty.slot`. SIP-94 §4 makes this a MUST:

> The decided value's `Version` selects how `DataSSZ` is decoded (`Gloas.BeaconBlock` when `Version >= DataVersionGloas`), and `Version` is leader-supplied. An operator MUST reject any decided value whose `Version` does not equal the fork scheduled at `duty.Slot`. Honest proposers always stamp `Version == fork(duty.Slot)`, so this rejects no honest value.

## Context / motivation

`value.version` is the only leader-controlled field that changes how every other operator interprets `data_ssz`, and Anchor currently trusts it everywhere it matters (refs at Anchor `b162b606`):

- `validate_block_proposal` (`ssv_types/src/consensus.rs:397-414`) derives `fork = ForkName::from(value.version)` and branches decode on it (`decode_block` for Gloas-or-later, blinded-then-contents fallback pre-Gloas). `do_validation` (`consensus.rs:325-384`) checks the duty's slot, role, pubkey, and index against our own value, but never checks `version`.
- Post-decide, `decode_decided_block` (`validator_store/src/lib.rs:1934-1953`) repeats the same version-keyed dispatch, so a wrong `version` that survives consensus steers how every operator decodes the block it then signs.

The check bites at all three QBFT validation points, since they all route through `QbftDataValidator::validate` -> `do_validation`: received proposals (`common/qbft/src/lib.rs:433`, `validate_message`), round-change justifications (`:488`, `justify_round_change_quorum`), and decided-message sanity checking (`:1207`, `received_decided`).

Why fork-schedule equality rather than trusting decode to fail: SSZ is not self-describing, and `from_ssz_bytes_for_fork` under a wrong fork arm fails with a shape-dependent `DecodeError` at best, or accepts crafted bytes at worst. Mixed go-ssv/Anchor clusters need a deterministic accept/reject rule that does not depend on which decode arm happens to error; `Version == fork(duty.Slot)` is that rule (agreed on the SIP thread, [discussion_r3542550168](https://github.com/ssvlabs/SIPs/pull/94#discussion_r3542550168)).

The honest path is unaffected: Anchor stamps its own proposal's version from the BN-returned block (`block.fork_name_unchecked().into()`, `validator_store/src/lib.rs:551`), which equals the scheduled fork at the duty slot for any honest BN.

## Suggested approach

`ssv_types/src/consensus.rs`:

- Add a `DataValidationError` variant styled after the existing mismatch variants:

```rust
#[error("wrong data version: expected fork {expected:?}, got {got:?}")]
VersionMismatch { expected: ForkName, got: ForkName },
```

- At the top of `validate_block_proposal`, before the decode-arm selection, compare against the fork schedule (the validator already holds `spec: Arc<ChainSpec>` and `E` via `PhantomData`; precedent for `fork_name_at_slot` at `qbft_manager/src/lib.rs:321`):

```rust
let expected = self.spec.fork_name_at_slot::<E>(value.duty.slot);
let got = ForkName::from(value.version);
if got != expected {
    return Err(DataValidationError::VersionMismatch { expected, got });
}
```

Checking before decode fails fast and makes the subsequent `fork` derivation trustworthy. `value.duty.slot` is already pinned to our own duty slot by the `SlotMismatch` check in `do_validation`, so the comparison is anchored to the duty, not to leader input.

## Acceptance criteria

- A Gloas-slot value stamped with a pre-Gloas `version` is rejected with `VersionMismatch`, not a decode error.
- A pre-Gloas-slot value stamped `DataVersionGloas` is rejected the same way (the rule is era-generic and hardens pre-Gloas slots too).
- A value whose `version` matches `fork_name_at_slot(duty.slot)` validates exactly as before (decode arms, slashing check, and error surfaces unchanged).
- The check runs regardless of `disable_slashing_protection`.
- The QBFT paths reject leader proposals, round-change data, and decided messages carrying a mismatched version (via the existing `validate` plumbing; no qbft-crate change).

## Tests

`cargo test -p ssv_types`. The existing Gloas proposer fixtures (`consensus.rs`, `test_proposer_duty()` + `DataVersion::from(ForkName::Gloas)` around `:3000-3160`) cover decode; add validator-level tests, which need a `ProposerConsensusDataValidator` instance (constructing one takes a `SlashingDatabase`; with `disable_slashing_protection: true` the DB is never touched, so a tempfile-backed one suffices):

- both mismatch directions reject with `VersionMismatch { expected, got }` carrying the right forks
- matching Gloas version at a Gloas slot passes through to decode (accept on well-formed bytes)
- matching pre-Gloas version at a pre-Gloas slot preserves the blinded-then-contents fallback behavior
- use a `ChainSpec` with `gloas_fork_epoch` set so `fork_name_at_slot` exercises both eras

## Notes

- Scope is the `BEACON_ROLE_PROPOSER` arm only, per SIP-94 §4. The legacy `BEACON_ROLE_AGGREGATOR` arm (`consensus.rs:367-372`) shares the version-trust pattern but is a pre-Boole path already slated for removal (`TODO(post-boole)` at `:363`); out of scope here.
- Do NOT generalize this to `EnvelopeConsensusData` (#1121): SIP-94 §6's value check deliberately ignores `version` for the envelope type.
- go-ssv PR #2901 now explicitly rejects a pre-Gloas `Version` on a Gloas duty slot before decoding. Anchor enforces exact fork-schedule equality in both directions, matching the SIP rule rather than relying on a wrong-fork decoder to reject the reverse mismatch.

## Addendum (2026-07-22): also pin the decoded block's slot to the duty slot

The version check makes the `data_ssz` decode arm trustworthy, but the decoded block's own `slot` remains unchecked against the duty. The block is signed under `header.slot` and slashing protection keys on `header.slot` (`validate_block_proposal`, epoch/domain derivation and `preliminary_check_block_proposal`), so a leader stamping a correct duty while embedding a block built for another slot S gets a valid signature for S that the duty-level checks never see: an equivocation vector, and an unbounded `header.slot` feeding the domain computation. go-ssv rejects exactly this in PR #2901 ("gloas block slot does not match duty slot"). Not an explicit SIP-94 MUST; adopted as slashing-safety hardening and reference-implementation parity.

Suggested change, in `validate_block_proposal` after the header decode and before the slashing gate (so it also runs under `disable_slashing_protection`):

```rust
if header.slot != value.duty.slot {
    return Err(DataValidationError::BlockSlotMismatch {
        expected: value.duty.slot,
        got: header.slot,
    });
}
```

with a `BlockSlotMismatch { expected: Slot, got: Slot }` variant styled after the existing mismatch variants. `value.duty.slot` is pinned to our own duty by the `SlotMismatch` check, same trust anchor as the version check.

Additional acceptance criteria:

- A value whose decoded block slot differs from the duty slot is rejected with `BlockSlotMismatch` in both the Gloas and pre-Gloas decode arms.
- The check runs regardless of `disable_slashing_protection`.
- Matching block slot validates exactly as before.

shane-moore:

`ProposerConsensusData.version` is leader-supplied and selects how every operator decodes `data_ssz`, during QBFT validation and again post-decide before signing. Anchor trusted it unchecked; SIP-94 §4 makes rejecting a version that mismatches the fork scheduled at the duty slot a MUST. SSZ is not self-describing, so relying on wrong-fork decode failure is shape-dependent and non-deterministic across implementations.

Closes #1129 (includes its addendum: the decoded block's own slot was similarly unchecked against the duty slot, letting a leader harvest a signature for a different slot than the duty; the slashing DB keys on the block's slot, so nothing tied it back).

## Change Overview (Required)

- New `DataValidationError` variants `VersionMismatch` and `BlockSlotMismatch`.
- `validate_block_proposal` now enforces, in order: version equals `fork_name_at_slot(duty.slot)`, then decode, then decoded block slot equals the duty slot, then the existing slashing gate. Both checks run regardless of `disable_slashing_protection` and are era-generic (they harden pre-Gloas slots too, per the issue's acceptance criteria).
- Tests: fixture helpers parameterized over spec and slashing flag, plus rejection/acceptance coverage through the public `do_validation` for both mismatch directions, both eras, and check-before-slashing ordering.

Read the production diff first (`do_validation` / `validate_block_proposal` / the error enum, ~40 lines); the rest is tests. Not changed: the QBFT plumbing (all three validation points already route through `QbftDataValidator::validate`), the legacy aggregator/sync arms (pre-Boole, slated for removal), and `EnvelopeConsensusData` (SIP-94 §6 deliberately ignores `version` there).

Anchor is stricter than go-ssv pre-Gloas: go-ssv's PR #2901 applies the version check and block-slot pin only on its Gloas path. The divergence only affects dishonest or buggy leader values (honest proposers stamp `version == fork(duty.slot)` from the BN block, refs in #1129), and the worst mixed-cluster outcome is a stalled round on a malformed value, never an unsafe signature. A misconfigured local BN producing off-schedule blocks now fails safe (no signature) instead of proceeding.

## Validation (Required)

`cargo test --release -p ssv_types` (115 unit + 2 doc-tests), `make lint` clean across the workspace, `make cargo-fmt-check` clean. Spec parity checked against SIP-94 §4 and go-ssv PR #2901.

Pure revert; validation-only, no wire-format, config, or data impact.

### DIFF
--- a/anchor/common/ssv_types/src/consensus.rs
+++ b/anchor/common/ssv_types/src/consensus.rs
@@ -372,7 +372,15 @@ impl<E: EthSpec> ProposerConsensusDataValidator<E> {
                 }
             }
             BEACON_ROLE_PROPOSER => {
-                self.validate_block_proposal(value)?;
+                // SIP-94 §4: `version` is leader-supplied and selects how `data_ssz` is decoded,
+                // so it must equal the fork scheduled at the duty slot. Keyed on our own duty's
+                // slot (equal to `value.duty.slot` after the `SlotMismatch` check above).
+                let expected = self.spec.fork_name_at_slot::<E>(our_value.duty.slot);
+                let got = ForkName::from(value.version);
+                if got != expected {
+                    return Err(DataValidationError::VersionMismatch { expected, got });
+                }
+                self.validate_block_proposal(value, got)?;
             }
             BEACON_ROLE_SYNC_COMMITTEE_CONTRIBUTION => {
                 // There is nothing special to check for sync committee contributions.
@@ -384,9 +392,12 @@ impl<E: EthSpec> ProposerConsensusDataValidator<E> {
         Ok(())
     }
 
+    /// `fork` is the value's version already validated against the duty-slot fork schedule
+    /// by [`Self::do_validation`].
     fn validate_block_proposal(
         &self,
         value: &ProposerConsensusData,
+        fork: ForkName,
     ) -> Result<(), DataValidationError> {
         // Decode the block header to ensure the value is decodable (even when slashing
         // protection is disabled). Under Gloas (EIP-7732), DataSSZ is decoded directly as a plain
@@ -394,8 +405,6 @@ impl<E: EthSpec> ProposerConsensusDataValidator<E> {
         // payload is decoupled from the block body. The outcome of `decode_blinded_block`
         // and `decode_block` are identical for this variant. The Pre-Gloas branch
         // preserves the existing try-blinded-then-full fallback.
-        let fork = ForkName::from(value.version);
-
         let header = if fork >= ForkName::Gloas {
             value
                 .decode_block::<E>()
@@ -455,6 +464,8 @@ pub enum DataValidationError {
         expected: ValidatorIndex,
         got: ValidatorIndex,
     },
+    #[error("wrong data version: expected fork {expected:?}, got {got:?}")]
+    VersionMismatch { expected: ForkName, got: ForkName },
     #[error("Block proposal would be slashable: {0}")]
     SlashableBlockProposal(NotSafe),
 }
@@ -3057,12 +3068,12 @@ mod tests {
     // ProposerConsensusData Gloas (EIP-7732) Tests
     // ═══════════════════════════════════════════════════════════════════════════════
 
-    /// Creates a minimal proposer `ValidatorDuty` at slot 0 for Gloas block tests.
-    fn test_proposer_duty() -> ValidatorDuty {
+    /// Creates a minimal proposer `ValidatorDuty` at `slot` for block proposal tests.
+    fn test_proposer_duty(slot: Slot) -> ValidatorDuty {
         ValidatorDuty {
             r#type: BEACON_ROLE_PROPOSER,
             pub_key: PublicKeyBytes::empty(),
-            slot: Slot::new(0),
+            slot,
             validator_index: ValidatorIndex(0),
             committee_index: 0,
             committee_length: 0,
@@ -3080,7 +3091,7 @@ mod tests {
         let block = BeaconBlock::Gloas(BeaconBlockGloas::<MainnetEthSpec>::empty(&spec));
 
         let consensus_data = ProposerConsensusData {
-            duty: test_proposer_duty(),
+            duty: test_proposer_duty(Slot::new(0)),
             version: DataVersion::from(ForkName::Gloas),
             data_ssz: VariableList::new(block.as_ssz_bytes())
                 .expect("Gloas block bytes should fit in DataSSZ"),
@@ -3103,7 +3114,7 @@ mod tests {
         let block = BeaconBlock::Gloas(BeaconBlockGloas::<MainnetEthSpec>::empty(&spec));
 
         let consensus_data = ProposerConsensusData {
-            duty: test_proposer_duty(),
+            duty: test_proposer_duty(Slot::new(0)),
             version: DataVersion::from(ForkName::Gloas),
             data_ssz: VariableList::new(block.as_ssz_bytes())
                 .expect("Gloas block bytes should fit in DataSSZ"),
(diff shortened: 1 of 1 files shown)


shane-moore:
@claude, please take a look

shane-moore:
@claude, agreed on all points. The two non-blocking notes are per design: the aggregator arm's unchecked version is the pre-Boole legacy path slated for removal (scoped out in #1129), and the local runs it could not execute are green (`cargo test --release -p ssv_types`, `make lint`, `make cargo-fmt-check`).

jnhsigmap (approved):
looks good! couldn't find any issues

PR #1169, opened 2026-07-24
feat(ssv_types): add envelope-signing QBFT value + structural value check

### LINKED ISSUE #1121: feat(ssv_types): add BlindedExecutionPayloadEnvelope and EnvelopeConsensusData with QbftData and value check
## Goal

Introduce the `BlindedExecutionPayloadEnvelope<E>` SSZ container as Anchor's internal envelope-QBFT value, the `EnvelopeConsensusData { duty, version, data_ssz }` QBFT value mirroring `ProposerConsensusData`, and an `EnvelopeConsensusDataValidator` implementing SIP-94 §6's `EnvelopeValueCheckF`: deliberately structural, with NO envelope-content validation and NO slashing-protection calls.

## Context / motivation

SIP-94 §6 runs QBFT over a blinded envelope so consensus messages stay bounded (hundreds of bytes instead of up to MBs). By SSZ root substitution `hash_tree_root(blinded) == hash_tree_root(full)`, so the post-consensus signature over the blinded signing root is valid for the full envelope Lighthouse publishes. This is an SSV consensus value, not a commitment to a beacon-APIs transport schema. ethereum/beacon-APIs#624 currently proposes removing the public `BlindedExecutionPayloadEnvelope` API object and returning a full `ExecutionPayloadEnvelope` from the envelope endpoint; Anchor should still keep the internal blinded QBFT value as long as SIP-94 and ssv-spec keep that consensus shape.

Internal field shape (SIP-94 §6 / go-ssv `types/gloas`; LH full envelope at `consensus/types/src/execution/execution_payload_envelope.rs:20-27` at pin `81d576943`):

```
payload_root: Hash256            // == hash_tree_root(envelope.payload)
execution_requests: ExecutionRequestsGloas<E>
builder_index: u64
beacon_block_root: Hash256
parent_beacon_block_root: Hash256
```

`EnvelopeConsensusData` mirrors `ProposerConsensusData` as the same `{duty, version, data_ssz}` triple while remaining a distinct type for `QbftDecidable` routing. SIP-94, ssv-spec#633, and go-ssv use this same shape; cross-client conformance vectors and golden hash tests are follow-up scope.

Anchor's current Lighthouse pin already uses `ExecutionRequestsGloas<E>` for the full envelope. That is the Gloas five-list request container: deposits, withdrawals, consolidations, builder deposits, and builder exits.

## Suggested approach

In `ssv_types/src/consensus.rs` (refs at `04e3ddc78abd`):

- `BlindedExecutionPayloadEnvelope<E: EthSpec>` deriving `Encode, Decode, TreeHash` (+ `Clone, Debug, PartialEq`), plus `impl SignedRoot` (marker over the derived `TreeHash`) so the sign path can call `.signing_root(domain)`. The `types::execution::ExecutionRequestsGloas<E>` field forces the generic (like `AggregatorCommitteeConsensusData<E>`) and is current-pin scope.
- `EnvelopeConsensusData` + `impl QbftData`, mirroring `ProposerConsensusData` (:244-284), with `data_ssz: VariableList<u8, ProposerConsensusDataLen>` and a `decode_blinded_envelope::<E>()` helper mirroring `decode_block` (:267).
- `pub const BEACON_ROLE_ENVELOPE_PROPOSER: BeaconRole = BeaconRole(9);` in the `BEACON_ROLE_*` block (:485+). The synthesized local duty stamps it as `r#type`; the value checker should stay aligned with SIP/go-ssv and not reject solely on embedded `duty.r#type`.
- `EnvelopeConsensusDataValidator<E>` mirroring the SIP/go-ssv value-check contract: SSZ decode of the blinded envelope; duty slot, validator index, and pubkey match; `builder_index == BUILDER_INDEX_SELF_BUILD` (`types::consts::gloas`, LH `core/consts.rs:29`); `beacon_block_root ==` the block-QBFT-decided root, supplied as a **constructor parameter** for this issue's tests. The follow-up decided-root hand-off issue wires the real provider. Contrast `validate_block_proposal` (:387-428): no slashability call here, envelope signing is not slashable.
- Do not reject solely on `version` or embedded `duty.r#type`. SIP-94 and go-ssv's `EnvelopeValueCheckF` ignore both for this type; they remain part of the opaque SSZ value and round-trip.
- Reuse `DataValidationError` (:436) for mirrored checks only if the display text is made generic. Do not let envelope decode failures log as "Unable to decode ssz in ProposerConsensusData". Add envelope-specific variants for at least not-self-build and decided-root mismatch.

## Acceptance criteria

- **Root equivalence (load-bearing):** for a fixture with `payload_root == envelope.payload.tree_hash_root()` and the other four fields verbatim, `blinded.tree_hash_root() == full.tree_hash_root()` against the LH pin's `ExecutionPayloadEnvelope`.
- SSZ round-trips for both types; `EnvelopeConsensusData.data_ssz` keeps the same 8 MiB bound as `ProposerConsensusData`; `QbftData::hash()` deterministic.
- Value check accepts iff decode succeeds, `duty.{slot, validator_index, pub_key}` match, `builder_index == BUILDER_INDEX_SELF_BUILD`, and `beacon_block_root` equals the supplied root; each failure a distinct error. No `version` or embedded `duty.r#type` rejection unless SIP/go-ssv add one.
- No `slashing_protection.*` call, no field-level content comparison.
- The implementation does not depend on a public beacon-APIs `BlindedExecutionPayloadEnvelope` schema. Beacon-APIs transport may return/publish full envelope forms while Anchor still uses the blinded form internally for QBFT.

## Tests

Root-equivalence fixture, Anchor-local SSZ round-trips for both types, max-`DataSSZ` bound, one rejection test per value-check condition, and a regression that envelope decode errors are not mislabeled as proposer-consensus decode errors. `cargo test -p ssv_types`.

## Notes

- **Internal value, not beacon-APIs transport:** ethereum/beacon-APIs#624 removes the public `BlindedExecutionPayloadEnvelope` API object from the draft API shape. That does not remove the SIP-94/go-ssv SSV consensus value. If SIP-94 or ssv-spec later renames or removes the internal blinded value, follow that upstream change then.
- **Merkleization parity is the contract.** The blinded type must tree-hash exactly as the LH pin's full envelope does (plain positional container). consensus-specs #4630 (EIP-7688) made the envelope a `ProgressiveContainer` on master (serialization unchanged, tree-hash different); when a future LH bump adopts it, the root-equivalence test breaks loudly and both types flip together. Do not pre-implement progressive hashing.
- **`ExecutionRequestsGloas` is current-pin scope:** do not regress to the older 3-list `ExecutionRequests<E>` shape. If a future Lighthouse bump changes the Gloas request container or envelope merkleization, the root-equivalence fixture should fail loudly.
- **Trust boundary:** `payload_root` is leader-decided and unchecked; the safety boundary is the protocol-level bid/`block_hash` check at envelope processing, matching the blinded-block trust model (SIP §6 rationale). Resist adding content checks.

jnhsigmap:

- ePBS (EIP-7732 / Gloas) splits block production into a beacon block and a separately-signed execution payload envelope.
- SIP-94 §6 runs an envelope-signing QBFT duty over a blinded envelope; this PR adds the value type and value check that duty consumes.
- Closes #1121.

## Change Overview

All changes were to `anchor/common/ssv_types/src/consensus.rs`.

- Added `BlindedExecutionPayloadEnvelope<E>`: holds `payload_root` in place of the full `payload`, other fields same as full envelope.
  - `from_full` builds it from `ExecutionPayloadEnvelope<E>`.
  - `impl SignedRoot`.
- Merkleization parity is load-bearing: `blinded.tree_hash_root() == full.tree_hash_root()`, so a signature over the blinded root is valid for the full envelope. Tested for.
- `EnvelopeConsensusData`: QBFT value wrapping the blinded envelope as `data_ssz` (bounded by `ProposerConsensusDataLen`, 8 MiB).
  - `decode_blinded_envelope` decodes it.
  - `impl QbftData` provides a deterministic SSZ-then-SHA256 `hash`.
- `EnvelopeConsensusDataValidator<E>`: structural value check accepting if
  - `data_ssz` decodes.
  - duty `{slot, validator_index, pub_key}` match.
  - `builder_index == BUILDER_INDEX_SELF_BUILD`.
  - `beacon_block_root ==` the decided block root.
- `EnvelopeValidationError`: distinct `thiserror` variant per failure (`DecodeError`, `SlotMismatch`, `IndexMismatch`, `PubKeyMismatch`, `NotSelfBuild`, `DecidedRootMismatch`); decode failures are attributed to `EnvelopeConsensusData`, not `ProposerConsensusData`.
- `BEACON_ROLE_ENVELOPE_PROPOSER = BeaconRole(9)`, matching the `Role::EnvelopeProposer` wire byte.
- The check does not reject on `version` or `duty.r#type`, makes no slashing-protection call, and does no field-level payload comparison.

- Additive types with no consumers yet; blast radius is one module.
- `payload_root` is leader-decided and unchecked (blinded-block trust model); the safety boundary is the protocol-level bid/block-hash check at envelope processing, out of scope here.
- A future Lighthouse envelope-merkleization change fails the root-parity test loudly and flips both types together.

## Validation

- `cargo test -p ssv_types`: root parity, SSZ round-trips (both types), 8 MiB `data_ssz` bound, deterministic `hash`, accept path, one rejection test per value-check condition, and a regression that decode failures are attributed to `EnvelopeConsensusData`.

## Rollback

- Additive, no consumers; revert the commits with no runtime or data impact.

## Blockers / Dependencies

- Base branch `epbs`. QBFT wiring, decided-root store, and signing land in follow-up PRs.

### DIFF
--- a/anchor/common/ssv_types/src/consensus.rs
+++ b/anchor/common/ssv_types/src/consensus.rs
@@ -3733,7 +3733,7 @@ mod tests {
 
         let full = envelope_test_full_envelope(BUILDER_INDEX_SELF_BUILD, decided_root);
         let blinded = BlindedExecutionPayloadEnvelope::from_full(&full);
-        let duty = envelope_test_duty(BEACON_ROLE_PROPOSER, pubkey, slot, validator_index);
+        let duty = envelope_test_duty(BEACON_ROLE_ENVELOPE_PROPOSER, pubkey, slot, validator_index);
         let value = envelope_consensus_data(duty, &blinded);
 
         (validator, value)
--- a/anchor/common/ssv_types/src/consensus.rs
+++ b/anchor/common/ssv_types/src/consensus.rs
@@ -3778,11 +3778,8 @@ mod tests {
         );
     }
 
-    /// Encode an EnvelopeConsensusData and decode it back. Confirm:
-    /// 1. The value.
-    /// 2. Its root.
-    /// 3. The nested blinded envelope inside data_ssz.
-    /// All come back identical.
+    /// Encode an EnvelopeConsensusData and decode it back, confirming the value, its root, and
+    /// the nested blinded envelope inside data_ssz all come back identical.
     #[test]
     fn envelope_consensus_data_ssz_round_trip() {
         // EnvelopeConsensusData whose data_ssz holds a real blinded envelope.
--- a/anchor/common/ssv_types/src/consensus.rs
+++ b/anchor/common/ssv_types/src/consensus.rs
@@ -615,8 +615,6 @@ pub const BEACON_ROLE_SYNC_COMMITTEE: BeaconRole = BeaconRole(3);
 pub const BEACON_ROLE_SYNC_COMMITTEE_CONTRIBUTION: BeaconRole = BeaconRole(4);
 pub const BEACON_ROLE_VALIDATOR_REGISTRATION: BeaconRole = BeaconRole(5);
 pub const BEACON_ROLE_VOLUNTARY_EXIT: BeaconRole = BeaconRole(6);
-/// Beacon role for ePBS envelope proposer duty (wire byte 9, SIP-94 §6). Matches
-/// `Role::EnvelopeProposer` in the SSV message type system.
 pub const BEACON_ROLE_ENVELOPE_PROPOSER: BeaconRole = BeaconRole(9);
 pub const BEACON_ROLE_UNKNOWN: BeaconRole = BeaconRole(u64::MAX);
 
@@ -3637,8 +3635,6 @@ mod tests {
     const ENVELOPE_TEST_VALIDATOR_INDEX: usize = 7;
     /// Alternative validator index used across the envelope fixtures.
     const ENVELOPE_OTHER_TEST_VALIDATOR_INDEX: usize = 8;
-    /// `ProposerConsensusDataLen` is `2^23` bytes (8 MiB); `data_ssz` shares this bound.
-    const ENVELOPE_DATA_SSZ_MAX_LEN: usize = 1 << 23;
 
     /// Deterministic validator pubkey for the "matching" side of the value check.
     fn envelope_test_pubkey(validator_index: usize) -> PublicKeyBytes {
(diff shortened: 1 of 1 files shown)


jnhsigmap:
@claude review this PR

shane-moore (approved):
Verified locally: root parity against the LH pin's `ExecutionPayloadEnvelope` (5-field shape, `BUILDER_INDEX_SELF_BUILD = u64::MAX`), value-check set and order against go-ssv's `NewEnvelopeChecker` (identical four checks), and the type shapes against ssv-spec #632/#633. Tests, fmt, and clippy clean on the PR head. LGTM.

One optional nit: `valid_envelope_setup` stamps the duty with `BEACON_ROLE_PROPOSER`, while the real synthesized duty will carry `BEACON_ROLE_ENVELOPE_PROPOSER` (the SSZ round-trip test uses it correctly). Since this fixture is what the follow-up wiring tests will likely copy from, worth making the canonical valid fixture representative. Coverage is unaffected, `envelope_value_check_ignores_version_and_duty_type` already proves `r#type` is ignored.

PR #1173, opened 2026-07-27
fix(ssv_types): cap Role::Proposer QBFT rounds at 2

### LINKED ISSUE #1119: fix: Proposer max_round is 6, should be 2 (match go-ssv)
## Summary

`Role::Proposer` has `max_round = Some(6)`; go-ssv caps the proposer QBFT round at 2. Anchor therefore accepts consensus messages at rounds 3-6 for a proposer duty that go-ssv drops.

## Current behavior

`common/ssv_types/src/msgid.rs:76`:

```rust
Role::Proposer | Role::SyncCommittee => Some(6),
```

## go-ssv parity (`epbs-gloas` `90df7fa0`)

`message/validation/consensus_validation.go` `maxRound`:

```go
case spectypes.RoleProposer, spectypes.RoleEnvelopeBuilder:
    return 2, nil
case ssvtypes.RoleSyncCommitteeContribution:
    return 6, nil
```

So Proposer = 2, SyncCommitteeContribution = 6. A message with `round > maxRound` is `ErrRoundTooHigh` (IGNORE).

## Why 2

A proposer QBFT that has not decided by round 2 has already missed its in-slot broadcast deadline (attestation deadline at 25% of the slot under Gloas), so the block is late regardless of whether QBFT decides later; rounds 3-6 produce nothing usable and only widen the accept surface. This is the same rationale SIP-94 §7 uses to pin the new `RoleEnvelopeProposer` (proposer-class) round cut-off at 2.

## Fix

Split the arm: `Role::Proposer => Some(2)`, keeping `Role::SyncCommittee => Some(6)` (matches go's `SyncCommitteeContribution`). Confirm anchor's `Role::SyncCommittee` is the QBFT contribution role before leaving it at 6. Add a test asserting round 3 is rejected for `Role::Proposer`.

Note: the forthcoming `RoleEnvelopeProposer` must also use 2 (SIP-94 §7, tracked in #1120); this issue is the separate pre-existing Proposer divergence.

Refs at anchor `upstream/epbs` `35191521`, go-ssv `epbs-gloas` `90df7fa0`.

jnhsigmap:

- `Role::Proposer` capped QBFT at round 6, so Anchor accepted proposer consensus messages at rounds 3-6 that go-ssv drops as `ErrRoundTooHigh`.
- go-ssv `epbs-gloas` `message/validation/consensus_validation.go` `maxRound` returns 2 for the proposer-class roles, 6 for sync-committee contribution, 12 for the committee and aggregator roles.
- A proposer that has not decided by round 2 has already missed its in-slot broadcast deadline, so rounds 3-6 only widen the accepted-message surface. go-ssv's round timer documents the same reasoning: the 2s quick timeout is deliberately not retimed for Gloas, making the proposer effectively round-1-must-succeed.
- Closes #1119

## Change Overview

- `Role::Proposer` now shares the round 2 cap with `Role::EnvelopeProposer`, mirroring how go-ssv groups the two proposer-class roles.
- `Role::SyncCommittee` keeps its round 6 cap in its own match arm.
- Updated `max_round` comment citing SSV SIP-94 (reference now cannot go stale).
- Unchanged: `Role::EnvelopeProposer`, committee and aggregator, the non-QBFT roles, round-timer durations, and timeout-mode selection.

- Proposer consensus messages at rounds 3-6 are now ignored instead of accepted, and a local proposer instance times out one round-change earlier. Both are the intended go-ssv parity.
- Anchor uses one value for the gossip gate and the local instance limit, while `ssv` keeps its local cut-off role-independent at 12. However, rounds above the validation cap cannot reach quorum because peers drop them, so stopping the local instance there forfeits no reachable consensus.
- Blast radius is one constant. No signature, wire-format, or control-flow change.

## Validation

- New unit test pins the proposer cap at 2 and the sync-committee cap at 6, so a future re-merge of the arms fails loudly.
- New message-validation tests assert a proposer message at round 3 is rejected with `RoundTooHigh` and one at round 2 is still accepted, covering both sides of the boundary.
- `cargo test -p ssv_types -p message_validator`: 220 passed, 0 failed.
- `make cargo-fmt-check` and `make lint`: clean.

## Rollback

- Revert the commit. The match arms return to a shared round 6 cap for `Role::Proposer` and `Role::SyncCommittee`. No config, data, or migration impact.

### DIFF
--- a/anchor/common/ssv_types/src/msgid.rs
+++ b/anchor/common/ssv_types/src/msgid.rs
@@ -76,11 +76,12 @@ impl Role {
     }
 
     pub fn max_round(self) -> Option<u64> {
-        // as per https://github.com/ssvlabs/ssv/blob/6382d4b52ea5e0efd9378a5a00ef481f39d6234f/message/validation/consensus_validation.go#L370
+        // Caps both the incoming consensus-message round gate and the local QBFT
+        // instance round limit. ePBS (Gloas) caps specified https://github.com/ssvlabs/SIPs/pull/94.
         match self {
             Role::Committee | Role::Aggregator | Role::AggregatorCommittee => Some(12),
-            Role::Proposer | Role::SyncCommittee => Some(6),
-            Role::EnvelopeProposer => Some(2),
+            Role::Proposer | Role::EnvelopeProposer => Some(2),
+            Role::SyncCommittee => Some(6),
             // These roles don't use QBFT consensus
             Role::ValidatorRegistration
             | Role::VoluntaryExit
@@ -517,6 +518,21 @@ mod tests {
             msg_id.duty_executor(),
             Some(DutyExecutor::Validator(pk)),
             "EnvelopeProposer resolves to a validator-scoped duty executor"
+        );
+    }
+
+    /// Pins the proposer QBFT and sync-committee round caps.
+    #[test]
+    fn proposer_and_sync_committee_max_rounds_are_pinned() {
+        assert_eq!(
+            Role::Proposer.max_round(),
+            Some(2),
+            "`Role::Proposer` must cap QBFT at round 2"
+        );
+        assert_eq!(
+            Role::SyncCommittee.max_round(),
+            Some(6),
+            "`Role::SyncCommittee` must maintain its round 6 cap"
         );
     }
 }
--- a/anchor/common/ssv_types/src/msgid.rs
+++ b/anchor/common/ssv_types/src/msgid.rs
@@ -77,7 +77,8 @@ impl Role {
 
     pub fn max_round(self) -> Option<u64> {
         // Caps both the incoming consensus-message round gate and the local QBFT
-        // instance round limit. ePBS (Gloas) caps specified https://github.com/ssvlabs/SIPs/pull/94.
+        // instance round limit. Values mirror go-ssv's maxRound:
+        // https://github.com/ssvlabs/ssv/blob/d2352a3dba3e7b309ef090b7a23f4cac1d9002d1/message/validation/consensus_validation.go#L434-L443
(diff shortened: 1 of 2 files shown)


jnhsigmap:
pretty sure local claude re-pushed that :roll_eyes: 
@claude, review this PR

jnhsigmap:
I updated the comment because it was referencing a commit and was stale. the SIP hasn't merged yet and the `ssv` `epbs` branch will likely change before it gets merged anywhere so just being conservative.

shane-moore:
Looks great! Small nit: SIP-94 §7 only pins the cut-off for the new roles; `Proposer` = 2 is pre-existing go-ssv behavior. Simpler to just cite go-ssv:

```rust
// Caps both the incoming consensus-message round gate and the local QBFT
// instance round limit. Values mirror go-ssv's maxRound:
// https://github.com/ssvlabs/ssv/blob/d2352a3dba3e7b309ef090b7a23f4cac1d9002d1/message/validation/consensus_validation.go#L434-L443
```

jnhsigmap:
> Looks great! Small nit: SIP-94 §7 only pins the cut-off for the new roles; `Proposer` = 2 is pre-existing go-ssv behavior. Simpler to just cite go-ssv:
> 
> ```rust
> // Caps both the incoming consensus-message round gate and the local QBFT
> // instance round limit. Values mirror go-ssv's maxRound:
> // https://github.com/ssvlabs/ssv/blob/d2352a3dba3e7b309ef090b7a23f4cac1d9002d1/message/validation/consensus_validation.go#L434-L443
> ```

yeah good point, fixed

shane-moore (approved):

Issue #1178, opened 2026-07-27
feat(message_receiver): quarantine Unknown-view Early RANDAO shares and promote on complete duty install


shane-moore:
**Blocked by:** #1177, #1183, and #1184

## Goal

Implement Anchor's Unknown-only subset of [SIP-101's optional duty-assignment retention](https://github.com/ssvlabs/SIPs/blob/82798537fea271d1a31d02fdc000d87b37fa3d98/sips/early_randao_preconsensus.md#L100-L108): keep an otherwise-valid Early RANDAO candidate locally when its proposer-duty view is Unknown, then fully revalidate it after a complete proposer schedule is installed. SIP-101 also permits retaining Known-unassigned candidates when duty assignment is their sole non-passing check, but Anchor initially does not admit that higher-supply class.

SIP-101 makes retention a receiver `MAY`, and permits implementation-specific earlier discard. Anchor intentionally includes Unknown retention in the Early RANDAO milestone because it can recover shares that arrived before the beacon-node duty response. Keeping Known-unassigned outside Anchor's initial cache is a conformant local policy choice. Once Anchor retains a candidate, the SIP's admission, isolation, and promotion safety requirements are mandatory.

## Dependency model

#1176 remains closed. Its current and next epoch duty-fetch work was absorbed by #1182. The remaining chain is:

1. #1183 retains and shape-validates complete proposer-duty responses.
2. #1184 atomically installs complete schedules and provides `proposer_duties_watch()`.
3. This issue consumes that watch to retry retained candidates.

The watch is therefore not lost when #1176 stays closed.

## Protocol classification

The initial gossip verdict for an otherwise-valid candidate under either Unknown or Known-unassigned is always IGNORE, whether retention is enabled or disabled. SIP-101 permits local retention for either state when duty assignment is the sole non-passing check, and a receiver that retains neither remains conformant.

Anchor initially admits only Unknown candidates. Known-unassigned candidates keep the same wire-visible IGNORE and are not retained. This narrows continuous operator-signature verification, cache-contention, and revalidation exposure while preserving recovery for missing duty views.

If Anchor retains the candidate, it must satisfy these SIP requirements:

- Before retention, the candidate passed operator-signature verification and every applicable message-validation rule other than duty assignment.
- Initial retention does not forward the message, penalize the peer, feed signature collection or a reconstruction cache, or mutate ordinary validation state.
- Promotion performs full validation again against the then-current duty view and applies every stateful rule at that time.
[...]

Issue #1177, opened 2026-07-27
feat(message_validator): early RANDAO receiver rules: eligibility predicate, earliness window, ordering exemption, duty tri-state


shane-moore:
**Blocked by:** #1183, #1184, and #1190

## Goal

Implement the mandatory receiver rules in [SIP-101 at 8279853](https://github.com/ssvlabs/SIPs/blob/82798537fea271d1a31d02fdc000d87b37fa3d98/sips/early_randao_preconsensus.md): qualifying-message classification, target-slot activation, the two-slot earliness window and dedicated clock tolerance, message-kind-scoped slot-ordering exemption, complete-schedule duty authority, and the correct gossip verdicts.

Optional local retention is implemented separately in #1178. SIP-101 permits retention when either Unknown or Known-unassigned is the sole non-passing check, but Anchor initially chooses an Unknown-only subset. This issue supplies the internal, candidate-specific Unknown outcome that #1178 may consume. Known-unassigned still receives the mandatory wire-visible IGNORE, but is not offered to retention under Anchor's current local policy.

## Protocol classification

The receiver behavior in this issue is SIP `MUST` behavior:

- Determine eligibility from the stamped proposal slot `S`, with `S >= EARLY_RANDAO_LEAD` and `epoch(S) >= GLOAS_FORK_EPOCH`.
- Enable the receiver path in time for a candidate targeting the first Gloas slot `F` to arrive at receiver-local `slot_start(F - 2) - EARLY_RANDAO_CLOCK_TOLERANCE`.
- Apply the candidate timing bound of `2 * SLOT_DURATION + 1000ms`, inclusive on the accept side. This replaces, rather than adds to, the generic tolerance for this rule.
- Exempt qualifying RANDAO candidates from the per-operator, per-`MessageID` high-water rule in both directions. A candidate below the high-water mark is not rejected for ordering, and accepting it does not advance the high-water mark applied to other proposer traffic.
- Treat Known-unassigned and Unknown duty views as IGNORE, never REJECT solely because of the duty view.
- Mutate ordinary validation state only on acceptance or successful promotion, never on IGNORE or initial retention.

[...]

PR #1182, opened 2026-07-28
fix(duties_tracker): validate proposer assignments against complete current/next views

### LINKED ISSUE #1142: fix(duties_tracker): validate proposer assignments against complete current/next duty views (SIP-94 §5/§7)
## Goal

Implement the SIP-94 §5/§7 receive-side guarantee for proposer-assignment validation without making proposer-duty freshness depend on Anchor's SSV validator snapshot.

Anchor already downloads the beacon node's complete proposer schedule. Retain that complete schedule for the current and next epoch, then validate `Role::ProposerPreferences` (8) assignments by the validator pubkey carried in the message ID. Return `NoDuty` only when a fetched, complete epoch view proves that the validator is not assigned at `proposal_slot`; an unfetched epoch remains unknown and is tolerated.

Expose the same pubkey-based assignment query for the §6 `RoleEnvelopeProposer` (9) arm in milestone 8.

## Context / motivation

SIP-94 was updated for this in [`0a8de9e`](https://github.com/shane-moore/SIPs/commit/0a8de9e276894706b90e616e93769a481debf340), following [iurii-ssv's devnet finding](https://github.com/ssvlabs/SIPs/pull/94#issuecomment-4984927696). A §5 preference partial is a deterministic one-shot gossip broadcast. Anchor derives the gossipsub message ID from the full message bytes and retains duplicate IDs for an epoch, so an identical rebroadcast cannot recover from an initial wrong drop.

The observed failure is:

1. A receiver fetches proposer duties before a validator registration wave.
2. The beacon node returns the complete epoch schedule, but Anchor filters it against `NetworkState::validator_indices()` at fetch time (`duties_tracker/src/duties_tracker.rs:205-257`).
3. A newly registered validator is therefore absent from the stored vector even though its beacon proposer assignment already existed.
4. `is_epoch_known_for_proposers` still returns `true`.
5. The role-8 arm from #1062 returns `ValidationFailure::NoDuty` -> `MessageAcceptance::Ignore`, permanently losing the honest partial.

The stale view is created by Anchor's local filtering, not by the beacon response. An epoch contains one proposer assignment per slot, so retaining the full response is small and makes the assignment view independent of SSV validator membership and index-update timing.

Pubkey lookup is also load-bearing. A newly registered validator can exist in `NetworkState` before its beacon index is populated: `get_committee_info_by_validator_pk` then returns committee membership with an empty `validator_indices` vector (`database/src/state.rs:374-389`). Role 8 must not turn that temporary index gap into `UnexpectedFailure` / REJECT when the complete proposer view can validate the assignment directly by pubkey.

### 1. Retain complete current and next epoch proposer views (`duties_tracker`)

Refactor `poll_beacon_proposers` to fetch both `current_epoch` and `current_epoch + 1`, matching the role-8 proposer lookahead.

For each epoch:

- call the existing proposer-duties endpoint;
- store `response.data` unchanged, without filtering against `network_state_rx` or `validator_indices()`;
- preserve each successful epoch independently if the other request fails;
- keep the existing historical pruning.

No `database` generation counter or validator-set invalidation protocol is needed: SSV registration changes no longer change the contents of a cached proposer view.

### 2. Add one atomic pubkey-based assignment query (`DutiesProvider`)

Keep the existing `is_epoch_known_for_proposers` and index-based `is_validator_proposer_at_slot` methods for the existing `Role::Proposer` path. Add:

```rust
fn proposer_assignment_at_slot(
    &self,
    slot: Slot,
    validator_pubkey: &PublicKeyBytes,
) -> Option<bool>;
```

Semantics:

- `Some(true)`: the epoch is fetched and the complete view assigns this pubkey at `slot`;
- `Some(false)`: the epoch is fetched and the complete view proves this pubkey is not assigned at `slot`;
- `None`: the epoch is not fetched, so assignment is unknown and must be tolerated.

Compute the result under one proposer-map read. Do not split freshness and assignment into separate predicate calls.

### 3. Consume the tri-state result in the role-8 arm (`message_validator`)

In the `Role::ProposerPreferences` arm introduced by #1062:

- extract the validator pubkey from `DutyExecutor::Validator` in the message ID;
- query `proposer_assignment_at_slot(slot, &validator_pubkey)`;
- return `ValidationFailure::NoDuty` only for `Some(false)`;
- accept `Some(true)` and tolerate `None`;
- do not require `committee_info.validator_indices.first()` for this role.

The validator is still required to exist in Anchor's registry and the signer is still required to belong to its committee through the earlier message-validation path. Only the proposer-assignment lookup stops depending on the locally resolved validator index.

## Acceptance criteria

- Proposer duties returned by the beacon node are retained even when their validator pubkeys or indices are absent from `NetworkState`.
- Successful current and next epoch fetches are both stored; failure for one epoch does not discard a successful fetch for the other.
- `proposer_assignment_at_slot` returns:
  - `Some(true)` for the assigned pubkey and slot;
  - `Some(false)` for an unassigned pubkey in a fetched epoch;
  - `None` for an unfetched epoch.
- A role-8 message for an assigned, registered validator is accepted even when its local validator index is still unresolved.
- A role-8 message against a fetched view where the pubkey is not assigned at `proposal_slot` returns `NoDuty` (IGNORE).
- A role-8 message for an unfetched epoch is accepted (existing unknown-view tolerance).
- Existing `Role::Proposer` (6), `Role::SyncCommittee`, and database behavior are unchanged.
- The future `RoleEnvelopeProposer` (9) assignment arm can reuse `proposer_assignment_at_slot` without adding validator-set freshness state.

## Tests

Use the tester-subagent before authoring or modifying tests.

- `duties_tracker`:
  - stores the complete proposer response when `NetworkState::validator_indices()` is empty;
  - stores current and next epoch views;
  - preserves one successful epoch when the other fetch fails;
  - covers the `Some(true)` / `Some(false)` / `None` assignment states.
- `message_validator` (extend #1062's role-8 harness):
  - assigned pubkey + unresolved local index -> accepted;
  - fetched epoch + unassigned pubkey -> `NoDuty`;
  - assigned pubkey -> accepted;
  - unknown epoch -> accepted.
- Regression-pin the existing index-based proposer path.

Verification: `make cargo-fmt && make lint`, then `cargo test -p duties_tracker -p message_validator`.

## Notes

- Stacks on #1062 / PR #1106, which introduces the role-8 arm and its anti-spam caps.
- Related: #1063 / PR #1125 (sign path), #1064 (Lighthouse service wiring), and #1131 (peer-aware duplicate classification).
- The emitter-side validator-added trigger and recommended registration grace are tracked by correcting #1064 separately. The current pinned Lighthouse service schedules once per `(epoch, dependent_root)`, so a newly discovered validator under an unchanged root is not emitted without an upstream fix.
- Out of scope: duty-view invalidation during a `dependent_root` reorg, the existing `Role::Proposer` (6) behavior, and the future role-9 validation arm itself.

### LINKED ISSUE #1147: refactor(message_validator): validate EnvelopeProposer assignment by pubkey via proposer_assignment_at_slot
**Blocked by:** #1120, #1142

## Goal

Replace the index-based proposer-duty check in the `Role::EnvelopeProposer` (9) arm of `validate_beacon_duty` with the pubkey-based tri-state query `proposer_assignment_at_slot` introduced by #1142.

## Context / motivation

#1120 specifies the role-9 duty check as `is_epoch_known_for_proposers(epoch) && !is_validator_proposer_at_slot(slot, validator_index)`, mirroring the existing `Role::Proposer` arm (`message_validator/src/lib.rs:786-818`). #1142 established that this predicate pair has two failure modes: a false `NoDuty` when the epoch view was fetched but filtered against the local validator set, and `UnexpectedFailure` when the validator exists in `NetworkState` but its beacon index is not yet resolved. Anchor's gossipsub message-ID dedup makes an initial wrong drop unrecoverable, and an envelope partial is a one-shot per proposal, so both modes apply to role 9 exactly as they do to role 8. #1142 fixes role 8 and exposes the query on `DutiesProvider` (`duties_tracker/src/lib.rs:113`) for this arm to reuse.

## Suggested approach

- In the role-9 arm, take the validator pubkey from the message ID's `DutyExecutor::Validator` (`ssv_types/src/msgid.rs:91-94`) and call `proposer_assignment_at_slot(slot, &pubkey)`.
- Return `ValidationFailure::NoDuty` only for `Some(false)`; accept `Some(true)`; tolerate `None` (unfetched epoch).
- Drop the `committee_info.validator_indices.first()` dependency for this role.
- No `duties_tracker` changes: consume the query exactly as landed by #1142.

## Acceptance criteria

- An envelope message for an assigned pubkey with an unresolved local validator index is accepted.
- A fetched epoch view where the pubkey is not assigned at the slot returns `NoDuty` (Ignore).
- An unfetched epoch is tolerated (accepted).
- `Role::Proposer` (6) and `Role::ProposerPreferences` (8) arms are unchanged.

## Tests

- Extend the role-9 duty harness from #1120 with the three tri-state cases above.
- Regression-pin that ordinary `Role::Proposer` still uses the index-based path.
- `cargo test -p message_validator`

## Notes

petarjuki7:

- The receive-side proposer-assignment check for the ProposerPreferences duty depended on Anchor's local SSV validator snapshot. A validator whose beacon proposer assignment already existed could be dropped from the retained schedule when it was absent from that snapshot at fetch time (for example during a registration wave), or present but with an unpopulated beacon index.
- A ProposerPreferences partial is a one-shot gossip broadcast whose message id is retained for an epoch, so an initial wrong drop is not recoverable by rebroadcast and an honest partial is lost permanently.
- Worth doing now: unblocks correct ProposerPreferences validation for the ePBS Proposer Preferences Duty milestone.
- Evidence: iurii-ssv devnet finding; SIP-94 was updated accordingly.
- Links: closes #1142, closes #1147; SIP-94 §5/§7.

## Change Overview

- Retain the beacon node's complete proposer schedule for the current and next epoch, without filtering against the local validator set.
- Validate ProposerPreferences assignments by the validator pubkey from the message id, using a single tri-state query: assigned, not-assigned-in-a-fetched-epoch, or unknown-epoch. `NoDuty` (Ignore) is returned only for the proven not-assigned case; an unfetched epoch is tolerated.
- Split the proposer-duty poll into a fetch step and a store step; each epoch is stored independently, and one epoch's fetch failure does not discard the other.
- Intentionally unchanged: the index-based `Role::Proposer` and `SyncCommittee` paths and all database behavior. The role-9 `EnvelopeProposer` assignment arm reuses the same query.

- Blast radius limited to proposer-duty retention and the ProposerPreferences/EnvelopeProposer validation arm.
- Trade-off: the retained schedule now holds all proposers for two epochs rather than only local validators. That is one assignment per slot, so the added memory is negligible.
- The slot-bound assignment check and the tri-state tolerance are covered by unit tests; the index-based path is regression-pinned.

## Validation

- Unit tests cover: the tri-state assignment query including the slot bind; the role-8 accept/ignore/tolerate outcomes; acceptance with an unresolved local validator index; the `NoDuty` -> Ignore mapping; and a regression pin that the `Role::Proposer` index path is unchanged.
- `make cargo-fmt`, `make lint`, and `cargo test -p duties_tracker -p message_validator` pass.
- Not unit-covered: the beacon-node fetch itself (the two-epoch download at the network boundary), as the workspace has no HTTP mock seam. The per-epoch store and error aggregation now sit inside that fetch loop, so they are not unit-covered either; #1183 reshapes this storage and establishes that coverage on the final response shape.

## Rollback

- Revert the commit. No config, data, schema, or operational impact.

## Blockers / Dependencies

- Stacks on the role-8 arm and the sign/service wiring already merged for this milestone.

- Also closes #1147: the `Role::EnvelopeProposer` (9) assignment check is handled by the same shared arm, so it moves to the pubkey-based tri-state query with no extra production code; a role-9 regression covers the unresolved-index acceptance through the full partial-signature pipeline.
- The proposer-assignment query returns a named `DutyAssignment { Assigned, NotAssigned, Unknown }` verdict rather than `Option<bool>`, so the verdict is self-describing at the call site and reusable for the role-7 PTC arm.
- Proposer duties are stored per epoch as each fetch returns, so a completed current-epoch view stays queryable during the next-epoch fetch. A timing regression test for that ordering needs a beacon-node test double and will land with that harness.


anchor/duties_tracker/src/duties_tracker.rs
```diff
@@ -207,53 +208,56 @@ impl<T: SlotClock + 'static> DutiesTracker<T> {
...
+        for epoch in [current_epoch, current_epoch + 1] {
+            let result = self
+                .beacon_nodes
+                .first_success(|beacon_node| async move {
+                    beacon_node.get_validator_duties_proposer(epoch).await
+                })
+                .await
+                .map(|response| response.data)
+                .map_err(|e| Error::FailedToPollProposers(e.to_string()));
+            fetched.push((epoch, result));
```
shane-moore: **[P1] Store the current view before starting the next request**

Could we store each successful epoch here before starting the next request? A successful current response currently stays in `fetched` while `first_success` for `current_epoch + 1` runs. If Lighthouse's separate proposer cache has the current duty but this tracker does not, Lighthouse can start the proposal duty during that wait. The absent epoch becomes `NoDuty`, and outgoing self-validation returns before gossiping the packet. The first-slot RANDAO exception does not cover proposer QBFT or post-consensus messages.

Per-epoch stores also become the seam the early RANDAO quarantine (#1178) hangs its promotion trigger on: retained shares should promote the moment their epoch's fetch completes, not after the sibling epoch's request resolves.

Please store each successful result inside the loop, before starting the next iteration. Preserve the existing error behavior and prune after both attempts. The regression test should keep the next-epoch request pending, assert that the current assignment is already queryable, then fail the next request and confirm the current view remains stored.

anchor/duties_tracker/src/lib.rs
```diff
@@ -122,4 +122,10 @@ pub trait DutiesProvider: Sync + Send + 'static {
     fn is_validator_proposer_at_slot(&self, slot: Slot, validator_index: ValidatorIndex) -> bool;
 
     fn get_voluntary_exit_duty_count(&self, slot: Slot, pubkey: &PublicKeyBytes) -> u64;
+
+    fn proposer_assignment_at_slot(
+        &self,
+        slot: Slot,
+        validator_pubkey: &PublicKeyBytes,
+    ) -> Option<bool>;
```
shane-moore: **[P2] Consider a named tri-state instead of `Option<bool>`**

Non-blocking. Raising it now only because the method is new, so the migration never gets cheaper than the two call sites it has today.

Three things push me toward a named enum:

- The call site needs its comment to be readable: `if let Some(false) = duty_provider.proposer_assignment_at_slot(slot, &pubkey)`. Nothing in the type says which layer is fetched-ness and which is assignment.
- SIP-94 §7 requires the same tri-state for `RolePTCAttester` assignment, and we have no role-7 arm yet. A role-agnostic type gets reused there; `Option<bool>` gets copy-pasted.
- `None` is going to mean more than "not fetched". The dependent-root wording in §7 folds root-stale, refetch-pending, refetch-failed and beacon-node disagreement into the same tolerated state, and we will want counters split by which one. There is nowhere to hang that on `Option<bool>`, so the signature changes then anyway.

Something like:

```rust
/// Whether a validator holds a duty at a slot, as one atomic verdict over the stored duty view.
pub enum DutyAssignment {
    /// A complete fetched view assigns the validator at this slot.
    Assigned,
    /// A complete fetched view proves the validator is not assigned at this slot.
    NotAssigned,
    /// The view cannot answer: no completed fetch for the slot's epoch.
    Unknown,
}
```

I would keep it payload-free here and let the reason land with the metrics that consume it, rather than ship a reason enum with one variant.

Context for why I care: I am building the early RANDAO receiver rules (#1177) on this method, and that PR moves `Role::Proposer` onto it as well, so it picks up two more call sites shortly. Happy to do this myself in a follow-up if you would rather not grow this PR.

shane-moore:
**Non-blocking: finish the #1147 closure**

This PR implements #1147's production behavior in the shared role-8/role-9 arm: `EnvelopeProposer` now uses the message-ID pubkey, consumes `proposer_assignment_at_slot`, and returns `NoDuty` only for `Some(false)`.

Could we also add a full role-9 partial-signature regression with `committee_info.validator_indices = vec![]` and `proposer_assignment = Some(true)`? The current role-9 tri-state tests call `validate_beacon_duty` directly with populated indices, so they do not cover #1147's unresolved-index failure mode through the full post-consensus path. I verified locally that this case passes with the current code, so no production change should be required.

Once that test is present, please add `Closes #1147` to the PR body so the implemented issue does not remain open.

shane-moore (approved):
LGTM!

shane-moore:
marked as ready-for-merge so i can build on top of this in new branch based off `epbs` branch

PR #1193, opened 2026-07-30
fix(message_sender): validate outbound messages statelessly

### LINKED ISSUE #1192: fix(message_sender): outbound validation commits state before network enqueue
## Goal

Make outbound self-validation stateless so a local message that fails before network enqueue does not leave phantom duplicate, ordering, or peer state that suppresses a later attempt for the same duty.

This is an Anchor sender-path correctness prerequisite for #1179. It is not an independent SIP-101 requirement, and it does not guarantee network redelivery of an identical Early RANDAO share.

## Context

At [`upstream/epbs` 7bad23f](https://github.com/sigp/anchor/tree/7bad23f2b37f08fb886b7d33be05963bf879b112), [`message_sender::network::do_send`](https://github.com/sigp/anchor/blob/7bad23f2b37f08fb886b7d33be05963bf879b112/anchor/message_sender/src/network.rs#L136-L193) runs the full inbound validator before slot extraction, subnet routing, channel enqueue, and eventual gossipsub publication. The inbound validator can commit duplicate and ordering state before any of those later steps succeed.

A later emission can then be suppressed by Anchor's own state even when the earlier message never reached the network. Early RANDAO exposes this on the recommended unconditional in-slot attempt, but the defect is general outbound behavior.

Removing sender-side state mutation fixes only that local suppression class. A successful earlier publication may already be present in the producer's or receivers' gossip duplicate caches, whose lifetimes can exceed the useful Early RANDAO window. A byte-identical retry may therefore be absorbed even when Anchor makes the publication attempt correctly.

## Suggested approach

- Add a stateless outbound validation entry point that checks decoding, structural invariants, and role parsing needed to prevent locally malformed messages from reaching enqueue.
- The outbound entry point must not touch the duty-state map, per-operator state, duplicate counters, high-water marks, epoch counters, or inbound duty verdicts. It should not re-run inbound RSA verification on bytes Anchor just produced.
- Make `do_send` use that stateless entry point instead of the full inbound `validate` path.
- Treat outbound validation failure as a local construction error and drop before enqueue.
- Document the ownership boundary: the receiver path is the stateful validator; sender-side sanity checks are non-mutating; enqueue and publication failures leave no inbound-validation-state trace.
- Preserve normal local consumption paths. QBFT and signature collection already consume locally produced values through their own paths rather than by depending on `do_send` to mutate inbound state.

## SIP-101 interaction

- #1177's candidate-specific duty handling and #1178's Unknown-only retention handoff apply to inbound receiver validation. They must not suppress an operator's own early share on the outbound path.
- #1179 can make the unconditional in-slot publication attempt after an early attempt without being blocked by sender-side duplicate state.
- This still does not promise a second network delivery. Gossip may deduplicate identical bytes after a successful earlier publication.

## Acceptance criteria

- Outbound validation mutates no inbound validation state.
- A byte-identical inbound copy after outbound validation is judged solely by inbound state and rules, not by a sender-side phantom entry.
- An outbound multi-signer message leaves peer operator state untouched.
- Failure during slot extraction, routing, channel enqueue, or downstream publication leaves no inbound-validation-state trace.
- A later local attempt for the duty is not suppressed by state committed by `do_send`.
- Undecodable or structurally invalid outbound messages are still dropped before enqueue.
- Inbound validation behavior remains unchanged.
- The implementation and dependent issues do not claim guaranteed receiver delivery or reliable recovery solely because this fix lands.

## Tests

Cover well-formed partial-signature and consensus messages, undecodable and structurally invalid bytes, outbound-then-inbound identical bytes, outbound multi-signer followed by a distinct inbound message with the same signer set, failure before enqueue followed by a successful retry, and assertions that every inbound validation-state structure is unchanged by the outbound entry point.

Add an Early RANDAO regression showing that the in-slot attempt reaches the publication path after an earlier local attempt without being blocked by sender-side state. The test should not assert that a gossip layer delivers the byte-identical message twice.

## Notes

- #1190 owns inbound proposer duty verdicts and the proposer QBFT dispatch allocation gate.
- #1177 owns the inbound Early RANDAO candidate result.
- #1179 owns producer scheduling and the unconditional in-slot attempt.

shane-moore:

`NetworkMessageSender::do_send` reused the inbound `Validator`, so outbound self-validation mutated shared duty and duplicate state before slot routing and network enqueue. If routing or enqueue then failed, a later re-emission could be suppressed even though the original message never reached the wire.

Closes #1192.

## Change Overview (Required)

- Add raw-byte `validate_outbound` and typed `validate_outbound_message` stateless preflight helpers. The sender validates its existing `SignedSSVMessage` before serialization, checks structural invariants and role, and returns the nested routing slot.
- Remove the stateful `Validator` and `DutiesProvider` coupling from `NetworkMessageSender`.
- Keep incoming validation behavior and pre-decode versus post-decode classification unchanged.
- Deliberately leave network, duty, timing, fork-role, signature-verification, and validation-state checks to outbound producers and the incoming validator.

- A semantically invalid but structurally valid locally produced message is no longer stopped by sender-side inbound validation. Peers still run the full incoming validator.
- #1201 fixes the type-3 pre-consensus packaging case and is already present on `epbs`. #1208 fixes the separate type-0 post-consensus case and is merged on `unstable`, but is not present on `epbs`. An Alan deployment containing this change must also include #1208 or an equivalent type-0 fix. Boole-only deployments do not require it.
- Structurally malformed messages are still dropped before routing or enqueue.

## Validation (Required)

- `cargo test -p message_validator -p message_sender`, 120 validator tests passed.
- `cargo check -p client`.
- `make cargo-fmt-check`.
- `make lint`, full workspace Clippy with warnings denied.
- `git diff upstream/epbs...HEAD --check`.

Revert the commit. There are no data migrations, configuration changes, or persistent state changes.

#1208, or an equivalent type-0 post-consensus fix, is conditionally required only for deployments whose active SSV fork is still Alan.

### DIFF
--- a/anchor/client/src/lib.rs
+++ b/anchor/client/src/lib.rs
@@ -562,7 +562,6 @@ impl Client {
                     network_tx: network_tx.clone(),
                     private_key: key.clone(),
                     operator_id: operator_id.clone(),
-                    validator: Some(message_validator.clone()),
                     is_synced: is_synced.clone(),
                     subnet_service: subnet_service.clone(),
                 },
--- a/anchor/message_sender/src/network.rs
+++ b/anchor/message_sender/src/network.rs
@@ -1,7 +1,7 @@
 use std::sync::Arc;
 
 use database::OwnOperatorId;
-use message_validator::{DutiesProvider, MessageAcceptance, TopicContext, Validator};
+use message_validator::validate_outbound;
 use openssl::{
     hash::MessageDigest,
     pkey::{PKey, Private},
@@ -15,7 +15,7 @@ use ssv_types::{
 use ssz::Encode;
 use subnet_service::SubnetService;
 use tokio::sync::{mpsc, mpsc::error::TrySendError, watch};
-use tracing::{debug, error, trace, warn};
+use tracing::{error, trace, warn};
 
 use crate::{Error, MessageCallback, MessageSender, SigningError};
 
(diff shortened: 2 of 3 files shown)


shane-moore:
@claude, please take a look

shane-moore:
@claude, thanks, good catch. Addressed in `4387e802`.

`do_send` now validates the existing typed `SignedSSVMessage` before serialization. The byte-oriented `validate_outbound` remains as the raw-input wrapper and delegates to `validate_outbound_message`, preserving outer decode errors without paying the encode/decode round trip on the sender path. Structural validation is retained for aggregated messages.

The #1052 deployment condition is tracked in milestone 9 and this PR’s Blockers section: any Alan deployment containing #1192 must include #1052 or an equivalent fix; Boole-only ePBS deployments do not require it.

Validation: 112 `message_validator` tests, `cargo check -p client`, `make cargo-fmt-check`, and `make lint`.

shane-moore:
local-testnet is failing cuz of #1050, so we need to resolve that first

jnhsigmap (approved):
this is a nice simplification, with the trade-offs as you have mentioned. all good if this is the desired direction 👍

jnhsigmap:
merge conflicts are trivial

Issue #1183, opened 2026-07-28
refactor(duties_tracker): retain proposer-duty v2 response metadata


shane-moore:
**Builds on:** #1182

## Goal

Retain the complete proposer-duty v2 response for each fetched epoch, including response metadata, and reject any response shape that cannot establish the complete proposer schedule required by [SIP-101 at 8279853](https://github.com/ssvlabs/SIPs/blob/82798537fea271d1a31d02fdc000d87b37fa3d98/sips/early_randao_preconsensus.md#L86-L98).

This issue owns response shape and metadata retention. #1184 owns authoritative-view installation semantics, proposer-assignment transitions, and the notification watch consumed by #1178.

## Protocol classification

SIP-101 requires the following behavior:

- A newly fetched schedule establishes a proposer-duty view only when it contains exactly `SLOTS_PER_EPOCH` duties, one for every distinct slot in the requested epoch.
- The complete schedule is authoritative for every validator in that epoch, independent of Anchor's local SSV registry.
- A missing-slot, duplicate-slot, out-of-epoch, failed, or otherwise malformed response establishes no new view.
- A failed or malformed refresh does not invalidate a previously retained complete view.

Using the beacon API v2 response and retaining fields such as `dependent_root` and `execution_optimistic` are Anchor implementation choices. Those fields are useful for diagnostics and refresh decisions, but neither field changes SIP-101 authority. In particular, `execution_optimistic` does not create a provisional proposer-duty state.

## Current code

At [`upstream/epbs` 7bad23f](https://github.com/sigp/anchor/tree/7bad23f2b37f08fb886b7d33be05963bf879b112):

- [`ProposerMap`](https://github.com/sigp/anchor/blob/7bad23f2b37f08fb886b7d33be05963bf879b112/anchor/duties_tracker/src/lib.rs#L87-L93) stores only `Vec<ProposerData>` per epoch.
[...]

PR #1199, opened 2026-07-31
fix: tolerate unknown proposer duties without spawning QBFT

### LINKED ISSUE #1190: fix: tolerate unknown proposer duty views and gate network-triggered QBFT spawn
## Goal

Migrate general `Role::Proposer` validation to the pubkey-keyed tri-state proposer assignment, tolerate Unknown rather than treating an absent duty view as negative evidence, and prevent network input from creating a proposer QBFT instance unless the dispatch-time verdict is authoritatively `Assigned`.

This is an Anchor correctness and resource-safety prerequisite. It is not an independent SIP-101 wire requirement. #1177 builds the candidate-specific Early RANDAO IGNORE handoff on top of this general proposer migration.

## Context

At [`upstream/epbs` 7bad23f](https://github.com/sigp/anchor/tree/7bad23f2b37f08fb886b7d33be05963bf879b112), [`Role::Proposer` validation](https://github.com/sigp/anchor/blob/7bad23f2b37f08fb886b7d33be05963bf879b112/anchor/message_validator/src/lib.rs#L800-L850) still resolves a beacon validator index from local committee data and has a special first-slot RANDAO path. The proposer-preferences and envelope roles already query the validator pubkey through `proposer_assignment_at_slot` and reject only an authoritative `NotAssigned` result.

An absent complete proposer schedule is not evidence that a validator has no duty. Treating it as `NoDuty` can drop honest inbound messages. Tolerating it through validation, however, must not let network traffic allocate a new proposer QBFT instance for an unverified duty.

## Duty-view contract

After #1183 and #1184:

- `Assigned` means a retained complete schedule contains the validator at exactly the queried slot.
- `NotAssigned` means a retained complete schedule exists and contains no such exact assignment.
- `Unknown` means no complete schedule is retained for that epoch.

Only `NotAssigned` is authoritative negative evidence. Failed or malformed refreshes, local validator changes, `execution_optimistic`, and reorg detection do not create extra Unknown causes when a prior complete schedule exists. The prior schedule remains Known until a complete replacement is retained.

For general Proposer traffic, `Assigned` and `Unknown` continue validation, while `NotAssigned` returns `NoDuty`. For an eligible Early RANDAO candidate, #1177 ensures both Unknown and NotAssigned map to the protocol IGNORE verdict. SIP-101 permits optional retention for either state when duty assignment is the sole non-passing check, but Anchor's current #1178 policy offers only an otherwise-valid Unknown candidate through the internal `UnknownProposerEpoch` reason. That retention handoff is not part of this issue's general mapping.

## Suggested approach

- Merge the `Role::Proposer` duty lookup into the existing pubkey-keyed proposer-related path:
  - extract the validator pubkey from the `MessageID`;
  - query `proposer_assignment_at_slot(slot, pubkey)`;
  - continue on `Assigned` or `Unknown`;
  - return `NoDuty` on `NotAssigned`.
- Delete the validator-index lookup, empty-index failure, first-slot RANDAO carve-out, and legacy proposer-duty boolean helpers once they have no callers.
- Keep this general validation mapping slot-independent and message-kind-independent. #1177 owns the narrower Early RANDAO candidate result.
- Make QBFT dispatch two explicit operations:
  - receive-or-spawn for an authoritative `Assigned` verdict;
  - receive-if-present for `Unknown`, delivering only to an already-running instance and never allocating one.
- Recompute the proposer assignment immediately before dispatch. A verdict can change between validation and dispatch.
- Drop without buffering when dispatch sees `NotAssigned`, or when `Unknown` has no existing processor. Existing local instances may receive Unknown traffic because they were created by local duty execution or an earlier authoritative assignment.
- Keep spawn-on-receive behavior for unrelated validator and committee QBFT roles unchanged.

## Acceptance criteria

- Missing local validator index data no longer causes a proposer message to be Rejected.
- `Assigned` and `Unknown` continue general Proposer validation; `NotAssigned` yields `NoDuty`.
- First-slot RANDAO uses the same tri-state path as every other slot.
- No proposer index resolution or legacy epoch-known boolean remains in this path.
- A proposer QBFT message can create an instance only when the dispatch-time verdict is `Assigned`.
- Under `Unknown`, a message can reach an existing instance but cannot create one.
- A validation-time `Assigned` result that becomes non-assigned before dispatch cannot spawn.
- A validation-time Unknown that becomes `Assigned` before dispatch may use receive-or-spawn.
- Other QBFT roles retain existing allocation behavior.
- The implementation does not introduce root-stale, optimistic, disputed, or cause-specific duty states.

## Tests

Cover all three general proposer message kinds under `Assigned`, `NotAssigned`, and `Unknown`; missing validator index data; identical first-slot and ordinary-slot behavior; pubkey extraction; validation-to-dispatch verdict changes; receive-or-spawn only under dispatch-time `Assigned`; receive-if-present under Unknown; no allocation under Unknown or `NotAssigned`; delivery to a pre-existing instance; and unaffected controls for another validator QBFT role and a committee role.

Preserve the duties-tracker tri-state tests, updated to the complete-schedule authority contract from #1183 and #1184.

## Notes

- #1192 separately removes stateful inbound validation from the outbound send path. This issue owns inbound duty verdicts and the proposer QBFT dispatch gate.
- #1177 owns Early RANDAO candidate-specific duty handling, and #1178 owns Anchor's Unknown-only retention policy. Do not weaken the general dispatch gate to implement retention.
- Position relative to #1183 and #1184 can be resolved when the PR is rebased, but the final combined behavior must use the complete-schedule contract above.

shane-moore:

An absent proposer-duty schedule is not evidence that a validator has no duty. The current proposer path can therefore reject honest inbound messages when local duty data or validator indices are missing. Simply accepting those messages is also unsafe if network input can allocate an unverified QBFT instance.

Closes #1190.

## Change Overview (Required)

- Validate proposer duties by message-ID pubkey with the existing `Assigned`, `Unknown`, and `NotAssigned` contract.
- Continue validation for `Assigned` and `Unknown`; return `NoDuty` only for authoritative `NotAssigned`.
- Recompute the assignment inside `QbftManager::receive_network_message`: `Assigned` may create an instance, `Unknown` can use only an existing instance, and `NotAssigned` drops before map access.
- Keep allocation behavior unchanged for non-proposer QBFT roles.
- Remove the legacy validator-index and first-slot RANDAO exception paths.

No wire, topic, fork-schedule, database, or configuration change is included.

- An `Unknown` message with no existing instance is accepted by gossip validation but dropped locally without buffering. Replay and count state has already been consumed, so the same message is not deferred for later delivery.
- `EnvelopeProposer` has no local QBFT initializer today, so its `Unknown` path is normally drop-only.
- The dispatch-time lookup is point-in-time rather than transactional with map insertion.

The manager owns the gate at the typed-map routing boundary. `ExistingOnly` performs one lookup without falling through to allocation, and tests cover both proposer roles, all verdict transitions, exact-key isolation, fork gating, and unchanged controls.

## Validation (Required)

- `cargo test -p duties_tracker -p message_validator -p qbft_manager -p message_receiver --no-fail-fast`
- `cargo check -p client`
- `make cargo-fmt-check`
- `make lint`
- `make test`

Revert this PR. It has no persistent-data, configuration, or wire-format migration.

Keep as draft until:

- #1193 merges. This branch is stacked on its stateless outbound-validation changes and should drop those duplicate commits after merge.
- #1183 lands. It establishes complete proposer schedules before missing assignments become authoritative `NotAssigned` evidence.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6307,6 +6307,7 @@ dependencies = [
  "bls",
  "dashmap",
  "database",
+ "duties_tracker",
  "ethereum_ssz",
  "fork",
  "futures",
--- a/anchor/client/src/lib.rs
+++ b/anchor/client/src/lib.rs
@@ -562,7 +562,6 @@ impl Client {
                     network_tx: network_tx.clone(),
                     private_key: key.clone(),
                     operator_id: operator_id.clone(),
-                    validator: Some(message_validator.clone()),
                     is_synced: is_synced.clone(),
                     subnet_service: subnet_service.clone(),
                 },
--- a/anchor/common/ssv_types/src/msgid.rs
+++ b/anchor/common/ssv_types/src/msgid.rs
@@ -98,6 +98,21 @@ impl Role {
         self.max_round().is_some()
     }
 
+    /// Returns true if this role's duty is bound to a single validator's proposal slot, so the
+    /// proposer-duty assignment for that slot decides whether a message can have a duty at all.
+    pub fn is_proposer_scoped(self) -> bool {
+        match self {
+            Role::Proposer | Role::ProposerPreferences | Role::EnvelopeProposer => true,
+            Role::Committee
+            | Role::Aggregator
+            | Role::SyncCommittee
+            | Role::ValidatorRegistration
+            | Role::VoluntaryExit
+            | Role::PTCAttester
+            | Role::AggregatorCommittee => false,
+        }
+    }
+
     /// monotonicSlotRole reports whether a role's signer advances through slots one at a time, so a
     /// message for a slot below the signer's max is stale and must be rejected. False for
     /// committee roles (state is slot-keyed across many validators) and for proposer
--- a/anchor/duties_tracker/src/duties_tracker.rs
+++ b/anchor/duties_tracker/src/duties_tracker.rs
@@ -10,7 +10,7 @@ use task_executor::TaskExecutor;
 use thiserror::Error;
 use tokio::{sync::watch, time::sleep};
 use tracing::{debug, error, trace, warn};
-use types::{ChainSpec, Epoch, Slot};
+use types::{ChainSpec, Slot};
 
 use crate::{
     Duties, DutiesProvider, DutyAssignment, MembershipKey,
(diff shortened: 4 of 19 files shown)


shane-moore:
closing for now: early RANDAO (#1190's driver) is out of scope, and without it the current `Role::Proposer` path is go-ssv parity with no local-participation gap. keeping #1190 open; this branch can be revived from there if pre-slot proposer messages come back.

PR #1201, opened 2026-08-03
fix: batch pre-Boole sync contribution proofs

### LINKED ISSUE #1050: v1.2.3: `InvalidPartialSignatureTypeCount` outbound-drops on multi-subnet sync-committee aggregators

shane-moore:
Closes #1050.

Ethereum sync committees are position-based, so the same validator may occupy positions on multiple subnets or appear more than once within one subnet. For example, positions `[0, 1, 128]` map to subnet 0 twice and subnet 1 once. Lighthouse therefore calls Anchor once per unique subnet, two callbacks in this example, while the deployed SSV shape used by go-ssv and ssv-spec requires one `ContributionProofs` envelope per validator and slot with one entry per raw position, three entries here. Anchor instead sent one envelope per callback, causing `InvalidPartialSignatureTypeCount` and wrong-root-count errors.

This is especially visible on concentrated devnets and testnets. Although Boole bypasses this path, Anchor's built-in mainnet, Hoodi, and Holesky schedules currently have no Boole activation configured, so hardening the Alan producer remains useful.

## Change Overview

- Preserve raw sync committee position multiplicity while retaining one callback per unique subnet.
- Give every callback a canonical descriptor of all roots and multiplicities.
- Eagerly admit one complete envelope per validator and slot.
- Keep failed synchronous admission retryable through a `Pending | Admitted` record.
- Enforce the 13-position producer bound and reject malformed assignments.

Runtime behavior changes only for pre-Boole sync committee contribution selection proofs. Regular sync committee messages, other duties, Boole committee batching, public wire types, and inbound validation behavior are unchanged.

## Design Choice

We considered an all-callback accumulator, buffering in ValidatorStore or the generic sender, and a broader Lighthouse batch API. Accepting one envelope per subnet was rejected because it preserves a wire shape that diverges from go-ssv and ssv-spec.

The eager full descriptor keeps ownership aligned: ValidatorStore derives Ethereum duty facts and roots, while the signature collector owns SSV envelope construction, admission, retry, and cleanup. Any usable callback can admit the complete envelope, avoiding an all-callback barrier and allowing a sibling to retry synchronous admission failure.

The blast radius is intentionally narrow. The new mutex covers only bounded synchronous signing, SSZ construction, and nonblocking queue admission, with no await or blocking send while held. Descriptor, retry, cleanup, mismatch, multiplicity, and concurrency tests cover the new state.

## Validation

Before the standalone rebase:

- `git diff --check`
- `make cargo-fmt-check`
- Focused release tests for `signature_collector`, `anchor_validator_store`, `message_validator`, and `ssv_types`
- `make lint`
- Full `make test`

After replaying the #1050 commit onto current `unstable`:

- `git range-diff` confirmed the rebased commit preserves the original patch identity
- `git diff --check`
- `make cargo-fmt-check`
- `cargo test --release -p signature_collector`, 20 passed
- `cargo test --release -p anchor_validator_store`, 40 passed
- `cargo test --release -p message_validator`, 57 passed

A mixed pre-Boole ssv-mini run used three candidate Anchor operators and one go-ssv operator. Ten validators each held seven raw positions across four unique subnets.

- 3,360 / 3,360 validator-slot pre-consensus quorums succeeded
- Every accepted envelope contained all seven entries
- Every quorum contained exactly four unique roots
- Zero type 3 wrong-root-count rejections
- Zero restarts, OOMs, monitor alerts, or descriptor mismatch logs

## Rollback

Revert the #1050 commit. There are no configuration, storage, migration, or wire-format changes.

## Dependencies

#1198 has merged. This branch has been replayed onto current `unstable` and now contains one standalone #1050 commit with only this PR's diff. There are no remaining open PR dependencies.

The ssv-mini run also exposed a separate pre-existing post-consensus multi-root batching issue. It is intentionally out of scope here. The owned descriptor and eager `Pending | Admitted` admission model can be extended with a phase-aware key, or a separate post-consensus record, to address that path next.


shane-moore:
@claude, please take a look

shane-moore:
@claude Thanks for the review. I checked the optional observations against the current head and am not changing #1201:

- The function-length concern is fair, but keeping `process_single_validator_batch` linear makes record validation, the mutex boundary, message construction, and admission/retry invariants easy to audit. A generic suppression helper would also obscure the distinct structured diagnostics. Any pure construction refactor can be separate.
- The 13-entry semantic cap intentionally belongs to ValidatorStore, with independent inbound enforcement. Duplicating it in SignatureCollector would split ownership; #1208 generalizes the same boundary to both phases.
- `committee_id_mismatch` and `validator_index_mismatch` are already exercised by `mismatched_callbacks_suppress_publication_but_inject_current_share`. The Codecov zero-hit lines are structured logging fields, not the branch and return. #1208 covers the genuinely missing defensive paths.

No changes are needed for these non-blocking observations.

jnhsigmap (approved):

PR #1203, opened 2026-08-03
fix(message_validator): classify proposer-preferences duplicate roots per peer

### LINKED ISSUE #1131: fix(message_validator): classify proposer-preferences duplicate roots per peer (SIP-94 §7)
## Goal

Implement SIP-94 §7 two-tier duplicate classification for `Role::ProposerPreferences` (8): a repeated signing root from the same propagation peer is REJECT, while a relay of a globally-seen root from a different peer is IGNORE. This requires threading the libp2p `propagation_source` into message validation, which currently drops it before the validator runs.

## Context / motivation

SIP-94 §7 pins the dedup rule: "a repeat of a seen root is a duplicate (same-peer REJECT, relayed IGNORE, the existing two-tier handling)" ([epbs_support.md at `938648d`](https://github.com/ssvlabs/SIPs/blob/938648d8b9613d8f29c77d10c07e1ff8d9b70cbc/sips/epbs_support.md#7-ssv-message-validation)). go-ssv implements it with per-peer state checked before world state ([ssvlabs/ssv#2901, `message/validation/partial_validation.go:300-320` at `679eb43d7`](https://github.com/ssvlabs/ssv/blob/679eb43d76a4021aeb8a4f1d9b3c672613dba68d/message/validation/partial_validation.go#L300-L320)).

Anchor cannot comply after #1106: `NetworkMessageReceiver::receive` has `propagation_source: PeerId` but calls `validator.validate(&message.data, &topic_context)` without it (`anchor/message_receiver/src/manager.rs:83` at the #1106 head); the peer is only attached to the post-verdict `Outcome` (`manager.rs:95`). The dedup state is a peer-blind `HashSet<Hash256>` (`SignerState::seen_preferences`, `anchor/message_validator/src/duty_state.rs:332`), so every repeated root returns `DuplicatedMessage`, which maps to Reject (`anchor/message_validator/src/lib.rs:236-262`, default arm) and applies the gossipsub P4 invalid-message penalty to whichever peer delivered it. Peer scoring is live (`anchor/network/src/behaviour.rs:151`) with the P4 weight sized against the graylist threshold (`anchor/network/src/scoring/topic_score_config.rs:209`).

The misclassification is reachable, not theoretical. Role-8 messages are valid up to `(1 + MIN_SEED_LOOKAHEAD)` epochs before `proposal_slot` plus 2 slots after, while the gossipsub duplicate cache only absorbs byte-identical copies for one epoch (`behaviour.rs:108`). Honest post-cache relays therefore reach the validator and get REJECTed. Our own emissions are affected too: the sender self-validates outbound messages through the shared `Validator` (`anchor/message_sender/src/network.rs:143`), committing the root at send time, so a peer relaying our own message back after the cache window is penalized for it.

## Suggested approach

Smallest-surface design, behaviorally equivalent to go-ssv's two tiers for this role:

- Add `received_from: Option<PeerId>` to `Validator::validate` (`lib.rs:422`) and `ValidationContext`. The receiver passes `Some(propagation_source)` (`manager.rs:83`); sender self-validation passes `None` (`message_sender/src/network.rs:143`). These are the only two production callers.
- Replace `seen_preferences: HashSet<Hash256>` with `HashMap<Hash256, Option<PeerId>>`, mapping each accepted root to its first deliverer (`None` = own emission).
- Classify in `update_for_partial_signature` (`duty_state.rs:100`), in order:
  1. root present and first deliverer equals `received_from` (both `Some`): `DuplicatedMessage`, Reject
  2. root present otherwise: new Ignore-class variant (e.g. `RelayedDuplicateMessage`)
  3. map at `MAX_PROPOSER_PREFERENCES_DISTINCT_ROOTS`: `TooManyDistinctSigningRoots`, Ignore
  4. otherwise insert and accept
- Membership before capacity, so a same-peer duplicate stays Reject after the cap. This subsumes the capacity-order review note on #1106.
- Keep the commit after RSA verification (`partial_signature.rs:61` verify, `:74` commit): invalid signatures and malformed packets must not populate the map. Inserting only on accept also bounds the map at 4 entries with no per-peer growth. go-ssv's `Peers` map has no count cap and lazily inserts entries for unverified traffic; do not copy that.
- Classification and mutation stay inside the existing `DashMap<MessageId, DutyState>` entry guard held across validation (`lib.rs:505`), so no new locking is needed.

Equivalence note: go-ssv records roots into per-peer and world state only at commit, commit only runs on accept, and a root is accepted at most once, so "peer P's set contains root R" is identical to "P first delivered R". The collapsed map preserves observable classification while staying bounded.

## Acceptance criteria

- Peer A sends root A: accepted.
- Peer A repeats root A: `DuplicatedMessage`, Reject.
- Peer B relays root A: Ignore, no peer penalty.
- Distinct roots accepted up to 4 per (`MessageId`, operator, `proposal_slot`); a further distinct root is `TooManyDistinctSigningRoots`, Ignore.
- Same-peer repeat after the cap is still Reject; a different-peer relay after the cap is still Ignore.
- Invalid RSA signatures and semantically invalid packets do not populate the map: a subsequent honest send of the same root is accepted.
- A root committed via sender self-validation and later relayed by any peer is Ignore, never Reject.
- Per-peer information lives on `SignerState` and shares the role-8 ring lifecycle (`lib.rs:543-549`, 128 slots on mainnet); no separate expiry mechanism.
- Existing message-validation behavior for all other roles is unchanged.

## Tests

- Extend the role-8 dedup tests (`partial_signature.rs:2520-2680`) with two deterministic `PeerId` values covering the matrix above, including cap-boundary orderings, the double-relay case (Ignore both times), and the poisoning-resistance cases.
- Full-path coverage per the reported defect: exercise `NetworkMessageReceiver::receive` through `Validator::validate` to `Outcome.action` with peers A, A, B asserting Accept, Reject, Ignore. `message_receiver` has no tests today; note `Outcome` is emitted before the not-interested share check (`manager.rs:93-97`), which keeps the harness light.
- Regression: existing `message_validator` suites pass unchanged.

## Notes

- File and line references are at the #1106 head (`7ec34fb4`); this issue stacks on #1106.
- The duty-limit classification mismatch (Reject vs Ignore) is tracked separately in #1118 and stays out of scope here.
- go-ssv applies peer/world two-tier handling to all message types, not only this role. Generalizing Anchor's other roles is a separate discussion and out of scope.
- ssv-spec contains no message-validation state for this rule (p2p validation is outside ssv-spec's scope); the SIP text and go-ssv are the alignment targets.

petarjuki7:

- The `Role::ProposerPreferences` (8) duplicate-root check was peer-blind: every repeat of a seen signing root returned `DuplicatedMessage` -> Reject, applying the gossipsub P4 invalid-message penalty to whichever peer delivered it. Honest relays of a globally-seen root — and peers relaying our own emission back after the duplicate-cache window — were penalized for it.
- Reachable, not theoretical: role-8 messages stay valid for roughly `(1 + MIN_SEED_LOOKAHEAD)` epochs before the proposal slot (plus a couple of slots after), while the gossipsub duplicate cache only absorbs byte-identical copies for one epoch. Post-cache honest relays reach the validator.
- Links: closes #1131; SIP-94 §7; aligned with go-ssv's per-peer/world two-tier handling for this role.

## Change Overview

- Thread the libp2p propagation peer into message validation: the receiver passes the delivering peer, and the sender's own self-validation passes none.
- Key the per-slot dedup on root -> first deliverer instead of a peer-blind set. A repeat from the same peer is `DuplicatedMessage` (Reject); a relay from a different peer, or of our own emission, is the new Ignore-class `RelayedDuplicateMessage`.
- Membership is checked before capacity, so a same-peer repeat stays Reject even once the distinct-root cap is full. Only accepted messages populate the map, so it stays bounded with no per-peer growth.
- Intentionally unchanged: all other roles, the distinct-root cap, and the commit-after-verify ordering (so invalid signatures and packets never populate the map).
- Two commits: the production fix, then the test coverage.

- Blast radius limited to role-8 duplicate classification and the plumbing of one optional parameter through the partial-signature validation path.
- Trade-off: one peer is stored per accepted root (bounded by the distinct-root cap), a negligible addition.
- The two-tier matrix (accept / same-peer Reject / relay Ignore, cap boundaries, poisoning resistance, own-emission relay) is covered by unit tests; existing role-8 dedup tests are preserved.

## Validation

- Unit tests cover every acceptance criterion at the validator level.
- `make cargo-fmt`, `make lint`, and `cargo test -p message_validator` pass.
- Deferred: the full-path `message_receiver` integration test (`receive` -> `Outcome`). That crate has no test scaffolding, and driving `receive` end to end needs heavy fixtures (processor loop, a fully populated `NetworkState`, etc.). The accept/reject/ignore matrix — including the A, A, B sequence that test would exercise — is fully covered at the validator level. Tracked in #1238.

## Rollback

- Revert the two commits. No config, data, schema, or operational impact.

## Blockers / Dependencies

- Stacks on the role-8 arm and anti-spam caps already merged for this milestone.

### DIFF
--- a/anchor/message_receiver/src/manager.rs
+++ b/anchor/message_receiver/src/manager.rs
@@ -80,7 +80,7 @@ impl<E: types::EthSpec, S: SlotClock + 'static, D: DutiesProvider> MessageReceiv
                 let span = debug_span!("message_receiver", msg=%message_id);
                 let _enter = span.enter();
 
-                let result = receiver.validator.validate(&message.data, &topic_context);
+                let result = receiver.validator.validate(&message.data, &topic_context, Some(propagation_source));
 
                 let mut action = MessageAcceptance::from(&result);
 
--- a/anchor/message_validator/src/duty_state.rs
+++ b/anchor/message_validator/src/duty_state.rs
@@ -3,6 +3,7 @@ use std::{
     collections::{HashMap, HashSet},
 };
 
+use libp2p::PeerId;
 use ssv_types::{
     CommitteeId, Epoch, OperatorId, Slot,
     consensus::{QbftMessage, QbftMessageType},
@@ -102,6 +103,7 @@ impl DutyState {
         partial_signature_messages: &PartialSignatureMessages,
         signer: &OperatorId,
         slots_per_epoch: u64,
+        received_from: Option<PeerId>,
     ) -> Result<(), ValidationFailure> {
         let operator_state = self.get_or_create_operator(signer);
         let message_slot = partial_signature_messages.slot;
@@ -132,12 +134,16 @@ impl DutyState {
                 .first()
                 .ok_or(ValidationFailure::NoPartialSignatureMessages)?
                 .signing_root;
-            // Duplicate identity takes precedence over capacity: an already-seen root is a
-            // Reject-class DuplicatedMessage even once the distinct-root set is full, whereas only
-            // a NEW root beyond the cap is the Ignore-class TooManyDistinctSigningRoots (SIP-94
-            // §7).
-            if signer_state.seen_preferences.contains(&root) {
-                return Err(ValidationFailure::DuplicatedMessage {
+            // Same-peer repeat is spam (Reject); a relay of a seen root from another peer, or of
+            // our own emission, is not (Ignore). Membership is checked before capacity so a
+            // same-peer repeat stays Reject even when the set is full.
+            if let Some(first_deliverer) = signer_state.seen_preferences.get(&root) {
+                if received_from.is_some() && *first_deliverer == received_from {
+                    return Err(ValidationFailure::DuplicatedMessage {
+                        got: format!("proposer-preferences root {root:?}"),
+                    });
+                }
+                return Err(ValidationFailure::RelayedDuplicateMessage {
                     got: format!("proposer-preferences root {root:?}"),
                 });
             }
@@ -149,7 +155,7 @@ impl DutyState {
                     ),
                 });
             }
-            signer_state.seen_preferences.insert(root);
+            signer_state.seen_preferences.insert(root, received_from);
         }
 
         // Record the partial signature (only once)
@@ -332,9 +338,10 @@ pub(crate) struct SignerState {
     pub(crate) proposal_hash: Option<[u8; 32]>,
     /// A set of CommitteeIds indicating which committees have already been seen.
     seen_signers: HashSet<CommitteeId>,
-    /// Distinct ProposerPreferences signing roots already seen for this
-    /// (MessageId, operator, slot), used to reject exact-duplicate resends.
-    seen_preferences: HashSet<Hash256>,
+    /// Accepted ProposerPreferences signing roots for this (MessageId, operator, slot), each
+    /// mapped to its first deliverer (`None` = our own emission), to classify a repeat as a
+    /// same-peer duplicate (Reject) or a relay (Ignore).
+    seen_preferences: HashMap<Hash256, Option<PeerId>>,
 }
 
 impl SignerState {
@@ -346,7 +353,7 @@ impl SignerState {
             message_counts: MessageCounts::default(),
             proposal_hash: None,
             seen_signers: HashSet::new(),
-            seen_preferences: HashSet::new(),
+            seen_preferences: HashMap::new(),
         }
     }
 
--- a/anchor/message_validator/src/duty_state.rs
+++ b/anchor/message_validator/src/duty_state.rs
@@ -138,6 +138,8 @@ impl DutyState {
             // our own emission, is not (Ignore). Membership is checked before capacity so a
             // same-peer repeat stays Reject even when the set is full.
             if let Some(first_deliverer) = signer_state.seen_preferences.get(&root) {
+                // `is_some()` guard: our own emission is stored as `None`, so without it a
+                // re-emission would match `None == None` and wrongly Reject itself.
                 if received_from.is_some() && *first_deliverer == received_from {
                     return Err(ValidationFailure::DuplicatedMessage {
                         got: format!("proposer-preferences root {root:?}"),
--- a/anchor/message_validator/src/partial_signature.rs
+++ b/anchor/message_validator/src/partial_signature.rs
@@ -1,6 +1,7 @@
 use std::{collections::HashMap, sync::Arc};
 
 use duties_tracker::DutiesProvider;
+use libp2p::PeerId;
 use slot_clock::SlotClock;
 use ssv_types::{
     OperatorId,
@@ -23,6 +24,7 @@ pub(crate) fn validate_partial_signature_message(
     validation_context: ValidationContext<impl SlotClock>,
     duty_state: &mut DutyState,
     duty_provider: Arc<impl DutiesProvider>,
+    received_from: Option<PeerId>,
 ) -> Result<ValidatedSSVMessage, ValidationFailure> {
     // Decode message directly to PartialSignatureMessages
     let messages = match PartialSignatureMessages::from_ssz_bytes(
@@ -75,6 +77,7 @@ pub(crate) fn validate_partial_signature_message(
         &messages,
         signer,
         validation_context.slots_per_epoch,
+        received_from,
     )?;
 
     Ok(ValidatedSSVMessage::PartialSignatureMessages(messages))
--- a/anchor/message_validator/src/partial_signature.rs
+++ b/anchor/message_validator/src/partial_signature.rs
@@ -2373,7 +2373,8 @@ mod tests {
         proposal_slot: Slot,
         signing_root: Hash256,
     ) -> SignedSSVMessage {
-        // Reuse the well-formed constructor, then swap in a signature from the wrong key.
+        // Signs the whole packet with an unrelated key, so RSA verification fails against the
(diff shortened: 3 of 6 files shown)


anchor/message_receiver/src/manager.rs
```diff
@@ -80,7 +80,7 @@ impl<E: types::EthSpec, S: SlotClock + 'static, D: DutiesProvider> MessageReceiv
                 let span = debug_span!("message_receiver", msg=%message_id);
                 let _enter = span.enter();
 
-                let result = receiver.validator.validate(&message.data, &topic_context);
+                let result = receiver.validator.validate(&message.data, &topic_context, Some(propagation_source));
```
shane-moore, line 83: This argument is the behavioral half of the fix, and no test reaches it: the new matrix tests all drive the validator directly with their own `PeerId`. If this regresses to `None`, `received_from.is_some()` is false for every inbound message, every role-8 duplicate classifies as the Ignore-class relay, and the whole suite stays green.

The full `receive` harness sketched in #1131 looks heavier than the issue implied (processor loop, qbft manager, signature collector), so no objection to deferring that. But a minimal guard asserting the receive path supplies `Some(propagation_source)` would be enough to pin this line.

anchor/message_validator/src/partial_signature.rs
```diff
@@ -2334,6 +2362,67 @@ mod tests {
...
+    /// validator_index) is well-formed, so validation reaches — and fails at — the RSA check,
+    /// which sits AFTER semantics/duty logic but BEFORE `update_for_partial_signature`. Used to
+    /// prove a signature-invalid packet never poisons `seen_preferences`.
+    fn create_bad_signature_proposer_preferences_message(
+        signer_id: OperatorId,
+        wrong_key: &Rsa<Private>,
+        proposal_slot: Slot,
+        signing_root: Hash256,
+    ) -> SignedSSVMessage {
+        // Reuse the well-formed constructor, then swap in a signature from the wrong key.
```
shane-moore: There is no swap: `wrong_key` signs the packet from the start. Either fix the comment (e.g. "signs the whole packet with an unrelated key, so RSA verification fails against the registered key") or drop this same-arity pass-through and call `create_signed_proposer_preferences_message` directly at the call site, keeping the doc comment there.

anchor/message_validator/src/partial_signature.rs
```diff
@@ -3124,6 +3265,519 @@ mod tests {
...
+            create_proposer_preferences_context(&signed, committee_info, map, proposal_slot);
+        validate_partial_signature_message(
+            context,
+            duty_state,
+            Arc::new(MockDutiesProvider::default()),
+            received_from,
+        )
+    }
+
+    /// Root-byte helper: distinct low byte -> distinct signing root, all other bytes zero.
```
shane-moore: `[tag; 32]` fills all 32 bytes with `tag`; no byte is zero and it is not just the low byte. Suggest: "Fills all 32 bytes with `tag`, so distinct tags give distinct signing roots."

anchor/message_validator/src/duty_state.rs
```diff
@@ -132,12 +134,16 @@ impl DutyState {
...
-            // Reject-class DuplicatedMessage even once the distinct-root set is full, whereas only
-            // a NEW root beyond the cap is the Ignore-class TooManyDistinctSigningRoots (SIP-94
-            // §7).
-            if signer_state.seen_preferences.contains(&root) {
-                return Err(ValidationFailure::DuplicatedMessage {
+            // Same-peer repeat is spam (Reject); a relay of a seen root from another peer, or of
+            // our own emission, is not (Ignore). Membership is checked before capacity so a
+            // same-peer repeat stays Reject even when the set is full.
+            if let Some(first_deliverer) = signer_state.seen_preferences.get(&root) {
+                if received_from.is_some() && *first_deliverer == received_from {
```
shane-moore, line 143: nit: worth a line in the comment above on why the `is_some()` guard exists, since it reads as redundant next to the equality check: our own emission is stored as `None`, so without it a re-emission would compare `None == None` and self-classify as Reject. Right now that reasoning only lives in the criterion-7 test comment.

anchor/message_validator/src/partial_signature.rs
shane-moore: Fixed in fe3b0c6.

anchor/message_validator/src/partial_signature.rs
shane-moore: Fixed in fe3b0c6.

anchor/message_validator/src/duty_state.rs
shane-moore: Fixed in fe3b0c6.

anchor/message_receiver/src/manager.rs
petarjuki7: True, that line is not tested right now, which is an important entry point. I tried a minimal guard but a message can't reach the dedup without passing RSA verify against the key seeded in NetworkState, and the fixtures that build that matched pair are stuck inside message_validator's test module, out of reach from message_receiver. Needs a test-utils feature there first, so I opened #1238 to track it / to get this PR merged and not clutter it too much.

shane-moore: agreed!

shane-moore (approved):

PR #1208, opened 2026-08-03
fix: batch pre-Boole post-consensus sync contributions

shane-moore:

- Before Boole, QBFT can decide multiple sync contributions for one validator, but Anchor publishes each post-consensus callback as a single-root type-0 envelope. go-ssv expects the complete decided root multiset and rejects these envelopes with `wrong expected roots count`.
- A prior mixed-client reproduction recorded 626 type-0 wrong-root-count rejections across 210 multi-root duties.
- #1201 fixed the analogous type-3 producer path and provides the batch reused here.

## Change Overview (Required)

- Prepare a canonical descriptor from the complete decided `Contributions`, sorted by subnet and root with identical entries retained as multiplicity.
- Reuse the eager `Pending`/`Admitted` single-validator batch for post-consensus messages, isolated from type 3 by phase and keyed by exact signing root.
- Keep the single-root path unchanged. Boole, wire types, inbound validation, and Lighthouse APIs are unchanged.
- Review `validator_store` preparation first, then `signature_collector` admission and retry behavior.

- The main risk is concurrent sibling callbacks publishing incomplete or duplicate envelopes. Focused tests cover admission, retry, injection, cleanup, stale work, impostor mode, and isolation boundaries.

## Validation (Required)

- 30-minute mixed run with three Anchor nodes and one go-ssv node: 1,035 type-0 duties, including 195 multi-root duties and exact two-, three-, and four-root envelopes; zero root-count, root, signature, descriptor, quorum, restart, or OOM failures; type 3 remained unchanged.
- After #1201 merged, this branch was replayed as one commit on current `unstable`. The only conflicts were in `validator_store` test scaffolding added by #1213, so the harness constructor now takes both a proposer delay and a mock consensus decider. `git range-diff` confirms the production diff is unchanged from the validated candidate.
- On the replayed commit: `git diff --check`, `make cargo-fmt-check`, `make lint`, and release tests covering 56 `validator_store`, 33 `signature_collector`, and 57 `message_validator` cases.

- Revert this commit. There are no config, data, API, or wire-format migrations.

### DIFF
--- a/anchor/signature_collector/src/lib.rs
+++ b/anchor/signature_collector/src/lib.rs
@@ -80,15 +80,32 @@ struct CommitteePartialSignatureBatch {
     for_slot: Slot,
 }
 
-/// One unique pre-Boole sync selection proof root and its original position multiplicity.
+/// One unique pre-Boole sync committee subnet and signing root pair.
 #[derive(Debug, Clone, PartialEq, Eq)]
-pub struct SyncSelectionProofDescriptor {
-    /// Unique sync committee subnet represented by this entry.
+pub struct SyncCommitteeBatchEntry {
+    /// Sync committee subnet represented by this entry.
     pub subnet_id: SyncSubnetId,
-    /// Ethereum signing root for the subnet's selection proof.
+    /// Ethereum signing root for this partial signature.
     pub signing_root: Hash256,
-    /// Number of original sync committee positions that map to this subnet and root.
-    pub position_count: usize,
+    /// Raw committee-position count for type 3, or repeated identical decided-entry count for
+    /// type 0.
+    pub multiplicity: usize,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+enum SingleValidatorBatchPhase {
+    ContributionProofs,
+    PostConsensus,
+}
+
+impl SingleValidatorBatchPhase {
+    fn from_kind(kind: PartialSignatureKind) -> Option<Self> {
+        match kind {
+            PartialSignatureKind::ContributionProofs => Some(Self::ContributionProofs),
+            PartialSignatureKind::PostConsensus => Some(Self::PostConsensus),
+            _ => None,
+        }
+    }
 }
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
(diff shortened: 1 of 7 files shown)


shane-moore:
@claude, please take a look

shane-moore:
@claude Verified your observations against the branch. Two corrections and one addition, no code changes from this review.

1. Observation 1's conclusion is right but the citation is wrong. `message_sender/src/network.rs:66` is inside the closure that runs later on the processor. The non-blocking primitive is `processor/src/senders.rs:66-68`, where `send_work_item` calls `tx.try_send`, reached from `network.rs:68-69`. So under the record lock `sign_and_send` only enqueues, and a full queue returns `Err(Error::Processor)`, which leaves the record `Pending` for a sibling callback to retry. The impostor sender performs no I/O at all.

2. The identical-descriptor property across siblings is structural rather than incidental. Sync contribution consensus runs one instance per validator and slot, `ProposerInstanceId { validator, duty: SyncCommitteeAggregator, instance_height: slot }`, so every subnet callback decodes the same decided value before building its descriptor.

3. Confirming the root-only descriptor guard in code, not just by argument: type 0 roots cover `contribution.subcommittee_index` through `ContributionAndProof`, and type 3 roots cover `SyncAggregatorSelectionData { slot, subcommittee_index }` at `validator_store/src/lib.rs:557-565`. Root identity is therefore injective per subnet in both phases.

shane-moore (commented):
This review was performed by gpt-5.6-sol with max reasoning effort.

No blocking correctness or safety findings at 21746a02.

The core change holds up:

- ValidatorStore derives a deterministic descriptor from the complete QBFT-decided contribution multiset.
- Phase, slot, and validator-public-key keying isolates type 0 from type 3.
- Exact-root identity plus the Pending and Admitted state preserves single-winner admission, sibling retry, and local injection.
- Focused tests cover concurrency, retry, cleanup, phase isolation, and multiplicity. All recorded CI checks at this head are green.

I investigated one minor hardening edge: K=1 sync committee duties take SingleValidator and therefore bypass the batch-only stale-work guards. Detached work delayed past collector cleanup can recreate short-lived state and attempt a no-longer-useful send. I do not recommend changing this PR for it. K=1 routing is deliberate and restores the historical path, staleness begins only after two slot advances, outgoing messages are self-validated, and the second sender queue hop is also non-expiring. A complete solution belongs to generic processor work expiry across both queue hops under #254, rather than routing K=1 back through the batch path.

One landing caveat: unstable is now at 120d4300 while this PR has merge-base 82bac14. The recorded checks are green, but they do not demonstrate current-base merge CI.

Verdict: ready from a code-review perspective.

PR #1210, opened 2026-08-04
ci: deploy docs site on release publish instead of every unstable push

### LINKED ISSUE #1209: ci(docs): deploy docs site on release publish instead of every unstable push
## Goal

Publish anchor.sigmaprime.io only when a release is published, so the live docs never describe CLI flags or download URLs that no public binary provides.

## Context and motivation

`docs.yml` currently deploys to the live S3 bucket (with `--delete`) on every push to `unstable`. Operators see docs for unreleased flags weeks before a binary exists, and during each release window the site advertises versioned download URLs that 404 (the version bump merges to `unstable` before the release is published). Separately, the Linux example in `installation.mdx` omits the architecture from the tarball name, so that wget 404s permanently regardless of timing.

## Suggested approach

- Trigger `docs.yml` on `release: types: [published]` plus a bare `workflow_dispatch:` escape hatch (manual redeploys; primary use is dispatching `--ref stable` after a docs hotfix is cherry-picked to `stable`).
- Guard prereleases in the existing job-level `if` (`github.repository_owner == 'sigp' && !github.event.release.prerelease`).
- Replace the per-ref concurrency group with a constant one (`cancel-in-progress: false`) so a release deploy and a manual dispatch cannot run `s3 sync --delete` concurrently.
- Fix the arch-less Linux download example in `installation.mdx`.
- Note the new publication timing in `docs/README.md`.

## Acceptance criteria

- [ ] Pushes to `unstable` no longer deploy the docs site.
- [ ] Publishing a (non-pre) release deploys docs built from the release tag, with the released version in download URLs.
- [ ] Manual `workflow_dispatch` from a chosen ref redeploys the site; concurrent deploys serialize.
- [ ] The Linux install example downloads an artifact name that release assets actually use.

## Tests

CI trigger changes are not integration-testable pre-merge: `actionlint` the workflow, then verify the first release publish after merge fires the deploy and the live site shows the released version and working download URLs.

## Notes

- Rollout: the new trigger and the dispatch button only take effect once the file reaches `stable` (default branch) at the next release; until then the site freezes at its current content (status quo). Merging the change to `unstable` does not fire a final unstable deploy (push events read the workflow file at the pushed commit).
- `push: branches: [stable]` was the runner-up trigger; rejected because the release pipeline can fail after the ff-merge, leaving 404 download URLs live indefinitely. `release: published` fails safe (old site stays up). Also considered and rejected: a staging bucket for unstable docs (new infra; file separately if wanted), building from the latest tag on unstable pushes (redundant deploys), "unreleased" banners (doesn't fix 404s).
- Adversarially reviewed (Codex, 2 rounds, converged): concurrency and prerelease hardening above came from that review; "verify the published release is the latest before deploying" was considered and rejected as complexity without a current problem (a stale-draft publish is recoverable via dispatch).
- Deliberate deviation from sigp/lighthouse, whose book deploys from `unstable`: Anchor's docs embed versioned download URLs and a generated CLI reference, which makes unstable-tracking docs actively misleading.

shane-moore:

- The docs workflow deploys anchor.sigmaprime.io on every push to `unstable`, so docs for unreleased CLI flags go live weeks before a binary exists, and during each release window the site advertises versioned download URLs that 404 (the version bump merges to `unstable` before the release is published).
- Operators are currently seeing docs for flags their binary rejects.
- Separately, the Linux install example uses an arch-less tarball name that no release asset matches, and the version-sync regex mangles arch-qualified names at build time.
- Closes #1209 (alternatives considered and rollout details are recorded there).

## Change Overview (Required)

- `docs.yml` now triggers on `release: published` (prerelease-guarded) plus a bare `workflow_dispatch` for manual redeploys; deploys serialize under a constant concurrency group with `cancel-in-progress: false` so an in-flight `s3 sync --delete` is never killed halfway.
- `sync-version.js` filename regexes drop the greedy optional suffix group that swallowed the architecture component; the installation example is now arch-qualified; the docs README documents the new publication timing.
- Reading order: `docs.yml` (the behavior change), then `sync-version.js` + `installation.mdx` (content fix), then `README.md`.
- Intentionally unchanged: the build itself, linkcheck and CLI-reference checks on PRs/unstable, and the S3 sync mechanics.

- The new trigger and the manual-dispatch button only take effect once this file reaches `stable` (default branch) at the next release; until then the live site freezes at its current content (status quo, self-corrects at next release publish).
- Merging this stops unstable-push deploys immediately (push events read the workflow file at the pushed commit; no final unstable deploy fires).
- Docs-only fixes between releases no longer publish automatically; path is cherry-pick to `stable` and `gh workflow run docs.yml --ref stable`.
- Publishing a stale draft release would redeploy that tag's docs; accepted as low-likelihood and recoverable via a manual redeploy from `stable`.

## Validation (Required)

- `actionlint` clean on the modified workflow; `node --check` clean on `sync-version.js`.
- Regex fix verified with a standalone node repro: arch-qualified names now survive version rewriting (previously `anchor-v1.1.0-x86_64-...` became `anchor-vX.Y.Z-...`).
- End-to-end trigger behavior is only observable at the next release publish; first post-release deploy should be checked against the live site.

- Revert the commit; the previous push-to-unstable deploy behavior returns on the next `unstable` push. No data or config impact; the S3 bucket is fully rewritten by any subsequent deploy.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,17 +1,19 @@
 name: docs
 
 on:
-  push:
-    branches:
-      - unstable
+  release:
+    types: [published]
+  workflow_dispatch:
 
+# All runs deploy to the same S3 bucket with --delete, so they must serialize,
+# and an in-flight sync must never be cancelled halfway through.
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
+  group: docs-deploy
+  cancel-in-progress: false
 
 jobs:
   build-and-upload-to-s3:
-    if: github.repository_owner == 'sigp'
+    if: github.repository_owner == 'sigp' && !github.event.release.prerelease
     runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@v6
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -25,17 +25,23 @@ jobs:
           cache: 'npm'
           cache-dependency-path: docs/package-lock.json
 
+      # Step timeouts on the network-bound steps keep a hang from holding the
+      # docs-deploy concurrency group (default job timeout is 6h). Deliberately
+      # no job-level timeout: it could cancel the S3 sync mid-run.
       - name: Install dependencies
         run: npm ci
         working-directory: docs
+        timeout-minutes: 10
 
       - name: Install Playwright browsers
         run: npx playwright install
         working-directory: docs
+        timeout-minutes: 10
 
       - name: Sync version and GitHub stars
         run: npm run sync-version
         working-directory: docs
+        timeout-minutes: 10
 
       - name: Build vocs
         run: npm run build
--- a/docs/README.md
+++ b/docs/README.md
@@ -65,6 +65,14 @@ To preview the built site:
 npm run preview
 ```
 
+## Deployment
+
+The live site ([anchor.sigmaprime.io](https://anchor.sigmaprime.io)) is deployed by the
+`docs` GitHub workflow when a release is published, so the published docs always match the
+latest released binary. Docs merged to `unstable` do not go live until the next release.
+For an out-of-band redeploy (e.g. a docs fix cherry-picked to `stable`), run the workflow
+manually: `gh workflow run docs.yml --ref stable`.
+
 ## Contributing
 
 ### Adding New Pages
--- a/docs/docs/pages/installation.mdx
+++ b/docs/docs/pages/installation.mdx
@@ -35,8 +35,8 @@ wget https://github.com/sigp/anchor/releases/download/<version>/anchor-<version>
 tar -xvf anchor-<version>-<platform>.tar.gz
 
 # Specific version example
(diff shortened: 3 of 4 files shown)


shane-moore:
@claude, please take a look

.github/workflows/docs.yml
jnhsigmap, line 31: low priority but I noticed no timeout on this job. Timeout is dangerous potentially with `aws s3 sync --delete` as we'd rather not corrupt the bucket in the case where the job gets axed as that is running. we could though add `timeout-minutes: 10` or similar to steps that have external fetches like `Install dependencies` and `Install Playwright browsers` independently to stop hangs blocking our other deploys.

jnhsigmap (dismissed):
all good otherwise

.github/workflows/docs.yml
shane-moore: Good call. Added `timeout-minutes: 10` to the three network-bound steps (npm ci, Playwright install, and the GitHub stats fetch in sync-version) in b2e6b9b. Left the job-level timeout off for the reason you gave: it could cancel the S3 sync mid-run. A hang now fails before the sync starts instead of holding the concurrency group for the 6h default.

jnhsigmap (approved):

Issue #1212, opened 2026-08-04
feat(client): fork-gated post-Gloas proposer delay (go-ssv ProposerDelayEPBS parity)


shane-moore:
## Problem

`--proposer-delay-ms` (#1207) is a single, fork-agnostic knob. go-ssv splits the same setting in two, because the post-Gloas proposal deadline is tighter and the safe ranges differ.

From their runner (`protocol/v2/ssv/runner/proposer.go:679-686`, go-ssv `90df7fa0a`):

> `proposerDelayForSlot` returns the fork-appropriate proposer delay: `proposerDelayEPBS` from the Gloas fork on, `proposerDelay` before it. They are separate knobs because ePBS retimes the proposal deadline (slot quarters), so their safe ranges differ.

Without this, an operator migrating from an SSV node loses a setting at Gloas, and any value tuned pre-Gloas silently continues to apply under a tighter deadline.

## What go-ssv does

| | `ProposerDelay` | `ProposerDelayEPBS` |
| --- | --- | --- |
| Applies | before Gloas | from Gloas on |
| Default | 0 | 0 |
| Cap | 1000ms, overridable via `AllowDangerousProposerDelay` | 1000ms, **refused unconditionally** |

Config field at `cli/operator/node.go:113`; cap enforcement at `cli/operator/node.go:889-897`, whose comment reads "the post-ePBS proposal deadline is tighter, so the cap is enforced unconditionally".

## Acceptance criteria

- [ ] `--proposer-delay-epbs-ms`, default `0` (opt-in), independent of `--proposer-delay-ms`.
- [ ] The value applied is selected by the fork at the duty slot: Gloas or later uses the ePBS value, earlier slots use the existing one.
- [ ] Hard cap of 1000ms on the ePBS value, refused at startup regardless of `--allow-dangerous-proposer-delay`. That flag must **not** gate it.
- [ ] Tests: fork selection either side of the Gloas boundary slot, unconditional cap refusal, and both values defaulting to 0.
- [ ] `docs/docs/pages/mev_configuration.mdx` gains the second value in its limits table and in the "Coming from an SSV node" mapping.

## Notes

[...]

Issue #1179, opened 2026-07-27
feat(validator_store): early RANDAO pre-sign service behind --early-randao


shane-moore:
**Blocked by:** #1177, #1184, #933, and #1192

**Anchor rollout companion:** #1178

## Goal

Add a flag-gated producer service that starts RANDAO partial-signature collection at the recommended time before a locally known eligible proposal slot, so Lighthouse's unchanged in-slot `randao_reveal` call can reuse an already completed reconstruction.

Early emission is optional under [SIP-101 at 8279853](https://github.com/ssvlabs/SIPs/blob/82798537fea271d1a31d02fdc000d87b37fa3d98/sips/early_randao_preconsensus.md#L51-L64). An implementation that never emits early remains conformant. Anchor chooses to implement the recommended `S - 1` policy and keep the existing in-slot path as the fallback.

## Protocol classification

For a locally known proposer duty at eligible target slot `S`:

- The producer **MAY** emit at any time in `[slot_start(S - 2), slot_start(S))`.
- If it emits early, it **MUST NOT** emit before `slot_start(S - 2)` by its own clock.
- If it elects to emit early and knows the duty by the recommended time, it **SHOULD** make its first local publication attempt at exact `slot_start(S - 1)`.
- It **SHOULD** still invoke the existing in-slot emission path at `S` regardless of its locally recorded early outcome.
- It **MAY** serve an already reconstructed reveal to the in-slot consumer without waiting for repeated local-share signing, envelope signing, or publication.
- It **MAY** emit immediately for a duty first discovered after the recommended time and **SHOULD** process multiple eligible duties in ascending slot order.

The two-slot value is the maximum receiver window, not the normal Anchor producer schedule. The 1000ms clock tolerance is receiver-side headroom and does not shift the recommended producer time. The SIP does not coordinate a second pre-slot attempt. Anchor adopts one normal attempt at `S - 1`, followed by the existing in-slot path at `S`.

Different producer schedules remain interoperable, but the common policies used by a cluster determine when a valid threshold becomes available and when the reveal becomes knowable.

Eligibility uses the stamped target slot: `S >= EARLY_RANDAO_LEAD` and `epoch(S) >= GLOAS_FORK_EPOCH`. For the first non-genesis Gloas slot `F`, the legal window begins at `slot_start(F - 2)`, while Anchor's recommended attempt is at `slot_start(F - 1)`. The service must not add a current-epoch, current-fork, or SSV-fork gate.

## Dependencies and rollout

[...]

PR #1213, opened 2026-08-04
feat(validator_store): add configurable proposer delay for MEV bids

### LINKED ISSUE #1207: feat(validator_store): configurable proposer delay before beacon block production
## Goal

Add an operator-configurable delay that holds the beacon block request until a target offset into the slot, so MEV-boost relays have longer to return a better bid. Behavioural parity with go-ssv's `ProposerDelay`.

## Context and motivation

A large operator reports this as the single blocker preventing them adopting Anchor. It is a per-operator knob with no wire or serialized surface.

go-ssv v2.4.3 (`31010e70b1`) semantics, verified at source:

- The wait sits in `ProcessPreConsensus` (`protocol/v2/ssv/runner/proposer.go:150-161`), after RANDAO quorum and reconstruction, before `GetBeaconBlock`.
- `remainingProposerDelay` (`:521-528`) returns `max(0, slot_start + delay - now)`. It is an absolute **floor from slot start**, not added latency. If pre-consensus already ran past the target there is no wait.
- Not re-entrant, guarded by `hasQuorum`.
- Default 0; refuses above 1s without `AllowDangerousProposerDelay`.
- Read by nothing else in go-ssv. No compensating timeout or QBFT timer adjustment anywhere.

## Suggested approach

Anchor's only hook between duty start and block fetch is `randao_reveal` (`anchor/validator_store/src/lib.rs:2241`); `Arc<AnchorValidatorStore>` goes straight to `BlockServiceBuilder` (`anchor/client/src/lib.rs:720-732`) with no wrapper. This placement is sound rather than incidental: `get_validator_blocks_v3_ssz`/`v4_ssz` take the reveal as a **required parameter** (`common/eth2/src/lib.rs:2481-2493`), so Lighthouse structurally cannot request the block first.

- Add a decision function returning `Disabled | TargetPassed | Waited(Duration) | ClockUnavailable`, not `Option<Duration>`, so metrics distinguish a no-op from missing instrumentation.
- Consume the `clock_slot` already captured at `:2263` before `collect_signature`. Never re-sample.
- `checked_add` on `slot_start + delay`, plus a hard upper bound refused regardless of the acknowledgement flag.
- Fail open if `start_of(slot)` returns `None`. This must never cause a missed proposal through its own error path.
- Call it at the **single success return point** of `randao_reveal`, skipped on the error path.
- `--proposer-delay-ms <u64>` (default 0) and `--allow-dangerous-proposer-delay` on `PayloadBuildingOptions` (`anchor/cli/src/cli.rs:466`). Gate in `config.rs::from_cli`, copying the acknowledgement-flag pattern at `:244-252`.
- 1s is a conservative policy threshold for operator parity, explicitly **not** a claimed safe ceiling. Anchor's round model cannot yield one: `max_round = 6` at `QUICK_TIMEOUT = 2s` already exceeds the publication window.

## Acceptance criteria

- Behaviour identical to go-ssv: absolute floor from slot start, no wait when the target has passed, at most once per proposal.
- Default 0 is a strict no-op with no added latency and no behaviour change.
- Values above the acknowledgement threshold are refused at startup unless the flag is set; values above the hard bound are refused either way.
- A checkpoint **and** a pre-wait span field are recorded before the wait, so `RANDAO_REVEAL_COMPLETED` stops attributing policy sleep to RANDAO work.
- A decision histogram records no-wait outcomes, not only waits.
- A histogram of successful RANDAO completion offset is recorded immediately after `collect_signature`, before any wait. This data does not exist today, because success emits only `trace!` (`:2298`) while `logfile_debug_level` defaults to `DEBUG`. It is what shows an operator whether the floor is biting at their configured value.
- Documentation states the recommended starting value of 300ms, matching go-ssv, and states plainly that the setting is a minimum offset from slot start rather than added latency.
- Regenerated `docs/docs/generated/cli-node-options.mdx`, or `cli-reference-check` fails.

## Tests

Zero delay; normal delay; target already passed; clock unavailable fails open; overflow; exact threshold boundaries. Config gate accepted, rejected, and acknowledged, plus hard bound refused either way. Delay applied only on the success path. Cancellation drops the wait cleanly. In-slot invocation contract regression test. Mixed-delay cluster: operators with differing delays initialise proposer QBFT at different times and followers buffer up to `MESSAGE_BUFFER_LIMIT = 100` (`anchor/qbft_manager/src/instance.rs:28`).

## Notes

- **Follow go-ssv's guidance: default 0, recommended starting value 300ms.** The absolute-floor semantics make the configured value portable across clients: the block request lands at `max(randao_completion, slot_start + delay)`, so the same value produces the same in-slot request time on go-ssv and Anchor. What differs between clients is the downstream budget after that point, not the delay.
- **Do not present go-ssv's ~1.2s as Anchor's ceiling.** That figure is derived from go-ssv's own latency budget (`qbftTime = 350ms`, `blockSubmissionTime = 1000ms`, `miscellaneousTime = 150ms`). Anchor's equivalents are unmeasured. Keep the 1s acknowledgement gate for operator parity, and validate our own headroom before endorsing values above it.
- Note that go-ssv's `docs/MEV_CONSIDERATIONS.md` budget formula treats `RANDAOTime + ProposerDelay` as additive, while `remainingProposerDelay` implements `max()`. RANDAO time is absorbed by the delay, not added to it, so their ~1.2s is slightly conservative rather than wrong. It errs safe.
- The completion-offset histogram is tuning support and headroom validation, not a gate on shipping. It is what lets an operator see whether the floor is actually biting at their configured value, and what we would use to justify raising guidance above 300ms.
- **Under Gloas the wait still executes and still consumes proposal budget**; only the builder-bid benefit is absent while Lighthouse always builds locally (`block_service.rs:624`). Warrants a startup or fork-transition warning, not documentation alone.
- Early RANDAO (#1179) does not break this seam: it adds a separate `early_randao_reveal(pubkey, proposal_slot)` and preserves Lighthouse's unchanged in-slot call. But it will restructure `randao_reveal`'s internals, hence the single-return-point requirement, and its eligibility is Gloas-gated so the synergy is Gloas-only.
- Anchor has no config file or environment variable support, so operators set a CLI flag rather than porting `ProposerDelay: 300ms` from YAML. Broader gap, tracked separately.
- **Alternatives rejected:** upstreaming a `proposal_request_offset` to Lighthouse's `BlockService` (architecturally cleaner and Lighthouse owns the true request boundary, but there is no precedent there for a knob trading missed-proposal risk for MEV, and the correct value is a function of SSV QBFT latency); sleeping in `sign_block` (too late, block already fetched); a relative sleep instead of a floor (removes late-duty recovery); sender-side buffering; a `humantime` dependency for `300ms` ergonomics.

shane-moore:

Anchor requests a block as soon as RANDAO pre-consensus finishes. Relays that hold out for a better bid have not answered yet, so operators leave MEV on the table

Closes #1207.

## Change Overview

`--proposer-delay-ms` is an **absolute floor from the start of the slot, not added latency**. The request lands at `max(randao_completion, slot_start + delay)`, so a duty whose pre-consensus already ran past the target is never delayed further. This matches go-ssv exactly, so an operator's existing value carries across clients.

The wait sits at the success return point of `randao_reveal`. That is the last point Anchor owns before block production: the reveal is a required parameter of the block request, so Lighthouse structurally cannot ask earlier.

Reading order: `cli.rs` for the two flags, `client/src/config.rs` for the startup gate, then `validator_store/src/lib.rs` for the decision and the wait. Everything else is plumbing, metrics, and docs.

**Unchanged:** default is `0`, so behaviour is identical for anyone who does not set the flag. No change to the signing path, slashing protection, QBFT, or publication. Not fork-gated.

The feature deliberately spends proposal headroom, so a large value can cost a proposal. That is the trade the operator opts into, and the docs say so rather than presenting it as free.

- Default `0`; above `1000` needs `--allow-dangerous-proposer-delay` (matching the SSV node's threshold); above `4000` is refused as a typo guard, not as a safe ceiling.
- Fails open: if the slot clock cannot say how far into the slot we are, the delay is skipped.
- Two histograms (`anchor_randao_reveal_completion_offset_seconds`, `anchor_proposer_delay_applied_seconds`) let an operator see whether the delay bites at all. The applied wait is measured after the sleep, not before, so it reports the real wait.

Known and deliberate: the floor is approximate under a backward clock adjustment, since elapsed time is sampled once and the wait runs on a monotonic timer. A re-check loop could turn a clock correction into extra proposal-critical delay, so failing open is preferred.

Post-Gloas this keeps working but buys something different, and go-ssv splits it into a second fork-gated knob. Tracked separately in #1212; out of scope here.

## Validation

- 13 new tests (8 policy, 5 config), all pure, no runtime or store needed.
- `make test` full suite green; `make lint`, `make cargo-fmt-check`, `make mdlint`, `make cli-reference-check` all pass.
- Adversarial review by Codex (gpt-5.6-sol, two rounds) against the diff, plus four cleanup passes. Findings applied include the applied-wait metric recording the planned rather than actual wait, and a docs claim that only the consensus leader requests a block (every operator does).

## Rollback

Revert the commit. The default is `0`, so no operator is affected unless they opted in, and there is no config, data, or on-disk migration.


shane-moore:
@claude, please take a look

shane-moore:
@claude Added the paused-clock coverage in 34faefdc.

`await_proposer_delay` only ever read `self.proposer_delay`, so it is now a free function like its two neighbours. That drops the store dependency and lets the tests run on a virtual clock, where `tokio::time::Instant` observes the same clock `sleep` advances. One test asserts a 300ms floor with 100ms already elapsed advances the timer by exactly 200ms; the other that all three no-wait cases (overdue, no clock, disabled) return without advancing it at all.

Also dropped an intra-doc link from a public item to a private one, which rustdoc warns about and CI does not check.

shane-moore:
@claude Leaving the global metric mutation as-is: isolating it would cost more machinery than the risk warrants, and as you note it proves the metric path runs on the no-wait branches too. Noted for anyone who later asserts on those histograms. CI green on 34faefdc.

jnhsigmap:
I found 1 that might deserve a look. Feel free to discard this is its not needed. There is currently no test that calls `randao_reveal`
- If a refactor moves `await_proposer_delay` above `collect_signature` or onto the error path, this will go undetected by the test suite,
- This moves the wait above `collect_signature` and creates added latency by spending proposal headroom twice.

Things that protect us a little bit -
- This proposer delay feature defaults to `off`.
- Accidentally introducing this regression will hit operators who have opted in.

You also note that early randao reveal in #1179 will restructure the internals of `randao_reveal`. Maybe for piece of mind we address this now?

Pre-setup, `anchor/validator_store/src/testing/common.rs`:
- This requires a bit of a refactor to `ValidatorStoreTestHarness` and `CapturedSignatureCall`
- Wraps existing `new_with_fork` body in `new_with_options` and creates second call for this case (proposer delay) that adds a `Duration`.
- Timestamp on captured mock signature calls.
- Shared clock on `ValidatorStoreTestHarness`.

```diff
-use tokio::sync::watch;
+use tokio::{sync::watch, time::Instant};
```

```diff
 pub(super) const TEST_SLOT: u64 = 1;
-const SLOT_DURATION_SECS: u64 = 12;
+pub(super) const SLOT_DURATION_SECS: u64 = 12;
```

```diff
 pub(super) struct CapturedSignatureCall {
     pub(super) requester: SignatureRequester,
     pub(super) validator_pubkey: PublicKeyBytes,
     pub(super) signing_root: Hash256,
+    /// When the call was made.
+    pub(super) captured_at: Instant,
 }
```

```diff
         self.captured.lock().push(CapturedSignatureCall {
             requester,
             validator_pubkey: signing_data.validator_pubkey,
             signing_root: signing_data.root,
+            captured_at: Instant::now(),
         });
```

```diff
 pub(super) struct ValidatorStoreTestHarness {
     pub(super) validator_store:
         Arc<AnchorValidatorStore<ManualSlotClock, MainnetEthSpec, MockConsensusDecider>>,
     committee_setups: Vec<CommitteeSetup>,
     pub(super) captured_calls: CapturedCalls,
+    /// Shares `current_time` with the store's clone, so tests can reposition the clock.
+    pub(super) slot_clock: ManualSlotClock,
     pub(super) is_synced_tx: watch::Sender<bool>,
     _slashing_db_dir: TempDir,
     _exit_signal: async_channel::Sender<()>,
 }
```

```diff
     pub(super) fn new_with_fork(
         committee_setups: Vec<CommitteeSetup>,
         our_operator_id: OperatorId,
         active_fork: Fork,
+    ) -> Self {
+        // No proposer delay by default: most tests assert timing-free behaviour.
+        Self::new_with_options(
+            committee_setups,
+            our_operator_id,
+            active_fork,
+            Duration::ZERO,
+        )
+    }
+
+    pub(super) fn new_with_proposer_delay(
+        committee_setups: Vec<CommitteeSetup>,
+        our_operator_id: OperatorId,
+        proposer_delay: Duration,
+    ) -> Self {
+        Self::new_with_options(
+            committee_setups,
+            our_operator_id,
+            Fork::Boole,
+            proposer_delay,
+        )
+    }
+
+    fn new_with_options(
+        committee_setups: Vec<CommitteeSetup>,
+        our_operator_id: OperatorId,
+        active_fork: Fork,
+        proposer_delay: Duration,
     ) -> Self {
```

```diff
             30_000_000,
             None,
             false,
-            Duration::ZERO, // no proposer delay: tests assert timing-free behaviour
+            proposer_delay,
             false,
             is_synced_rx,
             executor,
```

```diff
         Self {
             validator_store,
             committee_setups,
             captured_calls,
+            slot_clock,
             is_synced_tx,
             _slashing_db_dir: slashing_db_dir,
             _exit_signal: exit_signal,
         }
```
Then comes the actual test implementation:

#### 2. `anchor/validator_store/src/testing/mod.rs`

```diff
 mod committee_aggregate;
 mod committee_attestation;
+mod proposer_delay;
 mod sync_selection_proof;
```

#### 3. New file: `anchor/validator_store/src/testing/proposer_delay.rs`

```rust
//! Integration tests for the proposer delay applied by `randao_reveal()`.
//!
//! These pin where the wait sits in the method, not just the delay arithmetic. The
//! hazard defended against is hoisting the wait above `collect_signature`, which would turn
//! the floor into added latency on top of pre-consensus and make the error path block for
//! the full delay.
//!
//! The harness clock is positioned to 100ms into `TEST_SLOT` so the documented recommended
//! delay of 300ms (safely under both config bounds: the 1000ms acknowledgement gate and the
//! 4000ms hard cap) leaves a 200ms remainder.

use std::time::Duration;

use bls::PublicKeyBytes;
use slot_clock::ManualSlotClock;
use ssv_types::OperatorId;
use tokio::time::Instant;
use types::Epoch;
use validator_store::ValidatorStore;

use super::common::*;
use crate::Error;

const OUR_OPERATOR_ID: OperatorId = OperatorId(1);
const OPERATOR_IDS: [OperatorId; 4] = [OperatorId(1), OperatorId(2), OperatorId(3), OperatorId(4)];
const COMMITTEE_INDEX: usize = 0;
const VALIDATOR_INDEX: usize = 0;

/// The documented recommended `--proposer-delay-ms` value; a configurable, realistic setting.
const PROPOSER_DELAY: Duration = Duration::from_millis(300);
/// How far into `TEST_SLOT` the clock is repositioned before each test.
const ELAPSED_IN_SLOT_AT_CALL: Duration = Duration::from_millis(100);
/// The floor minus the elapsed time: 300ms - 100ms.
const EXPECTED_REMAINING_WAIT: Duration = Duration::from_millis(200);

/// Epoch of `TEST_SLOT` (slot 1) on the mainnet spec.
const SIGNING_EPOCH: Epoch = Epoch::new(0);

/// Places the shared clock `ELAPSED_IN_SLOT_AT_CALL` into `TEST_SLOT`, replacing the harness
/// default of 5s in, which the 300ms floor could never reach.
fn reposition_clock_early_in_test_slot(slot_clock: &ManualSlotClock) {
    slot_clock.set_current_time(
        Duration::from_secs(TEST_SLOT * SLOT_DURATION_SECS) + ELAPSED_IN_SLOT_AT_CALL,
    );
}

fn harness_with_proposer_delay() -> ValidatorStoreTestHarness {
    let committee = create_committee_setup(&OPERATOR_IDS, 1, 0);
    let harness = ValidatorStoreTestHarness::new_with_proposer_delay(
        vec![committee],
        OUR_OPERATOR_ID,
        PROPOSER_DELAY,
    );
    reposition_clock_early_in_test_slot(&harness.slot_clock);
    harness
}

/// The delay is a floor from slot start, applied *after* signature collection: with the clock
/// 100ms into the slot, the 300ms floor sleeps only the 200ms remainder. Time is paused, so
/// `Instant` observes virtual time and the mock collector's `captured_at` proves the signature
/// was collected before any sleep ran.
#[tokio::test(flavor = "current_thread", start_paused = true)]
async fn randao_reveal_success_waits_only_the_floor_remainder_after_collection() {
    let harness = harness_with_proposer_delay();
    let validator = harness.validator_metadata(COMMITTEE_INDEX, VALIDATOR_INDEX);
    let started = Instant::now();

    let result = harness
        .validator_store
        .randao_reveal(validator.public_key, SIGNING_EPOCH)
        .await;

    result.expect("randao reveal should succeed");
    assert_eq!(
        started.elapsed(),
        EXPECTED_REMAINING_WAIT,
        "delay must be a floor from slot start, sleeping only the remainder"
    );
    let captured = harness.captured_calls.lock();
    assert_eq!(captured.len(), 1, "expected one sign_and_collect call");
    assert_eq!(
        captured[0].captured_at, started,
        "signature must be collected before the proposer delay sleeps"
    );
}

/// A failing duty must not be held to the floor: the wait sits on the success path only, so an
/// unknown pubkey errors immediately with no virtual time spent.
#[tokio::test(flavor = "current_thread", start_paused = true)]
async fn randao_reveal_unknown_pubkey_fails_without_waiting() {
    let harness = harness_with_proposer_delay();
    let unknown_pubkey = PublicKeyBytes::deserialize(&[0xFF; 48]).expect("valid length");
    let started = Instant::now();

    let result = harness
        .validator_store
        .randao_reveal(unknown_pubkey, SIGNING_EPOCH)
        .await;

    assert!(
        matches!(result, Err(Error::UnknownPubkey(pk)) if pk == unknown_pubkey),
        "unknown pubkey should fail with UnknownPubkey"
    );
    assert_eq!(
        started.elapsed(),
        Duration::ZERO,
        "error path must not apply the proposer delay"
    );
}
```

jnhsigmap:
at `anchor/validator_store/src/lib.rs:294-298` - 
- "Slot clock unavailable" is one of the two causes that reach this outcome/branch.
- The comment inside `elapsed_in_slot` (`lib.rs:206-207`) provides the alternative: "a readable clock whose current time is before the slot started."
- Worth adding more detail to guide debugging.

Something like this?
```      
      if let ProposerDelayDecision::ClockUnavailable = decision {
          warn!(
              checkpoint = instrumentation::checkpoints::PROPOSER_DELAY_APPLIED,
              outcome,
              "Slot timing unreadable (clock unavailable or reported time before slot start), skipping configured \
               proposer delay"
          );
```

jnhsigmap (dismissed):

shane-moore:
Both applied in 591bc23c: your `randao_reveal` tests and harness refactor essentially verbatim (all 53 crate tests pass), and the expanded clock warn wording. Thanks for the ready-made drafts!

jnhsigmap (approved):

PR #1215, opened 2026-08-04
chore(deps): bump Lighthouse to unstable HEAD e58ec88fe and migrate sign_attestations to SingleAttestation

### LINKED ISSUE #1214: chore(deps): bump Lighthouse to unstable HEAD e58ec88fe and migrate sign_attestations to SingleAttestation
## Goal

Bump the Lighthouse pin on `epbs` from `81d576943` to unstable HEAD `e58ec88fe` and adapt Anchor so the workspace compiles and validator duties, including Gloas aggregation, work correctly against it.

## Context and motivation

The 59-commit pin window contains three breaking interface changes and one behavioral one:

1. Lighthouse #9578 reshaped the `ValidatorStore` attestation interface: `AttestationToSign` lost its generic and now carries `attester_index`, `pubkey`, `committee_index`, and bare `AttestationData`; `sign_attestations` now streams `Vec<SingleAttestation>`. Anchor's aggregation-bit plumbing (`add_signature`, `validator_committee_index`) has no counterpart in the new contract.
2. The EIP-7688 head commit requires mirroring Lighthouse's six `[patch.crates-io]` git pins into Anchor's root workspace (patches only apply from the root) and adds an `Attestation::Gloas` variant that breaks one exhaustive match.
3. Lighthouse #9573 added `ForkName::Heze`; #9580 renamed the `EnrExt` multiaddr accessors.
4. EIP-7688 progressive types are serialization-compatible with their positional counterparts, but change signing-root derivation at Gloas. Anchor's aggregate decode sites branch only Base-vs-Electra; at Gloas slots they decode successfully but merkleize positionally, and the now-7688-aware beacon nodes reject the resulting `SignedAggregateAndProof` at publish. SIP-94 (as amended in ssvlabs/SIPs#94) requires decoding the aggregate as the Gloas type and computing the signing root under Gloas merkleization. The block path needs no work: decided blocks are already decoded as Gloas types, and progressive roots come with the Lighthouse types.

## Suggested approach

- Mirror Lighthouse's `[patch.crates-io]` entries with a sync-obligation comment; add the Heze `DataVersion` arms (value 9) and the `EnrExt` rename.
- Migrate `sign_committee_attestations` to the new `AttestationToSign`: apply QBFT-decided mutations (root, source, target, and at Gloas the decided index) directly to `data`, build `SingleAttestation` at result assembly, retype `slashing_protection_attestations`.
- Add identity consistency checks at assembly, post-collection: compare the duty's `attester_index` and `committee_index` against Anchor's own metadata; mismatches are marked non-publishable but still flow through slashing protection so every emitted signature is recorded. Do not filter before collection: the committee partial-signature batch size is declared from `VotingAssignments` and the collector sends only on an exact count, so pre-collection filtering stalls the committee.
- Add two fork-aware decode helpers next to `DataVersion` (one for `Attestation`, one for `AggregateAndProof`) with an explicit match: Base, Electra, Gloas, and Heze fail-closed until a Heze wire shape is pinned upstream. Rewire the four aggregate decode sites through them and update the stale doc comment claiming bytes vary only Base/Electra.
- Bind `value.version` to `our_value.version` in both aggregator data validators before decoding, mirroring the proposer duty-slot-fork check.
- Single atomic commit for pin plus compile adaptations; decode helpers and tests may follow as separate review commits.

## Acceptance criteria

- Workspace compiles against `e58ec88fe`; `make lint`, `make test` (including ssv-spec fixtures), `cargo check --workspace --locked`, and `make audit` pass.
- `sign_attestations` streams `Vec<SingleAttestation>` per committee batch; decided-index and pre-Electra index-passthrough semantics are unchanged and covered by tests.
- An identity-mismatched duty reaches the slashing database but is absent from the returned batch, and the committee message still sends.
- At version Gloas, decided aggregates decode as Gloas types and their `AggregateAndProof` signing roots use progressive merkleization (differing from the positional root on identical bytes).
- Heze-version values and version mismatches against the local value are rejected.

## Tests

- Migrate `testing/common.rs`, `testing/committee_attestation.rs` (result alias), and `testing/committee_attestation_gloas.rs`.
- A focused `Signature` to `AggregateSignature` conversion test with a real keypair (the harness mock returns infinity signatures).
- Slashing-protection success, same-data, and conflicting-data cases for the migrated function.
- Decoder root tests: Gloas-typed decode yields the progressive root, differing from the positional root on identical bytes; Heze rejection; version-binding mismatches; maximum-size Gloas aggregate against `MaxAggregatedAttestationBytes`.
- The metadata-service test helper that labels an Electra-shaped attestation as Gloas becomes a real `Attestation::Gloas`.

## Notes

- Rollout: operators computing progressive signing roots and operators computing positional ones split threshold-signature reconstruction on identical decided bytes, so Gloas devnet fleets must upgrade in lockstep with their Lighthouse images. Pre-Gloas networks (mainnet, hoodi) are byte-identical and unaffected.
- No cross-client impact today: go-ssv cannot represent `DataVersion` beyond Fulu; root derivation is now pinned in SIP-94.
- Rejected alternatives: keeping an internal `Attestation<E>` and converting at the boundary (rebuilds deleted complexity); splitting metadata from data internally (churn with no consumer); erroring the whole committee on one identity mismatch (trades one bad duty for the committee's slot); filtering invalid duties pre-collection (stalls the exact-count collector); splitting the 7688 signing-root work into a follow-up PR (would leave Gloas aggregation broken against upgraded beacon nodes in between).
- Out of scope: the local-testnet `cl_image` bump, the share-time slashing insert, mixed-slot/duplicate-batch defensive hardening.

shane-moore:

- The epbs branch does not compile against the latest Lighthouse unstable. The pin window includes sigp/lighthouse#9578, which reshaped the attestation signing interface, and the EIP-7688 merge (sigp/lighthouse#9450), which changed how Gloas attestation types compute signing roots.
- Without the EIP-7688 part, aggregation breaks on a Gloas devnet: upgraded beacon nodes reject aggregates whose signing roots use the old positional merkleization. So the bump and the conformance work ship together.
- Closes #1214. Spec: SIP-94 (ssvlabs/SIPs#94, commit 72a18283d). Previous pin bump: #1123.

## Change Overview (Required)

- Bump the 16 Lighthouse crate revs to e58ec88fe and mirror Lighthouse's six EIP-7688 `[patch.crates-io]` git pins into our root workspace (Cargo only applies patches from the root workspace).
- Migrate `validator_store` to the new signing interface: `AttestationToSign` is no longer generic and carries only `AttestationData`; `sign_attestations` now returns `SingleAttestation` batches. Duty identity fields (`attester_index`, `committee_index`) are compared against our own metadata as a diagnostic only (warn/info log plus a reason-labelled counter); publication proceeds regardless, since the fields are not part of the signing root and the beacon node validates them authoritatively. A duty cannot be dropped before signature collection because the committee's partial-signature batch size is exact and dropping would stall the whole committee.
- Add two fork-aware decode helpers on `DataVersion` in `ssv_types` (one for `Attestation`, one for `AggregateAndProof`). Gloas decodes the EIP-7688 progressive shapes; Heze fails closed until a wire shape is pinned upstream. All four fork-conditional decode sites now go through them.
- Bind the leader-supplied `version` to our own candidate in both aggregator validators, matching the existing proposer check.
- Merkleize `BlindedExecutionPayloadEnvelope` as a progressive container so blinded/full root parity holds at the new pin (SIP-94 section 6).
- Smaller adaptations: Heze `DataVersion` arms, EnrExt `dialable_multiaddrs_*` rename.
- Reading order: `Cargo.toml`, then `ssv_types/src/consensus.rs` (helpers, version binding, envelope), then `validator_store/src/lib.rs` (migration), then tests.
- Not changed: SSV wire encodings, QBFT value hashes, and partial-signature messages are byte-identical (EIP-7688 progressive types serialize the same as their positional counterparts; only merkleization differs). Pre-Gloas networks behave exactly as before.

- On Gloas networks, operators computing progressive signing roots and operators computing positional ones would split threshold-signature reconstruction on identical decided bytes. Rollout must be lockstep on Gloas devnets. No cross-client risk: go-ssv cannot represent versions past Fulu.
- The patched SSZ crates apply workspace-wide, so they could in principle shift our own wire encodings. The ssv-spec fixture tests act as the tripwire and pass unchanged.
- The patch mirror must be re-diffed against Lighthouse's on every future pin bump (marked with a FIXME).

## Validation (Required)

- Full workspace test suite green, including ssv-spec fixtures.
- New tests: identity mismatch reaches the slashing DB, is published with the duty's identity fields echoed verbatim, and does not stall the committee; per-fork decode shape selection and Heze fail-closed; identical bytes decoded as Electra and Gloas shapes produce different roots (EIP-7495 serialization compatibility); version-binding rejections in both validators; max-size Gloas aggregate fits the 131,308-byte bound; blinded/full envelope root parity against the pinned Lighthouse type.
- `cargo fmt`, clippy, and `cargo check --workspace --all-targets` clean.

- Revert the single commit. No config, data, or DB migrations. Only relevant before any Gloas devnet is running.

- N/A

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1204,7 +1204,7 @@ dependencies = [
  "futures-lite",
  "parking",
  "polling",
- "rustix",
+ "rustix 1.1.4",
  "slab",
  "windows-sys 0.61.2",
 ]
@@ -1390,7 +1390,7 @@ checksum = "2af50177e190e07a26ab74f8b1efbfe2ef87da2116221318cb1c2e82baf7de06"
 [[package]]
 name = "beacon_node_fallback"
 version = "0.1.0"
-source = "git+https://github.com/sigp/lighthouse?rev=81d576943#81d576943835d443f43b6dda52769784ed82bd47"
+source = "git+https://github.com/sigp/lighthouse?rev=e58ec88fe#e58ec88fe9a8f755a0d92a80b25fd76100d08d49"
 dependencies = [
  "bls",
  "clap",
@@ -1490,7 +1490,7 @@ dependencies = [
 [[package]]
 name = "bls"
 version = "0.2.0"
-source = "git+https://github.com/sigp/lighthouse?rev=81d576943#81d576943835d443f43b6dda52769784ed82bd47"
+source = "git+https://github.com/sigp/lighthouse?rev=e58ec88fe#e58ec88fe9a8f755a0d92a80b25fd76100d08d49"
 dependencies = [
  "alloy-primitives 1.5.7",
  "blst",
@@ -2606,7 +2606,7 @@ dependencies = [
 [[package]]
 name = "eip_3076"
 version = "0.1.0"
-source = "git+https://github.com/sigp/lighthouse?rev=81d576943#81d576943835d443f43b6dda52769784ed82bd47"
+source = "git+https://github.com/sigp/lighthouse?rev=e58ec88fe#e58ec88fe9a8f755a0d92a80b25fd76100d08d49"
 dependencies = [
  "bls",
  "ethereum_serde_utils 0.8.0",
@@ -2830,7 +2830,7 @@ dependencies = [
 [[package]]
 name = "eth2"
 version = "0.1.0"
-source = "git+https://github.com/sigp/lighthouse?rev=81d576943#81d576943835d443f43b6dda52769784ed82bd47"
+source = "git+https://github.com/sigp/lighthouse?rev=e58ec88fe#e58ec88fe9a8f755a0d92a80b25fd76100d08d49"
 dependencies = [
  "bls",
  "context_deserialize",
@@ -2840,6 +2840,7 @@ dependencies = [
  "ethereum_serde_utils 0.8.0",
  "ethereum_ssz",
  "ethereum_ssz_derive",
+ "fork_choice",
  "futures",
  "futures-util",
  "mediatype",
@@ -2859,7 +2860,7 @@ dependencies = [
 [[package]]
 name = "eth2_config"
 version = "0.2.0"
-source = "git+https://github.com/sigp/lighthouse?rev=81d576943#81d576943835d443f43b6dda52769784ed82bd47"
+source = "git+https://github.com/sigp/lighthouse?rev=e58ec88fe#e58ec88fe9a8f755a0d92a80b25fd76100d08d49"
 dependencies = [
  "paste",
  "types",
@@ -2868,7 +2869,7 @@ dependencies = [
 [[package]]
 name = "eth2_interop_keypairs"
 version = "0.2.0"
-source = "git+https://github.com/sigp/lighthouse?rev=81d576943#81d576943835d443f43b6dda52769784ed82bd47"
+source = "git+https://github.com/sigp/lighthouse?rev=e58ec88fe#e58ec88fe9a8f755a0d92a80b25fd76100d08d49"
 dependencies = [
  "bls",
  "ethereum_hashing 0.8.0",
@@ -2881,7 +2882,7 @@ dependencies = [
 [[package]]
 name = "eth2_key_derivation"
 version = "0.1.0"
-source = "git+https://github.com/sigp/lighthouse?rev=81d576943#81d576943835d443f43b6dda52769784ed82bd47"
+source = "git+https://github.com/sigp/lighthouse?rev=e58ec88fe#e58ec88fe9a8f755a0d92a80b25fd76100d08d49"
 dependencies = [
  "bls",
  "num-bigint-dig",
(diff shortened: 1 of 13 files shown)


shane-moore:
@claude, please take a look

shane-moore:
@claude Applied the metrics nit in fa585db6: the identity-mismatch withheld branch now increments SIGNED_ATTESTATIONS_TOTAL with a dedicated `withheld` label instead of `other_error`.

shane-moore:
@claude, please review the new commit 147af278: the identity-mismatch action flips from withhold-from-publication to publish-with-diagnostic. The identity fields are not part of the signing root and the beacon node validates them authoritatively, so withholding could only suppress valid attestations when the slot-start snapshot went stale mid-slot. The `withheld` label from the earlier metrics nit is gone; mismatches now count in `anchor_attestation_duty_identity_mismatches_total{reason}` (warn for `attester_index`, info for committee drift). PR body updated to match.

jnhsigmap (approved):

Issue #1217, opened 2026-08-05
PTC payload attestations deadlock when a slot has multiple PTC duties (random duty order + serial threshold signing)


shane-moore:
## Summary

On a 4-operator Gloas devnet, every PTC duty in a slot fails with `SignatureCollectionFailed(QueueClosedError)` whenever the slot carries two or more PTC duties and the operators split evenly on which duty to sign first. The duty order is random per process (a `HashSet` iteration in the LH duties service), signing is serial per duty, and under Anchor each signature blocks on the 3-of-4 partial-signature threshold, so a 2-2 split on the first duty deadlocks both collections. Observation data is byte-identical across all four operators, and nothing gets published for the slot.

The code involved is Lighthouse's `PayloadAttestationService` / `DutiesService` (spawned by Anchor since #1078), but the bug can only manifest under Anchor: a plain VC signs locally and instantly, so list order is irrelevant. Filing here because Anchor owns the integration and the LH pin.

## Environment

- Anchor [`4a4cfcf6`](https://github.com/sigp/anchor/commit/4a4cfcf65e33f6e2ac5f7531db8164c2c852c48f), Lighthouse pin [`e58ec88fe`](https://github.com/sigp/lighthouse/commit/e58ec88fe9a8f755a0d92a80b25fd76100d08d49)
- 4 geo-distributed operators (Sydney, Montreal, Falkenstein x2), one cluster, threshold 3-of-4
- Self-genesis Gloas devnet (fork at epoch 3, 12s slots), `RUST_LOG=debug,signature_collector=trace,message_receiver=trace`

## Mechanism

[...]

PR #1221, opened 2026-08-05
fix(validator_store): bound PTC collection awaits to the duty slot end

### LINKED ISSUE #1218: fix(validator_store): bound PTC collection awaits to the duty slot end
## Goal

Bound the PTC payload-attestation signature-collection await to the end of the duty slot, so a no-quorum collection resolves while publishing is still possible instead of hanging until collector eviction.

## Context / motivation

`sign_payload_attestation` awaits `collect_signature` with no timeout (`anchor/validator_store/src/lib.rs:3500-3541`). A collection that never reaches threshold resolves only when the cleaner evicts it at the second slot boundary after the duty fires (`SIGNATURE_COLLECTOR_RETAIN_SLOTS = 1`), surfacing as `QueueClosedError` well after the [75%, 100%] publish window has closed. A slot-N payload attestation is consulted by fork choice only in slot N+1 and dropped on the wire once `data.slot != current_slot`, so any wait past slot end is pure loss. #1217 measured the blast radius under the current serial loop: ~15s for the first non-quorum validator plus ~12s per additional one, with reconstructed sibling messages held unpublished throughout.

The Lighthouse-side fix for #1217 (concurrent signing) removes the ordering deadlock but leaves a slot's batch publish coupled to its slowest collection. The residual case is complementary duty-set divergence: after a dependent-root flap with a 2-2 beacon-node-view split, operators 1,2 hold duties {A, X} and operators 3,4 hold {A, Y}; duty A reconstructs on all four operators while X and Y sit at 2 of 3 partials, every operator's concurrent join blocks on its phantom duty until eviction, and A's message misses the slot everywhere. With awaits bounded at slot end, the phantom awaits fail in time for A's batch to publish.

`sign_proposer_preferences` already ships this exact mitigation: `tokio::time::timeout` around the same `collect_signature` call, mapping elapse to `CollectionError::CollectionTimeout` (`lib.rs:3560-3585`, constant at `:104`). PTC is the only remaining Gloas signature-collection path with an unbounded await.

## Suggested approach

- Wrap the `collect_signature` await in `sign_payload_attestation` with `tokio::time::timeout_at` at the end of `data.slot`, via `self.get_instant_in_slot(data.slot + 1, Duration::ZERO)?` (`lib.rs:814`). An absolute deadline rather than ProposerPreferences' relative 2-slot constant, because PTC usefulness ends at a fixed point in the slot; a call arriving after slot end then fails fast, which is the desired behavior.
- Map elapse to `CollectionError::CollectionTimeout` exactly as ProposerPreferences does (`lib.rs:3580-3585`), so it flows through `report_ptc_collection_failure` (`lib.rs:873`) into the existing NoSignature bucket (`instrumentation.rs:54-61`) and `anchor_ptc_reconstruction_failures_total{reason="no_signature"}` with no new telemetry.
- Update the instrumentation comment that says `CollectionTimeout` is synthesized only by `sign_proposer_preferences` (`instrumentation.rs:54`).
- Bounding at exactly slot end (no subtracted margin) is deliberate: publishes slightly past the boundary can still land within `MAXIMUM_GOSSIP_CLOCK_DISPARITY` (500ms), and an earlier cutoff wastes reconstruction chances. Tune at PR time if evidence suggests otherwise.

## Acceptance criteria

- A PTC collection that never reaches threshold returns `CollectionTimeout` by the end of the duty slot instead of `QueueClosedError` at eviction.
- The timeout classifies as NoSignature (warn + `reason="no_signature"`), not infra; no new metric or log shape.
- A call made after the duty slot has ended fails fast instead of awaiting.
- Successful collections are unaffected; no wire-behavior change.

## Tests

Mirror the ProposerPreferences timeout tests (`anchor/validator_store/src/testing/proposer_preferences.rs:443-512`): a no-quorum collection surfaces `CollectionTimeout` via the bound and increments the reconstruction-failure metric; a threshold-reaching collection is unaffected. Add the fail-fast case (duty slot already over at call time).

## Notes

Follow-up from #1217 planning. Complements, and does not replace, the Lighthouse-side concurrent-signing fix: concurrency removes the ordering deadlock; this bound removes the residual coupling of a slot's publish to a genuinely sub-threshold collection. Related: #1079; surfacing no-quorum as `CollectionTimeout` at the PTC callsite narrows the QueueClosedError ambiguity #1079 describes, but collector-granularity divergence attribution remains its scope.

shane-moore:

- `sign_payload_attestation` awaited `collect_signature` with no bound; a no-quorum collection resolved only at collector eviction, surfacing as `QueueClosedError` well after the [75%, 100%] publish window. Peers drop a payload attestation once `data.slot` is no longer current, so any wait past slot end is pure loss.
- #1217 measured the blast radius on a Gloas devnet. The Lighthouse-side concurrent-signing fix (sigp/lighthouse#9756) removes the ordering deadlock but leaves a slot's batch publish coupled to its slowest collection; this bound removes that residual coupling.
- Closes #1218. Related: #1217, #1079.

## Change Overview

- Bound the PTC collection wait at the end of the duty slot: compute the remaining slot time with saturating slot-clock math, fail fast without starting a collection (or broadcasting a partial) when the slot is already over, otherwise time out at slot end. Elapse surfaces as `CollectionTimeout` through the existing NoSignature reporting path (`reason="no_signature"`); no new telemetry.
- The elapse-to-`CollectionTimeout` mapping now lives in a `collect_within` helper shared with `sign_proposer_preferences`; each caller keeps its own bound computation at the call site.
- Read `lib.rs` first (helper, then the PTC call site), then the comment updates in `instrumentation.rs`/`metrics.rs`, then the tests.
- Intentionally unchanged: wire behavior, ProposerPreferences' relative 2-slot bound, collector eviction semantics, metric and log shapes.

- Deadline is exactly slot end, no subtracted margin: maximizes reconstruction chances, and a boundary publish still lands within `MAXIMUM_GOSSIP_CLOCK_DISPARITY`. One-constant change if devnet evidence argues for a margin.
- The bound applies to the caller only: already-queued signing/broadcast work may still run after a timeout. Deliberate; our late partial can still complete a slower peer's collection (PTC partials are accepted up to 2 slots late).

## Validation

- New paused-clock tests: a hanging collection surfaces `CollectionTimeout` after exactly the remaining slot fraction (an accidental fixed relative wait fails the assertion), and an expired duty fails fast without starting a collection; both assert the no_signature/infra metric split.
- `cargo test -p anchor_validator_store` (68 passed), `make cargo-fmt-check`, and `cargo clippy --all-targets -- -D warnings` clean.

## Rollback

- Plain revert restores the unbounded await; no config, data, or wire impact.


shane-moore:
@claude, please take a look

anchor/validator_store/src/lib.rs
```diff
@@ -3514,18 +3535,36 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> ValidatorStore
...
-            .collect_signature(
+        // Peers drop a payload attestation once `data.slot` is no longer current, so bound the
+        // wait at slot end rather than hanging until collector eviction (#1218). The saturating
+        // slot/duration math cannot panic; `duration_to_slot` is avoided because its `None`
+        // would conflate an expired duty with a broken clock.
+        let now = self
+            .slot_clock
+            .now_duration()
+            .ok_or(SpecificError::SlotClock)?;
+        let slot_end = self
```
jnhsigmap, line 4006: if there is a case for checking this `data.slot` value to ensure that it is not misplaced (too far ahead), then we could add a check here
```
  // `data.slot` arrives unchecked and determines the duration of the await. More than one slot ahead
  // can indicate a broken or hostile beacon node. Reject it if it further ahead of the next slot.
  let current_slot = self
      .slot_clock
      .slot_of(now)
      .ok_or(SpecificError::SlotClock)?;

  if data.slot > current_slot + 1 {
      return Err(SpecificError::SlotTooFarAhead {
          data_slot: data.slot,
          current_slot,
      }
      .into());
  }
```

where `SpecificError::SlotTooFarAhead` can be added above as a new variant:
```
  /// Duty data names a slot more than one slot ahead of the local clock. Only a broken
  /// or hostile beacon node produces this.
  SlotTooFarAhead { data_slot: Slot, current_slot: Slot },
```

A test can be added too where the `SlotTooFarAhead` error is asserted where `data.slot = Slot::new(TEST_SLOT + 2)`.

jnhsigmap:
all good otherwise, just like my change it's got conflicts :sweat_smile:

anchor/validator_store/src/lib.rs
shane-moore: Good catch, and the case is stronger than "if there is a case": the timeout is sized from `data.slot`, and collector eviction keys off the same slot (`for_slot >= current - 1`), so a far-future value re-opens exactly the unbounded wait this PR removes. It would also broadcast a partial that peers reject under the strict no-future rule for PTC partial signatures.

Added in 03117a8a essentially as you proposed: `SlotTooFarAhead` as a pre-collection error (no collection started, no partial broadcast, same treatment as `MissingIndex`), your `TEST_SLOT + 2` rejection test, plus a companion pinning that `TEST_SLOT + 1` is still accepted so the headroom can't silently tighten.

jnhsigmap (approved):

Issue #1126, opened 2026-07-09
feat(validator_store): implement sign_execution_payload_envelope


shane-moore:
## Goal

Replace the `Unsupported` stub at `validator_store/src/lib.rs:3315` with the SIP-94 envelope-signing duty against the current Lighthouse full-envelope trait: blind the full `ExecutionPayloadEnvelope` that Lighthouse hands in for the internal QBFT value, collect post-consensus signatures under `Domain::BeaconBuilder`, and return the full `SignedExecutionPayloadEnvelope` only when the local full envelope blinds to the decided value.

## Context / motivation

Lighthouse already drives this duty inline from `BlockService`: after a Gloas block is signed and published, it fetches the envelope, calls `sign_execution_payload_envelope(pubkey, envelope)`, and POSTs every successful result (`block_service.rs:627-689`, `:698-715` at LH pin `81d576943`). The trait passes the full envelope and no separate duty or slot (`validator_client/validator_store/src/lib.rs:202-207`), so the slot derives from `envelope.slot()` (`payload.slot_number`, `execution_payload_envelope.rs:65-67`). The current pin's envelope fields are `payload`, `ExecutionRequestsGloas<E>`, `builder_index`, `beacon_block_root`, and `parent_beacon_block_root` (`execution_payload_envelope.rs:20-27`).

The intended shape is go-ssv parity at the SSV layer: no pre-consensus phase, QBFT over `EnvelopeConsensusData.DataSSZ = SSZ(BlindedExecutionPayloadEnvelope)`, value-check against the block-QBFT-decided root, post-consensus signing under `DOMAIN_BEACON_BUILDER`, and content-matched publication only by operators whose full local envelope blinds to the decided value. Content matching does not elect a unique publisher: multiple operators may hold byte-identical matching envelopes. One current Lighthouse seam differs from go-ssv: go-ssv can let non-builders skip publication and still mark the duty successful, but this LH trait POSTs every `Ok`, so Anchor must return a distinct non-publish error for non-matching operators. This is the only safe result with the current trait, but Lighthouse wraps every signer error as `BlockError::Recoverable` and logs it as a block-proposal failure even though the block was already published. It does not retry this path.

ethereum/beacon-APIs#624 currently proposes removing the public blinded-envelope transport from the API. That is fine for this issue: Anchor's internal QBFT value remains blinded, while the Lighthouse-facing method and publication path are full-envelope at the current pin.

## Suggested approach

Lead with symbols; line hints drift.

```rust
[...]

PR #1224, opened 2026-08-07
feat(validator_store): record block-QBFT-decided beacon block roots

### LINKED ISSUE #1124: feat(validator_store): thread the block-QBFT-decided root to the envelope value check
## Goal

The envelope value check must enforce `envelope.beacon_block_root ==` the root the block QBFT decided for the slot, but nothing carries that root out of the proposer flow. For Gloas-era decided data, record it at `decide_abstract_block`'s single success point, keyed `(validator_pubkey, slot)`, readable by the envelope sign path.

## Context / motivation

`decide_abstract_block` (`validator_store/src/lib.rs:542` at Anchor `ceaa8c00`) has one authoritative success tail (after `decode_decided_block`, :2002) that every participating operator reaches with the QBFT-decided value, not the local proposal. Its only current caller is `sign_block` (:2807). Lighthouse calls `fetch_sign_and_publish_payload_envelope` after `sign_and_publish_block` returns, then calls `sign_execution_payload_envelope` with the fetched envelope (`validator_client/validator_services/src/block_service.rs`). The root therefore already exists in the normal local sign path before envelope signing begins.

Since Lighthouse #9276 (merged), the envelope GET is keyed by `(slot, locally produced block root)`. That root serves only the local BN fetch: each operator retrieves the envelope its own BN built, which is what lets non-builders initialize and participate in envelope QBFT. It can differ from the decided root whenever another operator's proposal wins block QBFT. The value check must use the decided root recorded here; the two roots must remain visibly separate.

## Suggested approach

Add `decided_block_roots: Mutex<HashMap<DecidedBlockRootKey, Hash256>>` to `AnchorValidatorStore`, keyed `(validator_pubkey, slot)`. This is a protocol handoff store, not a cache: the decided root is not recomputable after loss, and its lifetime is the duty slot's liveness. Do not use `DashMap`, an LRU, a watch channel, or a QBFT-manager query.

**Write path.** After `Completed::Success`, successfully decode the decided block. Only when `ForkName::from(completed_data.version) >= ForkName::Gloas`, record the decoded Gloas `BeaconBlock`'s `canonical_root()`. The Gloas decoder yields `UnsignedBlock::Full(FullBlockContents::Block(block))`, so this path needs no Blinded/Full root dispatch. Do not record the local proposal root, and do not use `ProposerConsensusData::hash()`, which is the QBFT value hash over wrapper SSZ bytes. Compute the root outside the mutex; never hold the lock across an await.

Writes are first-write-wins (precedent: `VotingContext::remember_decided_vote`):

- same key, same root: idempotent `Ok`;
- same key, different root: hard conflict error. This deliberately fails the in-flight block duty before threshold signing. Two decided roots for one `(validator, slot)` means QBFT safety has failed or the handoff bookkeeping is corrupt; threshold-signing the second block would be the dangerous choice.

**Read path.** Expose narrow `record_decided_block_root` / `get_decided_block_root(&self, validator_pubkey, slot)` helpers. The getter reads `self.slot_clock` internally and rejects requests older than `MAX_DECIDED_ROOT_AGE_SLOTS` (suggest 4, exceeding the 3-slot EnvelopeProposer partial-signature TTL) before the map lookup. Do not accept a caller-supplied current slot: the consumer's slot derives from the envelope under check. Valid reads are non-destructive and return the root by value; the envelope value check receives a snapshot, never a live map reference. Missing root, stale request, and clock failure each surface a distinct error; no caller may substitute a default root.

**Bounding and staleness are two separate responsibilities:**

- evict-on-insert by slot (drop entries older than the max age) keeps memory bounded, matching go-ssv's `ProposedBlockRoots` retention;
- the clock-based check on read is what prevents a stale entry from being served. Eviction alone cannot: with no later inserts, old entries linger in any evict-on-insert scheme.

## Acceptance criteria

- After a block-QBFT decide, the envelope path reads the decided root for `(validator, slot)`. Distinct slots do not collide, and validator identity is explicit rather than relying on go-ssv's per-validator runner scoping.
- The stored root equals the decoded decided block's beacon block root, not `ProposerConsensusData::hash()` and not the local full proposal root.
- Missing, stale (past `MAX_DECIDED_ROOT_AGE_SLOTS` by the store's own slot clock), and clock-failure reads surface as distinct errors, never an accept.
- No code path substitutes `Hash256::zero()` or another default root.
- Writes are first-write-wins: a same-root rewrite returns `Ok`; a conflicting-root write is a hard error that aborts the block duty before threshold signature collection.
- Entries are bounded by slot-based eviction on insert; the read-side clock check, not eviction, is the stale-serve guard.
- Pre-Gloas decisions do not create root-store entries (holds with #1129 in place; see Notes) and the existing proposer flow remains unaffected.

## Tests

Standalone `validator_store` tests:

- Gloas record-then-read; missing root; cross-validator isolation; cross-slot isolation.
- Root-shape regression: the stored root is the decoded block root, not the consensus wrapper hash.
- Stale read is rejected while the entry is still physically present in the map (proves the read check, not eviction, provides stale-read protection).
- Exact `MAX_DECIDED_ROOT_AGE_SLOTS` boundary, and one slot beyond it.
- Clock-failure error path.
- Same-root rewrite is idempotent; conflicting-root write is a hard error that preserves the original entry.
- A conflicting write aborts block signing before `sign_abstract_block` / signature collection is entered (a map-level error test alone is insufficient).
- Out-of-order old insertion does not evict or disturb a newer live root.
- Pre-Gloas decision creates no entry (run with #1129's validator in place).

The consumer lands in the sign-path follow-up; if the getter is otherwise unused, scope a narrow `#[expect(dead_code)]` or exercise it test-only to pass lint.

## Notes

- **#1129 is a hard prerequisite, not just a related check.** The `version >= Gloas` record gate reads a leader-supplied field: without #1129, a Gloas-stamped decided value at a pre-Gloas duty slot could survive consensus, decode on the Gloas arm, and reach the record branch, violating the pre-Gloas-no-entry invariant. With #1129 rejecting `version` values that disagree with the fork scheduled at the duty slot, the gate is sound. Do not duplicate #1129's validation in this hand-off.
- Lighthouse context: #9276 and beacon-APIs#624 are merged. The epbs branch still pins Lighthouse `81d576943`; a pin bump is required before #1126 (the envelope GET/POST shapes changed) but not for this issue, since the trait passes no root either way.
- go-ssv parity: go-ssv's `ProposedBlockRoots` is the same shape conceptually (mutex, map, slot-window eviction with retention 4). Its store is slot-keyed because each runner is per-validator; Anchor's `AnchorValidatorStore` is multi-validator, so `(validator_pubkey, slot)` makes that scoping explicit, while the protocol still has only one proposer for a slot. This store is deliberately stricter than go-ssv on two axes: go-ssv overwrites silently on conflicting writes and has no read-side staleness check.
- Normal Lighthouse ordering means the root exists before local envelope signing (`sign_and_publish_block` completes, including block QBFT, before the envelope flow starts). If a future refactor permits overlap, a short wait can be considered, but timeout/missing root must remain an error.

jnhsigmap:

- The envelope milestone rejects any proposed envelope whose `beacon_block_root` differs from the root the cluster decided. Nothing records that root today, so the check has no value to compare against.
- Block QBFT is the only point where every participating operator holds the decided block. Record the root there, and all operators hold the same value when envelope signing starts.
- Closes #1124. The reader of this store lands in #1126.

## Change Overview (Required)

- `AnchorValidatorStore` gains a small handoff store: a mutex-guarded map from `(validator, slot)` to the decided block root, written at the success point of block consensus for Gloas-and-later duties.
- Writes are first-write-wins. A repeat write of the same root is idempotent. A different root for the same duty is a hard error that aborts the duty before threshold signing, because two decided roots for one duty means consensus safety failed.
- Reads reject slots older than 4 slots against the store's own clock. Eviction on insert bounds the map. A miss, a stale read, and a clock failure each return a distinct error. No path returns a default root.
- Read the production change in `validator_store` first (one store field, two private helpers, one hook in the block-decide path). The two new test modules are map-level contract tests and end-to-end `sign_block` tests, and they share a small fixture struct.
- Intentionally unchanged: the envelope sign stub, pre-Gloas proposer flow (creates no entries), `decode_decided_block`, and the store's public constructor signature.

- Blast radius is one crate. The field and helpers are private, so no consumer outside `validator_store` changes.
- Trade-off: a conflicting root fails the whole block duty instead of overwriting. Signing a second block for one slot is the dangerous outcome, so the duty aborts and an end-to-end test pins that no signature is attempted after a conflict.
- Trade-off: the conflict error payload is boxed to keep the error enum under clippy's size limit while it keeps both roots for diagnosis.
- The write path takes no clock check by design. The decided value passed slot validation upstream, and a late write for an old slot must still succeed. A test pins that such a write cannot evict a newer live root.

## Validation (Required)

- 11 map-level tests pin the store contract: exact round-trip, error on a miss (never a default), key isolation across validators and slots, idempotent rewrite, conflict keeps the first root, stale rejection while the entry is still present, the inclusive age boundary, and both eviction directions.
- 3 end-to-end tests drive the real `sign_block` path: the stored root is the decoded block's `canonical_root()` and not the QBFT wrapper hash, a conflict aborts before any threshold signature, and an Electra duty leaves the store empty.
- `cargo test -p anchor_validator_store` (79 passed), full workspace `make test` in release mode, `make cargo-fmt-check`, and `make lint` all pass.

- Revert the two commits. The store is in-memory only, private, and has no consumer yet, so a revert has no config, data, or operational impact.

- None to merge. #1126 consumes the read helper and removes its dead-code allowance.

- Known coverage limits: the clock-failure read arm is untestable with the current harness (genesis at time zero), and the pre-Gloas test exercises the decoded-shape guard rather than the version gate, because Electra decodes to a blinded block. Both limits are recorded as comments in the test files.


jnhsigmap:
@claude review this PR. Prioritise finding potential edge cases not covered and quality of tests. Ensure solution is correctly implemented against issue spec.

jnhsigmap:
addressed claude review in https://github.com/sigp/anchor/pull/1224/commits/852aedd861a7468d12defdb21a8c084c02fb29ed. also my bad with the source branch name :facepalm:. I will fix that up on my local after this merges.

shane-moore:
lgtm breh! will approve after you knock out the small merge conflicts

shane-moore (approved):

PR #1228, opened 2026-08-08
fix(fork): publish Boole lifecycle transitions at exact slots

### LINKED ISSUE #1191: fix(fork): align Boole lifecycle transitions and downstream state at exact slots
## Goal

Publish each Boole fork lifecycle transition only when its recorded slot has been reached, while keeping lifecycle-driven topic scoring and ENR state consistent with the new current fork.

The node must not publish a transition early, remain silently alive with a dead fork monitor, leave Boole topics unscored for the first postfork epoch, or advertise a stale ENR when startup crosses activation.

## Context / motivation

Verified against Anchor `unstable` at `c168c6e12ee6cfb6e3285725226fb611020d7b6b`.

`ForkMonitor::new` records the intended transition slots:

| Lifecycle | Recorded slot | Meaning |
| --- | ---: | --- |
| `WarmUp` | activation minus one epoch | Subscribe to current and upcoming topics |
| `GracePeriod` | activation | Make Boole current while retaining Alan topics |
| `Normal` | activation plus 32 slots | Drop Alan topics |

See [`ForkMonitor::new`](https://github.com/sigp/anchor/blob/c168c6e12ee6cfb6e3285725226fb611020d7b6b/anchor/common/fork/src/monitor.rs#L84-L118).

The runtime does not honor those boundaries. `sleep_until_slot` subtracts one slot from every target, sleeps using a whole-seconds duration, and returns without checking the clock again. `run` then publishes the lifecycle immediately.

See [`sleep_until_slot` and `run`](https://github.com/sigp/anchor/blob/c168c6e12ee6cfb6e3285725226fb611020d7b6b/anchor/common/fork/src/monitor.rs#L161-L232) and the [whole-seconds client call](https://github.com/sigp/anchor/blob/c168c6e12ee6cfb6e3285725226fb611020d7b6b/anchor/client/src/lib.rs#L446-L453).

This makes all three transitions one slot early:

- Early `WarmUp` adds Boole subscriptions sooner than scheduled.
- Early `GracePeriod` changes lifecycle-driven ENR, handshake, discovery, scoring, and peer state while messages for that slot still use Alan.
- Early `Normal` removes Alan subscriptions before the configured grace window has elapsed.

Outgoing routing and incoming validation already derive the fork from the message slot and correctly switch at activation:

- [Outgoing slot-derived topic routing](https://github.com/sigp/anchor/blob/c168c6e12ee6cfb6e3285725226fb611020d7b6b/anchor/message_sender/src/network.rs#L158-L186)
- [Incoming slot-derived fork and domain validation](https://github.com/sigp/anchor/blob/c168c6e12ee6cfb6e3285725226fb611020d7b6b/anchor/message_validator/src/lib.rs#L607-L665)

### Exact activation exposes a topic-scoring dependency

During `WarmUp`, Boole topics are subscribed while Alan remains `fork_to_score`. The Boole `Subscribe` events therefore carry no message rate:

- [Lifecycle and subscription reconciliation](https://github.com/sigp/anchor/blob/c168c6e12ee6cfb6e3285725226fb611020d7b6b/anchor/subnet_service/src/subscriptions.rs#L117-L214)
- [Conditional message-rate attachment](https://github.com/sigp/anchor/blob/c168c6e12ee6cfb6e3285725226fb611020d7b6b/anchor/subnet_service/src/subscriptions.rs#L248-L275)

The network subscribes to those topics but skips `set_topic_params` when the rate is absent:

- [Network topic-event handling](https://github.com/sigp/anchor/blob/c168c6e12ee6cfb6e3285725226fb611020d7b6b/anchor/network/src/network.rs#L577-L650)

At exact activation, `WarmUp -> GracePeriod` changes `fork_to_score` from Alan to Boole but does not change the subscribed fork set. No subscription or rate event is emitted. The existing lifecycle test explicitly expects no event:

- [WarmUp-to-GracePeriod test](https://github.com/sigp/anchor/blob/c168c6e12ee6cfb6e3285725226fb611020d7b6b/anchor/subnet_service/src/subscriptions.rs#L588-L601)

Scoring updates otherwise run only at epoch boundaries:

- [Subnet-service select loop](https://github.com/sigp/anchor/blob/c168c6e12ee6cfb6e3285725226fb611020d7b6b/anchor/subnet_service/src/subscriptions.rs#L88-L115)
- [Scoring-rate update path](https://github.com/sigp/anchor/blob/c168c6e12ee6cfb6e3285725226fb611020d7b6b/anchor/subnet_service/src/scoring.rs#L12-L61)

Neither activation-boundary ordering installs Boole parameters:

1. If the scoring branch wins, it refreshes Alan and schedules the next epoch.
2. If the lifecycle branch wins, the pending epoch sleep is cancelled and the next loop schedules the following epoch.

Boole topics consequently have no topic score parameters, including the invalid-message P4 penalty, for the first 32 postfork slots. Validation still rejects invalid messages, but the delivering peer does not accumulate the intended topic-level penalty.

The old one-slot-early `GracePeriod` accidentally hid this dependency by changing `fork_to_score` before the activation epoch timer fired. Restoring early lifecycle publication is not the solution.

### Long sleeps do not follow later wall-clock corrections

`SystemTimeSlotClock` calculates a duration from wall time, while `tokio::time::sleep` waits against a monotonic deadline. A long one-shot sleep therefore does not react promptly if wall time advances across the target.

For example, if the monitor schedules a ten-slot sleep at slot 90 and wall time is then corrected to slot 100, the lifecycle can remain stale until the original monotonic timer expires.

### Monitor failure is not currently fail-closed

If a runtime clock reading becomes unavailable or is observed moving backwards, merely returning from the monitor task leaves Anchor running with the last lifecycle value.

Slot-derived sender and validator behavior can then advance to Boole while subscriptions, ENR, discovery, handshake, and peer state remain on Alan. A log followed by silent monitor completion is not a fail-closed node state.

### Startup can permanently miss an ENR transition

`Discovery::new` snapshots the lifecycle domain before awaiting discv5 startup and bootnode ENR requests:

- [Discovery ENR construction and awaits](https://github.com/sigp/anchor/blob/c168c6e12ee6cfb6e3285725226fb611020d7b6b/anchor/network/src/discovery.rs#L201-L297)

If activation occurs during those awaits, `Network::try_new` later reads the new lifecycle using `borrow_and_update`, marks it seen, and records it as `prev_lifecycle` without reconciling the already-built ENR:

- [Network construction](https://github.com/sigp/anchor/blob/c168c6e12ee6cfb6e3285725226fb611020d7b6b/anchor/network/src/network.rs#L95-L149)
- [Runtime lifecycle handling](https://github.com/sigp/anchor/blob/c168c6e12ee6cfb6e3285725226fb611020d7b6b/anchor/network/src/network.rs#L180-L207)

The node can then filter and handshake using Boole while continuing to advertise Alan.

### Exact and bounded monitor waits

- Pass `Slot` directly to the wait helper.
- Remove `seconds_per_slot` plumbing and the `as_secs()` conversion.
- Use the slot clock's full `Duration` without truncating fractional values.
- Never publish until a fresh slot-clock observation is at or beyond the target.
- Bound each sleep to at most one slot, then re-read the clock. This bounds lateness after a forward wall-clock correction.
- Carry the last trusted observation across successive transitions so a backwards movement between transition waits is not silently forgotten.
- Treat an unavailable or observed-backwards runtime clock as a client-fatal clock error. Request global shutdown instead of allowing only the fork-monitor task to exit.

### Reconcile scoring when the current fork changes

After applying a lifecycle update and reconciling subscriptions:

- detect whether `fork_to_score` changed;
- if topic scoring is enabled, immediately emit `RateUpdate` events for every already-subscribed topic of the new current fork;
- do not wait for the next epoch timer;
- do not publish `GracePeriod` early or score dormant Boole topics by changing lifecycle semantics.

### Reconcile ENR state during network construction

After `AnchorBehaviour::new` completes:

- read and mark the latest lifecycle;
- reconcile Discovery's local ENR domain to that lifecycle before moving the behavior into the swarm;
- retain later lifecycle updates as unseen watch changes for `Network::run`;
- fail startup rather than returning a network whose ENR is known to disagree with its lifecycle.

## Acceptance criteria

- `WarmUp`, `GracePeriod`, and `Normal` are never published before their recorded slots.
- A transition publishes when the slot clock reaches or passes its target.
- Continuous operation and restart at the same slot produce the same lifecycle.
- Fractional slot durations are preserved without integer-second truncation.
- An early Tokio wake cannot publish a transition while the slot clock remains before the target.
- A forward clock correction crossing the target is detected within one bounded resampling interval.
- An unavailable or observed-backwards clock cannot leave Anchor running indefinitely with a stale lifecycle.
- With topic scoring enabled, `WarmUp -> GracePeriod` immediately emits Boole `RateUpdate` events for existing Boole subscriptions without advancing to the next epoch.
- A network construction that straddles activation returns with an ENR matching the latest lifecycle.
- Existing slot-derived sender and validator fork selection remains unchanged.
- Existing transition-window constants and lifecycle types remain unchanged.
- This issue does not broaden NodeInfo handshake admission or change discovery candidate compatibility. #1185 and PR #1187 remain separate.

## Tests

Add deterministic coverage for:

1. Exact target publication, including `target - 1`, exact target, and already-past target.
2. Fractional slot duration.
3. Early Tokio timer wake while the slot clock remains behind.
4. Forward wall-clock jump across the target while the original monotonic deadline remains in the future.
5. Unavailable and backwards clock handling, including the client-fatal outcome.
6. A scoring-enabled `WarmUp -> GracePeriod` transition that immediately emits Boole rate updates without an epoch advance.
7. Network construction where lifecycle activation occurs after Discovery snapshots Alan but before `Network::try_new` records the latest lifecycle.
8. Existing complete lifecycle sequence, handshake, discovery, network, and subnet-service regressions.

Run at least:

```bash
cargo test -p fork
cargo test -p subnet_service
cargo test -p network handshake
cargo test -p network
cargo check -p client
make cargo-fmt-check
make lint
make test
git diff --check
```

## Notes

- The defect is deterministic whenever a Boole fork is scheduled. The built-in mainnet, Hoodi, and Holesky schedules currently contain Alan only, so it is dormant on those stock configurations.
- Existing Boole implementation tracking lives in #768.
- Post-Boole cross-client discovery compatibility is tracked separately in #1185 and PR #1187.
- Adjacent-domain handshake tolerance is reasonable transition-window hardening, but it is not a substitute for exact lifecycle timing and is outside this issue.
- Old and fixed Anchor versions disagree during the final pre-fork slot, so the fix should be deployed well before activation.
- SIP transition semantics: https://github.com/ssvlabs/SIPs/blob/0e84e19a7564c74f488d6aa6047996537fc93ef1/sips/network_topics_minhash.md#L125-L178

shane-moore:

The fork monitor never honored the transition slots it computed. `sleep_until_slot` subtracted a slot from every target and published immediately on waking, so `WarmUp`, `GracePeriod`, and `Normal` all landed one slot early. Measured on `unstable` in ssv-mini with Boole at epoch 6: transitions fired at slots **159 / 191 / 223** instead of 160 / 192 / 224.

Early `GracePeriod` is the one that matters on the wire: it flips ENR, handshake, discovery, and scoring to the new fork while messages for that slot are still Alan, since sender routing and validation derive the fork from the message slot and switch correctly at activation.

Three further defects sat behind the same code:

- **Clock failure was not fail-closed.** An unreadable clock made `sleep_until_slot` return instantly, so the loop published every remaining transition back to back regardless of the actual slot. A node restarted long before the fork would immediately behave as if Boole were live.
- **Long sleeps ignored wall-clock corrections.** A single pre-computed sleep runs against a monotonic deadline, so an NTP step during a multi-slot sleep left the lifecycle stale until the original timer expired.
- **Startup could permanently miss the ENR transition.** `Discovery::new` snapshots the lifecycle domain before awaiting discv5 startup and bootnode requests. A fork activating inside that window left a stale ENR that never self-corrected, since later lifecycle changes kept the same domain.

Exact activation also exposed a scoring dependency the early flip had been masking: Boole topics are subscribed rate-less during `WarmUp`, and at activation the fork set does not change, so no subscribe or rate event was emitted and Boole topics carried no score parameters (including the P4 invalid-message penalty) for the first 32 post-fork slots.

Closes #1191

## Change Overview (Required)

`ForkSchedule::lifecycle_at(slot, slots_per_epoch)` is now the single source of truth: the same slot always maps to the same lifecycle. The monitor task loops on a fresh clock observation each slot, derives the lifecycle, and publishes only when it changed. Exact-slot publication, publishing just the current state after a multi-boundary jump, and restart re-derivation all follow from that shape instead of from index bookkeeping.

**Reading order:** start with `schedule.rs` (`lifecycle_at`, the window definitions), then `monitor.rs` (`classify_observation` plus the ~40-line `run` loop). `network.rs` and `subscriptions.rs` are small consumer changes. `monitor.rs` is effectively a rewrite, so it reads far better as a file than as a diff; roughly 85% of the insertions are tests.

Consumer changes: the network reconciles the ENR domain during `try_new` and fails closed if a runtime ENR update fails (previously logged and never retried); the subnet service installs scoring parameters for the activating fork's already-subscribed topics at activation.

**What did not change:** the fork schedule math, window constants, `ForkLifecycle` itself, the watch-channel distribution model, per-transition operator logs, and all slot-derived sender routing and validation.

- **Fail-closed on clock failure is new behavior.** An unreadable clock or a backwards clock crossing a lifecycle boundary now requests a client shutdown. Published ENR, handshake, and scoring state cannot be rolled back, so the alternative is a silent split brain. Post-genesis, `SystemTimeSlotClock::now()` only returns `None` when the wall clock reads before genesis, so this is not a transient condition. Backwards movement *within* a window is tolerated with a warning.
- **The guards are process-scoped.** The highest observed slot is not persisted, so a supervisor restart with a still-rolled-back clock re-derives the earlier lifecycle without error. Documented in the module docs; watermark persistence is a filed follow-up.
- **The monitor no longer exits** after the final transition. Cost is one clock read and one pure computation per slot; measured below.
- **Deployment note:** during the final pre-fork slot, an old and a fixed node disagree about the lifecycle. This is the intended correction, and interop is governed by slot-derived routing on both sides.

## Validation (Required)

Unit: `fork` 44, `network` 63, `subnet_service` 65, plus the full release suite, `cargo fmt`, and `clippy -D warnings`.

Live validation on ssv-mini against go-ssv `stage` (874ba31), 2 Anchor + 2 go-ssv, 10 managed validators, Boole at epoch 6. Four runs, all clean-pass:

| Run | Test | Result |
|---|---|---|
| 1 | Clean control | All 6 transitions exact (2 nodes x 3), 1.4-11 ms into slot |
| 2 | Freeze across all boundaries | Node paused at slot 159, resumed 225: exactly one publish (`Normal{Boole}`), no replay, correct ENR, zero anomalies |
| 3 | Restart timing | Restart at 189 still flipped at exactly 192; restart at 200 derived `GracePeriod` and advertised Boole immediately |
| 4 | Base A/B | Pre-fix image reproduces the bug at 159 / 191 / 223 |

The A/B is the controlled comparison. The base image fired **4.9 s and 6.4 s** into its (wrong) slots and the two nodes drifted ~1.5 s apart from each other, because each slept a computed duration from its own start time. This branch landed **1.4-11 ms** into the correct slot with the two nodes agreeing within 40 microseconds. It also confirms the check discriminates rather than always passing.

Supporting: ENR flipped `0x00000000` to `0x00000001` at activation; zero `Failed to set topic score params` across the 128-topic burst; **zero handshake rejections on all four nodes in every run**; finality advanced and duties continued throughout.

Resource impact of the never-exiting monitor: post-transition CPU across candidate runs (0.74-3.73%) varies more than any candidate-versus-base difference (base 0.89%), and the base image's memory slope is *higher* (157 KB/slot vs 107-130). No spinning, no growth.

## Rollback (Required)

Revert the commit. No config, CLI, database, or wire-format changes. Reverting restores the one-slot-early behavior and the non-fail-closed clock handling.

Follow-ups filed, all in the Boole milestone and none of them fork blockers:

- #1230 detect slot-clock regression at startup. The fail-closed guards added here are process-scoped, so a restart with a still-rolled-back clock re-derives the earlier lifecycle. Filed as client-wide detection rather than a fork watermark, because a backwards clock breaks duties and attestations too, not just the fork lifecycle.
- #1231 tolerate both fork domains during the transition window, plus the 32 versus 34 slot `SUBSEQUENT_WINDOW_SLOTS` parity question with go-ssv.
- #1232 move the monitor out of `common/fork`, share the shutdown helper across its three call sites, and make an overlapping fork schedule a construction error.

The `db.changed()` arm is no longer deferred: it is fixed in this PR. It bound `changed()` with `_`, which matches `Err`, so a dropped database sender left the arm enabled and firing forever rather than disabling it. That is a different failure mode from the lifecycle arm's (a hot loop rather than a select-arm-disable panic), and it is pinned by a test verified to fail against the previous binding.

Not covered by live testing: the fail-closed paths themselves, which need a libfaketime image. They have unit coverage, and the runs did exercise the dangerous direction, since multi-minute forward jumps produced zero anomalies and no spurious shutdowns.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3303,6 +3303,7 @@ checksum = "00b0228411908ca8685dba7fc2cdd70ec9990a6e753e89b6ac91a84c40fbaf4b"
 name = "fork"
 version = "0.1.0"
 dependencies = [
+ "futures",
  "serde",
  "slot_clock",
  "ssv_types",
@@ -5403,6 +5404,7 @@ dependencies = [
  "ssz_types",
  "subnet_service",
  "task_executor",
+ "tempfile",
  "thiserror 2.0.18",
  "tokio",
  "tracing",
--- a/anchor/client/src/lib.rs
+++ b/anchor/client/src/lib.rs
@@ -448,7 +448,6 @@ impl Client {
             fork_schedule.clone(),
             slot_clock.clone(),
             E::slots_per_epoch(),
-            spec.get_slot_duration().as_secs(),
             executor.clone(),
         )?;
 
--- a/anchor/common/fork/Cargo.toml
+++ b/anchor/common/fork/Cargo.toml
@@ -5,6 +5,7 @@ authors = ["Sigma Prime <contact@sigmaprime.io>"]
 edition = { workspace = true }
 
 [dependencies]
+futures = { workspace = true }
 serde = { workspace = true }
 slot_clock = { workspace = true }
 ssv_types = { workspace = true }
--- a/anchor/common/fork/src/lifecycle.rs
+++ b/anchor/common/fork/src/lifecycle.rs
@@ -1,13 +1,13 @@
 //! Fork lifecycle state management.
 //!
 //! Provides [`ForkLifecycle`] for tracking the current fork transition state
-//! across all components via a `tokio::sync::watch` channel. The
-//! [`ForkMonitor`](crate::monitor) is the sole writer (sender); all other
-//! components hold receivers.
+//! across all components via a `tokio::sync::watch` channel. The fork monitor
+//! task ([`monitor::spawn`](crate::monitor::spawn)) is the sole writer
+//! (sender); all other components hold receivers.
 
 use crate::ForkConfig;
 
-/// Fork lifecycle state. Updated only by ForkMonitor.
+/// Fork lifecycle state. Updated only by the fork monitor task.
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub enum ForkLifecycle {
     /// Operating on a single fork. No transition in progress.
--- a/anchor/common/fork/src/monitor.rs
+++ b/anchor/common/fork/src/monitor.rs
@@ -1,292 +1,315 @@
 //! Fork transition monitoring.
 //!
-//! This module provides a standalone task that monitors and logs fork transitions,
-//! giving operators visibility into:
-//! - Current active fork at startup
-//! - Entering the preparation window before a fork
-//! - Fork activation when it occurs
+//! A standalone task derives the fork lifecycle from every fresh slot-clock
+//! observation via [`ForkSchedule::lifecycle_at`] and publishes changes on a
+//! watch channel. Because published state is always a pure function of the
+//! observed slot, transitions land at exact slot boundaries, a jump across
+//! several boundaries publishes only the state for the current slot, and a
+//! restart in any window re-derives the same state.
 //!
-//! The monitor pre-computes all fork transition points at creation time from the
-//! deterministic fork schedule. The `run()` method simply sleeps until each
-//! transition slot and sends the corresponding lifecycle update.
-//!
-//! The monitor exits automatically when all scheduled forks have activated.
+//! Sleeps are bounded to one slot so wall-clock corrections are noticed
+//! promptly. Clock failures fail closed by requesting a client-wide shutdown,
+//! as does a backwards clock that crosses a lifecycle boundary: published
+//! state (ENR, handshake, scoring) cannot be rolled back. Both guards live as
+//! long as the process but no longer: the highest observed slot is not
+//! persisted, so a restart with a still-rolled-back clock re-derives and
+//! re-advertises the earlier lifecycle without error.
 
 use std::{sync::Arc, time::Duration};
 
+use futures::channel::mpsc::Sender;
 use slot_clock::SlotClock;
-use task_executor::TaskExecutor;
+use task_executor::{ShutdownReason, TaskExecutor};
 use tokio::sync::watch;
-use tracing::{debug, error, info};
+use tracing::{error, info, warn};
 use types::{Epoch, Slot};
 
 use crate::{FORK_PREPARATION_EPOCHS, Fork, ForkLifecycle, ForkSchedule, SUBSEQUENT_WINDOW_SLOTS};
 
-/// A pre-computed fork transition at a specific slot.
-#[derive(Debug, Clone, PartialEq, Eq)]
-struct ScheduledTransition {
-    slot: Slot,
-    lifecycle: ForkLifecycle,
-}
-
-/// Pre-computed fork monitor with all transition points determined at creation.
+/// Classification of a fresh slot observation against the highest slot seen.
 ///
-/// The full transition timeline is deterministic from the fork schedule, so all
-/// `(slot, ForkLifecycle)` pairs are computed up front. The `run()` method becomes
-/// a simple sleep-and-send loop.
-pub(crate) struct ForkMonitor {
-    /// Lifecycle to send immediately
-    initial: ForkLifecycle,
-    /// Future transitions sorted by slot.
-    transitions: Vec<ScheduledTransition>,
+/// Pure decision logic for the monitor's backwards-clock policy: all effects
+/// (publishing, warning, shutdown) live in [`run`].
+#[derive(Debug, PartialEq, Eq)]
+enum ClockObservation {
+    /// The clock moved forward or held steady.
+    Advanced,
+    /// The clock moved backwards, but the lifecycle derived for the observed
+    /// slot matches the published one; tolerated with a warning.
+    BackwardsWithinWindow,
+    /// The clock moved backwards across a lifecycle boundary. Published state
+    /// cannot be rolled back, so the node must fail closed.
+    BackwardsAcrossBoundary,
 }
 
-impl ForkMonitor {
-    /// Create a new fork monitor by pre-computing all transition points.
-    ///
-    /// For each non-genesis fork (epoch > 0), three transitions are generated:
-    /// - `WarmUp` at the preparation window start (`fork_epoch - 1` epochs)
-    /// - `GracePeriod` at fork activation (`fork_epoch`)
-    /// - `Normal` after the grace period (`fork_epoch + SUBSEQUENT_WINDOW_SLOTS` slots)
-    ///
-    /// All transitions are sorted by slot, then split at `current_slot`:
-    /// - The last transition at or before `current_slot` becomes the initial state.
-    /// - Remaining future transitions are stored for the run loop.
-    ///
-    /// This handles all restart scenarios uniformly — whether the node starts
-    /// before any fork, during a warm-up window, during a grace period, or
-    /// after all forks have activated.
-    fn new(schedule: &ForkSchedule, current_slot: Slot, slots_per_epoch: u64) -> Self {
-        let current_epoch = current_slot.epoch(slots_per_epoch);
-        let current_fork_config = schedule.active_fork_config(current_epoch);
-        info!(fork = %current_fork_config.fork, epoch = %current_epoch, "Fork monitor started");
+/// Classify `now` against the highest slot observed so far.
+fn classify_observation(
+    schedule: &ForkSchedule,
+    slots_per_epoch: u64,
+    highest_observed_slot: Slot,
+    now: Slot,
+) -> ClockObservation {
+    if now >= highest_observed_slot {
+        ClockObservation::Advanced
+    } else if schedule.lifecycle_at(now, slots_per_epoch)
+        == schedule.lifecycle_at(highest_observed_slot, slots_per_epoch)
+    {
+        ClockObservation::BackwardsWithinWindow
+    } else {
+        ClockObservation::BackwardsAcrossBoundary
+    }
+}
 
-        // Step 1: Generate all transitions for every non-genesis fork.
-        //
-        // Each fork produces three transitions at deterministic slots:
-        //   prep_slot ──────> activation ──────> grace_end
-        //   [WarmUp]          [GracePeriod]      [Normal]
-        let mut all_transitions: Vec<ScheduledTransition> = Vec::new();
-
-        for &fork in Fork::all() {
-            let Some(fork_config) = schedule.config(fork) else {
-                continue;
-            };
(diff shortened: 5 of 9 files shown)


shane-moore:
@claude please take a look

shane-moore:
@claude the follow-ups you flagged are now filed, all in the Boole milestone:

- **#1232** covers both the duplicated shutdown idiom and the log-only `error_on_overlapping_windows`, bundled with the monitor's crate placement because the helper's home depends on where the monitor lives.
- **#1230** covers the process-scoped rollback watermark, filed as client-wide clock-regression detection rather than a fork watermark. A backwards clock also breaks duty scheduling and attestation targets, so a fork-only guard would leave the node broken elsewhere while appearing covered.
- **#1231** covers go-ssv transition parity (handshake domain tolerance and the 32 versus 34 slot window), from the same review pass rather than yours.

Two commits landed after your review, both in `subnet_service`, so the file you looked at has changed. The `db.changed()` arm bound `changed()` with `_`, which matches `Err`, so a dropped database sender left the arm enabled and firing forever rather than disabling it. That is a different failure mode from the lifecycle arm you reviewed, a hot loop rather than a select-arm-disable panic, and it is pinned by a test verified to fail against the previous binding.

petarjuki7 (approved):
LGTM! Went through the window math, the run loop, the backwards-clock handling and the ENR/scoring changes. All checks out, and the risks are well documented in the description.

Only note: the fail-closed shutdown can also trigger a few minutes after each transition if the clock steps back across the boundary, not just at the transition slot.

All in all looks good!

PR #1229, opened 2026-08-08
fix(validator_store): sign the decided AggregatorCommittee worklist once per slot

### LINKED ISSUE #1227: Boole AggregatorCommittee can omit decided roots from post-consensus batches
## Description

At Boole and later, one `AggregatorCommitteeConsensusData` decision can contain both aggregate-and-proof and contribution-and-proof objects.

Anchor computes the expected local committee batch size from the complete decided value. However, it creates signing requests only from the local Lighthouse callback vector for the current object class.

If a decided object is absent from that callback vector, Anchor counts the object but never adds its local partial signature. The committee batch can remain below its expected size and never broadcast.

This was observed in a mixed-client no-fault run and is reproduced by a focused validator-store regression.

## Version

Deterministic local reproduction:

- Anchor `unstable`: `82bac14ec635bc28a68a6bba9799bd603ec0c082`
- Rust: `rustc 1.91.0 (f8297e351 2025-10-28)`
- Platform: macOS arm64

Original mixed-client observation:

- Anchor: `fd3c7f9677b2bbb04fe7d7c2d4eca81313db7b79`
- Embedded Lighthouse diagnostic revision: `b263df596671a2bd42bf1034e1cdc8188ba8a9b0`
- go-ssv: `874ba316266aec6215bc21c6e2e929e195f08d6b`
- ssv-mini: `fef3b7d791384911078b2ce977100f87679e5c46`
- Topology: two Anchor operators plus two go-ssv operators in one 3-of-4 committee
- Fork: Boole at slot 192
- Validators: 74 active validators

Current remote heads were also checked:

- Anchor `unstable`: `82bac14ec635bc28a68a6bba9799bd603ec0c082`
- Anchor `epbs`: `36950b69f6f8e6ca362b46edd608b6b1037793f7`
- go-ssv `stage`: `a9008f3d0542f83291f2e9d18babf3edd3abd7b1`

The current ePBS branch has the same callback-vector outer loops. PR #1128 changes proposal inputs, but it does not remove this post-consensus worklist gap.

### Mixed-client evidence

In the no-fault run at slot 230, the decided AggregatorCommittee value contained three post-consensus objects:

- One aggregate for validator 68.
- Two sync contributions for validator 65.

go-ssv operators 3 and 4 each signed all three objects. Anchor operators 1 and 2 each signed only the two validator-65 contributions.

The validator-68 aggregate received only two shares. Both go-ssv runners classified the combined AggregatorCommittee duty as incomplete at slot end.

Ordinary attestations remained included, both beacon nodes stayed on the same chain, and finality advanced. The failure was isolated to AggregatorCommittee post-consensus completeness.

Relevant Anchor excerpt from operator 1:

```text
requester=Committee { validator_partial_signature_batch_size: 3,
  base_hash: 0x08750ba150bc48eafb9ed326bf49073d82a0e18671dd79eb80de197b0cb09c03 }
root=0x6ef13e08ec0ba710a73f3966dbad3fcae9083f2b853e016b1b35978371d85797
index=ValidatorIndex(65)

requester=Committee { validator_partial_signature_batch_size: 3,
  base_hash: 0x08750ba150bc48eafb9ed326bf49073d82a0e18671dd79eb80de197b0cb09c03 }
root=0x22b763e9a915bcdb2351accba31fef28e5f141795be2576805293440d7a09095
index=ValidatorIndex(65)

# No local sign_and_collect call for validator 68 / root:
0x7c4e5f2ea1031c52ee2068e3eb66adfd98d5a64cb99a0309553bbcd8f1b47765
```

The same Anchor log received the validator-68 root from go-ssv operators 3 and 4. It never produced a local share for that root.

Relevant go-ssv excerpt:

```text
got post-consensus message signer=3 validators=[68,65,65]
got post-consensus message signer=4 validators=[68,65,65]
duty did not complete before slot end (likely stuck)
```

Both Anchor nodes showed the same omission.

The preserved local report SHA-256 is:

```text
9d6351de4b183f413faf7580ee3ac39ec5a861cda4e743da1bb7c814150451a9
```

### Deterministic regression

A focused test uses one SSV committee with two locally owned validators:

- Validator 0 owns one decided aggregate.
- Validator 1 owns two decided contributions for subcommittees 0 and 1.
- Only the contribution callback is invoked.
- The signature collector mock records each `SignatureRequester` and signing root.

Expected: three signing calls with three unique roots. Every request must use batch size 3 and the same decided-data hash.

Actual: two signing calls. Both requests claim that the local batch size is 3.

Command:

```bash
cargo test -p anchor_validator_store contribution_callback_requests_complete_mixed_decided_batch -- --nocapture
```

Output on current `unstable`:

```text
running 1 test

thread testing::committee_post_consensus::contribution_callback_requests_complete_mixed_decided_batch panicked at
anchor/validator_store/src/testing/committee_post_consensus.rs:75:5:
assertion `left == right` failed: the decided batch has one aggregate root and two contribution roots
  left: 2
 right: 3

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 53 filtered out
```

The production signature collector does not send a committee batch until the local count equals the requested count. Therefore, a two-of-three local batch remains buffered:

https://github.com/sigp/anchor/blob/82bac14ec635bc28a68a6bba9799bd603ec0c082/anchor/signature_collector/src/lib.rs#L295-L344

### Source-level mechanism

The contribution callback iterates only its local `contributions` vector, but computes the expected size from all locally owned decided aggregators and contributors:

https://github.com/sigp/anchor/blob/82bac14ec635bc28a68a6bba9799bd603ec0c082/anchor/validator_store/src/lib.rs#L1372-L1495

The aggregate callback has the same asymmetry. It iterates only its local `aggregates` vector, but uses the complete decided-data count:

https://github.com/sigp/anchor/blob/82bac14ec635bc28a68a6bba9799bd603ec0c082/anchor/validator_store/src/lib.rs#L1830-L1886

go-ssv handles this differently. `AggregatorCommitteeRunner.ProcessConsensus` runs once for the decision, enumerates both object classes from the decided value, filters them by local share and duty, then sends one post-consensus message:

https://github.com/ssvlabs/ssv/blob/a9008f3d0542f83291f2e9d18babf3edd3abd7b1/protocol/v2/ssv/runner/aggregator_committee.go#L696-L824

## Expected Behaviour

After AggregatorCommittee QBFT decides, Anchor must derive the complete local post-consensus signing worklist from the decided value.

Any non-empty aggregate or contribution callback for the slot and committee must be able to trigger every locally signable decided root. Callback order must not determine which local partial signatures are emitted.

The aggregate callback must still return only requested aggregates. The contribution callback must still return only requested contributions.

Concurrent or later callbacks must reuse the same execution. They must not sign or broadcast the complete decision twice.

## Steps to resolve

A likely minimal direction is:

1. Build one heterogeneous worklist from all decided aggregators and contributors.
2. Resolve local validator metadata, usable shares, and cluster state for each decided entry.
3. Preserve one entry per `(validator_index, signing_root)`. Validator index alone is not unique because one validator can have several contribution roots.
4. Sign the complete local worklist with one `CollectionMode::Committee` batch.
5. Use `base_hash = decided_data.hash()` and `validator_partial_signature_batch_size = prepared.len()`.
6. Make the execution once-only and cancellation-safe for `(slot, committee_id, decided_hash)`.
7. Let both callback classes join the same result, then filter returned objects by callback class and requested identity.
8. Reject a conflicting decided hash for the same slot and committee without starting a second batch.

For ePBS compatibility, isolate this coordinator in `AnchorValidatorStore` or a small validator-store module. Avoid coupling the cache to `AggregationAssignments` or the ePBS metadata and `SlotVote` changes.

The aggregate decoder must use the decided data version. The ePBS branch already has `DataVersion::decode_attestation`, which supports the Gloas shape:

https://github.com/sigp/anchor/blob/36950b69f6f8e6ca362b46edd608b6b1037793f7/anchor/validator_store/src/lib.rs#L1786-L1820

Changing the expected batch size from 3 to 2 is not a valid fix. That would let callback order produce incomplete and inconsistent local committee batches.

## Acceptance criteria

- A contribution-only callback signs one decided aggregate plus all decided contributions owned by the operator.
- The mirror aggregate-only case also emits all locally signable decided roots.
- Concurrent aggregate and contribution callbacks emit one local committee message.
- A validator with several contribution subcommittees retains every distinct root.
- Foreign, missing, or unusable local entries are excluded from the expected local count.
- Dropping the first callback future does not abandon or duplicate local partial emission.
- A late callback reuses cached outcomes without another signature or network send.
- One failed root does not erase successful outcomes or restart already emitted partials.
- Existing pre-Boole behavior remains unchanged.
- The same design ports to ePBS without importing unrelated metadata, `SlotVote`, PTC, or Gloas work.
- A mixed two-Anchor plus two-go-ssv replay reconstructs the previously missing aggregate.

## Scope boundary

This fix can start when at least one Lighthouse callback class is non-empty. It does not independently schedule or publish decided objects when both callback vectors are empty.

Full callback-independent scheduling and direct publication remain a separate architectural follow-up.

Related test-coverage issue: #929.

shane-moore:

At Boole, one QBFT decision can carry both attestation aggregates and sync contributions. Anchor
signed these from two separate Lighthouse callbacks. Each callback signed only its own object type.
But both sized the shared partial-signature batch from the full decided value.

If the decided value held an object type that the local Lighthouse never asked for, the batch never
filled. The operator then never sent its one post-consensus message. The duty failed for the whole
cluster, not just for that operator.

We first saw this on a mixed Anchor and go-ssv network. On ssv-mini it breaks about 1.2% of
post-fork `AggregatorCommittee` duties, so it is a steady loss on any mixed network.

Closes #1227.

## Change Overview (Required)

Anchor now builds the signing list from the decided value instead of from what Lighthouse asked
for. Local state (share, cluster status, metadata) only filters that list.

The slot pipeline starts one signing job per committee when it publishes the decided value. This is
the only place that starts a job, so one committee sends exactly one message per slot. The two
callbacks now only read the result.

The batch size is the number of entries Anchor actually submits, so local filtering can always
reach it.

Read `aggregator_post_consensus.rs` first, then the trigger in `update_aggregation_assignments`,
then the two callbacks.

The signature collector is untouched, and `collect_signature` is byte-identical to before, so
pre-Boole paths cannot change.

Two behavior changes to be aware of:

- QBFT for `AggregatorCommittee` now starts at 2/3 slot from the pipeline. Before, it started when
  the first callback arrived.
- A per-root timeout now counts as `OTHER_ERROR` instead of `TIMEOUT`. This matches the existing
  BeaconVote path, and it is visible on dashboards.

Blast radius is `validator_store` plus one deleted helper in `ssv_types`.

Anchor signs every decided entry that it holds a share for. go-ssv also filters by local duty and
signs fewer. Signing extra roots is safe here, because peers do not check completeness for this
duty, and one validator stays inside the 5-root cap.

## Validation (Required)

66 unit tests, plus `make lint`, `make cargo-fmt-check`, and a full release `make test`, all clean.

Live A/B on ssv-mini with 2 Anchor and 2 go-ssv operators at a 3-of-4 threshold. Everything was
identical between the two arms except the Anchor build.

| | Old code | This PR |
|---|---|---|
| Post-fork `AggregatorCommittee` duties | 667 | 1996 |
| Failed duties | **8** | **0** |

Every one of the 8 failures looks the same. Only the two go-ssv operators sent a post-consensus
message. Both Anchor operators sent nothing, which left 2 of 4 partials against a threshold of 3.
On this branch that never happened once.

We also restarted an operator after the fork, because this PR keeps the signing jobs in memory. The
node misses 4 slots and then recovers on its own, with no stuck batch.

Revert the commits. There are no config, CLI, database, or wire-format changes, and nothing new is
written to disk.

None.

### DIFF
--- a/anchor/common/ssv_types/src/consensus.rs
+++ b/anchor/common/ssv_types/src/consensus.rs
@@ -526,34 +526,6 @@ pub struct AggregatorCommitteeConsensusData<E: EthSpec> {
         VariableList<SyncCommitteeContribution<E>, MaxSyncContributions>,
 }
 
-impl<E: EthSpec> AggregatorCommitteeConsensusData<E> {
-    /// Counts the total number of expected post-consensus signatures for this committee.
-    ///
-    /// This includes both aggregators and contributors that match the given filter.
-    /// Used to ensure all signatures are batched into a single message rather than
-    /// being split across multiple messages.
-    ///
-    /// Returns: count(aggregators) + count(contributors) filtered by committee membership
-    pub fn post_consensus_signature_count<F>(&self, is_in_committee: F) -> usize
-    where
-        F: Fn(&ValidatorIndex) -> bool,
-    {
-        let aggregator_count = self
-            .aggregators
-            .iter()
-            .filter(|agg| is_in_committee(&agg.validator_index))
-            .count();
-
-        let contributor_count = self
-            .contributors
-            .iter()
-            .filter(|contrib| is_in_committee(&contrib.validator_index))
-            .count();
-
-        aggregator_count + contributor_count
-    }
-}
-
 impl<E: EthSpec> QbftData for AggregatorCommitteeConsensusData<E> {
     type Hash = Hash256;
 
--- a/anchor/validator_store/src/aggregator_post_consensus.rs
+++ b/anchor/validator_store/src/aggregator_post_consensus.rs
@@ -1,0 +1,542 @@
+//! Post-consensus signing for Boole+ `AggregatorCommittee` duties.
+//!
+//! One QBFT decision carries both attestation aggregates and sync contributions for an SSV
+//! committee, and the operator must emit exactly one committee partial-signature message per
+//! `(committee, slot)` covering everything it can sign. Lighthouse asks for signatures through two
+//! independent callbacks, one per object class, so neither callback can own that message: whichever
+//! one fires, the other class still has to be signed (issue #1227).
+//!
+//! The signing set is therefore a property of the duty, not of any callback. The slot pipeline
+//! starts one execution per committee when it publishes the decided value at 2/3 slot, and the
+//! callbacks only look up their own results. This is the ssv-spec-normative shape, where the
+//! decided value drives what gets signed and local state only filters.
+
+use std::{
+    collections::{HashMap, HashSet, hash_map::Entry},
+    future::Future,
+    sync::Arc,
+};
+
+use bls::Signature;
+use database::{NonUniqueIndex, UniqueIndex};
+use futures::{
+    FutureExt,
+    future::{BoxFuture, Shared},
+};
+use qbft_manager::ConsensusDecider;
+use slot_clock::SlotClock;
+use ssv_types::{
+    Cluster, CommitteeId, ValidatorIndex, ValidatorMetadata,
+    consensus::{AggregatorCommitteeConsensusData, AssignedAggregator, DataVersion, QbftData},
+    msgid::Role,
+    partial_sig::PartialSignatureKind,
+};
+use ssz::Decode;
+use tokio::time::Instant;
+use tracing::{debug, error, warn};
+use types::{
+    AggregateAndProof, Attestation, AttestationBase, AttestationElectra, ContributionAndProof,
+    Domain, EthSpec, ForkName, Hash256, SelectionProof, SignedRoot, Slot,
+};
+
+use crate::{
+    AggregationAssignments, AnchorValidatorStore, CollectionMode, Error, SigningRequest,
+    SpecificError,
+};
+
+/// Epochs to retain finished `AggregatorCommittee` post-consensus executions.
+///
+/// Matches the message validator's per-signer duplicate-detection window, so a very late Lighthouse
+/// callback finds the execution it belongs to rather than nothing. Each entry is one small future
+/// per committee-slot.
+const AGGREGATOR_POST_CONSENSUS_RETAIN_EPOCHS: u64 = 2;
+
+/// A result shared between the detached task producing it and any number of joiners.
+type SharedResult<T> = Shared<BoxFuture<'static, Result<T, Arc<Error>>>>;
+
+pub(crate) type AggregatorPostConsensusShared<E> =
+    SharedResult<Arc<AggregatorPostConsensusOutcome<E>>>;
+
+/// One decided root this operator signs, with the handle to its signature.
+///
+/// The underlying `collect_signature` call runs in its own task, so the local partial reaches the
+/// committee batch even when no callback ever awaits the signature.
+pub(crate) struct PreparedRoot<M> {
+    pub(crate) request: SigningRequest<M>,
+    pub(crate) signature: SharedResult<Signature>,
+}
+
+/// Everything this operator signs for one decided `AggregatorCommittee` round, keyed by the
+/// identities the Lighthouse callbacks look up: validator index for aggregates,
+/// `(validator index, subcommittee index)` for contributions.
+pub(crate) struct AggregatorPostConsensusOutcome<E: EthSpec> {
+    pub(crate) aggregates: HashMap<ValidatorIndex, PreparedRoot<AggregateAndProof<E>>>,
+    pub(crate) contributions: HashMap<(ValidatorIndex, u64), PreparedRoot<ContributionAndProof<E>>>,
+}
+
+/// The signing worklist derived from one decided `AggregatorCommitteeConsensusData`.
+struct DecidedWorklist<E: EthSpec> {
+    aggregates: HashMap<ValidatorIndex, SigningRequest<AggregateAndProof<E>>>,
+    contributions: HashMap<(ValidatorIndex, u64), SigningRequest<ContributionAndProof<E>>>,
+}
+
+fn post_consensus_aborted() -> Arc<Error> {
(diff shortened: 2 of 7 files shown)


shane-moore:
@claude, please take a look

shane-moore:
@claude, both points addressed in 09b02103.

**1. Idempotent registration.** Fixed. I traced the Phase 3 loop (\`metadata_service.rs:334-352\`): it sleeps \`duration_to_next_slot + 2/3 slot\` each iteration, so a same-slot double publish is not reachable today, and an overrun skips a slot rather than repeating one. But the objection holds regardless of reachability: `insert` moved the exactly-once guarantee out of this module and into another module's timing loop, where a future re-publish path would break it silently and expensively. Registration is now vacant-entry-only.

Worth noting this was a regression I introduced: the earlier callback-triggered version of this code used `or_insert_with`, and the property was lost when the trigger moved to the slot pipeline.

**2. Stale field doc.** Fixed, it now describes the pipeline-triggered design and states that callbacks only read the map.

**On the missing test coverage**, which was the most useful part of the review: added `republished_assignments_do_not_resubmit`. I verified it actually catches the regression rather than merely passing, by reverting to `insert` and confirming it fails, then restoring. 66 tests pass, fmt and lint clean.

petarjuki7 (approved):
Looks good!

PR #1234, opened 2026-08-10
fix(fork): reject duplicate domains in explicit schedules

shane-moore:
## Summary

- Reject explicit fork schedules that reuse a signing domain.
- Report the shared domain and both conflicting forks.
- Guard domain uniqueness across every scheduled fork in the built-in network matrix.
- Clarify that `ForkSchedule::new` is for Alan-only defaults and test schedules.

Duplicate domains make forks or networks cryptographically indistinguishable. Anchor now rejects per-schedule collisions at startup and catches built-in cross-network collisions in tests.

## Testing

- `cargo test -p fork -p ssv_network_config -p global_config`
- `make cargo-fmt-check`
- `make lint`

### DIFF
--- a/anchor/common/fork/src/schedule.rs
+++ b/anchor/common/fork/src/schedule.rs
@@ -96,7 +96,11 @@ pub struct ForkSchedule {
 }
 
 impl ForkSchedule {
-    /// Create a new fork schedule with the given fork active from epoch 0.
+    /// Create a fork schedule with every fork up to `fork` active from epoch 0.
+    ///
+    /// Every included fork uses `baseline_domain_type`. This constructor is intended for
+    /// Alan-only network defaults and test schedules. Explicit production fork schedules should
+    /// use [`Self::from_fork_configs`] so ordering and domain uniqueness are validated.
     pub fn new(fork: Fork, baseline_domain_type: DomainType, network_name: &str) -> Self {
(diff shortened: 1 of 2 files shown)


shane-moore:
pr 1234 lol

shane-moore:
@claude, please take a look

shane-moore:
@claude One correction: `ForkSchedule::new(Fork::Boole, ...)` creates Alan and Boole at epoch zero with the same domain, rather than a single-fork schedule. Its only production caller uses `Fork::Alan`; Boole callers are tests. I clarified that contract and added a built-in cross-network domain-uniqueness test in `1013d73e`.

shane-moore (commented):
PR reviewed by gpt-5.6-sol max.

No blocking correctness or safety findings. The change is ship-shaped.

Non-blocking feedback:

1. ForkSchedule::new can still construct a duplicate-domain multi-fork schedule while from_fork_configs enforces uniqueness. Its only production caller is Alan-only, so this is a design hardening point rather than a current runtime regression.

2. test_builtin_network_fork_domains_are_globally_unique hardcodes mainnet, holesky, and hoodi separately from SsvNetworkConfig::constant. A future built-in network could therefore be omitted from the guard.

All current non-queue checks pass. The branch is four commits behind unstable, so synthetic queue CI remains the final integration check.

Issue #1223, opened 2026-08-07
Lighthouse proposer scheduler can skip an assigned slot after an early timer wake


shane-moore:
## Description

Anchor can skip a locally assigned proposal when the pinned Lighthouse proposer-duty timer wakes at a slot boundary but the Lighthouse slot clock still returns the previous slot.

This was observed on one Anchor operator during a mixed 2 Anchor plus 2 go-ssv stress run. The affected operator processed slot 78 again, sent a stale slot-78 notification, and then slept until the next boundary without ever checking slot 79. The other three operators reached the 3-of-4 threshold and produced canonical block 79.

The bug is most likely owned by Lighthouse `validator_services`. It is tracked here because Anchor pins and embeds that code, needs a dependency integration path, and can ship a temporary patched Lighthouse revision if required. The observation occurred before the Boole activation in this run, so the failure is not Boole-specific.

## Version

- Anchor `unstable`: `fd3c7f9677b2bbb04fe7d7c2d4eca81313db7b79`
- Anchor Lighthouse pin: `b263df596671a2bd42bf1034e1cdc8188ba8a9b0`
- Lighthouse current `unstable` checked on 2026-08-06: `02e8732a81a68784ce105609942570470a7989f2`, with the same scheduler behavior still present
- ssv-mini: `fef3b7d791384911078b2ce977100f87679e5c46`
- go-ssv `stage`: `874ba316266aec6215bc21c6e2e929e195f08d6b`
- Project Docker Rust toolchain: `1.93.1`
- Runtime image: `node/anchor:boole-proposer-diagnostic-fd3c7f9-b263df5`
- Image ID: `sha256:5a459c08cf1d8fc3d6a4b06f76249c6273f07c1d782f2be5a38d6c288a2aa7ee`
- Anchor binary SHA-256: `41c7addc3ef97b8be130e0e34690b43c05fa03714114209c9ed1c108d13303d3`

Anchor's pin is visible in [Cargo.toml](https://github.com/sigp/anchor/blob/fd3c7f9677b2bbb04fe7d7c2d4eca81313db7b79/Cargo.toml#L77-L92).

## Test topology

- 2 Anchor operators, operator IDs 1 and 2
- 2 go-ssv operators, operator IDs 3 and 4
- 10 managed validators, indices 64 through 73
- 3-of-4 threshold
- 74 total preregistered validators
- Fulu at genesis
- Boole at epoch 6
- 12-second slots
[...]

PR #1235, opened 2026-08-10
fix(validator_store): publish VotingAssignments 50ms past slot boundary

shane-moore:

- `MetadataService` Phase 1 sleeps to the slot boundary and then reads the wall clock. A timer wake a few milliseconds early reads the previous slot, republishes it, and the next iteration sleeps past the new slot. Waiters fail with `MetadataSlotPassed`, and the Phase 2/3 cascade costs the operator three slots of committee participation.
- Observed on a 512-slot scale-100 devnet soak: 6 skipped slots, 3 of them quorum-losing across a 4-operator committee, with wakes measured 1 to 5 ms before the boundary.
- Addresses the Anchor side of #1223. The proposer-scheduler half of that issue lives in Lighthouse and is tracked separately.

## Change Overview

- Phase 1 now sleeps `duration_to_next_slot() + 50ms`, the same idiom Phase 3 already uses (its 2/3-slot offset makes it immune to this race) and the same shape Lighthouse uses for slot-keyed work. 50 ms mirrors the message validator's `CLOCK_ERROR_TOLERANCE`, the inter-node clock error the network already budgets for, and is 10-50x the measured drift.
- The loop is extracted into a small `run_slot_start_publisher` helper so the race is unit-testable.
- Reading order: the `VOTING_ASSIGNMENTS_PUBLISH_DELAY` doc comment, then the helper, then the two tests.
- Intentionally unchanged: `update_voting_assignments` itself, the waiter contract, and Phase 2/3 scheduling.

- Assignments publish 50 ms later. Every consumer was traced: the selection-proof flows hold an absolute 2/3-slot deadline and the voting-context build triggers no earlier than a head event. Nothing reads the data inside the first 50 ms; block proposals never read it at all.
- A backwards clock step larger than 50 ms during one sleep produces a duplicate publication of the previous slot. Harmless under the waiter contract, which only fails a waiter when a newer slot arrives, and it self-corrects at the next boundary. All other clock-fault behavior is unchanged from the previous code.

## Validation

- New deterministic regression test drives the early wake under paused time by holding the wall clock short of the boundary when the timer fires; it fails with the offset removed (mutation-checked) and passes with it.
- Steady-state test asserts one publication per slot, strictly increasing.
- Full `make test` (including ssv-spec tests): 622 passed, 0 failed. fmt and clippy clean.
- Validated on a 4-operator, 100-validator ssv-mini devnet (same harness class that reproduced the race): in a 128-slot scored window under Boole, every node published exactly once per slot with strictly increasing slots, zero `MetadataSlotPassed`, and 120 of 123 produced blocks carried sync participation of 500+/512 on a finalizing chain. The two slots without publications coincided with a measured 34-second host-VM stall (all containers silent), not publisher behavior.

## Rollback

- Revert the single commit. No config, database, or wire-format impact.

### DIFF
--- a/anchor/validator_store/src/metadata_service.rs
+++ b/anchor/validator_store/src/metadata_service.rs
@@ -63,6 +63,19 @@ const WAD_SOFT_TIMEOUT: Duration = Duration::from_secs(1);
 const WAD_HARD_TIMEOUT: Duration = Duration::from_secs(3);
 const BLOCK_SLOT_LOOKUP_TIMEOUT: Duration = Duration::from_millis(500);
 
+/// Publish `VotingAssignments` this long after the slot boundary.
+///
+/// The boundary sleep runs on the monotonic timer but the slot is read from the
+/// wall clock, and the two drift apart by a few milliseconds over a full-slot
+/// sleep. A wake landing marginally before the boundary would read the previous
+/// slot, republish it, and skip the new one (issue #1223). This margin absorbs
+/// that drift. 50 ms mirrors the message validator's `CLOCK_ERROR_TOLERANCE`,
+/// the clock error the SSV network already budgets for between nodes. Nothing
+/// consumes the assignments this early: the soonest consumers are the
+/// selection-proof flows (deadline 2/3 slot) and the voting-context build
+/// (triggered no earlier than a head event).
+const VOTING_ASSIGNMENTS_PUBLISH_DELAY: Duration = Duration::from_millis(50);
+
 /// Builds each validator's per-subnet position counts from raw sync committee positions.
 fn build_sync_validator_assignments<E, I, P>(
     duties: I,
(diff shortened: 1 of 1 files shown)


shane-moore:
@claude, please take a look

shane-moore:
@claude On the error-branch clock source: the two durations are equal by construction. The production clock is built at `client/src/lib.rs:338` as `SystemTimeSlotClock::new(spec.genesis_slot, genesis_time, spec.get_slot_duration())`, its only construction site, so `slot_clock.slot_duration()` and the spec value cannot diverge without that line changing. The helper reads the clock so it stays self-contained.

shane-moore:
**ssv-mini validation results**

Setup: 4 Anchor operators, 100 managed validators on a full 4-operator committee, Boole active, 12s slots, image built from 42bf00b1.

Scored window of 128 slots (736-863):
- Every node published `VotingAssignments` exactly once per slot, slots strictly increasing, zero `MetadataSlotPassed` in the window.
- The #1223 failure signature (previous-slot republish followed by a skipped slot) did not occur anywhere in the run.
- Downstream health on the path this issue was breaking: 123/128 slots had blocks and 120 of those carried sync participation of 500+/512, with finality advancing throughout.
- The two slots without publications coincided with a measured 34-second host-VM stall (all containers silent), an environment artifact rather than publisher behavior. On resume the publisher picked up at the current slot as designed.

shane-moore:
The 50 ms offset remains necessary for the `VotingAssignments` early-wake failure. Merged #1245 resolves the separate Boole aggregate-publication snapshot race by publishing from Anchor, so it does not supersede this fix.

Review generated by `gpt-5.6-sol` with max reasoning.

Issue #1240, opened 2026-08-11
test(duties_tracker): cover poll_beacon_proposers fetch path via a beacon-node seam


petarjuki7:
Deferred from #1183.

`ProposerSchedule::from_response` is fully unit-tested, but nothing drives `poll_beacon_proposers` itself: the crate has no way to fake a beacon-node response (`BeaconNodeFallback` is constructed over a real HTTP client).

Untested as a result:
- the fetch uses `get_validator_duties_proposer_v2` (not v1)
- a failed fetch leaves the previously stored schedule in place
- a malformed response is rejected at the poll site and leaves the previously stored schedule in place

Needs either a mock HTTP server for the fallback client or a fetch abstraction the tests can substitute. Same shape as #1238.

Issue #1184, opened 2026-07-28
fix(duties_tracker): authoritative complete-schedule transitions for proposer assignment verdicts


shane-moore:
**Blocked by:** #1183

## Goal

Make proposer-assignment verdicts derive only from retained complete proposer schedules, and expose an atomic view-install notification that Anchor's optional Early RANDAO retention path (#1178) and late-discovered-duty producer scheduling (#1179) can consume.

This replaces the earlier provisional and cause-specific evidence model. SIP-101 defines complete-schedule authority, not optimistic, root-stale, disputed, or retention-eligibility states.

## Protocol classification

Under [SIP-101 at 8279853](https://github.com/ssvlabs/SIPs/blob/82798537fea271d1a31d02fdc000d87b37fa3d98/sips/early_randao_preconsensus.md#L86-L108), an epoch has two semantic view states:

- **Unknown:** Anchor retains no complete proposer schedule for the epoch.
- **Known:** Anchor retains a schedule containing exactly `SLOTS_PER_EPOCH` duties, one for every distinct slot in the epoch. It is authoritative for every validator, independent of the local SSV registry.

Within a Known view, an exact validator and slot match is `DutyAssignment::Assigned`; every other validator and slot query is `DutyAssignment::NotAssigned`. Only `NotAssigned` is authoritative negative evidence. A failed or malformed refresh, local validator-set change, `execution_optimistic`, or detected reorg does not revoke a retained complete view. A newly retained complete schedule atomically replaces it.

The notification watch is not a SIP wire requirement. It is an Anchor requirement because Anchor chooses to implement optional retention in #1178 and the optional immediate attempt for a duty first discovered after the recommended producer time in #1179.

## Current code

At [`upstream/epbs` 7bad23f](https://github.com/sigp/anchor/tree/7bad23f2b37f08fb886b7d33be05963bf879b112), [`proposer_assignment_at_slot`](https://github.com/sigp/anchor/blob/7bad23f2b37f08fb886b7d33be05963bf879b112/anchor/duties_tracker/src/duties_tracker.rs#L343-L361) returns `Unknown` when the epoch has no retained entry, `Assigned` for an exact pubkey and slot match, and `NotAssigned` otherwise. #1183 makes the retained entry a shape-validated complete response. This issue makes installation and transition semantics explicit and supplies the watch needed by #1178 and #1179.

## Suggested approach

- Store at most one validated complete schedule per epoch. Use #1183's validated response type.
[...]

PR #1241, opened 2026-08-11
refactor(duties_tracker): retain proposer-duty v2 response metadata

### LINKED ISSUE #1183: refactor(duties_tracker): retain proposer-duty v2 response metadata

petarjuki7:

- The tracker stored whatever proposer-duty list the beacon node returned, with no completeness check. A partial or empty response became the view for its epoch, and `proposer_assignment_at_slot` then answered `NotAssigned` for validators the response simply failed to include. SIP-101 allows a schedule to establish a view only when it is complete: one duty for every slot of the epoch.
- The response metadata was discarded. The v2 endpoint's `dependent_root` is the fork-aware decision root (v1 returns the legacy one), which later refresh logic needs.
- Closes #1183. Builds on #1182.

## Change Overview

- New `ProposerSchedule` type with private fields, built only through `from_response`, which accepts a response only if it has exactly `SLOTS_PER_EPOCH` duties, all inside the requested epoch, each slot exactly once. Anything else is rejected with a typed error naming the defect.
- The poll switches to `get_validator_duties_proposer_v2` and stores the validated schedule together with `dependent_root` and `execution_optimistic`. Neither field affects whether a schedule is accepted; they are retained for refresh decisions and diagnostics (#1184).
- A failed or malformed refresh leaves the previously stored schedule in place; an epoch with no prior schedule stays absent.
- Intentionally unchanged: the `DutiesProvider` trait and `DutyAssignment` semantics, per-epoch insert ordering, current+next polling, retention pruning, and no filtering by the local registry.

- Scope is proposer-duty storage in this crate; dependent crates compile unchanged.
- Behavior change: malformed responses no longer replace a stored view. This is the intended SIP-101 behavior, not a regression.
- Requires the beacon node to serve `/eth/v2/validator/duties/proposer`; the pinned Lighthouse does.

## Validation

- Unit tests cover the acceptance matrix: complete out-of-order schedule accepted, metadata retained verbatim, and wrong-length, empty, out-of-epoch, and duplicate-slot responses rejected with the exact error. With the length pinned at `SLOTS_PER_EPOCH` and every slot in range, a missing slot can only appear as a duplicate, so that case covers both.
- `cargo +nightly fmt --check`, `clippy --all-targets -D warnings`, and `cargo test -p duties_tracker` (11 tests) pass.

## Rollback

## Blockers / Dependencies

- #1184 builds on the stored type introduced here.

- The poll loop itself (v2 selection; prior view kept across a failed or malformed refresh) has no direct test: the crate cannot fake a beacon-node response today. The validation logic is a plain function and fully covered; fetch-path coverage is tracked in #1240.

### DIFF
--- a/anchor/duties_tracker/src/duties_tracker.rs
+++ b/anchor/duties_tracker/src/duties_tracker.rs
@@ -13,7 +13,7 @@ use tracing::{debug, error, trace, warn};
 use types::{ChainSpec, Epoch, Slot};
 
 use crate::{
-    Duties, DutiesProvider, DutyAssignment, MembershipKey,
+    Duties, DutiesProvider, DutyAssignment, MembershipKey, ProposerSchedule,
     voluntary_exit_tracker::VoluntaryExitTracker,
 };
 
@@ -205,7 +205,8 @@ impl<T: SlotClock + 'static> DutiesTracker<T> {
         Ok(())
     }
 
-    /// Download the proposer duties for the current epoch.
+    /// Download the proposer duties for the current and next epoch, retaining only
+    /// responses that validate as complete schedules.
     async fn poll_beacon_proposers(&self) -> Result<(), Error> {
         let current_slot = self.slot_clock.now().ok_or(Error::UnableToReadSlotClock)?;
         let current_epoch = current_slot.epoch(self.slots_per_epoch);
@@ -215,18 +216,26 @@ impl<T: SlotClock + 'static> DutiesTracker<T> {
             match self
                 .beacon_nodes
                 .first_success(|beacon_node| async move {
-                    beacon_node.get_validator_duties_proposer(epoch).await
+                    beacon_node.get_validator_duties_proposer_v2(epoch).await
                 })
                 .await
-                .map(|response| response.data)
             {
-                Ok(proposer_duties) => {
+                Ok(response) => {
                     trace!(
-                        num_proposer_duties = proposer_duties.len(),
+                        num_proposer_duties = response.data.len(),
                         "Downloaded proposer duties"
                     );
 
-                    self.duties.proposers.write().insert(epoch, proposer_duties);
+                    match ProposerSchedule::from_response(epoch, self.slots_per_epoch, response) {
+                        Ok(schedule) => {
+                            self.duties.proposers.write().insert(epoch, schedule);
+                        }
+                        Err(e) => warn!(
+                            %epoch,
+                            error = %e,
+                            "Discarding malformed proposer duties; retaining prior view"
+                        ),
+                    }
                 }
                 Err(e) => last_err = Some(Error::FailedToPollProposers(e.to_string())),
             };
@@ -328,8 +337,8 @@ impl<T: SlotClock + 'static> DutiesProvider for DutiesTracker<T> {
             .proposers
             .read()
             .get(&epoch)
-            .map(|proposers| {
-                proposers.iter().any(|proposer_data| {
+            .map(|schedule| {
+                schedule.duties().iter().any(|proposer_data| {
                     proposer_data.slot == slot && proposer_data.validator_index == validator_index
                 })
             })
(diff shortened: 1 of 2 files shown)


shane-moore (approved):
very nice!

PR #1244, opened 2026-08-12
feat(client): fork-gated post-Gloas proposer delay

### LINKED ISSUE #1212: feat(client): fork-gated post-Gloas proposer delay (go-ssv ProposerDelayEPBS parity)

shane-moore:

`--proposer-delay-ms` (#1213) is a single, fork-agnostic knob. go-ssv splits the setting because ePBS retimes the proposal deadline, so the safe ranges differ: `ProposerDelay` before Gloas, `ProposerDelayEPBS` from it on, the latter hard-capped at 1s with deliberately no override. Without the split, an operator migrating from an SSV node loses a setting at Gloas, and a value tuned pre-Gloas silently keeps applying under a tighter deadline.

Closes #1212.

## Change Overview

- New `--proposer-delay-epbs-ms` (default 0), independent of `--proposer-delay-ms` with no fallback in either direction. Refused above 1000ms at startup regardless of `--allow-dangerous-proposer-delay`, matching go-ssv.
- The two values travel as one `ProposerDelays` pair from config into the validator store. Each proposer duty selects the value by the fork at its slot, and the duty span records which value applied so a post-fork flat-zero metric is diagnosable.
- Reading order: the validator store (pair + selection + the one call site in the RANDAO seam), then config validation, then tests and docs; the CLI and client wiring are mechanical.
- Unchanged: the delay's floor-from-slot-start semantics, metric names and outcome labels, and the ValidatorStore trait.
- One deliberate deviation from the issue's Note: selection keys on `signing_epoch` rather than the captured clock slot. Lighthouse passes the duty slot's epoch there, and `fork_name_at_slot` is defined as `fork_name_at_epoch(slot.epoch())`, so this is exactly "the fork at the duty slot" while removing clock-vs-duty skew at the boundary. Still no trait change.

- An operator running only `--proposer-delay-ms` loses the delay at Gloas by design (parity). Mitigated by a startup warning on networks where Gloas is scheduled, the span's `proposer_delay_source` field, and the docs' migration mapping.
- The 1000ms cap is go-ssv's policy value, not an Anchor-measured ceiling; the docs keep presenting the limits as guard rails, not safe maxima.

## Validation

- Unit tests: selection either side of the Gloas activation epoch, unscheduled Gloas, and zero-on-the-selected-side pins for the no-fallback semantic; config tests cover both defaults, the cap boundary, and refusal with the acknowledgement flag set.
- Virtual-time integration tests: Gloas-at-genesis applies the ePBS value, pre-Gloas ignores it, and a mid-schedule activation pins that the duty's real epoch drives selection.
- Full workspace suite (825 tests), clippy, fmt, and the docgen check all pass. go-ssv semantics verified against its `epbs-gloas` branch (key names, env var, cap boundary, flag scoping).

## Rollback

Revert the commit. The flag is additive and defaults to 0, which preserves current behavior; no data or config migration.

### DIFF
--- a/anchor/cli/src/cli.rs
+++ b/anchor/cli/src/cli.rs
@@ -528,8 +528,8 @@ pub struct PayloadBuildingOptions {
         help = "Counterpart of --proposer-delay-ms applied from the Gloas (ePBS) fork onward: \
                 each proposer duty uses whichever value matches the fork at its slot, with no \
                 fallback between them. Post-ePBS slots leave less headroom after the block \
-                request, so this value has a hard cap that --allow-dangerous-proposer-delay \
-                does not raise.",
+                request, so this value has a hard 1000ms cap that \
+                --allow-dangerous-proposer-delay does not raise.",
         display_order = 0
     )]
     pub proposer_delay_epbs_ms: u64,
--- a/anchor/cli/src/cli.rs
+++ b/anchor/cli/src/cli.rs
@@ -515,16 +515,31 @@ pub struct PayloadBuildingOptions {
                 the start of the slot, not extra latency: if the RANDAO pre-consensus round \
                 already finished later than this, no additional wait happens. A recommended \
                 starting value is 300. Higher values increase the risk of missing the block \
-                proposal.",
+                proposal. Applies to slots before the Gloas (ePBS) fork; from that fork on, \
+                --proposer-delay-epbs-ms applies instead.",
         display_order = 0
     )]
     pub proposer_delay_ms: u64,
 
     #[clap(
         long,
+        value_name = "MILLISECONDS",
+        default_value_t = 0,
+        help = "Counterpart of --proposer-delay-ms applied from the Gloas (ePBS) fork onward: \
+                each proposer duty uses whichever value matches the fork at its slot, with no \
+                fallback between them. Post-ePBS slots leave less headroom after the block \
+                request, so this value has a hard cap that --allow-dangerous-proposer-delay \
+                does not raise.",
+        display_order = 0
+    )]
+    pub proposer_delay_epbs_ms: u64,
+
+    #[clap(
+        long,
         help = "Permit a --proposer-delay-ms above the safety threshold. Delays this large \
                 significantly increase the risk of missed block proposals. Only set this if you \
-                understand the trade-off.",
+                understand the trade-off. Does not apply to --proposer-delay-epbs-ms, which has \
+                a hard cap instead.",
         display_order = 0
     )]
     pub allow_dangerous_proposer_delay: bool,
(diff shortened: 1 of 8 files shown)


shane-moore:
@claude, please take a look

shane-moore:
Reviewed by GPT-5.6 Sol at max reasoning effort. No correctness, safety, or performance issues found at `29d08018`. The two minor documentation findings are fixed in the current head. LGTM.

PR #1245, opened 2026-08-13
fix(validator_store): publish decided AggregatorCommittee aggregates from the metadata service

shane-moore:
Rebased on `unstable` (#1229 merged as `c332d6a7`); the three commits here are the full review scope.

Post-Boole, Anchor completes every attestation selection proof but most aggregates are never published. Anchor's selection proofs are distributed signing rounds finishing ~250-500ms into the slot, while embedded Lighthouse's attestation service clones its duty view (including `selection_proof: Option`) when attestation production starts and never re-reads it. All-`None` clones make `produce_and_publish_aggregates` return `Ok` silently. Both Lighthouse trigger paths (head event and timer) share the clone, so delaying the head event cannot close the race.

- Scale-100 ssv-mini evidence: 13/200 expected unique aggregates published (amplified build), 73/200 (PR-1229-only control), while proofs, QBFT, and post-consensus signing all succeeded.
- The failure is silent: no log, no metric, on the skip path.
- go-ssv is not affected: it publishes decided aggregates from the SSV node itself with no validator-client snapshot in its path.

## Change Overview (Required)

Make Anchor the single authoritative Boole+ aggregate publisher, working from the decided value instead of Lighthouse's snapshot:

- The Boole+ branch of `sign_aggregate_and_proofs` now yields one empty `Ok` batch (Lighthouse's publish loop drops empty batches silently, verified at pin b263df5) and logs the discarded request set, and `sign_committee_aggregate_and_proofs` is deleted.
- `start_aggregator_post_consensus` returns its vacant-only insertions; the metadata service spawns one detached publisher per slot that joins each new execution under the existing two-slot deadline and publishes **each decided aggregate the moment its threshold signature reconstructs, one POST per aggregate** (go-ssv's publication shape): a beacon-node-rejected aggregate cannot fail a sibling's POST, and a no-quorum root cannot delay its committee's other aggregates.
- The v1/v2 endpoint and fork header derive from the decided value's `DataVersion` (the fork its payloads were decoded under), so they can never diverge from the payload variant.
- New `anchor_aggregator_committee_publish_total{result}` metric: `consensus_error`/`no_aggregates` count committees, `success`/`http_error`/`no_signatures` count aggregates; per-aggregate publish logs reproduce Lighthouse's field set (`type="aggregated"`).

Reading order: `aggregator_post_consensus.rs` (registration, per-committee join, per-root publish), then `metadata_service.rs` (spawn + POST), then the `lib.rs` deletion, then tests.

Did not change: pre-Boole per-validator aggregate signing (Lighthouse still publishes), the contributions callback, PR #1229's exactly-once registration and detached signing, and all head-event/timer plumbing.

- Publication logic for one object class moves into Anchor. Accepted deliberately for go-ssv parity: the decided value, reconstructed signature, and submission now live in one place, and correctness no longer depends on Lighthouse's internal scheduling at each pin.
- All four operators publish the same reconstructed aggregate (as in go-ssv); duplicate POSTs to shared beacon nodes were exercised in two mixed-client runs with zero errors, including with per-aggregate POSTs on both clients.
- One POST per aggregate raises request count (~3-16 per committee-slot vs 1); exercised at 250-validator load (~8 aggregates per slot per node) with zero publish failures.
- The empty-batch behavior of Lighthouse's publish loop is a pin-specific fact; re-verify on pin bumps (tripwire noted in the code comment).
- Cold Boole-only code path; zero lines and zero latency on attestation production.

## Validation (Required)

- 74 crate tests (was 66), including recorder-based per-root publisher tests, fork-propagation from the decided `DataVersion`, vacant-only registration, quorum-failure, deadline, and partial-quorum paths (one root misses quorum, the sibling still publishes), and the previously uncovered pre-Boole per-validator path. `make cargo-fmt-check`, `make lint` clean.
- Seven ssv-mini campaigns (evidence and per-run verdicts recorded in the campaign notes). Batch-shape build (runs 1-4):
  - Scale-100 all-Anchor: **200/200** unique aggregates (control: 73/200) with the underlying race still firing (48-52 of 62 proof batches after Lighthouse's snapshot start).
  - Mixed 2 Anchor + 2 go-ssv: **200/200** under a harsher race (63/64 late batches); publishes from go-led decided values; duplicate POSTs harmless.
  - Fault matrix: single-operator restarts x3 (continuous publication on 3/4), two-operator quorum loss (correct null result, no hangs, instant recovery), hung-BN pause (zero publish failures, instant resume).
  - 250 validators, 513-slot soak: every decided aggregate published, zero publish failures, flat resources; a ~3% eligible-assignment shortfall traced to selection-proof collection timeouts under load (upstream signing pipeline, pre-existing class).
- Final per-root build (runs 5-7; pre-rebase commits `32534555` + `c2e886c4`, now `9233cf56` + `c8c291bb`):
  - Scale-100 all-Anchor: **200/200**, zero publish failures.
  - Mixed 2 Anchor + 2 go-ssv: unique **200/200** under the harshest race profile (63/64 late batches), per-aggregate duplicate POSTs on both clients, zero Anchor publish failures.
  - 250 validators, two 64-slot windows: **500/500 in both**, zero publish failures, zero missed-signature roots.
- Known limitation: the `no_signatures` outcome (a decided root that never reaches quorum) is covered by unit tests including the partial-quorum case, but was not reproducible live (local reconstruction outruns slot-precise fault injection).

Three commits that revert together; reverting restores the Lighthouse callback path (and its snapshot race). No config, schema, or wire-format changes; the new metric disappears on revert.

- None: #1229 merged (`c332d6a7`) and this branch is rebased on it.

- Follow-up to be filed: `AggregatorCommitteeDataValidator` version-vs-duty-slot-fork validation (mirrors b6809645; bounds the decided `DataVersion` the publish fork is derived from). An upstream Lighthouse report of the duty-snapshot race may follow separately; it no longer affects Anchor.
- Alternatives considered and rejected: head-event readiness gate (structurally partial: the timer path shares the snapshot; couples attestation liveness to aggregate readiness), fixed post-fork head-event delay (constant-dependent), Lighthouse fork (pin cadence). A redundant dual-publisher rollout was debated and declined: it forfeits the callback-path deletion and makes the publish metric unmeasurable under races. Batch-per-committee publication was implemented first and validated (runs 1-4), then replaced by per-root publication for exact go-ssv parity and failure isolation (runs 5-7).

### DIFF
--- a/anchor/validator_store/src/aggregator_post_consensus.rs
+++ b/anchor/validator_store/src/aggregator_post_consensus.rs
@@ -141,7 +141,9 @@ async fn publish_one_root<E: EthSpec, F, Fut>(
             metrics::inc_publish_result(metrics::NO_SIGNATURES);
             return;
         }
-        // Deadline expired before this root's quorum; only this root is withheld.
+        // Deadline expired before this root's quorum; only this root is withheld. Labelled
+        // `timeout` to match the committee-level join arm, now that expiry is distinguishable
+        // from a collection failure.
         Err(_) => {
             warn!(
                 pubkey = ?root.request.validator.public_key,
@@ -150,7 +152,7 @@ async fn publish_one_root<E: EthSpec, F, Fut>(
             );
             validator_metrics::inc_counter_vec(
                 &validator_metrics::SIGNED_AGGREGATES_TOTAL,
-                &[metrics::OTHER_ERROR],
+                &[metrics::TIMEOUT],
             );
             metrics::inc_publish_result(metrics::NO_SIGNATURES);
             return;
--- a/anchor/validator_store/src/aggregator_post_consensus.rs
+++ b/anchor/validator_store/src/aggregator_post_consensus.rs
@@ -2,11 +2,14 @@
 //!
 //! One QBFT decision carries both attestation aggregates and sync contributions for an SSV
 //! committee, and the operator must emit exactly one committee partial-signature message per
-//! `(committee, slot)` covering everything it can sign. No consumer can own that message:
-//! contributions are returned through a Lighthouse callback that may or may not fire (issue
-//! #1227), and aggregates are published by the metadata service's publisher because Lighthouse's
-//! duty snapshot is cloned before slot-start selection proofs finish and would silently skip
-//! them.
+//! `(committee, slot)` covering everything it can sign. The decided value drives both what gets
+//! signed and what gets published: Anchor owns Boole+ aggregate publication outright, through
+//! the metadata service's publisher, because the protocol's unit of agreement is the decided
+//! value and no local Lighthouse view is part of it. (Lighthouse's duty snapshot in particular
+//! is cloned before slot-start selection proofs finish, so routing publication through it would
+//! silently skip late-installed proofs.) Contributions are returned through a Lighthouse
+//! callback that may or may not fire (issue #1227), so no callback can own the committee
+//! message either.
 //!
 //! The signing set is therefore a property of the duty, not of any consumer. The slot pipeline
 //! starts one execution per committee when it publishes the decided value at 2/3 slot; the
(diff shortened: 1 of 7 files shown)


shane-moore:
@claude, please take a look

shane-moore:
@claude, good catch on the deadline-expiry label: adopted in 6605b34e (the arm now carries `timeout`, matching the committee-level join; the old `other_error` was a leftover from the collective drain, where expiry and collection failure were indistinguishable).

petarjuki7 (approved):

PR #1248, opened 2026-08-13
feat(validator_store): add envelope-signing groundwork

jnhsigmap:
Lands the shared groundwork for the SIP-94 envelope-signing duty, ahead of the duty implementation.

- The envelope-signing duty method is still an `Unsupported` stub. The duty implementation needs error variants, a QBFT value-check factory, and test-harness support first.
- Landing the groundwork separately keeps both PRs small and independently reviewable.
- Part of #1126 (PR 1 of 2).

## Change Overview

- Three new duty error variants: a pre-Gloas rejection, a non-self-build rejection, and a sentinel for consensus deciding an envelope that another operator built.
- A factory that builds the QBFT value check for envelope consensus data, staged behind a dead-code expectation until the duty wires it in.
- Test harness: the mock consensus decider can now force a decided envelope value and records every decide call. A new envelope test module adds the shared fixtures and four tests.
- Suggested reading order: the error variants and the factory, then the harness extensions, then the new test module.
- No behavior changed: the duty stub still returns `Unsupported`, and no production path uses the additions yet.

- Blast radius is minimal: every addition is either test-only or unused by production code.
- The staged dead-code expectations are lint-enforced, so the follow-up PR cannot forget to remove them.
- The sentinel variant models an intentional non-publish as an error because the caller publishes every `Ok`. The follow-up PR classifies it separately in logs and metrics.

## Validation

- `cargo test -p anchor_validator_store`: 108 tests pass (104 existing + 4 new).
- Format check and clippy with `-D warnings` are clean.
- The new tests pin the mock's forced-decision and capture behavior and the factory's parameter wiring. The value-check rejection matrix is already covered by `ssv_types` tests.

## Rollback

## Blockers / Dependencies

- The follow-up PR (the duty implementation) depends on this PR.

### DIFF
--- a/anchor/validator_store/src/lib.rs
+++ b/anchor/validator_store/src/lib.rs
@@ -47,8 +47,9 @@ use ssv_types::{
         AggregatorCommitteeConsensusData, AggregatorCommitteeDataValidator, BEACON_ROLE_AGGREGATOR,
         BEACON_ROLE_PROPOSER, BEACON_ROLE_SYNC_COMMITTEE_CONTRIBUTION, BeaconVote,
         BeaconVoteValidator, Contribution, ContributionWrapper, Contributions, DataVersion,
-        ForkDecodeError, GloasBeaconVote, GloasBeaconVoteValidator, ProposerConsensusData,
-        ProposerConsensusDataValidator, QbftData, SelectionProofBatchId, ValidatorDuty,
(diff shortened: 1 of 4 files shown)


jnhsigmap:
@claude, review this PR

shane-moore (approved):
lgtm breh! like the idea of splitting into a couple pr's for this one, makes sense

PR #1249, opened 2026-08-13
fix(duties_tracker): pin authoritative complete-schedule transitions

### LINKED ISSUE #1184: fix(duties_tracker): authoritative complete-schedule transitions for proposer assignment verdicts

petarjuki7:

- Built to the reduced scope agreed on #1184: the early RANDAO feature is dropped, so there is no watch channel, no generation counter, and none of the notify-ordering work or tests.
- What remains is the retention contract, which still matters because the message validator rejects ProposerPreferences and EnvelopeProposer messages on `NotAssigned`.
- No acceptance criterion needed a behavior change; #1241 already made them all true. But several were true only because the opposing code does not exist yet: nothing revokes on a reorg because the crate has no reorg handling, the local registry cannot affect the schedule because the proposer path never reads it, and `execution_optimistic` cannot create a provisional state because nothing reads it. None of that was pinned, so any of it could regress silently.
- Closes #1184. Follows #1241.

## Change Overview

- Extracted the validate-and-install step out of the poll loop into `install_proposer_schedule`. Behavior is identical, but the transition path is now reachable from tests, and the poll loop loses a level of nesting. The early return on a validation error is what preserves the previously retained schedule.
- Documented the contract on `DutyAssignment`: a retained schedule is replaced only by another complete schedule, never revoked by a rejected refresh, a registry change, `execution_optimistic`, or a reorg. Also noted that `is_epoch_known_for_proposers` now means a complete schedule is retained.
- Six tests covering replacement, a rejected refresh with and without a prior view, optimistic metadata, a changed dependent root, and authority over validators that are not in the local registry. The unfetched-epoch test now checks every slot in the epoch rather than one.
- Intentionally unchanged: the `DutiesProvider` trait, `DutyAssignment` semantics, polling and pruning, and the index-keyed `is_validator_proposer_at_slot` (that is #1190).

- Production risk is low: one method moved, no logic altered, dependent crates compile unchanged.
- The extracted method is private, so the crate's public surface is the same.

## Validation

- `cargo +nightly fmt --check`, `clippy --all-targets -D warnings`, and `cargo test -p duties_tracker` (17 tests) pass. `message_validator`, `eth`, and `client` compile unchanged.
- The two load-bearing tests were mutation-checked: removing the insert fails the replacement test, and making the error path insert anyway fails the preservation test.

## Rollback

- A transport failure and the v2-versus-v1 endpoint choice are still not covered, since the crate cannot fake a beacon-node response. Tracked in #1240. For the same reason the registry test pins that the schedule is never filtered by the local registry, rather than driving a registry change through the database.

### DIFF
--- a/anchor/duties_tracker/src/duties_tracker.rs
+++ b/anchor/duties_tracker/src/duties_tracker.rs
@@ -3,6 +3,7 @@ use std::{future::Future, sync::Arc};
 use beacon_node_fallback::BeaconNodeFallback;
 use bls::PublicKeyBytes;
 use database::NetworkState;
+use eth2::types::{DutiesResponse, ProposerData};
 use safe_arith::ArithError;
 use slot_clock::SlotClock;
 use ssv_types::ValidatorIndex;
@@ -13,7 +14,7 @@ use tracing::{debug, error, trace, warn};
 use types::{ChainSpec, Epoch, Slot};
 
 use crate::{
-    Duties, DutiesProvider, DutyAssignment, MembershipKey, ProposerSchedule,
+    Duties, DutiesProvider, DutyAssignment, MembershipKey, ProposerSchedule, ProposerScheduleError,
     voluntary_exit_tracker::VoluntaryExitTracker,
 };
 
@@ -226,15 +227,12 @@ impl<T: SlotClock + 'static> DutiesTracker<T> {
                         "Downloaded proposer duties"
                     );
 
-                    match ProposerSchedule::from_response(epoch, self.slots_per_epoch, response) {
-                        Ok(schedule) => {
-                            self.duties.proposers.write().insert(epoch, schedule);
-                        }
-                        Err(e) => warn!(
+                    if let Err(e) = self.install_proposer_schedule(epoch, response) {
+                        warn!(
                             %epoch,
                             error = %e,
                             "Discarding malformed proposer duties; retaining prior view"
-                        ),
+                        );
                     }
                 }
                 Err(e) => last_err = Some(Error::FailedToPollProposers(e.to_string())),
@@ -247,6 +245,19 @@ impl<T: SlotClock + 'static> DutiesTracker<T> {
             .retain(|&epoch, _| epoch + HISTORICAL_DUTIES_EPOCHS >= current_epoch);
 
         last_err.map_or(Ok(()), Err)
+    }
+
+    /// Validate `response` and install it as the schedule for `epoch`, replacing any schedule
+    /// already retained. A response that fails validation is discarded, leaving the previously
+    /// retained schedule in place.
+    fn install_proposer_schedule(
+        &self,
+        epoch: Epoch,
+        response: DutiesResponse<Vec<ProposerData>>,
+    ) -> Result<(), ProposerScheduleError> {
+        let schedule = ProposerSchedule::from_response(epoch, self.slots_per_epoch, response)?;
+        self.duties.proposers.write().insert(epoch, schedule);
+        Ok(())
     }
 
     pub fn start(self: Arc<Self>, executor: TaskExecutor) {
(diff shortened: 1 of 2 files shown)


shane-moore (approved):
lgtm! extraction checks out as behavior identical and the transition tests pin the #1184 contract cleanly

PR #1261, opened 2026-08-17
fix(message_validator): ignore, not reject, messages with an unresolved validator index

### LINKED ISSUE #1259: fix(message_validator): unresolved validator index yields Reject instead of Ignore
## Description

When a hosted validator is present in the local database but its beacon index is not yet resolved (metadata sync lagging after startup or activation), `CommitteeInfo::validator_indices` is empty. The `Role::Proposer` and `Role::SyncCommittee` branches of `validate_beacon_duty` then fail on `.first().ok_or(ValidationFailure::UnexpectedFailure { .. })`, and `UnexpectedFailure` maps to `MessageAcceptance::Reject`.

That penalizes an honest peer for a gap in the local node's data. go-ssv returns `ErrNoShareMetadata` in this case, which is a non-reject (Ignore) error. Anchor already has `ValidationFailure::NoShareMetadata` mapped to `Ignore`, but nothing returns it.

## Proposed change

Return `NoShareMetadata` instead of `UnexpectedFailure` when the validator index is unresolved in the Proposer and SyncCommittee duty checks.

Committee roles are unaffected: an empty index list only lowers the duty limit, and `TooManyDutiesPerEpoch` is already Ignore.

shane-moore:

When a hosted validator is in the local database but its beacon index has not been synced yet (right after startup or activation), `CommitteeInfo::validator_indices` is empty. The `Proposer` and `SyncCommittee` branches of `validate_beacon_duty` then fail with `UnexpectedFailure`, which maps to `MessageAcceptance::Reject`. An honest peer relaying a valid message gets penalized for a gap in our own view.

go-ssv returns `ErrNoShareMetadata` here, a non-reject (Ignore) error. Anchor already has `ValidationFailure::NoShareMetadata` mapped to Ignore, but nothing returned it.

Closes #1259

## Change Overview (Required)

- The two identical index lookups in `validate_beacon_duty` collapse into one helper that returns `NoShareMetadata` when the index is absent.
- Read `lib.rs` first (the helper and its two call sites), then the tests in `partial_signature.rs`.
- Nothing else changes: the Ignore mapping, the duty checks that follow, and every other role's path are untouched. Committee roles never take this lookup (an empty index list only lowers their duty limit, and `TooManyDutiesPerEpoch` is already Ignore).

- Blast radius is one verdict on one already-rare path: Reject becomes Ignore. The message is still not forwarded or processed.
- No wire, topic, or scoring-parameter change.

## Validation (Required)

- New tests: Proposer and SyncCommittee with empty `validator_indices` return `NoShareMetadata` and map to `Ignore`; a Committee-role control shows that path never hits the lookup. Both new assertions fail on the pre-fix code with the old `UnexpectedFailure`.
- `cargo test -p message_validator`: 60 passed, 0 failed.
- `make cargo-fmt-check`, `make lint`: clean.

Revert the commit. No config, data, or operational impact.

N/A

N/A

### DIFF
--- a/anchor/message_validator/src/lib.rs
+++ b/anchor/message_validator/src/lib.rs
@@ -762,6 +762,22 @@ fn verify_message_signatures(
     Ok(())
 }
 
+/// Returns the single validator index of a non-committee duty message.
+///
+/// The index is absent when the validator is known locally but its beacon metadata has not been
+/// synced yet. That is a gap in the local view, not a fault of the sending peer, so it maps to
+/// [`ValidationFailure::NoShareMetadata`] (Ignore) rather than a Reject.
+fn single_validator_index(
+    validation_context: &ValidationContext<impl SlotClock>,
+) -> Result<ValidatorIndex, ValidationFailure> {
+    validation_context
+        .committee_info
+        .validator_indices
+        .first()
+        .copied()
+        .ok_or(ValidationFailure::NoShareMetadata)
+}
+
 /// Validates if a validator is assigned to a specific duty
 pub(crate) fn validate_beacon_duty(
     validation_context: &ValidationContext<impl SlotClock>,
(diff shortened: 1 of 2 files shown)


shane-moore (commented):
Reviewed the final diff. The unresolved validator-index paths now consistently return `Ignore`, including Committee messages that produce `ExcessiveDutyCount`. The regression test covers the acceptance mapping. I found no remaining correctness or safety issues.

Reviewed by gpt-5.6-sol (max).

PR #1263, opened 2026-08-18
feat(validator_store): sign self-build execution payload envelopes

### LINKED ISSUE #1126: feat(validator_store): implement sign_execution_payload_envelope

jnhsigmap:

- After the Gloas fork, a self-build proposer must sign and publish an execution payload envelope (EIP-7732 / SIP-94).
- Anchor's validator store rejected this duty as unsupported, so a self-build proposal could not complete after the block was published.
- PR #1248 added the groundwork: typed errors, consensus value validation, and test fixtures. This PR completes the self-build envelope proposer duty path.
- Closes #1126.

## Change Overview (Required)

- The duty follows the same shape as the other consensus duties. The store blinds the local envelope, reaches QBFT consensus over it, and signs the decided envelope under the builder domain.
- Every operator signs first, so the builder can reconstruct the threshold signature. Only the operator whose local envelope matches the decided value returns it for publication. All other operators return a typed sentinel error, because the caller publishes every success result.
- A new counter classifies each completed attempt as published, not built locally, or failed. Tracing mirrors the sibling duties: a span per duty, warnings on failures, and an info line for the expected non-publish.
- Read the duty body first, then the metrics, then the tests.
- What did not change: pre-consensus rejections (not synced, pre-Gloas slot, external builder, missing decided block root) stay typed errors and never touch the counter. No other duty path changed.

- The blast radius is the envelope duty only, and the path is active only after the Gloas fork.
- The beacon node logs the sentinel at error level on every operator that did not build the decided envelope. Anchor logs the same event at info level with both roots and marks it as expected. The outcome metric separates this case from real failures.
- Trade-off: signing before the content gate spends a signature share on envelopes the operator will not publish. This is required so the builder can reconstruct the threshold signature.

## Validation (Required)

- Format and workspace lint are clean.
- `cargo test -p anchor_validator_store`: 130 passed, 0 failed, with 12 new envelope tests.
- The new tests cover the happy path, the signing root and domain, the pre-consensus gates, content mismatches, per-label metric deltas, a slashing-database probe, and an end-to-end block-then-envelope flow.
- The full workspace test suite passes.

- Revert the single commit. The duty returns to the unsupported stub.
- No configuration, data, or migration impact. The new metric stops being exported.

- None.

- Devnet validation of the duty is deferred indefinitely. Coverage limited to unit tests against the mocked consensus and collection layers.
- Follow-up candidates: bound the signature collection await to the duty slot end, and pin the consensus value version inside the shared value validator.

### DIFF
--- a/anchor/validator_store/src/lib.rs
+++ b/anchor/validator_store/src/lib.rs
@@ -4067,6 +4067,10 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> ValidatorStore
                     builder_index: envelope.builder_index,
                 }));
             }
+            let current_slot = self.slot_clock.now().ok_or(SpecificError::SlotClock)?;
+            if slot > current_slot {
+                return Err(Error::GreaterThanCurrentSlot { slot, current_slot });
+            }
 
             let decided_block_root = self.get_decided_block_root(validator_pubkey, slot)?;
             let local_blinded = BlindedExecutionPayloadEnvelope::from_full(&envelope);
--- a/anchor/validator_store/src/lib.rs
+++ b/anchor/validator_store/src/lib.rs
@@ -30,8 +30,9 @@ use openssl::{
 use parking_lot::Mutex;
 use qbft::Completed;
 use qbft_manager::{
-    AggregatorCommitteeInstanceId, CommitteeInstanceId, ConsensusDecider, ProposerInstanceId,
-    QbftError, QbftManager, TimeoutMode, ValidatorDutyKind,
+    AggregatorCommitteeInstanceId, CommitteeInstanceId, ConsensusDecider,
+    EnvelopeProposerInstanceId, ProposerInstanceId, QbftError, QbftManager, TimeoutMode,
+    ValidatorDutyKind,
 };
 use safe_arith::{ArithError, SafeArith};
 use signature_collector::{
@@ -45,11 +46,12 @@ use ssv_types::{
     OperatorId, ValidatorIndex, ValidatorMetadata,
     consensus::{
         AggregatorCommitteeConsensusData, AggregatorCommitteeDataValidator, BEACON_ROLE_AGGREGATOR,
-        BEACON_ROLE_PROPOSER, BEACON_ROLE_SYNC_COMMITTEE_CONTRIBUTION, BeaconVote,
-        BeaconVoteValidator, Contribution, ContributionWrapper, Contributions, DataVersion,
-        EnvelopeConsensusDataValidator, ForkDecodeError, GloasBeaconVote, GloasBeaconVoteValidator,
-        ProposerConsensusData, ProposerConsensusDataValidator, QbftData, SelectionProofBatchId,
-        ValidatorDuty,
+        BEACON_ROLE_ENVELOPE_PROPOSER, BEACON_ROLE_PROPOSER,
+        BEACON_ROLE_SYNC_COMMITTEE_CONTRIBUTION, BeaconVote, BeaconVoteValidator,
+        BlindedExecutionPayloadEnvelope, Contribution, ContributionWrapper, Contributions,
+        DataVersion, EnvelopeConsensusData, EnvelopeConsensusDataValidator, ForkDecodeError,
+        GloasBeaconVote, GloasBeaconVoteValidator, ProposerConsensusData,
+        ProposerConsensusDataValidator, QbftData, SelectionProofBatchId, ValidatorDuty,
     },
     msgid::Role,
     partial_sig::PartialSignatureKind,
@@ -64,6 +66,7 @@ use tokio::{
     time::{Instant, sleep},
 };
 use tracing::{Instrument, Span, debug, error, field, info, info_span, trace, warn};
+use tree_hash::TreeHash;
 use types::{
     AbstractExecPayload, Address, AggregateAndProof, Attestation, BeaconBlock, BeaconBlockRef,
     BlindedPayload, ChainSpec, Checkpoint, ContributionAndProof, Domain, Epoch, EthSpec,
@@ -74,6 +77,7 @@ use types::{
     SignedValidatorRegistrationData, SignedVoluntaryExit, SingleAttestation, Slot, SlotData,
     SyncAggregatorSelectionData, SyncCommitteeContribution, SyncCommitteeMessage,
     SyncSelectionProof, SyncSubnetId, ValidatorRegistrationData, VoluntaryExit,
+    consts::gloas::BUILDER_INDEX_SELF_BUILD,
 };
 use validator_metrics::IntCounterVec;
 use validator_store::{
@@ -945,7 +949,6 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> AnchorValidator
     /// with no later insert, an old entry stays in the map.
     ///
     /// Reads are non-destructive and return the root by value.
-    #[cfg_attr(not(test), expect(dead_code))] // no non-test caller yet
     fn get_decided_block_root(
         &self,
         validator_pubkey: PublicKeyBytes,
@@ -1369,7 +1372,6 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> AnchorValidator
     }
 
     /// Constructs the QBFT data validator for envelope-signing duties.
-    #[cfg_attr(not(test), expect(dead_code))] // no non-test caller yet
     fn create_envelope_consensus_data_validator(
         &self,
         validator_pubkey: PublicKeyBytes,
@@ -3052,7 +3054,7 @@ pub enum SpecificError {
         builder_index: u64,
     },
     /// Consensus decided an envelope another operator built. This is an intentional
-    /// non-publish, not a failure: the partial signature was already contributed.
+    /// non-publish, not a failure.
     EnvelopeNotBuiltLocally {
         local_root: Hash256,
         decided_root: Hash256,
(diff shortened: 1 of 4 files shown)


jnhsigmap:
@claude, review this PR

jnhsigmap:
In response to claude feedback:

Added some tests for error path with a `failing_envelope` function that causes every `EnvelopeConsensusData` decide call to result in `failure`. This imitates envelope consensus that times out or errors:
  - `undecodable_decided_envelope_counts_as_failed_and_signs_nothing`: garbage `data_ssz` in a forced decision. expects `InvalidQbftData`.
  - `consensus_timeout_counts_as_failed_and_signs_nothing`: Expects `SpecificError::Timeout`.
  - `consensus_error_counts_as_failed_and_signs_nothing`: Expects `QbftError::QueueClosedError` mapped through `SpecificError::from`.
  - `future_slot_envelope_is_rejected_before_consensus`: Envelope at `TEST_SLOT + 1`, clock at `TEST_SLOT`. Expects the guard error with both captures empty. Without the guard this test would see `DecidedRootUnavailable`. Pins the gate's position too.

More importantly, in `sign_execution_payload_envelope`: Duty reads the slot clock and rejects `envelope.slot() > current_slot` with `Error::GreaterThanCurrentSlot { slot, current_slot }`. This was added after the self-build gate and before the decided-root lookup. Fires before the outcome counter or consensus timer exist. Equal slots pass.

shane-moore (approved):
very nice, approved!

Issue #1236, opened 2026-08-11
fix(validator_store): continue beacon fallback on missing Boole sync contributions


shane-moore:
## Goal

Prevent a temporary `404 Not Found` from one beacon node from prematurely dropping a sync committee subnet from Anchor's Boole `AggregatorCommitteeConsensusData` proposal when another configured beacon node can supply the contribution.

## Context / motivation

Checked on 2026-08-10 against Anchor `unstable` at [`82bac14`](https://github.com/sigp/anchor/commit/82bac14ec635bc28a68a6bba9799bd603ec0c082). Reverify mutable refs at implementation time.

During Boole `AggregatorCommittee` proposal construction, Anchor fetches each required sync committee contribution through `BeaconNodeFallback::first_success` in [`MetadataService::fetch_sync_contributions`](https://github.com/sigp/anchor/blob/82bac14ec635bc28a68a6bba9799bd603ec0c082/anchor/validator_store/src/metadata_service.rs#L1021-L1131).

The Beacon API returns `404 Not Found` when a contribution is not available. Lighthouse maps that response to `Ok(None)`, but [`first_success`](https://github.com/sigp/lighthouse/blob/b263df596671a2bd42bf1034e1cdc8188ba8a9b0/validator_client/beacon_node_fallback/src/lib.rs#L650-L698) returns on any `Ok`, including `Ok(None)`. Anchor therefore does not query the remaining configured beacon nodes or perform the helper's second pass. The `Ok(None)` case is handled only after `first_success` has returned.

References:

- [Anchor sync contribution fetch](https://github.com/sigp/anchor/blob/82bac14ec635bc28a68a6bba9799bd603ec0c082/anchor/validator_store/src/metadata_service.rs#L1021-L1131)
- [Lighthouse optional-response handling](https://github.com/sigp/lighthouse/blob/b263df596671a2bd42bf1034e1cdc8188ba8a9b0/common/eth2/src/lib.rs#L247-L256)
- [Lighthouse `first_success`](https://github.com/sigp/lighthouse/blob/b263df596671a2bd42bf1034e1cdc8188ba8a9b0/validator_client/beacon_node_fallback/src/lib.rs#L650-L698)
- [Beacon API 404 behavior](https://github.com/ethereum/beacon-APIs/blob/ba859dbca9076ba19dedc095c59afa7b05a66819/apis/validator/sync_committee_contribution.yaml#L14-L15)

[...]

PR #1264, opened 2026-08-18
fix(validator_store): publish decided Boole sync contributions from the metadata service

### LINKED ISSUE #1260: fix(validator_store): publish decided Boole sync contributions from the metadata service
## Goal

Extend the #1245 authoritative publisher so decided sync contributions are POSTed by Anchor itself, removing the last dependency on Lighthouse's snapshot-based contribution callback.

## Context and motivation

#1245 moved Boole+ aggregate publication to the metadata service but deliberately left contributions on the Lighthouse callback path, on the premise that "sync selection proofs are precomputed a slot ahead". That premise is wrong for Anchor's config: with `lookahead_slot: 1`, Lighthouse's `fill_in_aggregation_proofs` starts signing slot N's proof at slot N's boundary (`start_slot = current + 1`, `proof_slot = slot - 1`), and post-Boole the proof then waits on slot-start `VotingAssignments` plus the shared `AggregatorCommitteePartialSig` batch (measured p50 ~307-323ms, p95 ~445-487ms at scale-100).

Two consequences:

- **Now (pin `e58ec88fe`):** Lighthouse snapshots the aggregators map at 4s, while our proof deadline is 2/3 slot. Any proof reconstructing in (4s, 8s] is silently dropped from the snapshot.
- **At the next pin bump past sigp/lighthouse#9675** (head `ce9403c84` at analysis time): the sync service triggers on head events at ~200ms and freezes the aggregators map then. An empty map spawns zero contribution tasks with no log. Same mechanism that lost 187 of 200 expected aggregates before #1245. Signing is already callback-independent on `unstable` (`spawn_root_collection` covers the outcome's `contributions` map); only the POST is missing.

Not wiring the 9675 head receiver instead would forfeit the win #1135 was closed for (post-consensus batches flush only when complete, so attestation partials on sync-overlap committees are held to >=4s today).

## Suggested approach

- Gate at the outer `sign_sync_committee_contributions` mirroring `sign_aggregate_and_proofs`: Boole+ returns one empty `Ok(Vec::new())` item before grouping; delete the per-committee helper. Lighthouse's publish loop treats an empty batch silently; pre-Boole untouched.
- Replace the aggregate-only early return in `publish_committee_aggregates` (it currently discards contribution-only outcomes) with independent class handling: both empty = skip; otherwise drain aggregate and contribution roots concurrently under the shared deadline, extending the existing per-root `FuturesUnordered` (Either/enum wrap fine; concurrency is the requirement).
- Assemble `SignedContributionAndProof` from the decided message plus the reconstructed threshold signature, one POST per `(validator_index, subcommittee_index)`, via `first_success` `post_validator_contribution_and_proofs` (Lighthouse's own publication policy). go-ssv parity: its contribution runner submits from the SSV node (`sync_committee_contribution.go:374`).
- Move `SIGNED_SYNC_COMMITTEE_CONTRIBUTIONS_TOTAL` accounting from the deleted callback to the publish path; add a per-contribution publish log for the campaign analyzer. No new counters; do not overload the aggregate-only `AGGREGATOR_COMMITTEE_PUBLISH_TOTAL`; define what a contribution-only outcome increments (today it counts as `NO_AGGREGATES`).
- Fix the "precomputed a slot ahead" comment; document the guarantee as process-scoped exactly-once (a restart can lose an in-flight slot).

## Acceptance criteria

- A contribution absent from Lighthouse's callback snapshot is still published.
- Contribution-only, aggregate-only, and mixed outcomes publish independently; a hanging root in one class does not delay the other.
- No duplicate POSTs when the Lighthouse callback fires with a non-empty snapshot.
- Pre-Boole behavior unchanged.

## Tests

- Metadata-level publisher tests (not only the generic closure) for contribution-only, aggregate-only, mixed; invert the existing contribution-only never-published test.
- Outer Boole+ callback yields exactly one silent empty item across multiple committees; cross-class independence; pre-Boole no-op; multi-subcommittee assembly.
- Campaign: extend the analyzer to contribution publish logs; scale-100 with head monitor enabled; unique attempted and successful publications approximately equal locally signable resolved roots that reached signature quorum.

## Notes

- Known limiter, out of scope: the metadata contribution fetch treats a first-node `Ok(None)` as success and never falls back, so a contribution can be absent from consensus data entirely; nothing downstream can publish it (#1236 describes this).
- Eligibility validation of decided contributors is deliberately out of scope, matching #1245.
- Rejected alternatives: passing `None` for the 9675 sync head receiver (forfeits the #1135 win, keeps the (4s, 8s] tail); an embedder-side readiness gate (retired for attestations, structurally partial, and dominated now that signing is callback-independent); an upstream Lighthouse late-read fix (decided against; the Anchor fix is self-sufficient).
- Adversarial review (3 rounds + subtraction): converged on this shape; cut a partial `is_aggregator` check, API renames, new counters, and a five-count campaign funnel as speculative or out-of-PR.

shane-moore:

At Boole+, one QBFT decision per SSV committee carries both attestation aggregates and sync contributions. #1245 made Anchor the authoritative publisher for aggregates but left contributions on the Lighthouse callback, on the premise that "sync selection proofs are precomputed a slot ahead".

That premise is wrong for Anchor's configuration. `lookahead_slot: 1` is effective *zero* lookahead: `fill_in_aggregation_proofs` uses `start_slot = current_slot + 1` and `proof_slot = slot - 1`, so slot N's proof signing begins at slot N's boundary, in the same slot-start partial-signature batch as the attestation proofs. Lighthouse's duty snapshot can therefore be cloned before the proofs exist, and `get_duties_for_slot` silently omits any aggregator whose proof is not installed. An empty aggregators map spawns zero tasks and logs nothing.

Measured on a scale-100 ssv-mini devnet (4 Anchor operators, 100 managed validators, head-event-triggered sync duties via an unmerged Lighthouse PR backported onto the current pin), 238 post-Boole slots, this branch against an identical build without it:

| | with this PR | without |
|---|---|---|
| subnet-slots published | 952/952 (100%) | 367/952 (38.6%) |
| slots publishing nothing | 0 | 144 (60.5%) |
| contribution identities published | 9260 | 0 |
| publish failures / quorum skips | 0 / 0 | 0 / 0 |

The pre-fix build lost 60% of slots silently: no error, no warning, no metric movement. The cause is confirmed from Anchor's own metrics rather than inferred: `voting_context_triggers_total{trigger="head_event"} = 143` versus `{trigger="timer"} = 10`, with head events arriving inside 250ms at the median.

The same race is reachable on the current pin without any Lighthouse change, because the timer arm snapshots at 4s while Anchor's proof deadline is 2/3 slot.

Closes #1260. Follows #1245.

## Change Overview (Required)

The metadata service's publisher now drains both classes of a decided outcome. Aggregates keep their existing path; contributions are assembled from the decided message plus the reconstructed threshold signature and POSTed one call per `(validator, subcommittee)` identity. The Lighthouse contribution callback becomes inert at Boole+, exactly like the aggregate gate in #1245.

Reading order: `lib.rs` first (the outer gate and the deleted per-committee helper), then `aggregator_post_consensus.rs` (`publish_one_contribution` and the both-class drain), then `metadata_service.rs` (the POST helper).

Contributions bound their signature wait at slot end rather than sharing the aggregate two-slot deadline. Beacon nodes accept a contribution only during its own slot (`sync_committee_verification.rs` floors at the current slot, unlike aggregates' one-epoch tolerance), so a later quorum could only produce a rejected POST. The campaign observed exactly that on Lighthouse's pre-Boole path: `PastSlot { message_slot: Slot(141), earliest_permissible_slot: Slot(142) }`.

Unchanged: pre-Boole behavior, `AGGREGATOR_COMMITTEE_PUBLISH_TOTAL` stays aggregate-only, no new metrics, no Lighthouse-side changes, no eligibility validation.

Removing the callback also removed the last readers of several structures, which this PR deletes rather than leaves dangling: the post-consensus retention map's values (now a key set, so a finished outcome is freed when its publisher completes instead of being pinned for two epochs), the `ConsensusDataNotFound` variant, `drain_signatures`' deadline parameter, and the metadata service's own fork-schedule handle. Both callback gates and the publisher's registration gate now call one predicate, so "exactly one path publishes per duty" is structural rather than argued in three doc comments.

- The gate and the publisher must land together. Separating them leaves a state that either double-publishes or publishes nothing.
- Publication stays process-scoped in-memory exactly-once with no durable record: a restart can lose an in-flight slot, the same trade already accepted for aggregates.
- Committee-level join failures count once per committee on both class counters: an undercount against multiple lost roots, an overcount for a class the committee had no duties in. Documented in code and pinned by a test. Making it exact means threading per-class counts through the publisher for a signal that only moves during already-alarming QBFT failures.
- A contribution missing from consensus data (first-node `Ok(None)` fetch, #1236) still cannot be published by any downstream mechanism. Out of scope here.
- Known growth point: the publisher carries four generic parameters for its two publish operations. A third duty class would double that again; a small trait is the natural fix, deliberately kept out of a behavior-change PR.
- Peak request count rises: one POST per contribution identity rather than one batch per subnet. This matches go-ssv, and it isolates failures (a bad identity no longer fails its siblings' POST).

## Validation (Required)

- `cargo nextest run -p anchor_validator_store`: 79 passed, 0 failed.
- `cargo clippy -p anchor_validator_store -p client --all-targets -- -D warnings`: clean.
- New `testing/committee_contribution.rs` pins the inert-callback contract. The previous "contributions-only committee never publishes" assertion is inverted. Added coverage for mixed outcomes, contribution-only committees, a late publisher reading the cached execution, cross-class quorum-miss independence, dedup, and multi-subcommittee assembly.
- Devnet A/B above, 238 post-Boole slots each, identical topology and Lighthouse pin, image as the only variable.
- go-ssv parity checked field by field against the Boole runner (`protocol/v2/ssv/runner/aggregator_committee.go` at `48d4f3ac5`): same message contents, same `ContributionAndProof` domain at the duty slot's epoch, one contribution per POST, continue-past-failure on submit errors.

Revert the commit. No config, data, or operational impact. Reverting restores the silent loss under head-triggered sync duties.

None for merge. The acute form of the bug activates when Anchor's Lighthouse pin gains head-event-triggered sync committee duties; the milder form is reachable on the current pin.

- The measurement has one trap worth knowing for future campaigns: the `Signed ContributionAndProof` debug line sits inside the old callback pre-fix and inside the publisher post-fix, so it counts different things in each build. The A/B above uses `(slot, subnet)` pairs, a denominator neither build can bias.
- `run_committee_signing` drops to a single caller with this change.

### DIFF
--- a/anchor/client/src/lib.rs
+++ b/anchor/client/src/lib.rs
@@ -780,7 +780,6 @@ impl Client {
             beacon_nodes.clone(),
             executor.clone(),
             spec.clone(),
-            fork_schedule.clone(),
             config.with_weighted_attestation_data,
         );
 
--- a/anchor/validator_store/src/aggregator_post_consensus.rs
+++ b/anchor/validator_store/src/aggregator_post_consensus.rs
@@ -3,31 +3,30 @@
 //! One QBFT decision carries both attestation aggregates and sync contributions for an SSV
 //! committee, and the operator must emit exactly one committee partial-signature message per
 //! `(committee, slot)` covering everything it can sign. The decided value drives both what gets
-//! signed and what gets published: Anchor owns Boole+ aggregate publication outright, through
-//! the metadata service's publisher, because the protocol's unit of agreement is the decided
-//! value and no local Lighthouse view is part of it. (Lighthouse's duty snapshot in particular
-//! is cloned before slot-start selection proofs finish, so routing publication through it would
-//! silently skip late-installed proofs.) Contributions are returned through a Lighthouse
-//! callback that may or may not fire (issue #1227), so no callback can own the committee
-//! message either.
+//! signed and what gets published: Anchor owns Boole+ publication of both classes outright,
+//! through the metadata service's publisher, because the protocol's unit of agreement is the
+//! decided value and no local Lighthouse view is part of it. Lighthouse's duty snapshots are
+//! cloned before slot-start selection proofs finish (both classes share one slot-start
+//! partial-signature batch), so routing publication through either callback would silently skip
+//! late-installed proofs, and a callback may not fire at all (issue #1227).
 //!
 //! The signing set is therefore a property of the duty, not of any consumer. The slot pipeline
-//! starts one execution per committee when it publishes the decided value at 2/3 slot; the
-//! contributions callback and the aggregate publisher only look up their own results. This is
-//! the ssv-spec-normative shape, where the decided value drives what gets signed and local state
-//! only filters.
+//! starts one execution per committee when it publishes the decided value at 2/3 slot, and the
+//! publisher drains that execution's roots of both classes itself; the Lighthouse callbacks
+//! return empty batches. This is the ssv-spec-normative shape, where the decided value drives
+//! what gets signed and published, matching go-ssv's runner-owned post-consensus submission.
 
 use std::{
     collections::{HashMap, HashSet, hash_map::Entry},
     future::Future,
-    sync::Arc,
+    sync::{Arc, LazyLock},
 };
 
 use bls::Signature;
 use database::{NonUniqueIndex, UniqueIndex};
 use futures::{
     FutureExt, StreamExt,
-    future::{BoxFuture, Shared},
+    future::{BoxFuture, Either, Shared},
     stream::FuturesUnordered,
 };
 use qbft_manager::ConsensusDecider;
@@ -43,19 +42,21 @@ use tokio::time::Instant;
 use tracing::{Instrument, debug, error, info, info_span, warn};
 use types::{
     AggregateAndProof, Attestation, AttestationBase, AttestationElectra, ContributionAndProof,
-    Domain, EthSpec, ForkName, Hash256, SelectionProof, SignedAggregateAndProof, SignedRoot, Slot,
+    Domain, EthSpec, ForkName, Hash256, SelectionProof, SignedAggregateAndProof,
+    SignedContributionAndProof, SignedRoot, Slot,
 };
+use validator_metrics::IntCounterVec;
 
 use crate::{
     AggregationAssignments, AnchorValidatorStore, CollectionMode, Error, SigningRequest,
     SpecificError, metrics,
 };
 
-/// Epochs to retain finished `AggregatorCommittee` post-consensus executions.
+/// Epochs to retain started `AggregatorCommittee` post-consensus keys.
 ///
-/// Matches the message validator's per-signer duplicate-detection window, so a very late Lighthouse
-/// callback finds the execution it belongs to rather than nothing. Each entry is one small future
-/// per committee-slot.
+/// Matches the message validator's per-signer duplicate-detection window, so a repeated
+/// `(committee, slot)` inside the window finds its registered key rather than starting a second
+/// execution. Each entry is one key, not a handle; the executions live in their detached tasks.
 const AGGREGATOR_POST_CONSENSUS_RETAIN_EPOCHS: u64 = 2;
 
 /// A result shared between the detached task producing it and any number of joiners.
@@ -74,8 +75,8 @@ pub(crate) struct PreparedRoot<M> {
 }
 
 /// Everything this operator signs for one decided `AggregatorCommittee` round, keyed by signing
-/// identity: validator index for aggregates (read by the aggregate publisher),
-/// `(validator index, subcommittee index)` for contributions (read by the Lighthouse callback).
+/// identity: validator index for aggregates, `(validator index, subcommittee index)` for
+/// contributions. Both classes are read by the metadata service's publisher.
 pub(crate) struct AggregatorPostConsensusOutcome<E: EthSpec> {
     /// Fork the decided value's SSZ payloads were decoded under (its `DataVersion`). The
     /// publisher derives its HTTP endpoint choice and fork header from this, so they cannot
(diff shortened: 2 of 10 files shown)


shane-moore:
@claude, please take a look

shane-moore:
Thanks @claude — all three notes are fair. Two are already documented trade-offs (the `no_aggregates` label shift and the join-failure double-count); the third caught a real process miss, so I ran the mandated `tester-subagent` retrospectively over the test diff. It found a gap worth closing, pushed in b8040813 (tests only, production code untouched):

**The per-class contribution deadline had zero coverage.** Every root-level failure test makes signature collection resolve to an error, which reaches `await_root_signature`'s `other_error` arm — no test ever reached a per-root *deadline expiry* for either class. Collapsing contributions back onto the aggregate's two-slot deadline, undoing the reason this asymmetry exists, passed all 79 tests.

`contributions_stop_at_slot_end_while_aggregates_keep_waiting` now pins it: with the clock between the two deadlines and collection hanging, the contribution roots must give up and count `timeout` while the aggregate must still be waiting. I verified it fails when the deadline is reverted, so it pins the behavior rather than merely executing it. That needed a hang mode on the mock collector, since failing and hanging are different (failure resolves to an error; hanging leaves the reader's deadline to decide).

Also added the contribution half of POST-failure containment, and renamed two tests whose vocabulary described the design this branch replaced (there is no execution cache, and registration is a `HashSet` insert rather than a vacant entry).

Deliberately left as follow-ups: fork-boundary coverage of `lighthouse_owns_publication` (every harness pins all forks at epoch 0, so no test varies the epoch), retention-window pruning (all tests publish at `TEST_SLOT = 1`, so the cutoff saturates and nothing is ever pruned), and the `drain_signatures` error arm (pre-existing). On coverage: `metadata_service.rs` sits at 0% because the POST helpers need HTTP-mock infrastructure the repo doesn't have — same trade #1245 made, and the reason the publisher is generic over the publish operation so tests can inject recorders.

81 tests pass, clippy clean.

shane-moore (commented):
Reviewed at `e777b110`. I traced the Boole publication ownership gate, decided sync contribution construction and signing, failure isolation, pre-Boole behavior, pinned Lighthouse integration, and the added deadline and POST-failure regression coverage. I found no actionable issues.

Reviewed by gpt-5.6-sol max.

Issue #1271, opened 2026-08-24
fix(message_validator): derive per-epoch duty counts from the slot ring (SIP-94 §7 retention)


shane-moore:
## Goal

Make `Role::ProposerPreferences` duty-count retention comply with SIP-94 §7 by deriving per-epoch duty counts from the existing slot ring and deleting the two-bucket epoch counters (`max_epoch`, `curr_epoch_duties`, `prev_epoch_duties`) from `OperatorState`.

## Context / motivation

SIP-94 §7 now requires validation state (recorded signing roots, duty counts) to be retained for a message's full validity window ([added in `d03a2ed`](https://github.com/shane-moore/SIPs/blob/d03a2ed2a561a6a62035efd6d05135c88fef384e/sips/epbs_support.md#L366), from [this review comment](https://github.com/ssvlabs/SIPs/pull/94#discussion_r3838479623)). For role 8 the window is `(1 + MIN_SEED_LOOKAHEAD) * SLOTS_PER_EPOCH + 2` slots, and acceptable message slots span up to four consecutive epochs at any instant: from wall-clock slot `T` the acceptable range is `[T - 2, T + 64]`, whose far edge is always in `epoch(T) + 2` and whose near edge reaches `epoch(T) - 1` during an epoch's first two slots.

Anchor already complies on the recorded-roots half: the role-scoped ring is 128 slots for role 8 (`stored_slot_count`, `anchor/message_validator/src/lib.rs:750-759` at `8a508ae8`), pinned by `test_proposer_preferences_ring_avoids_lookahead_collision` (`partial_signature.rs:4362-4484`).

The duty-count half does not comply. `OperatorState` keeps only `curr_epoch_duties` / `prev_epoch_duties` keyed off a rolling `max_epoch` (`anchor/message_validator/src/duty_state.rs:194-198`), `get_duty_count` returns 0 for anything older than `max_epoch - 1` (`duty_state.rs:218-224`), and the writer relabels buckets on epoch advance (`duty_state.rs:306-321`). With four concurrent role-8 epochs this mis-tracks two ways:

- An accepted message for epoch `E+2` sets `max_epoch = E+2` and relabels epoch `E`'s count as `E+1`'s; a later still-valid message for `E` then reads count 0.
- The `Ordering::Less` arm conflates all older epochs into `prev_epoch_duties`, so `E` and `E+1` share one bucket.

[...]

Issue #1238, opened 2026-08-11
test(message_receiver): pin receive() propagation-source threading via a message_validator test-utils feature


petarjuki7:
## Goal

Pin the behavioral half of the per-peer duplicate fix from #1131 / PR #1203: `NetworkMessageReceiver::receive` must pass `Some(propagation_source)` (not `None`) into `Validator::validate` (`message_receiver/src/manager.rs:83`). If that regresses to `None`, `received_from.is_some()` is false for every inbound message, so every role-8 duplicate classifies as the Ignore-class relay and the whole existing suite still passes.

## Why it isn't covered today

The per-peer matrix tests in #1203 call `validate_partial_signature_message` directly with their own `PeerId`, so they never exercise the `receive` seam. A `receive`-level guard needs a real role-8 `SignedSSVMessage` that survives full validation — including RSA verification against the operator pubkey seeded into `NetworkState`. That requires a matched pair: the message signed by the same key whose pubkey is in the DB. The fixtures that build this (`create_signed_proposer_preferences_message`, `create_operator_pub_keys`, `MockDutiesProvider`, `create_committee_info`) are `pub(crate)` inside `message_validator`'s `#[cfg(test)]` module and are unreachable from `message_receiver`; `message_validator` has no `test-utils` feature to expose them.

## Suggested approach

1. Add a `test-utils` feature to `message_validator` that exposes the role-8 message / committee / duties fixtures, mirroring `database`'s existing `test-utils` feature.
2. Add a `message_receiver` test (dev-dep on `message_validator` with `test-utils` and `database` `test-utils`): construct a `NetworkMessageReceiver`, spawn a processor, `is_synced = true`, seed a committee/share into `NetworkState`, then call `receive` twice from the same peer with the same message and assert the second `Outcome.action` is `Reject`.

## Acceptance criteria

- The new test fails if `manager.rs:83` is changed to `None`.
- No production logic change beyond gating existing test helpers behind the `test-utils` feature.

## Notes

- Follow-up to #1131 / PR #1203, split out to keep that PR focused (per review discussion).

shane-moore:
Obsolete after #1272, not superseded.

This issue's acceptance criterion is "the new test fails if `manager.rs:83` is changed to `None`". #1272 makes role-8 classification peer-agnostic, so `Some(peer)` and `None` now produce identical output at every observable boundary. The criterion cannot be satisfied by any behavioural test, and the `test-utils` feature proposed here loses its justification along with it.

[...]

PR #1272, opened 2026-08-24
fix(message_validator): ignore repeated proposer-preference roots regardless of peer

### LINKED ISSUE #1254: fix(message_validator): ignore repeated proposer-preference roots regardless of peer
## Goal

Align Anchor `Role::ProposerPreferences` duplicate handling with SIP-94: once a signing root has been accepted, every repeat of that root is IGNORE regardless of the propagation peer.

## Context / motivation

PR #1203 implemented the previous peer-aware rule: same-peer repeats are REJECT, while relays from another peer are IGNORE.

[SIP-94 now requires both cases to be IGNORE](https://github.com/ssvlabs/SIPs/blob/5d1df2b25f61b305f0ac384eb81d88518f1f6177/sips/epbs_support.md#L375-L379) because an honest retry or restart can repeat an accepted root after the gossip duplicate cache expires. Repetition alone does not prove peer fault.

Anchor currently compares the propagation peer for an already-recorded role-8 root and returns Reject-class `DuplicatedMessage` for a same-peer repeat (`anchor/message_validator/src/duty_state.rs:140-150`).

## Suggested approach

- When a proposer-preference root was recorded previously, return the existing Ignore-class duplicate result without comparing propagation peers.
- Keep the existing peer-state shape and `received_from` plumbing unchanged.
- Preserve membership-before-cap ordering, the four-distinct-root limit, and insertion only after all preceding validation succeeds.
- Update only the comments and tests that encode the previous peer-specific verdict.

## Acceptance criteria

- The first four distinct roots remain eligible for ACCEPT.
- Any previously accepted root is IGNORE, independent of propagation peer.
- A fifth distinct root is IGNORE and is not recorded.
- A repeated root remains IGNORE when the distinct-root set is full.
- Invalid messages do not populate the accepted-root set.
- Duplicate handling for every other role is unchanged.

## Tests

- Update the role-8 peer matrix so same-peer and different-peer repeats both expect IGNORE.
- Retain coverage for repeats at capacity, fifth-root retries, and invalid-message non-poisoning.
- Run `cargo test -p message_validator proposer_preferences`, `make cargo-fmt-check`, and `make lint`.

## Notes

- Implement after #1193 lands.
- Supersedes #1238, which would pin the previous same-peer REJECT behavior.
- Revisit #1148 separately: its peer-penalty rationale is stale, although suppressing redundant broadcasts may remain an optimization.
- Scope is Anchor only. No SIP, go-ssv, ssv-spec, or #1199 changes are included.

shane-moore:

- SIP-94 §7 classifies any repeat of a recorded `RoleProposerPreferences` signing root as IGNORE regardless of the propagation peer. Anchor currently returns Reject-class `DuplicatedMessage` for a same-peer repeat, so an honest operator retry or restart (re-sending a deterministic partial after the recipient's gossip duplicate cache expires) draws a gossipsub peer-score penalty against an honest cluster peer.
- The SIP rule was updated on 2026-08-17 for exactly this reason: repetition does not prove peer fault ([SIP-94 §7 dedup rule](https://github.com/shane-moore/SIPs/blob/5d1df2b25f61b305f0ac384eb81d88518f1f6177/sips/epbs_support.md#L375-L379)).
- Only bites while synced: the receiver already downgrades Reject to Ignore during sync.
- Closes #1254. Supersedes #1238. Follow-up to #1203, which implemented the previous peer-aware rule.

## Change Overview (Required)

- The role-8 membership branch in `DutyState::update_for_partial_signature` now returns Ignore-class `RelayedDuplicateMessage` for any recorded-root repeat; the peer comparison and the `None`-own-emission guard are deleted.
- Reading order: `duty_state.rs` first (the behavior change, one small branch), then `partial_signature.rs` (five test expectations flipped, stale peer-rationale comments swept, one new assertion), then `lib.rs` (variant doc comment only).
- Intentionally unchanged: membership-before-capacity ordering, record-only-on-accept, the 4-distinct-root cap, the `seen_preferences` map shape and `received_from` plumbing (kept per the issue's scope), and `DuplicatedMessage` handling for every other role (`message_counts.rs` untouched).

- Reject-to-Ignore removes peer-score pressure against same-peer repeat flooding. Bounded: gossipsub's seen-cache absorbs immediate repeats, Ignore is not re-forwarded, and only committee operators pass RSA signature validation at all; no new cost class, since the expensive verify already ran for other-peer relays under the old rule.
- Same-peer repeats and relays become indistinguishable in logs (both `RelayedDuplicateMessage`). The variant doc records the widened meaning; renaming it was considered and skipped to keep the diff minimal.

## Validation (Required)

- Tests-first: the five flipped expectations failed against the previous code, then passed after the branch change.
- New fifth-root retry assertion pins the SIP's MUST-NOT-record clause: a retried over-cap root hits `TooManyDistinctSigningRoots` again rather than the membership branch.
- `cargo test -p message_validator`: 123 passed. `make cargo-fmt-check` and `make lint` clean.

- Single-commit revert. In-memory verdict classification only; no config, data, or wire impact.

- None. #1271 (duty-count retention) is sequenced after this and rebases over it.

### DIFF
--- a/anchor/message_validator/src/duty_state.rs
+++ b/anchor/message_validator/src/duty_state.rs
@@ -29,9 +29,10 @@ use crate::{FIRST_ROUND, ValidationFailure, message_counts::MessageCounts};
 /// preference inputs change between emissions — chiefly a `dependent_root` shift under reorg (the
 /// SIP-94 §5 re-emission trigger), and also `target_gas_limit` / `fee_recipient` config changes
 /// across operator restarts. Per SIP-94 §7 the cap is policy headroom for a few realistic
-/// reorg-driven corrections while bounding spam — not a safety/consensus bound. That is why
-/// exceeding it is an Ignore (`TooManyDistinctSigningRoots`), whereas a repeat of an already-seen
-/// root stays a Reject-class duplicate.
+/// reorg-driven corrections while bounding spam, not a safety/consensus bound. Both exceeding
+/// the cap (`TooManyDistinctSigningRoots`) and repeating an already-recorded root
+/// (`RelayedDuplicateMessage`) are Ignore, regardless of the propagation peer; rationale at the
+/// enforcement site in `update_for_partial_signature`.
 const MAX_PROPOSER_PREFERENCES_DISTINCT_ROOTS: usize = 4;
 
 /// Test-only, crate-visible mirror of the private cap so pipeline tests in sibling modules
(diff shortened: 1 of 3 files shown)


shane-moore (commented):
Reviewed at a1c84a21 against epbs at a4eb597a. I traced the peer-agnostic role-8 duplicate classification, validation-before-record ordering, membership-before-capacity behavior, the comment cleanup in cbfe930e, and the current regression coverage. I found no blocking issues.

The remaining received_from and HashMap cleanup is already accepted as a non-blocking follow-up. All required CI checks, including cargo-udeps and the local testnet, pass.

Reviewed by gpt-5.6-sol max.

PR #1274, opened 2026-08-25
fix: resolve trait-solver recursion overflow on recent nightly

shane-moore:

`cargo-udeps` fails on **every** branch, not just new PRs. It is the only nightly job in CI, so nothing else catches this.

- Last green: nightly `e71c0f1e3` (2026-08-18), run [32275248291](https://github.com/sigp/anchor/actions/runs/32275248291).
- Every run from 2026-08-24 onward fails on nightly `fb6531d55` (2026-08-23), including three unrelated dependabot runs ([32697425240](https://github.com/sigp/anchor/actions/runs/32697425240), [32698198029](https://github.com/sigp/anchor/actions/runs/32698198029), [32700599160](https://github.com/sigp/anchor/actions/runs/32700599160)) and #1272.
- Cause: rustc's next solver now tracks recursion depth accurately during trait solving, roughly doubling the depth the same code needs ([rust-lang/rust#159228](https://github.com/rust-lang/rust/issues/159228)).

Reproduced locally on `nightly-2026-08-24` (= the exact CI compiler). Two symptoms with **different** causes:

| | `eth/src/sync.rs:598` | `validator_store/src/metadata_service.rs` |
|---|---|---|
| severity | hard `E0275`, fails the build | future-incompat warning |
| `#![recursion_limit]` | no effect at 256 / 1024 / 4096 | fixes it |
| real cause | genuine solver cycle | depth; default 128 just short (144 clears it) |

`fetch_logs` returned `impl Future + Send` and called `subdivide_fetch_logs(..).boxed()`, which called `fetch_logs` again. Proving one future is `Send` requires proving the other is, so no limit resolves it. Pinning the toolchain would only defer something the new solver will never accept once it reaches stable.

No issue filed for this; I searched before opening.

## Change Overview

Both crates now cut the proof chain structurally instead of raising a budget.

**Start with `anchor/eth/src/sync.rs`** — the mutual recursion between `fetch_logs` and `subdivide_fetch_logs` is replaced by an explicit worklist. `fetch_logs` holds a stack of pending `(from, to)` ranges; on a response-size error it pushes the two halves rather than recursing. `subdivide_fetch_logs` is deleted: its only caller was the recursion, and `subdivision_factor` was hardcoded `2` at that one site. Net −21 lines, no `.boxed()`, no allocation, and the returned future no longer captures `&self`.

**Then `metadata_service.rs`** — two `.boxed()` calls plus a comment. The `Send` proof for the spawned phase-3 task walked from `executor.spawn` down through `tokio::join!`'s `MaybeDone` nesting into both fetch futures; half the depth was join! internals. `BoxFuture` is unconditionally `Send`, so the walk stops there.

Deliberately unchanged: fetch order, RPC call pattern, error classification and messages, metric labels and timer semantics, and every public signature.

**Blast radius is the main risk.** `sync.rs` is the sole production path by which SSV contract events (`OperatorAdded`, `ValidatorAdded`, `ClusterLiquidated`, …) reach the database — `process_logs` has exactly two production callers, both in this file. `process_logs` buffers per block and flushes on a block-number change, so it **assumes ascending log order**, and the subdivision path is exactly where ordering could break.

Mitigated by differential-testing the old and new splitting algorithms over 53,104 scenarios (see Validation). Zero mismatches on order, call pattern, or error outcome.

**Known gap:** the subdivision path has no test, before or after this PR. The existing `anchor/eth/tests/` suites call `process_logs` directly with pre-built logs and never go through `fetch_logs`. The differential test validates a transcription of the algorithm, not the shipped Rust.

**Behavior change, intentional:** per-recursion-level `#[instrument]` spans are gone, since there is only one function now. `from`/`to` moved onto the `debug!`/`info!` events, so the range is still in the logs, but the span tree is flatter. The old messages carried no range of their own and relied entirely on span context.

**Trade-off:** boxing costs two heap allocations per phase-3 run, twice per slot, against a 2s beacon-API timeout in the same function.

## Validation

On `nightly-2026-08-24` (`fb6531d55`), the exact failing compiler:

- `cargo check --workspace --tests --all-targets` — clean; zero `E0275`, zero `recursion_depth_exceeding_limit`. Confirmed failing on this branch's parent.
- `cargo fmt --all --check` — clean.
- `cargo clippy --workspace --tests -- -D warnings -D clippy::allow_attributes` — clean.
- `cargo test --workspace` — all green (requires `git submodule update --init` for `spec_tests`).

Differential test of the two splitting algorithms, comparing the full sequence of attempted ranges, the sequence of successful ranges, the returned log order, and the error outcome including which range failed:

- exhaustive over every subset of failing sub-ranges for spans 1–7
- 20k randomised size-threshold oracles (the realistic "provider caps response size" case)
- 20k randomised non-monotone flaky oracles

53,104 scenarios, 0 mismatches.

Filter builder reordering verified against `alloy-rpc-types-eth-1.7.3/src/filter.rs:499,506,574,588` — `block_option`, `address`, and `topics` are independent fields.

## Rollback

Plain revert; no config, data, or operational impact. Reverting restores the nightly CI failure.

## Blockers / Dependencies

`epbs` carries the identical `sync.rs`, so it stays red until an `unstable` → `epbs` merge follows this. That is what unblocks #1272.

Worth adding a test for the subdivision path, either by extracting the split decision as a pure function or by mocking the provider to force `ErrorResp` on large ranges. Left out here to keep this focused on unbreaking CI.

### DIFF
--- a/anchor/eth/src/sync.rs
+++ b/anchor/eth/src/sync.rs
@@ -1,5 +1,5 @@
 use std::{
-    cmp::{max, min},
+    cmp::min,
     collections::{HashMap, VecDeque},
     sync::Arc,
 };
@@ -13,7 +13,7 @@ use alloy::{
     transports::{RpcError, TransportErrorKind},
 };
 use database::{NetworkDatabase, SlashingProtection};
-use futures::{FutureExt, Stream, StreamExt, stream::FuturesOrdered};
+use futures::{Stream, StreamExt, stream::FuturesOrdered};
 use reqwest::Url;
 use sensitive_url::SensitiveUrl;
 use ssv_network_config::SsvNetworkConfig;
@@ -545,7 +545,11 @@ impl SsvEventSyncer {
         Ok(())
     }
 
-    // Construct a future that will fetch logs in the range from_block..to_block
+    // Construct a future that will fetch logs in the range from_block..to_block.
+    //
+    // RPC providers cap the size of a `getLogs` response, so a range that is too busy is rejected
+    // rather than returned. When that happens the range is split in half and each half is fetched
+    // separately, repeating down to a single block before giving up.
     #[instrument(skip(self, deployment_address, events), level = "debug")]
     fn fetch_logs(
         &self,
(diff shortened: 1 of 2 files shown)


shane-moore (commented):
Reviewed at `642db052`.

Ship-shaped, no blocker. The explicit worklist preserves the old half-range boundaries, lower-first RPC and append order, terminal error behavior, and both historical and live callers. Boxing the phase-3 futures preserves concurrent polling and their individual deadlines. `cargo-udeps` and the aggregate test suite pass at this head.

One minor observability mismatch: when a request subdivides, `EXECUTION_LOG_FETCH_TIME` now records only the failed RPC attempt for the parent range. Previously, that parent timer stayed live across the recursive child fetches, so it measured the full subtree. Per-attempt timing may be cleaner, but the PR currently says timer semantics are unchanged. I would update that description, or explicitly preserve or document the old metric contract if downstream dashboards depend on it.

Reviewed by gpt-5.6-sol max.

Issue #1273, opened 2026-08-24
refactor(message_validator): delete the unused received_from plumbing and peer-keyed preference map


shane-moore:
## Goal

Delete the `received_from: Option<PeerId>` parameter and convert `SignerState::seen_preferences` from `HashMap<Hash256, Option<PeerId>>` back to `HashSet<Hash256>`. Behaviour-neutral.

## Context and motivation

PR #1272 made role-8 duplicate signing-root classification peer-agnostic per SIP-94 §7, deleting the peer comparison at `anchor/message_validator/src/duty_state.rs:143`. That removed the only read of the stored propagation peer. The value is now written at `duty_state.rs:156` and read nowhere.

`received_from` is a pure pass-through across five signatures, starting at the public `MessageValidator::validate` (`lib.rs:473`), through `validate_decoded_message` (`lib.rs:499`), `validate_ssv_message` (`lib.rs:767`), `validate_partial_signature_message` (`partial_signature.rs:27`), into `DutyState::update_for_partial_signature` (`duty_state.rs:107`). One production call site (`manager.rs:83`). Passing `Some(peer)` or `None` now produces identical output at every observable boundary.

#1238 states the consequence in its own words, written while the read still existed: "If that regresses to `None`, `received_from.is_some()` is false for every inbound message, so every role-8 duplicate classifies as the Ignore-class relay and the whole existing suite still passes." That is now the specified behaviour in every case, so #1238's acceptance criterion is unsatisfiable and it should close as obsolete.

The justification is maintenance clarity, not runtime cost. The field's presence teaches every reader, and taught #1238, that peer identity affects role-8 validation. State is tightly capped at 4 roots per signer-slot, so there is no meaningful memory argument.

Role 8, `received_from`, and `seen_preferences` exist only on `epbs`; none are present on `unstable`.

## Suggested approach

- Drop the parameter from the five signatures and four forwarding sites listed above, and update `manager.rs:83`.
- `seen_preferences` becomes `HashSet<Hash256>` (`duty_state.rs:343`, `:355`, `:156`), reverting to the shape it had before #1131 / PR #1203.
- Drop `use libp2p::PeerId` from `duty_state.rs:6`, `lib.rs:17`, `partial_signature.rs:4`. Keep `libp2p` as a dependency: `lib.rs:18` re-exports `MessageAcceptance`.
[...]

PR #1276, opened 2026-08-26
fix(message_validator): derive per-epoch duty counts from the slot ring (SIP-94 §7 retention)

shane-moore:

SIP-94 §7 requires duty counts to be retained for every epoch with slots inside a role's message acceptance window. Role 8 (`ProposerPreferences`) accepts slots spanning up to four concurrent epochs, but `OperatorState` kept only two rolling epoch buckets: accepting an epoch E+2 message relabeled E's count as E+1's and zeroed reads for E, and the older-epoch arm conflated all past epochs into one bucket. Fail-open admits nothing extra for role 8, but the fail-closed direction (a conflated bucket hitting the limit) drops honest preference shares on small devnets where one validator holds many proposal slots, which presents as a protocol bug during testing.

- Issue: #1271 (last open item in the Proposer Preferences Duty milestone)
- Spec: SIP-94 §7 retention requirement (ssvlabs/SIPs #94)
- Precursor: #1254 / #1272 (same file)

## Change Overview

- Delete the three counter fields and their relabeling writer; `get_duty_count(epoch, slots_per_epoch)` now derives the count by probing the ring at each of the epoch's slot keys. The ring already records exactly one occupied entry per counted duty slot, so counts gain a single source of truth that cannot drift from the dedup state.
- `validate_duty_count` checks `is_first_message_for_duty` before deriving (the count is only meaningful for first messages, and this keeps the walk off the per-message path); takes `&OperatorState`.
- The two now-identical setters merge into one `set_signer_state` (the round-change path's `max_slot` guard was always a no-op).
- Ring sizing: derivation is exact only if the ring also covers a full epoch beyond the acceptance window, so the default grows 64 -> 67 (`2*spe + LATE_SLOT_ALLOWANCE + 1`); role 8 stays at 128 (already exact with headroom). `stored_slot_count` is now exhaustive over `Role` so adding a role forces an explicit sizing decision instead of silently under-sizing.
- Intentionally unchanged: verdict classifications (`ExcessiveDutyCount` stays Ignore), the 4-root dedup rule, validation ordering, and all wire behavior.

Reviewer note: two acceptance criteria in #1271 are deliberately deviated from. "Duty-limit behavior for every other role is unchanged" holds as no-reachable-regression, not byte-identical: derived counts fix the same conflation edges for all limited roles (including VoluntaryExit, whose observed count, unlike its limit, comes from these counters). "No change to ring sizing" was dropped in favor of the 64 -> 67 exactness fix; one existing guard assertion updates accordingly.

Rejected alternative: widening to four buckets keeps the relabeling arithmetic (the bug class being fixed) and a second state that can drift from the ring; go-ssv retains that structure upstream, which is deliberate SIP-backed divergence with no interop surface (Ignore-class decisions only, both implementations).

- The count walk is at most `slots_per_epoch` vector probes, gated to first messages, and sits next to provider lookups and an RSA verify that dominate it. The whole check-then-record sequence runs under the existing per-MessageId shard guard, so no new interleavings.
- Larger default rings cost 3 extra `Option<SignerState>` slots per operator state and can newly (correctly) Ignore a message at a saturated epoch tail that the undersized ring would have admitted.
- Under exact counting, `ExcessiveDutyCount` is unreachable for the two roles whose limit equals `slots_per_epoch` (a full epoch leaves no first message to reject); the limit remains as the SIP-specified backstop.

## Validation

- `cargo test -p message_validator`: 128 passed; `make lint`, `make cargo-fmt-check`, `make test` all clean.
- The two module regressions and the pipeline saturation test all fail under a temporary revert to the bucket code (the saturation test with exactly `ExcessiveDutyCount { got: 32, limit: 32 }`, the fail-closed scenario from the issue).
- Three explicit mutations are pinned by tests, verified by applying each and confirming failure: dropping the slot-equality filter (stale-entry miscounts), deleting the first-message gate (follow-ups of the last allowed duty wrongly rejected), and a count stuck at zero.
- ssv-mini: a mixed 2 Anchor + 2 go-ssv Boole run at this branch head (fork + 6.7 epochs: zero ExcessiveDutyCount, zero errors, duties healthy across all classes, memory flat), plus an all-Anchor Gloas run with role 8 live (preference batches signed, published, and cross-validated by all four operators across epochs 2-5 with zero validation failures).

## Rollback

Single revert; all state is in-memory and rebuilt from gossip after restart. No config, schema, or wire impact.

## Blockers / Dependencies

#1273 removes `received_from` through the same `update_for_partial_signature` signature; whichever lands second rebases (mechanical).

None; no follow-up work is planned from this PR.

### DIFF
--- a/anchor/message_validator/src/consensus_message.rs
+++ b/anchor/message_validator/src/consensus_message.rs
@@ -59,11 +59,8 @@ pub(crate) fn validate_consensus_message(
         validation_context.operator_pub_keys,
     )?;
 
-    duty_state.update_for_consensus_message(
-        validation_context.signed_ssv_message,
-        &consensus_message,
-        validation_context.slots_per_epoch,
-    );
+    duty_state
+        .update_for_consensus_message(validation_context.signed_ssv_message, &consensus_message);
 
     // Return the validated message
     Ok(ValidatedSSVMessage::QbftMessage(consensus_message))
@@ -465,11 +462,11 @@ pub(crate) fn validate_qbft_message_by_duty_logic(
 
     // Rule: valid number of duties per epoch
     for &signer in signed_ssv_message.operator_ids() {
-        let signer_state = duty_state.get_or_create_operator(&signer);
+        let operator_state = duty_state.get_or_create_operator(&signer);
         validate_duty_count(
             validation_context,
             msg_slot,
-            signer_state,
+            operator_state,
             duty_provider.clone(),
         )?;
     }
@@ -1683,7 +1680,7 @@ mod tests {
             vec![],
             vec![],
         );
-        duty_state.update_for_consensus_message(&dummy_signed_msg, &dummy_qbft, 32);
+        duty_state.update_for_consensus_message(&dummy_signed_msg, &dummy_qbft);
 
         // Now validate a consensus message for height 1 (which is "old")
         let result = validate_qbft_message_by_duty_logic(
--- a/anchor/message_validator/src/duty_state.rs
+++ b/anchor/message_validator/src/duty_state.rs
@@ -1,7 +1,4 @@
-use std::{
-    cmp::Ordering,
-    collections::{HashMap, HashSet},
-};
+use std::collections::{HashMap, HashSet};
 
 use libp2p::PeerId;
 use ssv_types::{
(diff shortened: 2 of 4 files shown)


shane-moore (commented):
Reviewed at `d76f4557`. No findings survived adversarial self-audit. The ring-derived count, retention sizing, first-message gate, and state-update ordering match SIP-94 §7, and all current CI checks pass.

Reviewed by gpt-5.6-sol max.

Issue #1279, opened 2026-08-26
feat(client): add builder definitions config for direct builder connections


shane-moore:
## Goal

Give operators a per-node builder configuration file (`<data_dir>/builder_definitions.yml`) feeding the produce-time `BuilderConfig` and the ahead-of-time preferences submission.

## Context / motivation

Anchor's node options are CLI-flag-driven (`client/src/config.rs`), network identity comes from the network-config YAML (`ssv_network_config`), and per-validator values like fee recipient and graffiti come from the SSV contract. There is no operator-editable per-node config file today (no `--config-file`; Lighthouse's `validator_definitions.yml` and graffiti files are not consumed), and per-builder entries fit none of the existing surfaces: they are operator-local config, too structured for flags and not network-shared. The builder-API flow needs per-builder entries (`url`, `auth_data`, `builder_pubkeys`, `max_execution_payment`, `min_bid`, `builder_boost_factor`, the keymanager-APIs #88 vocabulary). Per SIP-94, entries MUST be byte-identical across all operators of a cluster, `auth_data` included: a one-byte divergence silently drops that builder (signing roots split below threshold), visible only in reconstruction telemetry.

## Suggested approach

Prefer reusing Lighthouse's `builder_store` crate (reads `builder_definitions.yml`) over hand-rolling parsing; `serde_yaml` and the `ForkScheduleFile` shape (`ssv_network_config/src/lib.rs:27-35`) are the fallback precedent. Add a `DataDir` accessor (`global_config/src/data_dir.rs:63-91` pattern) and an optional `--builder-definitions <PATH>` override in `PayloadBuildingOptions` (`cli/src/cli.rs:466-546`, `--key-file` pattern). Run docgen (enforced). Rewrite `docs/docs/pages/mev_configuration.mdx`, which currently says relays are configured on the beacon node.

## Acceptance criteria

- Entries parse and validate (zero-length `auth_data` rejected; `auth_data` defaults to the URL bytes).
- Config load rejects more than 8 builder entries per validator (the SIP-94 §5 cap message validation enforces as `MAX_REQUEST_AUTH_DISTINCT_ROOTS`): a 9th entry fails fast at startup rather than having its auth root silently Ignored on gossip with no signal.
[...]

Issue #1278, opened 2026-08-26
feat(validator_store): sign builder request auth via the Lighthouse builder-API pin


shane-moore:
## Goal

Bump the Lighthouse pin to the Gloas builder-API stack and implement the one required trait addition, `sign_request_auth_v1`, as a distributed signing round producing `SignedRequestAuth` per configured builder per proposal slot.

## Context / motivation

sigp/lighthouse#9803/#9804 (merged) plus #9805/#9806/#9807 (open, stacked) implement beacon-APIs #630. The trait seam is clean: from the current pin (`e58ec88fe`, Cargo.toml:76-92) to the stack head, `ValidatorStore` gains exactly one required method and no signature changes to the 23 methods Anchor implements (verified by migration probe 2026-08-26: one production compile error total). The pin bump and the implementation must be one PR to compile. Builds on a feature branch pinned to the stack head; the final pin lands after the stack merges to unstable.

Beyond the trait method, the pin has a runtime obligation the compiler does not catch: `BlockServiceBuilder::build()` now errors unless `.configured_builders(BuilderStore)` and `.request_auth_cache(RequestAuthCache)` are supplied, so client wiring (plus `builder_types`/`builder_store` deps and a DataDir accessor for `builder_definitions.yml`) is part of this PR.

## Suggested approach

[...]

PR #1281, opened 2026-08-26
feat(message_validator): admit builder request auth partial signatures

### LINKED ISSUE #1277: feat(message_validator): admit builder request auth partial signatures
## Goal

Accept and forward `RequestAuthPartialSig` (kind 9) traffic on `Role::ProposerPreferences` without emitting any: add the ssv_types variant, widen the role's kind gate, and give the new kind its own distinct-signing-root budget, per SIP-94 §5/§7 (ssvlabs/SIPs#94).

## Context / motivation

SIP-94's builder request auth extension adds a second partial-signature kind to the proposer-preferences role: one single-entry container per distinct `BuilderRequestAuth` signing root, up to 8 distinct roots per (`MessageID`, signer, `proposal_slot`), tracked separately from the 4-root preferences budget. This issue is deliberately emission-free and lands first: a container with an unknown kind fails SSZ decode and is REJECT-classed (today as `UnknownMessageSlot`, since topic validation's `extract_slot` decodes the container before the partial-signature validator runs), which penalizes the forwarding peer's gossip score, so the network must decode kind 9 before any operator emits it. Bundled as one PR because the two wildcard-less `match PartialSignatureKind` blocks in `message_counts.rs` compile-break on the new variant, and a kind-only PR would still REJECT kind 9 at the role gate (no rollout value alone).

## Suggested approach

- `common/ssv_types/src/partial_sig.rs`: add `RequestAuth = 9`; extend `TryFrom<u64>`; the invalid-variant sentinel tests currently assert 9 fails and bump to 10.
- `message_validator/src/partial_signature.rs`: role-8 arm becomes `kind == ProposerPreferences || kind == RequestAuth` (the disjunction shape of the arms below it).
- `message_validator/src/message_counts.rs`: `RequestAuth` joins the `ProposerPreferences` no-op arms, and both arm comments are updated to name both kinds. Not the shared `pre_consensus` arm: `MAX_MESSAGES_PER_ROUND = 1` would cap it at one packet per slot and defeat the budget.
- `message_validator/src/duty_state.rs`: `MAX_REQUEST_AUTH_DISTINCT_ROOTS = 8` beside `MAX_PROPOSER_PREFERENCES_DISTINCT_ROOTS`; sibling `seen_request_auth` map on `SignerState`; the `ProposerPreferences` kind check in `update_for_partial_signature` becomes a match selecting (map, cap), keeping the membership-before-capacity ordering so a recorded root stays IGNORE when the set is full. (#1276, merged, refactored this file; locate by symbol.)
- `message_validator/src/lib.rs` (doc-only): generalize the `RelayedDuplicateMessage` doc comment, currently written as proposer-preferences-specific, to cover both role-8 kinds.
- Everything else is role-keyed and needs no edits: fork gate, duty limit (kind-9 packets on existing duty slots do not increment it), retention ring, one-entry rule, earliness/lateness, monotonic-slot exemption, proposer-assignment check, doppelganger freshness exemption.

## Acceptance criteria

- `PartialSignatureKind::RequestAuth as u64 == 9`; SSZ round-trips as `[9,0,0,0,0,0,0,0]`.
- Role 8 admits kinds 8 and 9; every other role REJECTs kind 9 (`PartialSignatureTypeRoleMismatch`).
- Budget independence: 4 recorded type-8 roots at one (`MessageID`, signer, `proposal_slot`) consume none of the 8 type-9 roots, and vice versa.
- Repeat of a recorded type-9 root is IGNORE regardless of peer; a 9th distinct root is IGNORE.
- One-entry rule and pre-Gloas fork gate apply to kind-9 containers with no code change.

## Tests

- Extend the four existing exhaustive enum test lists (round-trip, byte-layout table, TryFrom-all-valid, tree-hash uniqueness) with the new variant; bump both invalid sentinels 9 to 10.
- Admission: kind 9 decodes and role 8 accepts it, exercised through the topic-validation path where today's `UnknownMessageSlot` REJECT occurs.
- Role gate: every other role rejects kind 9, swept via the test-only `EnumIter` derive on `Role`.
- Budget state machine: 8 distinct roots accept; the 9th is IGNORE (`TooManyDistinctSigningRoots`); retrying the rejected 9th still IGNOREs (proves it was not recorded); repeating an accepted root with the set full is `RelayedDuplicateMessage` (membership before capacity, mirroring the existing kind-8 tests).
- One interleaved budget-independence test covering both directions.
- Deliberately omitted as coverage padding (role-keyed code already pinned by kind-8 tests): kind-9-specific pre-Gloas, one-entry, doppelganger, and cross-peer duplicate tests.

## Notes

Deploy well before any emitter. Constants match ssv-spec ssvlabs/ssv-spec#632 (`RequestAuthPartialSig = 9`); that PR is still open, re-verify the value at implementation time.

Adversarial review outcome (2026-08-26): the 8-root budget was challenged as speculative because this issue ships no emitter; it is retained because admission itself opens the ACCEPT path for any registered operator key, and SIP-94 §7 couples the budget to admission (dropping it would force a second network-visible classification change later). The budget bounds ACCEPT amplification and retained state, not signature-verification CPU (RSA verify precedes the budget check); a pre-verification budget check was considered and rejected.

shane-moore:

- SIP-94's builder request auth extension adds a second partial-signature kind to the proposer-preferences role: `RequestAuthPartialSig = 9`, one single-entry container per distinct `BuilderRequestAuth` signing root. Today a kind-9 container fails SSZ decode during topic validation (`extract_slot` returns `None`, so the packet REJECTs as `UnknownMessageSlot`), penalizing the forwarding peer's gossip score on synced nodes.
- Per SIP-94 §5's deployment note, the wire variant must be decodable network-wide before any operator emits it, so this admission-only change lands first and soaks; the emitters build on it separately.
- Closes #1277. Spec: SIP-94 §5/§7 (ssvlabs/SIPs#94); wire constant pinned by ssv-spec ssvlabs/ssv-spec#632. Milestone sequencing: #1278/#1279/#1280 depend on this propagating.

## Change Overview (Required)

- `PartialSignatureKind::RequestAuth = 9` added; the role-8 gate admits kinds 8 and 9; kind 9 gets its own budget of 8 distinct signing roots per (`MessageId`, operator, `proposal_slot`), tracked separately from the 4-root preferences budget (SIP-94 §7: neither consumes the other). Membership is checked before capacity so a recorded root stays IGNORE when the set is full; duplicate-root and over-cap are both Ignore-class.
- Both kinds' root sets now live in one lazily-boxed `SigningRootBudgets` on `SignerState`, allocated on the first role-8 packet, so the other roles' ring entries shrink from an inline map to one pointer.
- Reading order: `ssv_types/partial_sig.rs` (wire kind) -> `partial_signature.rs` (role gate) -> `duty_state.rs` (budget + storage) -> `message_counts.rs` / `lib.rs` (compile-forced no-op arms, doc generalization).
- Intentionally unchanged: every other role-8 rule is role-keyed and inherited with zero edits (fork gate, earliness/lateness, duty limit, one-entry rule, proposer-assignment check, monotonic-slot exemption, doppelganger freshness exemption). No code emits kind 9. Kind-8 classification behavior is unchanged (error `got` strings now use the kind's Debug name).

- Rollout: nodes without this change REJECT kind-9 traffic and penalize forwarding peers, which is why this ships and propagates well before any emitter.
- Accepted kind-9 roots spawn passive signature collectors on share-holding nodes (same shape as kind 8 today); bounded by the 8-root budget and the collector slot-cutoff cleaner, and reaching the ceiling requires a registered operator's key.
- The 8 cap is the SIP §5 policy bound (wire container allows 64); provenance is documented at the constant, and #1279 carries a config-load fail-fast acceptance criterion so a >8-entry config cannot silently lose auth roots on gossip.

## Validation (Required)

- New tests pin the full request-auth state machine: 8 distinct roots accept, the 9th is IGNORE and stays unrecorded on retry, a repeat of an accepted root at a full set is IGNORE as a duplicate (membership before capacity), a root recorded under kind 8 is not a duplicate under kind 9, and an interleaved test proves the two budgets are independent in both directions.
- Role sweep asserts every non-role-8 role rejects kind 9, with a compile-time exhaustiveness guard on the role list; SSZ layout test pins `[9,0,0,0,0,0,0,0]`.
- `make cargo-fmt-check` and `make lint` clean; `cargo test -p ssv_types --lib` 142 passed; `cargo test -p message_validator --lib` 132 passed; `cargo check --workspace` green.
- Reviewed against SIP-94 §7 rule-by-rule (role table, dedup bullet, deployment note) plus four independent review lanes; the SIP's "type-9 adds no distinct duty slots" line is implemented via the stricter shared-ring-entry reading, noted at the duty-limit arm.

- Revert the commit. No config, data, or operational impact: no emitter exists, and reverted nodes simply return to REJECTing kind-9 traffic.

- None to merge. #1278/#1280 (emitters) must not ship until this has propagated; re-verify the ssv-spec #632 constant (still an open PR) before any emitter lands.

### DIFF
--- a/anchor/common/ssv_types/src/partial_sig.rs
+++ b/anchor/common/ssv_types/src/partial_sig.rs
@@ -44,6 +44,10 @@ pub enum PartialSignatureKind {
     // ProposerPreferences object (validator-scoped, non-QBFT; the reconstructed
     // SignedProposerPreferences is gossiped on the proposer_preferences topic)
     ProposerPreferences = 8,
+    // RequestAuthPartialSig is a standalone single-validator partial signature over a
+    // BuilderRequestAuth object (validator-scoped, non-QBFT; rides Role::ProposerPreferences
+    // as the role's second kind per SIP-94 §5's builder request auth extension)
+    RequestAuth = 9,
 }
 
 impl TryFrom<u64> for PartialSignatureKind {
@@ -60,6 +64,7 @@ impl TryFrom<u64> for PartialSignatureKind {
             6 => Ok(PartialSignatureKind::AggregatorCommitteePartialSig),
             7 => Ok(PartialSignatureKind::PTCAttester),
             8 => Ok(PartialSignatureKind::ProposerPreferences),
+            9 => Ok(PartialSignatureKind::RequestAuth),
             _ => Err(()),
         }
     }
@@ -198,6 +203,7 @@ mod tests {
             PartialSignatureKind::AggregatorCommitteePartialSig,
             PartialSignatureKind::PTCAttester,
             PartialSignatureKind::ProposerPreferences,
+            PartialSignatureKind::RequestAuth,
         ];
 
         for variant in variants {
@@ -232,6 +238,7 @@ mod tests {
             (PartialSignatureKind::AggregatorCommitteePartialSig, 6u64),
             (PartialSignatureKind::PTCAttester, 7u64),
             (PartialSignatureKind::ProposerPreferences, 8u64),
+            (PartialSignatureKind::RequestAuth, 9u64),
         ];
 
         for (variant, expected_value) in test_cases {
@@ -249,7 +256,7 @@ mod tests {
 
     #[test]
     fn partial_signature_kind_ssz_decode_invalid_variant() {
-        let invalid_value = 9u64.to_le_bytes();
+        let invalid_value = 10u64.to_le_bytes();
         let result = PartialSignatureKind::from_ssz_bytes(&invalid_value);
         assert!(matches!(result, Err(DecodeError::NoMatchingVariant)));
     }
@@ -321,11 +328,15 @@ mod tests {
             PartialSignatureKind::try_from(8u64).unwrap(),
             PartialSignatureKind::ProposerPreferences
         );
+        assert_eq!(
+            PartialSignatureKind::try_from(9u64).unwrap(),
+            PartialSignatureKind::RequestAuth
+        );
     }
 
     #[test]
     fn partial_signature_kind_try_from_u64_invalid_values() {
-        assert!(PartialSignatureKind::try_from(9u64).is_err());
+        assert!(PartialSignatureKind::try_from(10u64).is_err());
         assert!(PartialSignatureKind::try_from(100u64).is_err());
         assert!(PartialSignatureKind::try_from(u64::MAX).is_err());
     }
@@ -354,6 +365,7 @@ mod tests {
             PartialSignatureKind::AggregatorCommitteePartialSig,
             PartialSignatureKind::PTCAttester,
             PartialSignatureKind::ProposerPreferences,
+            PartialSignatureKind::RequestAuth,
         ];
 
         let hashes: Vec<_> = variants.iter().map(|v| v.tree_hash_root()).collect();
--- a/anchor/message_validator/src/duty_state.rs
+++ b/anchor/message_validator/src/duty_state.rs
@@ -29,19 +29,23 @@ use crate::{FIRST_ROUND, ValidationFailure, message_counts::MessageCounts};
 /// reorg-driven corrections while bounding spam, not a safety/consensus bound. Both exceeding
 /// the cap (`TooManyDistinctSigningRoots`) and repeating an already-recorded root
 /// (`RelayedDuplicateMessage`) are Ignore, regardless of the propagation peer; rationale at the
-/// enforcement site in `update_for_partial_signature`.
-const MAX_PROPOSER_PREFERENCES_DISTINCT_ROOTS: usize = 4;
+/// enforcement site in `update_for_partial_signature`. Crate-visible so pipeline tests in
+/// sibling modules reference the real value instead of hardcoding it.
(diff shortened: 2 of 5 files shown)


shane-moore:
@claude, please take a look

shane-moore:
@claude, re-checked ssv-spec#632 at head de34c611: `RequestAuthPartialSig = PartialSigMsgType(9)` still pinned by its golden test, and the workspace already sets `rust-version = "1.91.0"` so the `get_or_insert_default` MSRV note is covered. no changes needed.

shane-moore (commented):
Reviewed at `4a4a82dd`.

Ship-shaped. I found no correctness, safety, security, or specification-alignment issues. The kind-9 wire value, role binding, independent per-kind root budgets, Ignore classifications, and update-after-signature-verification ordering match the current SIP-94 and ssv-spec #632 heads.

Non-blocking: two added test comments in `anchor/message_validator/src/partial_signature.rs` (lines 3997 and 4075) use U+2014 punctuation, which conflicts with the repository style rule.

Reviewed by gpt-5.6-sol max.

PR #1282, opened 2026-08-27
feat(validator_store): sign builder request auth

### LINKED ISSUE #1278: feat(validator_store): sign builder request auth via the Lighthouse builder-API pin

shane-moore:
Closes #1278. First in the stack #1282 -> #1285 -> #1283. Merge the complete stack before releasing or deploying it. This PR alone can activate produce-time direct builders but does not provide cache pruning or SSV configuration constraints.

Gloas direct builders require RequestAuth signatures from the validator, so Anchor must reconstruct these signatures from operator shares rather than use Lighthouse's local-key signing path.

## Change Overview

Update Lighthouse to merged unstable revision `fdcc8658b3b917181855fce7a10a6147061419e4`, replacing the previous builder-stack revision. Implement `sign_request_auth_v1` using the fixed RequestAuth domain, exact opaque auth bytes, proposal slot, existing signature collector and proposer-preferences role. Add bounded future/current-slot collection and an immediate past-slot decline. Introduce the required builder store/cache dependencies and block-service wiring. Pass `None` to Lighthouse's new PTC event-receiver argument, preserving Anchor's deadline-driven scheduling.

The selected Lighthouse block path imposes a 200 ms outer deadline for a cold current-slot auth. Anchor's one-second direct-call backstop is not its effective proposal delay. Future preferences still need Anchor's two-slot bound because Lighthouse processes proposers sequentially. Collection failures have separate insufficient-shares and infrastructure metrics; expired requests do not inflate either failure class.

The update retains epbs #1299/#1300 and #1301, including same-slot knowledge frozen at QBFT instance start. Progressive SSZ/tree-hash patches remain unchanged. The lockfile also updates rustls to 0.23.45 and webpki to 0.103.15 for the current advisory check.

A separate dependency prerequisite was rejected: it would require the same trait implementation and store/cache wiring, splitting this coupled integration without making a separately deployable unit.

Lighthouse now defaults omitted builder auth to lowercase ASCII hostname bytes, not the complete URL. [SIP-94 commit 27fb510](https://github.com/ssvlabs/SIPs/commit/27fb5104eff53479e600559c5c388b1f7289e34d) now specifies the upstream hostname rule. The inspected go-ssv implementation still used the full-URL default; the SIP update does not itself establish client interoperability. The stack's operator documentation requires explicit identical auth bytes during migration or mixed-client operation. Explicit auth remains byte-for-byte unchanged. Do not claim mixed-client omitted-default compatibility.

sigp/lighthouse#10036 is still open. This PR does not enable stateless production or implement the SIP envelope fold. RequestAuth batch reception is handled separately in #1302; PTC endpoint amendments remain separate.

## Validation

- `cargo test -p anchor_validator_store --lib request_auth --locked`: 4 passed, including the known signing-root vector, exact collector context, fork-independent domain, deadline boundaries and failure metrics.
- `cargo test --release -p anchor_validator_store --lib request_auth --locked`: 4 passed on the integrated stack.
- `cargo check --workspace --all-targets --locked`: passed at this PR's local tip and both downstream tips.
- `make cargo-fmt-check`: passed.
- Integrated stack `make lint`, `make sort`, `make audit-CI`: passed; audit reported 12 allowed warnings under the existing repository policy.

These are local results for the refreshed code. Hosted CI must pass on the new head; checks on the previous head do not validate this update. Full-stack four-operator runtime evidence is described in #1285; the receiving stub does not independently validate BLS signatures or supply bids.

## Rollback

Before deployment, revert the dependent stack in reverse order. For a deployed rollback, restore a previously validated complete build and its compatible builder configuration. This PR adds no database migration. Coordinate explicit auth bytes across clients when changing versions; do not deploy an intermediate stack head.

## Blockers / Dependencies

The original Lighthouse builder stack is merged, so that prerequisite is cleared. Review and merge order is #1282 -> #1285 -> #1283, on epbs base `cc5c4144d5121b9d635634a1388af3babedad1ba`. The latter two PRs supply cache lifecycle and SSV configuration limits. Keep the complete stack as one deployment unit.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,26 @@
 version = 4
 
 [[package]]
+name = "account_utils"
+version = "0.1.0"
+source = "git+https://github.com/sigp/lighthouse?rev=44f442479d628ee4dcdcbd53db04791ec368057c#44f442479d628ee4dcdcbd53db04791ec368057c"
+dependencies = [
+ "bls",
+ "eth2_keystore",
+ "eth2_wallet",
+ "filesystem",
+ "rand 0.9.2",
+ "regex",
+ "rpassword 5.0.1",
+ "serde",
+ "tracing",
+ "types",
+ "validator_dir",
+ "yaml_serde",
+ "zeroize",
+]
+
+[[package]]
 name = "adler2"
 version = "2.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -363,7 +383,7 @@ dependencies = [
  "proptest",
  "rand 0.8.5",
  "ruint",
- "rustc-hash",
+ "rustc-hash 2.1.1",
  "serde",
  "sha3",
  "tiny-keccak",
@@ -392,7 +412,7 @@ dependencies = [
  "rand 0.9.2",
  "rapidhash",
  "ruint",
- "rustc-hash",
+ "rustc-hash 2.1.1",
  "serde",
  "sha3",
 ]
@@ -784,6 +804,7 @@ dependencies = [
  "async-channel 1.9.0",
  "beacon_node_fallback",
  "bls",
+ "builder_types",
  "database",
  "eth2",
  "ethereum_ssz",
@@ -1464,7 +1485,7 @@ checksum = "2af50177e190e07a26ab74f8b1efbfe2ef87da2116221318cb1c2e82baf7de06"
 [[package]]
 name = "beacon_node_fallback"
 version = "0.1.0"
-source = "git+https://github.com/sigp/lighthouse?rev=e58ec88fe#e58ec88fe9a8f755a0d92a80b25fd76100d08d49"
+source = "git+https://github.com/sigp/lighthouse?rev=44f442479d628ee4dcdcbd53db04791ec368057c#44f442479d628ee4dcdcbd53db04791ec368057c"
 dependencies = [
  "bls",
  "clap",
@@ -1564,7 +1585,7 @@ dependencies = [
 [[package]]
 name = "bls"
 version = "0.2.0"
-source = "git+https://github.com/sigp/lighthouse?rev=e58ec88fe#e58ec88fe9a8f755a0d92a80b25fd76100d08d49"
+source = "git+https://github.com/sigp/lighthouse?rev=44f442479d628ee4dcdcbd53db04791ec368057c#44f442479d628ee4dcdcbd53db04791ec368057c"
 dependencies = [
  "alloy-primitives 1.5.7",
  "blst",
@@ -1659,6 +1680,44 @@ dependencies = [
  "memchr",
  "regex-automata",
  "serde",
+]
+
+[[package]]
+name = "builder_store"
+version = "0.1.0"
+source = "git+https://github.com/sigp/lighthouse?rev=44f442479d628ee4dcdcbd53db04791ec368057c#44f442479d628ee4dcdcbd53db04791ec368057c"
+dependencies = [
+ "account_utils",
+ "bls",
+ "builder_types",
+ "filesystem",
+ "futures",
+ "hex",
+ "parking_lot",
+ "serde",
+ "ssz_types",
+ "tracing",
+ "types",
+ "yaml_serde",
+]
+
+[[package]]
+name = "builder_types"
+version = "0.1.0"
+source = "git+https://github.com/sigp/lighthouse?rev=44f442479d628ee4dcdcbd53db04791ec368057c#44f442479d628ee4dcdcbd53db04791ec368057c"
+dependencies = [
+ "bls",
+ "context_deserialize",
+ "ethereum_serde_utils 0.8.0",
+ "ethereum_ssz",
+ "ethereum_ssz_derive",
+ "sensitive_url",
+ "serde",
+ "ssz_types",
+ "tree_hash",
+ "tree_hash_derive",
+ "typenum",
+ "types",
 ]
 
 [[package]]
@@ -1879,6 +1938,7 @@ version = "1.3.1"
 dependencies = [
  "anchor_validator_store",
  "beacon_node_fallback",
+ "builder_store",
  "clap",
  "cli",
  "database",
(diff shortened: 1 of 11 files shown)


shane-moore:
@claude, please take a look

shane-moore (commented):
Reviewed at `e4f1fea9`.

I traced the request-auth domain and root, kind-9 and role-8 wire path, proposal-slot timing, collection mode, BuilderStore and RequestAuthCache wiring, and tests against the current builder spec, SIP-94, ssv-spec, go-ssv, and the pinned Lighthouse stack. I found no correctness, safety, or specification defect in the implementation.

Two follow-ups before this draft is ready:

1. `check-fmt` currently fails on the new doc comment around `anchor/validator_store/src/lib.rs:1391`. `make cargo-fmt` applies the required reflow.
2. This pin range includes [Lighthouse #9869](https://github.com/sigp/lighthouse/pull/9869), a lockfile-only update from `blst` 0.3.16 to 0.3.17 for the Pippenger divide-by-zero fixed in [blst #283](https://github.com/supranational/blst/pull/283) and reproduced on a 96-core host. Anchor's downstream lock remains on 0.3.16, so the pin bump does not inherit that hardening. Please carry 0.3.17 during the final repin. I did not establish that a normal Anchor duty reaches the affected large-point branch, so I view this as dependency hardening, not a demonstrated duty-path bug.

The existing draft blocker on Lighthouse #9805 through #9807 landing and the pin being repointed remains valid. At this head, the local testnet and every completed code, test, MSRV, dependency, Docker, and CLI check pass except formatting; the aggregate suite skips because of that failure.

Reviewed by gpt-5.6-sol max.

shane-moore:
addressed both in d1a56cdd: fmt reflow applied, blst carried to 0.3.17 now rather than at repin (full release suite green on the bump)

Issue #1280, opened 2026-08-26
feat(client): wire builder preferences service and produce-time builder config


shane-moore:
## Goal

Turn the feature on end to end: spawn Lighthouse's `BuilderPreferencesService` and hand the resolved `BuilderConfig` to block production.

## Context / motivation

`BuilderRequestAuth` is fully deterministic (config plus proposer lookahead), so Lighthouse's service is usable as-is with Anchor's `ValidatorStore`, unlike registration, which needed an Anchor-owned service for timestamp normalization (`registration_service.rs:1-13`). The one silent-miss risk is the `BlockServiceBuilder`: Anchor already omits three existing setters, so the new `BuilderConfig` setter defaults to `None` unnoticed unless wired.

## Suggested approach

Spawn `BuilderPreferencesService` inside the existing `is_gloas_scheduled` block beside `ProposerPreferencesService` (`client/src/lib.rs:838-859`). Wire the `BuilderConfig` setter on the `BlockServiceBuilder` (`client/src/lib.rs:724-736`) from the builder definitions. No `Eth-Builder-Url` work: stock Lighthouse echo behavior conforms to SIP-94's SHOULD-level owner-match rule.

## Acceptance criteria

- With entries configured: ahead-of-time submission fires for known proposal duties (current and next epoch) and the produce POST body carries entries with reconstructed auths; sub-quorum builders are omitted and the request still succeeds.
- Without entries: behavior identical to before this milestone.
- The service calls `RequestAuthCache::prune(current_slot)` each tick (it is the cache's only prune caller at the Lighthouse pin; without it the cache shared with the block service is insert-only). Share the single `BuilderStore`/`RequestAuthCache` instances constructed in client wiring (#1278) via clones rather than constructing new ones.

## Tests

- Service-spawn gating test (Gloas-scheduled only), produce-body wiring test with and without config.

Issues are directionally correct, not prescriptive; verify symbols at PR time.

shane-moore:
Closed as completed by #1285 (merged 2026-09-22 into epbs; its body says "Closes #1280"). Not auto-closed because the merge target is not the default branch.

PR #1283, opened 2026-08-27
feat(client): validate SSV builder definitions constraints at startup

### LINKED ISSUE #1279: feat(client): add builder definitions config for direct builder connections

shane-moore:
Closes #1279. Based on `epbs`, which now carries #1282 and #1285; this PR is the last piece of that stack. Deploy only the complete stack.

Lighthouse accepts up to 64 wire entries, but SSV permits only eight configured builders per validator. Its new per-validator overrides would bypass the original global-only Anchor validation. Excess auth roots can fragment signing quorum even when compilation and startup succeed.

## Change Overview

Validate at startup before services run:

- At most eight enabled global entries and eight entries in each explicit validator override.
- An absent per-validator list inherits the bounded global list; an empty list disables direct builders.
- Explicit empty auth and more than 64 builder public keys in a global entry fail, including in disabled entries.
- Disabled placeholder URLs need no hostname-derived auth. Active URL validation remains owned by Lighthouse.
- Startup errors keep the parser's line and column and the filesystem error kind, which are structural and cannot echo input, while still omitting URLs, auth bytes and parser messages.
- Lighthouse's `builder_store` crate is added to the logging allowlist, so a builder omitted at proposal time (for example a request-auth signing timeout on a cache miss) is logged rather than dropped silently.

Reuse Lighthouse's exported definition types and YAML parser. A small private file wrapper remains because the public store API cannot enumerate every raw override. Validation rereads the file after the store loads it; this existing external-write race is documented, not claimed eliminated.

Startup errors preserve safe categories plus position and error-kind metadata without rendering credentials or auth data. A parser-generated regression test also prevents a second-read YAML error from reflecting a sensitive scalar while asserting the position survives. Signing logs record auth length rather than contents.

Document per-validator configuration, inheritance/disabling and the hostname-default migration. For mixed clients or rolling upgrades, configure explicit identical auth bytes on every operator. Explicitly configuring the standard hostname bytes is valid without negotiating a new custom token; arbitrary custom values still require builder agreement. [SIP-94 commit 27fb510](https://github.com/ssvlabs/SIPs/commit/27fb5104eff53479e600559c5c388b1f7289e34d) aligns the normative default with this Lighthouse pin. That specification update does not replace mixed-client implementation and runtime verification.

## Validation

- `cargo test -p client --lib builder_definitions::tests --locked`: 11 passed.
- `cargo test --release -p client --lib builder_definitions::tests --locked`: 11 passed.
- New diagnostic, inactive-URL and parser-redaction tests failed before their respective fixes and passed afterward.
- Tests cover 8/9 entry boundaries, global inheritance, empty overrides, explicit bytes, hostname defaults, disabled policies, builder-key bounds and upstream-store round trips.
- `cargo check --workspace --all-targets --locked`, `make cargo-fmt-check`, `make lint`, `make sort`, `make audit-CI`, `git diff --check`: passed.

No keymanager mutation route is exposed by Anchor, so file startup is the reachable configuration boundary. No runtime reload, auth normalization in the signer, batching, stateless production or envelope-fold implementation is added. See #1285 for the integrated local runtime results and their limitations.

## Rollback

Removing this validation reopens the SSV entry-cap and silent-filtering failure modes. Do not deploy a partial revert with direct builders configured; use a previously validated complete build and compatible configuration. This PR adds no database migration; configuration edits apply on restart.

## Blockers / Dependencies

#1282 and #1285 are merged. Release the complete stack together. Hosted CI must pass on the refreshed head. The local stub run is not proof of real-builder signature acceptance, bid selection, delivery or mixed-client interoperability.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,26 @@
 version = 4
 
 [[package]]
+name = "account_utils"
+version = "0.1.0"
+source = "git+https://github.com/sigp/lighthouse?rev=44f442479d628ee4dcdcbd53db04791ec368057c#44f442479d628ee4dcdcbd53db04791ec368057c"
+dependencies = [
+ "bls",
+ "eth2_keystore",
+ "eth2_wallet",
+ "filesystem",
+ "rand 0.9.2",
+ "regex",
+ "rpassword 5.0.1",
+ "serde",
+ "tracing",
+ "types",
+ "validator_dir",
+ "yaml_serde",
+ "zeroize",
+]
+
+[[package]]
 name = "adler2"
 version = "2.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -363,7 +383,7 @@ dependencies = [
  "proptest",
  "rand 0.8.5",
  "ruint",
- "rustc-hash",
+ "rustc-hash 2.1.1",
  "serde",
  "sha3",
  "tiny-keccak",
@@ -392,7 +412,7 @@ dependencies = [
  "rand 0.9.2",
  "rapidhash",
  "ruint",
- "rustc-hash",
+ "rustc-hash 2.1.1",
  "serde",
  "sha3",
 ]
@@ -784,6 +804,7 @@ dependencies = [
  "async-channel 1.9.0",
  "beacon_node_fallback",
  "bls",
+ "builder_types",
  "database",
  "eth2",
  "ethereum_ssz",
@@ -1464,7 +1485,7 @@ checksum = "2af50177e190e07a26ab74f8b1efbfe2ef87da2116221318cb1c2e82baf7de06"
 [[package]]
 name = "beacon_node_fallback"
 version = "0.1.0"
-source = "git+https://github.com/sigp/lighthouse?rev=e58ec88fe#e58ec88fe9a8f755a0d92a80b25fd76100d08d49"
+source = "git+https://github.com/sigp/lighthouse?rev=44f442479d628ee4dcdcbd53db04791ec368057c#44f442479d628ee4dcdcbd53db04791ec368057c"
 dependencies = [
  "bls",
  "clap",
@@ -1564,7 +1585,7 @@ dependencies = [
 [[package]]
 name = "bls"
 version = "0.2.0"
-source = "git+https://github.com/sigp/lighthouse?rev=e58ec88fe#e58ec88fe9a8f755a0d92a80b25fd76100d08d49"
+source = "git+https://github.com/sigp/lighthouse?rev=44f442479d628ee4dcdcbd53db04791ec368057c#44f442479d628ee4dcdcbd53db04791ec368057c"
 dependencies = [
  "alloy-primitives 1.5.7",
  "blst",
@@ -1659,6 +1680,44 @@ dependencies = [
  "memchr",
  "regex-automata",
  "serde",
+]
+
+[[package]]
+name = "builder_store"
+version = "0.1.0"
+source = "git+https://github.com/sigp/lighthouse?rev=44f442479d628ee4dcdcbd53db04791ec368057c#44f442479d628ee4dcdcbd53db04791ec368057c"
+dependencies = [
+ "account_utils",
+ "bls",
+ "builder_types",
+ "filesystem",
+ "futures",
+ "hex",
+ "parking_lot",
+ "serde",
+ "ssz_types",
+ "tracing",
+ "types",
+ "yaml_serde",
+]
+
+[[package]]
+name = "builder_types"
+version = "0.1.0"
+source = "git+https://github.com/sigp/lighthouse?rev=44f442479d628ee4dcdcbd53db04791ec368057c#44f442479d628ee4dcdcbd53db04791ec368057c"
+dependencies = [
+ "bls",
+ "context_deserialize",
+ "ethereum_serde_utils 0.8.0",
+ "ethereum_ssz",
+ "ethereum_ssz_derive",
+ "sensitive_url",
+ "serde",
+ "ssz_types",
+ "tree_hash",
+ "tree_hash_derive",
+ "typenum",
+ "types",
 ]
 
 [[package]]
@@ -1879,6 +1938,7 @@ version = "1.3.1"
 dependencies = [
  "anchor_validator_store",
  "beacon_node_fallback",
+ "builder_store",
  "clap",
  "cli",
  "database",
@@ -2376,6 +2436,24 @@ dependencies = [
 ]
 
 [[package]]
+name = "deposit_contract"
+version = "0.2.0"
+source = "git+https://github.com/sigp/lighthouse?rev=44f442479d628ee4dcdcbd53db04791ec368057c#44f442479d628ee4dcdcbd53db04791ec368057c"
+dependencies = [
+ "alloy-dyn-abi",
+ "alloy-json-abi",
+ "alloy-primitives 1.5.7",
+ "bls",
+ "ethereum_ssz",
+ "hex",
+ "reqwest",
+ "serde_json",
+ "sha2",
+ "tree_hash",
+ "types",
+]
+
+[[package]]
 name = "der"
 version = "0.7.10"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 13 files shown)


shane-moore:
@claude, please take a look. just the top commit, as described in pr description

shane-moore (commented):
Reviewed the top commit, `85596357`, against its direct parent, `d1a56cdd`.

The SSV entry cap, zero-length auth validation, startup ordering, parser drift guard, and derivation vectors match the current SIP-94, go-ssv, and pinned Lighthouse behavior. I found two issues to resolve before undrafting:

1. `builder_pubkeys` is missing its startup bound. [Anchor's new pass](https://github.com/sigp/anchor/blob/85596357a928d95a2b30ee1e6ee81621b0f34339/anchor/client/src/builder_definitions.rs#L124-L144) does not inspect the list length, and [Lighthouse's load validator](https://github.com/sigp/lighthouse/blob/44f442479d628ee4dcdcbd53db04791ec368057c/validator_client/builder_store/src/builder_definitions.rs#L189-L228) also leaves the `Vec` unbounded. A definition with 65 valid keys passes startup; [`builder_config`](https://github.com/sigp/lighthouse/blob/44f442479d628ee4dcdcbd53db04791ec368057c/validator_client/builder_store/src/lib.rs#L100-L105) later logs and omits that builder because the wire list is capped at 64. Please enforce this in Lighthouse and re-pin, or check it here, then add a 65-key test and document the limit.
2. The [security guidance](https://github.com/sigp/anchor/blob/85596357a928d95a2b30ee1e6ee81621b0f34339/docs/docs/pages/mev_configuration.mdx#L133-L135) should not say `auth_data` is non-secret or imply credentials in URLs are safe. The merged keymanager schema explicitly permits `auth_data` to be [a shared secret](https://github.com/ethereum/keymanager-APIs/blob/5039201ac140fdde568763dd6af038bf56a0e840/types/builder_entry.yaml#L77-L95), while the pinned store [logs raw builder URLs](https://github.com/sigp/lighthouse/blob/44f442479d628ee4dcdcbd53db04791ec368057c/validator_client/builder_store/src/lib.rs#L90-L104) and [`BuilderUrl::Display` does not redact them](https://github.com/sigp/lighthouse/blob/44f442479d628ee4dcdcbd53db04791ec368057c/common/builder_types/src/builder_url.rs#L105-L108). Treat custom auth data and URL credentials as sensitive. Until those Lighthouse logs are redacted, advise operators not to embed credentials in builder URLs.

All CI checks pass at this head, including the local testnet. The existing draft gates on the Lighthouse stack and issue #1280 remain.

Reviewed by gpt-5.6-sol max.

shane-moore:
Both review items addressed in 4ca6d8b4: startup now rejects any entry with more than 64 `builder_pubkeys` (`TooManyBuilderPubkeys`, all entries, bound tracked structurally via `MaxBuilderPubkeys::USIZE`, with a 64-pass/65-fail boundary test through the real store), and the docs no longer call `auth_data` non-secret; they now treat agreed auth data as a credential and advise keeping credentials out of builder URLs, since logs print them verbatim. Moving the pubkeys bound into Lighthouse's load validation stays on the re-pin ask list; the local check becomes deletable then.

shane-moore (commented):
**Review by Claude Fable 5.1** (Claude Code), against head `3f01d4a8` with `epbs` at `6c98a389` and the Lighthouse pin `fdcc8658`. Findings were folded into the merge commit `9e0d3f65`, so this is the record rather than a request.

**Scope checked**

- Global entries: at the pin, Lighthouse `BuilderConfigFile::validate` skips disabled entries and, for enabled ones, checks only URL shape and scheme, derivable default auth, and `(url, auth)` duplicates. It does not reject an explicit empty `auth_data` and does not bound `builder_pubkeys`. The per-entry loop here is unfiltered by `enabled`, so both are caught for every global entry.
- Per-validator overrides: Lighthouse `ValidatorBuilderConfig::validate` already bounds the list at 64, rejects oversized pubkeys and empty auth, and dedups. This PR adds only the SSV cap of 8 on present lists; `None` inherits the already-capped enabled globals and `Some([])` disables, matching `resolved_for`.
- SIP-94 at PR #94 head `fc4f965d`: 8 entries per validator, lowercase ASCII hostname default, zero-length data invalid. go-ssv #2901 head `6fe7a951`: still defaults auth to the raw URL bytes and caps at 8, so the mixed-client migration section is accurate as of today.
- Docs claims against the pin: template file on first start, 2048-byte URL and 4096-byte auth bounds, no `enabled` on override entries, override replaces globals, disabled globals need no derivable hostname. All hold.
- `yaml_serde` resolves to a single `0.10.4` on both sides. Startup placement is unconditional, as `open_or_create` already was on the base, so no new failure class on non-Gloas networks.

No correctness or safety defects found. The two items from the 08-27 review (`builder_pubkeys` bound, `auth_data` treated as a credential) are addressed at this head.

**Three minor items, fixed in `9e0d3f65`**

1. Startup errors redacted safe metadata along with the sensitive payload: parse errors lost the parser's line and column, open errors lost the `io::ErrorKind`. Both are structural and cannot echo input, so they are now rendered while URLs, auth bytes, and parser messages stay out.
2. The docs told operators not to rely on Lighthouse `builder_store` omission logs being visible. That was true because the crate was missing from the logging allowlist, and one path had no other logger: the 200 ms request-auth sign timeout on a cache miss at proposal time, which dropped the builder silently. `builder_store` is now on the allowlist (its only log sites are `error!`) and the caveat is removed.
3. `bls` was added to `[dev-dependencies]` while already a regular dependency.

**Considered and dropped**

- Entry cap versus the 8-root gossip budget under overrides: an override list replaces globals, so a validator resolves to at most 8 entries and at most 8 roots.
- The two-read window between the store load and this validation: unchanged from the earlier review; no Anchor writer, no runtime reload, and only a Lighthouse accessor over loaded definitions would close it. The module doc already names that as the deletion path for the wrapper.

PR #1284, opened 2026-08-27
refactor(message_validator): delete the unused received_from plumbing and peer-keyed preference map

### LINKED ISSUE #1273: refactor(message_validator): delete the unused received_from plumbing and peer-keyed preference map

shane-moore:

- #1272 made role-8 duplicate signing-root classification peer-agnostic per SIP-94 §7, which deleted the only read of the stored propagation peer. Since then `received_from` is written into the per-kind root maps and read nowhere; passing `Some(peer)` or `None` produces identical output at every observable boundary.
- The dead field actively misleads: it teaches readers that peer identity affects role-8 validation (it produced #1238, closed as obsolete for exactly this reason), and #1281's own doc comment says the deliverer "is retained only because issue #1254 scopes the peer plumbing as unchanged", i.e. it deferred this cleanup here.
- Closes #1273. Context: #1131/#1203 (introduced the peer-keyed map), #1254/#1272 (made the verdict peer-agnostic), #1281 (added the second budgeted kind).

## Change Overview (Required)

- Drop `received_from: Option<PeerId>` from the five-signature validation chain (`validate` -> `validate_decoded_message` -> `validate_ssv_message` -> `validate_partial_signature_message` -> `update_for_partial_signature`) and its one production call site in `message_receiver::manager`.
- Narrow both `SigningRootBudgets` maps from `HashMap<Hash256, Option<PeerId>>` to `HashSet<Hash256>`; rewrite the two doc comments that described the stored deliverer.
- Tests: delete the peer fixtures, remove the trailing peer argument everywhere, collapse tests whose only remaining distinction was the deliverer identity, and delete three older tests that became strict subsets of the dedup criteria tests once the peer axis vanished.
- Reading order: `duty_state.rs` (the only semantic hunks: set narrowing + enforcement site), then `lib.rs`/`partial_signature.rs`/`manager.rs` (mechanical arity changes), then the test module.
- Intentionally unchanged: classification behavior and error variants, membership-before-capacity ordering, the exhaustive `root_budget` match, the lazy boxed budget allocation, and `manager.rs`'s use of `propagation_source` for the gossipsub `Outcome` report. `libp2p` stays a dependency (`MessageAcceptance` re-export and conversions).

- Main risk is an accidental semantic change smuggled in via the test collapse. Mitigation: the acceptance criterion was that every surviving test passes with argument removals, renames, and comment fixes only; zero expectation changes were needed. The `RelayedDuplicateMessage -> Ignore` mapping pin, the cap/poisoning/budget-independence tests, and the ring-collision test all survive.
- `Validator::validate` is a public signature, so this is a source-level API change; the single in-repo caller is updated in the same diff.
- Trade-off declared: net test count drops 132 -> 126. Each deleted test was verified as a strict behavioral subset of a surviving criteria test before deletion.

## Validation (Required)

- `cargo test -p message_validator`: 126 passed, 0 failed.
- `make lint`, `make cargo-fmt-check`, `make check-benches`, `cargo check -p message_receiver`: clean.
- `grep -rn received_from anchor/`: zero hits.
- Rejected alternative for the record: keeping the plumbing to seed a go-ssv-style two-tier peer/world state was withdrawn when #1273 was filed; a real port is a redesign this parameter would not meaningfully advance.

- Pure revert; no config, data, or wire impact. Runtime verdicts are unchanged, so mixed deployments are unaffected.

- N/A. #1276 and #1281 are already in the base; no overlap with open #1282.

### DIFF
--- a/anchor/message_receiver/src/manager.rs
+++ b/anchor/message_receiver/src/manager.rs
@@ -80,7 +80,7 @@ impl<E: types::EthSpec, S: SlotClock + 'static, D: DutiesProvider> MessageReceiv
                 let span = debug_span!("message_receiver", msg=%message_id);
                 let _enter = span.enter();
 
-                let result = receiver.validator.validate(&message.data, &topic_context, Some(propagation_source));
+                let result = receiver.validator.validate(&message.data, &topic_context);
 
                 let mut action = MessageAcceptance::from(&result);
 
--- a/anchor/message_validator/src/consensus_message.rs
+++ b/anchor/message_validator/src/consensus_message.rs
@@ -608,7 +608,6 @@ mod tests {
                 voluntary_exit_duty_count: expected_duty_count,
                 ..Default::default()
             }),
-            None,
         );
 
         assert_qbft_message_accepted(result, "Expected successful validation");
@@ -659,7 +658,6 @@ mod tests {
                 voluntary_exit_duty_count: 0,
                 ..Default::default()
             }),
-            None,
         );
 
         assert_validation_error(
@@ -717,7 +715,6 @@ mod tests {
                 voluntary_exit_duty_count: 0,
                 ..Default::default()
             }),
-            None,
         );
 
         assert_validation_error(
@@ -772,7 +769,6 @@ mod tests {
                 voluntary_exit_duty_count: 0,
                 ..Default::default()
             }),
-            None,
         );
 
         assert_validation_error(
@@ -2078,7 +2074,6 @@ mod tests {
                 voluntary_exit_duty_count: 0,
                 ..Default::default()
             }),
-            None,
         )
     }
 
--- a/anchor/message_validator/src/duty_state.rs
+++ b/anchor/message_validator/src/duty_state.rs
@@ -1,6 +1,5 @@
 use std::collections::{HashMap, HashSet};
 
-use libp2p::PeerId;
 use ssv_types::{
     CommitteeId, Epoch, OperatorId, Slot,
     consensus::{QbftMessage, QbftMessageType},
@@ -96,7 +95,6 @@ impl DutyState {
         &mut self,
         partial_signature_messages: &PartialSignatureMessages,
         signer: &OperatorId,
-        received_from: Option<PeerId>,
     ) -> Result<(), ValidationFailure> {
         let operator_state = self.get_or_create_operator(signer);
         let message_slot = partial_signature_messages.slot;
@@ -130,7 +128,7 @@ impl DutyState {
             // recipient's gossip duplicate cache expires, so repetition does not prove peer
             // fault. Membership is checked before capacity so a recorded root stays IGNORE
             // even when the set is full.
-            if seen_roots.contains_key(&root) {
+            if seen_roots.contains(&root) {
                 return Err(ValidationFailure::RelayedDuplicateMessage {
                     got: format!("{kind:?} root {root:?}"),
                 });
@@ -140,7 +138,7 @@ impl DutyState {
                     got: format!("{kind:?} distinct roots exceed cap {cap}"),
                 });
             }
-            seen_roots.insert(root, received_from);
+            seen_roots.insert(root);
         }
 
         // Record the partial signature (only once)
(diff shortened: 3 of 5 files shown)


shane-moore:
@claude, please take a look

shane-moore (commented):
Reviewed at `dc80d806`.

No findings survived self-audit. The `received_from` value was write-only, and narrowing the root budgets from `HashMap<Hash256, Option<PeerId>>` to `HashSet<Hash256>` preserves membership, capacity, insertion, and per-kind budget semantics. The parameter removal is complete, while `propagation_source` remains available for the gossipsub outcome.

The surviving tests retain coverage of first-root acceptance, duplicate Ignore mapping, capacity and non-poisoning behavior, independent per-kind budgets, and ring retention. All 126 `message_validator` tests passed locally, `message_receiver` compiled, and current CI is green.

Ship-shaped.

Reviewed by gpt-5.6-sol max.

PR #1285, opened 2026-08-27
feat(client): spawn the builder preferences service

### LINKED ISSUE #1280: feat(client): wire builder preferences service and produce-time builder config

shane-moore:
Closes #1280. Stacked on #1282; followed by #1283. GitHub's base remains `epbs`, so the full diff includes #1282. Review this service-wiring change against #1282 head `508d64714fe36ff876c622583b86d4a24bb99249`. No release or deployment from an intermediate head.

Produce-time builder authentication is too late to supply preferences reliably ahead of proposals. Without the builder-preferences service, the request-auth cache also lacks its pruning owner.

## Change Overview

Start Lighthouse's `BuilderPreferencesService` under Anchor's existing Gloas-scheduled gate. Share the same Arc-backed BuilderStore and RequestAuthCache with block production. This keeps ahead-of-time signing, produce-time reuse and cache lifetime consistent without adding an Anchor-specific scheduler or cache.

The merged Lighthouse pin now skips elapsed duties, uses the target epoch's fork header, resolves per-validator configuration and signs builders within a proposer concurrently. Proposers are still sequential and publication follows collection. Therefore a below-quorum auth can delay other preferences; Anchor's two-slot future collection bound remains necessary. Successful-only marking and chunk retries are retained. Builders must tolerate replay, including after restart.

## Validation

- `cargo check --workspace --all-targets --locked`: passed at this local stack boundary.
- Integrated stack formatting and lint: passed.
- Four Anchor operators, quorum three, ran with Lighthouse `fdcc8658`. Initial runs observed ahead-of-time preferences across the Gloas boundary and reuse of reconstructed auths.
- A common auth on all four operators plus a distinct second auth on only two operators delivered common preferences before all ten managed duties checked in epochs 53/54, including adjacent slots 1711/1712 and 1729/1730. The second auth produced no preference receipts. This demonstrates bounded progress in that scenario, not universal scheduling fairness.
- Restart scenarios included past/current/future duties. Early startup auth timeouts were observed and not attributed solely to head-of-line blocking.
- Empty configuration produced zero preference receipts over 75 seconds while managed attestations continued.
- All original configurations were restored byte-identically. Host clamshell sleep interrupted the longer observation and post-restoration health interval; no uninterrupted soak claim is made.

The receiver was a threaded HTTP stub. It proves receipt timing but does not verify BLS signatures, identify individual sending operators or return bids. Whole-chunk retries produced extra receipt records during transient forwarding timeouts. The runtime image preceded final diagnostic, inactive-URL, test and documentation changes; the exercised signing, service and populated valid-config paths are unchanged. The interrupted run did not establish post-sleep finality recovery. August's old-pin run remains historical evidence only.

The fixed pin can echo a locally produced candidate's builder URL after Anchor selects another candidate. SIP section 4 treats forwarding as a latency optimization with gossip fallback, not a liveness prerequisite. This refresh does not add a new publication subsystem. Real-builder signature acceptance and bid selection remain separate runtime evidence gaps. Hosted CI must pass on this refreshed head. These runs did not include go-ssv, so mixed-client readiness remains unproven.

## Rollback

Reverting this service alone removes ahead-of-time publication and the auth cache's pruning owner. Do not deploy that state with direct builders enabled. Roll back the complete builder stack to a validated build and compatible configuration; no new database or wire format is introduced by this PR.

## Blockers / Dependencies

Depends on #1282 and must be followed by #1283 before deployment. The merged Lighthouse pin clears the old upstream-stack blocker. #1302's batch receiver and the future stateless/envelope-fold integration remain separate.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,26 @@
 version = 4
 
 [[package]]
+name = "account_utils"
+version = "0.1.0"
+source = "git+https://github.com/sigp/lighthouse?rev=44f442479d628ee4dcdcbd53db04791ec368057c#44f442479d628ee4dcdcbd53db04791ec368057c"
+dependencies = [
+ "bls",
+ "eth2_keystore",
+ "eth2_wallet",
+ "filesystem",
+ "rand 0.9.2",
+ "regex",
+ "rpassword 5.0.1",
+ "serde",
+ "tracing",
+ "types",
+ "validator_dir",
+ "yaml_serde",
+ "zeroize",
+]
+
+[[package]]
 name = "adler2"
 version = "2.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -363,7 +383,7 @@ dependencies = [
  "proptest",
  "rand 0.8.5",
  "ruint",
- "rustc-hash",
+ "rustc-hash 2.1.1",
  "serde",
  "sha3",
  "tiny-keccak",
@@ -392,7 +412,7 @@ dependencies = [
  "rand 0.9.2",
  "rapidhash",
  "ruint",
- "rustc-hash",
+ "rustc-hash 2.1.1",
  "serde",
  "sha3",
 ]
@@ -784,6 +804,7 @@ dependencies = [
  "async-channel 1.9.0",
  "beacon_node_fallback",
  "bls",
+ "builder_types",
  "database",
  "eth2",
  "ethereum_ssz",
@@ -1464,7 +1485,7 @@ checksum = "2af50177e190e07a26ab74f8b1efbfe2ef87da2116221318cb1c2e82baf7de06"
 [[package]]
 name = "beacon_node_fallback"
 version = "0.1.0"
-source = "git+https://github.com/sigp/lighthouse?rev=e58ec88fe#e58ec88fe9a8f755a0d92a80b25fd76100d08d49"
+source = "git+https://github.com/sigp/lighthouse?rev=44f442479d628ee4dcdcbd53db04791ec368057c#44f442479d628ee4dcdcbd53db04791ec368057c"
 dependencies = [
  "bls",
  "clap",
@@ -1564,7 +1585,7 @@ dependencies = [
 [[package]]
 name = "bls"
 version = "0.2.0"
-source = "git+https://github.com/sigp/lighthouse?rev=e58ec88fe#e58ec88fe9a8f755a0d92a80b25fd76100d08d49"
+source = "git+https://github.com/sigp/lighthouse?rev=44f442479d628ee4dcdcbd53db04791ec368057c#44f442479d628ee4dcdcbd53db04791ec368057c"
 dependencies = [
  "alloy-primitives 1.5.7",
  "blst",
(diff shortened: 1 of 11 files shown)


shane-moore:
@claude, please take a look

shane-moore (commented):
Reviewed only the top commit, `89d01ca5`, against its direct parent, `d1a56cdd`.

The Anchor wiring itself is correct: the block service and builder-preferences service share one Arc-backed store and request-auth cache, and the new service starts under the existing Gloas gate. I found one upstream behavior worth noting during soak:

1. The [pinned service](https://github.com/sigp/lighthouse/blob/44f442479d628ee4dcdcbd53db04791ec368057c/validator_client/validator_services/src/builder_preferences_service.rs#L190-L209) derives one `current_fork`, gathers duties from both the current and next epochs, then [submits every entry with that current fork](https://github.com/sigp/lighthouse/blob/44f442479d628ee4dcdcbd53db04791ec368057c/validator_client/validator_services/src/builder_preferences_service.rs#L260-L290). In epoch `gloas_fork_epoch - 1`, it skips the current Fulu duties, gathers next-epoch Gloas duties, and sends their Gloas-only bodies with `Eth-Consensus-Version: fulu`. The merged validator API says the header names [the version the submitted preferences belong to](https://github.com/ethereum/beacon-APIs/blob/159622d983a703eb03a8a37bb1edeab7ffc3b6bc/apis/validator/builder_preferences.yaml#L16-L24), while the builder spec defines a [Gloas request body](https://github.com/ethereum/builder-specs/blob/b5ec384aeddba1aa2fea257a7492b363dfbc1a52/apis/builder/builder_preferences.yaml#L25-L50) and says the header names [the fork of the body's type](https://github.com/ethereum/builder-specs/blob/b5ec384aeddba1aa2fea257a7492b363dfbc1a52/specs/gloas/builder.md#L153-L155). A builder that dispatches decoding by this required header can reject every pre-fork attempt, so the first Gloas epoch loses its intended ahead-of-time publication window. This behavior is in the pinned dependency, not the Anchor wiring reviewed here.

All CI checks, including `run-local-testnet`, pass at this head. The existing draft and soak gates remain.

Reviewed by gpt-5.6-sol max.

PR #1288, opened 2026-09-03
feat: admit one envelope dissemination per signer and sign by binding

shane-moore:

A single committee member could cost the cluster its self-build reveal by winning a race.

SIP-94 §7 admitted one `EnvelopeDissemination` per `(MessageId, slot)`, signer-independent, and §6 signed whichever one arrived first. The first structurally valid carrier was therefore the only one an operator would forward or consider. A member that broadcast a well-formed but decision-unbound envelope ahead of the builder made every non-builder abstain, and the payload never revealed. Nothing about that requires a sophisticated attacker: one misbehaving member, once, is enough.

Both halves of the rule come from review of the SIP by Iurii (ssvlabs), whose asks were classified MUST where they are observable on the SSV wire. These two are: a go-ssv implementation that dedups per slot instead of per signer would drop disseminations Anchor admits, and the two would disagree about which envelope to sign. The SIP text edit follows this PR rather than preceding it, so the rule is proven implementable before it is written down.

Builds on #1286.

## Change Overview (Required)

Admit one dissemination per `(MessageId, signer, slot)`, and have the non-builder task sign the first candidate whose envelope satisfies the decided block's four bindings rather than the first that arrives. Forwarding stays bounded because only committee members pass validation, so the candidate list is capped by committee size.

Read it in this order:

1. `message_validator/src/duty_state.rs` - the dedup record moves from a signer-independent ring on `DutyState` to a flag on the signer's own `SignerState`, which is already the per-signer per-slot record and already carries the max-slot and per-epoch duty-count effects the old record reproduced by hand.
2. `message_validator/src/dissemination.rs` - the validation rule itself.
3. `common/dissemination_store/src/lib.rs` - the store keeps every accepted candidate per `(validator, slot)` behind an `Arc`, and `wait_matching` walks them once each, running the caller's predicate outside the lock.
4. `validator_store/src/lib.rs` - `sign_disseminated_envelope` supplies that predicate.
5. `message_receiver/src/manager.rs` and `operator_doppelganger/src/service.rs` - mechanical pattern updates for the reshaped variant.

What did not change: the wire format, the four bindings themselves, the builder's own publish path, and the deliberate decision to leave `payload_root` unchecked. No Lighthouse changes.

**The accepted residual is unchanged and still real.** `payload_root` is not among the checked bindings, so a forgery that passes all four bindings while carrying a different payload root splits operators across two roots. Arrival order is node-local, so each root collects some shares and neither reaches threshold. The symptom is a signature collection timeout, never a wrong payload. This is documented at the selection site.

**A store entry now holds a list rather than one value.** Bounded on the validated path by committee size, at most one candidate per member per slot. A cap was proposed in review and declined: it cannot fire unless some other invariant is already broken, and it would then drop a candidate that may be the honest one, trading bounded transient memory for a missed reveal.

**The dedup flag is never cleared.** That rests on role 9 having no consensus path, since `OperatorState::update` is the only writer that replaces a `SignerState` outright and it is reachable only for QBFT roles. A future role that both disseminates and runs QBFT would need the bit carried across the replacement. Stated at the field.

## Validation (Required)

Unit: 12 in `dissemination_store`, 139 in `message_validator`, 168 in `anchor_validator_store`, workspace otherwise green, `make cargo-fmt-check` and `make lint` clean.

Two guards were documented in prose but pinned by no test, found by mutation testing. Making `record_dissemination` always replace the `SignerState` passed the entire suite, and is a live budget-reset bug: gossip does not order a signer's dissemination against its own partial signature and the two share one per-signer-per-slot entry. Dropping the ring's stored-slot filter also passed, and would silently ignore an honest dissemination exactly one ring length later. Both now have tests, each confirmed to fail against its mutation.

Devnet: 4-operator all-Anchor Gloas run on ssv-mini, operators split across two beacon nodes so each proposal has both builder and non-builder operators. 11 managed proposals, 11 of 11 envelopes revealed on chain and served by the BN. The accounting invariant held exactly on every operator (`signed_other + published == received`, and `delegated == lh_sentinel == signed_other`). Zero dissemination timeouts, zero binding mismatches, zero skipped candidates, zero validation failures, zero restarts. Reveal timing 0.15s to 0.39s into the slot.

That topology also exercises the §7 change directly, which is worth stating because it is not obvious: two operators share the builder's beacon node and build the identical block, so both broadcast, putting two disseminations from two distinct signers on one key every proposal. Under the old per-slot rule the second would have been rejected at validation on every receiving node.

Not exercised, and not claimed: a candidate that fails the bindings. Both candidates in a healthy cluster pass, so the skip-and-continue path never runs. Demonstrating it needs a patched operator broadcasting a binding-failing envelope.

Revert the four commits. No config, no schema, no wire change, so a reverted node interoperates with a non-reverted one except that it reverts to admitting one dissemination per slot.

The SIP-94 §6/§7 text edit follows this PR.

The review found eight points a second implementer could resolve differently, which will go into the SIP text rather than staying here. Two are sharp: recording a claimed signer before verifying its signature would be a censorship primitive, and the committee-size bound on the candidate list only holds if the text requires the signer to be a committee member rather than merely a known operator.

https://claude.ai/code/session_01SJP7vu2EGTFknFswZpjNjJ

### DIFF
--- a/anchor/common/dissemination_store/Cargo.toml
+++ b/anchor/common/dissemination_store/Cargo.toml
@@ -10,3 +10,7 @@ parking_lot = { workspace = true }
 ssv_types = { workspace = true }
 tokio = { workspace = true }
 types = { workspace = true }
+
+[dev-dependencies]
+# `start_paused` in the wait tests: the deadline is the only thing they wait on.
+tokio = { workspace = true, features = ["test-util"] }
--- a/anchor/common/dissemination_store/src/lib.rs
+++ b/anchor/common/dissemination_store/src/lib.rs
@@ -1,13 +1,14 @@
 //! Handoff store for SIP-94 §6 envelope disseminations.
 //!
 //! The message receiver writes every validator-accepted `EnvelopeDissemination` for a
-//! `(validator, slot)`; the envelope duty runner scans them in arrival order for the first
-//! whose envelope binds to its own §4 decision. Message validation admits at most one
-//! dissemination per (`MessageId`, signer, slot) (SIP-94 §7) and only from committee members,
-//! so the candidate list is bounded by committee size on the validated path. The store itself
-//! enforces neither: it is a handoff, and validation is its only production writer.
+//! `(validator, slot)`; the envelope duty runner scans them in arrival order and selects one by
+//! content (see `sign_disseminated_envelope` for why selection cannot be by arrival). Message
+//! validation admits at most one dissemination per (`MessageId`, signer, slot) (SIP-94 §7) and
+//! only from committee members, so the candidate list is bounded by committee size on the
+//! validated path. The store enforces neither: it is a handoff, and validation is its only
+//! production writer.
 
-use std::collections::HashMap;
+use std::{collections::HashMap, sync::Arc};
 
 use bls::PublicKeyBytes;
 use parking_lot::Mutex;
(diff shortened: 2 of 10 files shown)


shane-moore (commented):
Reviewed `571590c1` against `epbs` at `e874c1ea`.

No actionable findings. Per-signer admission is recorded only after all validation, signer state preserves prior partial-signature accounting, the candidate handoff is committee-bounded and lost-wakeup safe, and the non-builder path skips unbound candidates while preserving the payload-due deadline. The residual unchecked `payload_root` behavior is explicit and tested.

Focused local tests (319), formatting, Clippy with warnings denied, and diff checks passed. All exact-head GitHub checks, including the local testnet, are green.

Reviewed by gpt-5.6-sol xhigh.

shane-moore:
Closing: SIP-94 now carries `payload_root` in the decided value and folds the envelope signature into the proposer post-consensus packet (ssvlabs/SIPs@0f7fe152), which removes the dissemination round this PR aligns. Until both clients cut over, mixed clusters on the current format keep the per-signer versus per-slot dedup difference.

PR #1290, opened 2026-09-04
fix(validator_store): avoid non-leader blinded publish

### LINKED ISSUE #1289: fix(validator_store): avoid the guaranteed-failure blinded publish on non-leader operators
## Goal

Stop non-leader operators from issuing a blinded publish that their own beacon node cannot satisfy, on every pre-Gloas proposal.

## Context / motivation

`select_publish_block` (`anchor/validator_store/src/lib.rs:4366`) reconstructs a full block only when this operator produced the matching local block. Otherwise it returns the blinded block via `PUBLISH_PATH_BLINDED_BLOCK_NOT_LEADER` (`anchor/validator_store/src/lib.rs:4363`, used at `:4380`), and `sign_block` hands that to Lighthouse to publish (`:3462`). A non-leader's beacon node did not build that payload, so pre-Gloas it cannot unblind it and returns `500 INTERNAL_SERVER_ERROR: Blind block proposal failed: NoPayloadBuilder`, which Lighthouse logs at ERROR.

Observed on a 4-operator all-Anchor ssv-mini Gloas run with operators split across two beacon nodes. All four operators published at all six pre-Gloas managed proposals (slots 48, 51, 52, 55, 60, 63), and the two operators on the beacon node that did not build the decided block failed every time, alternating with which node led. Blocks still landed, because the leader's pair published successfully. The failures stop at the Gloas fork, where blocks publish full and there is nothing to unblind.

Two consequences:

- Redundant publish is lost pre-Gloas. If the leader fails to publish, no other operator can cover for it, because their publish is guaranteed to fail.
- Every operator running its own beacon node, which is the normal production topology, emits an ERROR per proposal for a condition the code already names and instruments as an expected publish path.

## Suggested approach

Decide what a non-leader should do pre-Gloas: skip the publish, or reconstruct the full block from its own execution payload when the bid matches. Whichever is chosen, the expected case should not surface as an ERROR. Anchor cannot lower Lighthouse's log level, so the fix has to be to stop returning a block that is known to be unpublishable locally.

## Acceptance criteria

- No `NoPayloadBuilder` publish failures on a cluster whose operators use separate beacon nodes, pre-Gloas.
- Post-Gloas publishing behavior is unchanged.
- `PUBLISH_PATH_*` metric accounting still distinguishes the leader path from the non-leader path.

## Tests

Unit coverage on `select_publish_block` for the non-leader pre-Gloas case. Existing post-Gloas coverage unchanged.

## Notes

Found while devnet-validating #1288. Unrelated to that change, which only touches the envelope dissemination path and has no pre-Gloas reachability.

ke7haav:
## Problem

Before Gloas, an operator only retains a full block for its own local proposal. If QBFT decides another operator's proposal, the resulting blinded block does not match that local block.

Publishing it through a separate beacon node is guaranteed to fail when that node lacks the matching cached payload and payload builder:

`Blind block proposal failed: NoPayloadBuilder`

Fixes #1289.

## Change overview

- Skip unmatched blinded-block publishing for pre-Gloas duties.
- Preserve matching leader full-block reconstruction.
- Preserve blinded leader publishing.
- Leave Gloas and later publishing unchanged.
- Derive the fork boundary from the locally requested duty block.
- Add the `skipped_blinded_block_not_leader_pre_gloas` publish-path label.
- Preserve successful signing and publish-path telemetry.
- Add focused leader, non-leader, Gloas, path-selection, and crossed-fork tests.

Full reconstruction was not added because the retained local full block belongs to a different proposal. Pairing its payload or blobs with the decided block could invalidate the signed body root. Lighthouse already reconstructs using an exact matching cached payload when one exists.

## Risks

Lighthouse has no explicit successful no-publish result in its `ValidatorStore` API. The change uses its existing `UnknownPubkey` cancellation path, which avoids publishing but emits a misleading missing-pubkey warning.

## Validation

- `make cargo-fmt && make cargo-fmt-check`
- `cargo +nightly test -p anchor_validator_store`
- `cargo +nightly test --release -p anchor_validator_store`
- `cargo +nightly clippy --workspace --tests --features "" -- -D warnings -D clippy::allow_attributes`
- `cargo +nightly test --release --features ""`

## Rollback

Revert this change to restore the previous non-leader blinded publish attempt. There are no configuration, database, or migration effects.

### DIFF
--- a/anchor/validator_store/src/lib.rs
+++ b/anchor/validator_store/src/lib.rs
@@ -2513,7 +2513,7 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> ValidatorStore
                     "Publish path selected"
                 );
 
-                Ok(publish_decision.signed_block)
+                Ok(publish_decision)
             }
             .await;
 
@@ -2542,12 +2542,21 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> ValidatorStore
         }
         .instrument(span);
 
-        run_and_update_metrics(
+        let publish_decision = run_and_update_metrics(
             BLOCK_LOG_NAME,
             &validator_metrics::SIGNED_BLOCKS_TOTAL,
             future,
         )
-        .await
+        .await?;
+
+        // Lighthouse's `ValidatorStore` API has no explicit "signed, do not publish" outcome.
+        // Its block service treats `UnknownPubkey` as a benign cancellation and returns without
+        // publishing, whereas every other error is surfaced as a block-production failure. Keep
+        // this integration signal outside the signing instrumentation so the completed threshold
+        // signature and the selected non-leader path are recorded as successful.
+        publish_decision
+            .signed_block
+            .ok_or(Error::UnknownPubkey(validator_pubkey))
     }
 
     async fn sign_validator_registration_data(
@@ -3125,7 +3134,7 @@ impl<T: SlotClock, E: EthSpec, C: ConsensusDecider<E> + 'static> ValidatorStore
 }
 
 struct PublishDecision<E: EthSpec> {
-    signed_block: SignedBlock<E>,
+    signed_block: Option<SignedBlock<E>>,
     proposal_matched: bool,
     publish_path: &'static str,
 }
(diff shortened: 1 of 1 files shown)


shane-moore:
Closing as not planned. We’ll retain the existing publication fallback, including attempts that can fail with `NoPayloadBuilder`. Suppressing all nonmatching proposals would also remove potentially useful publication paths, and changing Lighthouse’s error handling is outside our scope.

The proposed implementation follows the issue’s suggested approach, but we’ve reconsidered that approach and don’t want to change publication behavior solely to reduce this error noise.

Issue #1295, opened 2026-09-06
The proposer preferences duty and the execution payload envelope duty have no tolerance for a stale duty view


ovidiu-ssv-labs:
## Summary

The Go SSV node does not apply the assignment check for the proposer preferences duty and the
execution payload envelope duty if its own duty view for that epoch is stale. Anchor has no concept of
a stale view. Anchor applies the check against the old data and ignores the message with `NoDuty`.

After a change of the validator set, this difference can stop the first proposer preferences duty of a
new validator for ever.

We did not see this defect in a live run, because our test network registers all validators before the
start. We report it from the code, and we give the test that can reproduce it.

## Test environment

- Network: `ssv-mini`, a local Kurtosis testnet with 4 SSV operators. The quorum is 3 of 4.
- Operators 1 and 2 run Anchor, commit `52ba589` on branch `epbs` (`Anchor/v1.3.1-52ba589`).
- Operators 3 and 4 run the Go SSV node, commit `613dd52c9` on branch `epbs-gloas`.
- Consensus client: Lodestar `v1.46.0/aa8303e`.
- Gloas fork epoch: 4 (slot 128). The Boole fork is not scheduled.
- 10 validators, registered before the start. Thus the count of validator set changes was 0.

## The difference

Go SSV node, in `message/validation/common_checks.go`, for the proposer preferences duty and the
execution payload envelope duty:

```go
if mv.dutyStore.Proposer.IsEpochSet(epoch) && !mv.dutyStore.Proposer.IsEpochStale(epoch) &&
    mv.dutyStore.Proposer.ValidatorDuty(epoch, slot, validatorIndex) == nil {
    return ErrNoDuty
}
```

There are two conditions: the epoch data is present, **and** the epoch data is fresh.

Anchor, in `anchor/message_validator/src/lib.rs` (about lines 911 to 928):

```rust
if duty_provider.proposer_assignment_at_slot(slot, &validator_pubkey) == DutyAssignment::NotAssigned {
    return Err(ValidationFailure::NoDuty);
}
```

`DutyAssignment` has three values: `Assigned`, `NotAssigned` and `Unknown` (about lines 1674 to 1680).
`Unknown` means that the epoch data is not present. There is no value for old data. A search for
`stale` in `anchor/message_validator/` gives no result about the age of the epoch data.

## Why this can stop a quorum

1. An operator adds a validator. The duty data of the other operators becomes old.
2. The Go SSV operators see that their data is old. They do not apply the check. They accept the
[...]

PR #1299, opened 2026-09-08
fix: validate PTC duty assignments

shane-moore:

Addresses the PTC assignment portion of #1295. Incoming PTC partial signatures currently have no duty-assignment check. [SIP-94 section 7](https://github.com/ssvlabs/SIPs/blob/791dfe36cf7cfa72fd1bb20497288592700e4c35/sips/epbs_support.md#L404-L413) requires ignoring messages for known unassigned slots while tolerating unknown duty views.

## Change Overview

- Poll current-epoch PTC duties for all locally known indexed validators, including validators whose messages this operator only relays.
- Track the exact queried indices in an atomically replaced snapshot. A queried validator omitted from the response is known unassigned; an index learned during the request remains unknown until queried. Failed or malformed responses retain the prior snapshot, and the previous epoch remains available for late messages.
- Apply the existing `NoDuty`/`Ignore` result only to known unassigned PTC messages. Unknown views continue through the existing validation checks. Proposer preferences, envelope duties, and Lighthouse are unchanged; #1295 remains open for its other concerns.

Querying all known indices adds beacon-node work. Lighthouse's producer cache covers local shares and does not record the query coverage needed for this receive-side check. The tracker therefore owns this snapshot. Full-network query capacity and a hard one-slot freshness bound are not established. Transition tests cover concurrent registration, failed refreshes, and epoch retention.

## Validation

- `cargo test -p duties_tracker -p message_validator --locked`: 161 passed, including eight new HTTP, snapshot, and admission regression tests.
- `make cargo-fmt-check`, `make lint`, `make sort`, and `git diff --check`: passed.
- Latest `make test`: 312 passed before two unchanged public-RPC tests failed. Direct probes confirmed hostname-resolution failures for their endpoints, `eth.merkle.io` and `ethereum-rpc.publicnode.com`.
- Local ssv-mini with one Anchor, three Go operators, Lodestar/Geth, and ten managed validators: 108 Go PTC partials across 36 validator duties matched Anchor's post-admission traces, with no new PTC polling or invalid-schedule errors. Runtime coverage is ordinary PTC receive traffic; known-negative rejection and concurrent registration are covered by regression tests. The runtime used the implementation commit; subsequent cleanup changes only tests and dev dependencies.

## Rollback

Revert this PR. The snapshot is in memory; no configuration or database migration is required.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5217,7 +5217,6 @@ dependencies = [
  "processor",
  "safe_arith",
  "sensitive_url",
- "serde_json",
  "sha2",
  "slot_clock",
  "ssv_types",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2614,6 +2614,7 @@ checksum = "92773504d58c093f6de2459af4af33faa518c13451eb8f2b5698ed3d36e7c813"
 name = "duties_tracker"
 version = "0.1.0"
 dependencies = [
+ "axum",
  "beacon_node_fallback",
  "bls",
  "dashmap",
@@ -2623,6 +2624,7 @@ dependencies = [
  "parking_lot",
  "safe_arith",
  "sensitive_url",
+ "serde_json",
  "slot_clock",
  "ssv_types",
  "task_executor",
@@ -5200,10 +5202,13 @@ dependencies = [
 name = "message_validator"
 version = "0.1.0"
 dependencies = [
+ "axum",
+ "beacon_node_fallback",
  "bls",
  "dashmap",
  "database",
  "duties_tracker",
+ "eth2",
  "ethereum_ssz",
  "fork",
  "hex",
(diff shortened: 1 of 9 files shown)


shane-moore (commented):
Reviewed `b9880834` against `epbs` at `ff74ca08`.

No findings survived the three-pass review. The snapshot records exactly the queried indices, preserves unknown coverage during registration changes, and replaces assignments atomically. The PTC gate ignores known-unassigned messages without consuming accepted-duty state; unknown views continue through the remaining checks.

Verified CI logs for the PR merge checkout: 952 tests passed in both debug and release, including all eight new PTC tests. Tests were not rerun locally for this review.

Full-network polling cost and a hard one-slot refresh bound remain unmeasured, as documented in the PR. No blocker found within this PTC-only scope; the remaining proposer/envelope concerns in #1295 are outside it.

Reviewed by gpt-6-astra high.

Issue #1291, opened 2026-09-06
The payload attestation duty always stops: the request for the payload attestation data uses the wrong URL format


ovidiu-ssv-labs:
## Summary

Anchor puts the slot in the URL path when it asks the beacon node for the payload attestation data.
Lodestar gives this endpoint with the slot as a query parameter. Lodestar answers with HTTP 404.
Anchor then writes `No block received for slot, skipping payload attestation` and stops the duty.

As a result, Anchor never sends a partial signature for the payload attestation duty. In a committee
that contains two Anchor operators, this duty never gets a quorum.

## Test environment

- Network: `ssv-mini`, a local Kurtosis testnet with 4 SSV operators. The quorum is 3 of 4.
- Operators 1 and 2 run Anchor. Operators 3 and 4 run the Go SSV node.
- Anchor: branch `epbs`, commit `52ba589`, version string `Anchor/v1.3.1-52ba589`.
  It pins `sigp/lighthouse` at rev `e58ec88fe`.
- Go SSV node: branch `epbs-gloas`, commit `613dd52c9`.
- Consensus client: Lodestar `v1.46.0/aa8303e`. All 4 operators use the same beacon node.
- Execution client: go-ethereum, `glamsterdam-devnet-8` image.
- Gloas fork epoch: 4 (slot 128). The Boole fork is not scheduled.
- 10 validators, indices 64 to 73.

## What we saw

We ran the network for 3 epochs after the Gloas fork.

- Anchor, on each of the two nodes: 36 x `Producing payload attestations`, and then
  36 x `No block received for slot, skipping payload attestation`.
  0 x `Successfully published payload attestations`.
- Lodestar: 73 x `warn: Route GET:/eth/v1/validator/payload_attestation_data/<slot> not found`.
  72 of these agree with the 72 Anchor skips. The times agree to the millisecond. The last one is
  our own manual test request.
- Go SSV node: 30 payload attestation duties. Each duty got only 2 partial signatures, from the two
  Go SSV operators. The quorum is 3. All 30 duties failed with
  `duty did not reach signature quorum before slot end`.
- The committee published 0 payload attestations.

## The two URL formats

```
GET /eth/v1/validator/payload_attestation_data/240        -> HTTP 404
GET /eth/v1/validator/payload_attestation_data?slot=240   -> HTTP 200
{"data":{"beacon_block_root":"0x8bff1f43...","slot":"240","payload_present":false,
 "blob_data_available":false},"version":"gloas"}
```

[...]

PR #1300, opened 2026-09-09
fix: admit preferences for locally assigned proposers

shane-moore:

Anchor can reject proposer-preference shares when its complete proposer tracker says a validator is unassigned, even though the local Lighthouse duty producer currently assigns that validator to the same slot. That disagreement can prevent locally scheduled signing work from collecting quorum. A receiver/collector regression reproduces the failure.

This is a narrow follow-up to #1295's proposer-view investigation. It does not implement general stale-view detection.

## Change Overview

Allow preference-role shares through the assignment check when the local producer's live cache contains the exact epoch, slot and validator assignment. Reuse that cache through client wiring rather than maintaining another schedule. Most of the diff is regression coverage.

The exception includes RequestAuth partials, which share the preference role. Envelope and other duty policies, remaining signature/timing/budget checks, and the Lighthouse pin are unchanged.

A local positive is not proof that Lighthouse has the newer view. An obsolete positive can admit obsolete shares, consume existing budgets, or complete already-active signing work. Admission alone does not start signing. The lookup reads the current cache, so it does not preserve an in-flight snapshot that Lighthouse has already replaced.

The exception is bounded to preference-role messages with an exact local assignment. Missing local evidence preserves the existing rejection. This deliberately avoids introducing freshness ordering between independently updated caches.

## Validation

- [x] Receiver regression fails before the change and completes a verified 3-of-4 BLS signature afterward, using real RSA validation, receiver delivery and collection.
- [x] Controls cover exact live-cache matching, missing/wrong epoch/slot/pubkey, invalid RSA, RequestAuth, and envelope exclusion.
- [x] `cargo test -p duties_tracker -p message_validator -p message_receiver -p client --lib --locked`: 174 passed.
- [x] `make test`: 959 passed, zero failed or ignored.
- [x] `make cargo-fmt-check`, `make lint`, `make sort`, `make audit-CI`: passed. Rust review and simplify found no remaining actionable changes.

ssv-mini, one Anchor plus three Go operators over four complete Gloas epochs: all 19 preferences published; 24 early collection timeouts recovered when Go began signing. Conflicting caches were not forced in this run.

Broader Gloas acceptance **failed**: all 17 managed proposals were missing and aggregates were rejected with the old Lodestar v1.44.0 image. Source checks and a block-root probe confirmed API/hashing incompatibilities; [Go SSV #3021](https://github.com/ssvlabs/ssv/issues/3021#issuecomment-5512708827) records the same image problem and an upgrade rerun. A compatible-image mixed-client rerun remains pending. PTC skips are separately tracked by #1291. These results do not claim general Gloas interoperability.

## Rollback

Revert this commit to restore the complete tracker's assignment policy for all preference messages. No configuration or data migration is required.


shane-moore (commented):
Reviewed at 037a61e0. No actionable findings survived the three-pass review.

The exception requires an exact epoch, slot, and validator match in the live local proposer cache. It covers preferences and RequestAuth while preserving envelope assignment policy and the remaining signature, timing, and budget checks. The receiver regression verifies completed 3-of-4 BLS reconstruction.

CI debug and release suites each passed 957 tests, including all five added tests. Formatting, lint, and local-testnet CI also passed.

The documented limits remain: local positive evidence does not establish which cache is newer, preserve a replaced signing snapshot, or prove general Gloas interoperability.

Reviewed by gpt-6-astra high.

PR #1301, opened 2026-09-10
feat: reject same-slot FULL votes the head event already rules out (SIP-94 §2)

shane-moore:

[SIP-94 section 2](https://github.com/ssvlabs/SIPs/blob/5d6addeb229e12a4025090a988a54c7995c45bba/sips/epbs_support.md#L110) now says an operator should reject a Gloas beacon vote with `AttestationDataIndex = 1` when its own view already establishes that the block at `BlockRoot` has slot `duty.slot`, evaluated against knowledge fixed at instance start. A block proposed in the current slot has no payload yet, so a FULL vote on it is invalid and the beacon node gossip-rejects every attestation the committee signs over it. SSV Labs QA measured exactly that on ssv-mini with a leader flipping the index ([SIPs#94 comment](https://github.com/ssvlabs/SIPs/pull/94#issuecomment-5621992247)); go-ssv tracks its side as ssvlabs/ssv#3035. Anchor today range-checks the index to `{0, 1}` and nothing more.

## Change Overview

- The metadata service already learns the needed fact: when a head event for the current slot triggers the voting context, that event names a block whose slot is the current slot, and a reorg never changes a block's slot. `VotingContext` gains `same_slot_head_root`, set from that event only when the event's slot equals the slot the context is built for (the two clock reads can straddle a boundary). Timer-triggered contexts carry `None`.
- The Gloas vote validator takes the root at construction, which happens once per committee instance, so proposals, round-change justifications, and decided messages in that instance evaluate the same knowledge. After the existing range check, index 1 on that root is rejected with a new `SameSlotFullIndex` error; every other vote is unchanged.
- No new task, channel, lock, or cache. With `--disable-beacon-head-monitor` the root is always `None` and behavior is unchanged. Pre-Gloas validation is untouched.

Coverage is a deliberately partial snapshot: a second same-slot head arriving after the first event, or a head event landing after the attestation-due timer, is unknown and the check is skipped, which the SIP permits. A rejection has no false positive, since an honest beacon node never yields index 1 for a same-slot block; it returns from `receive` and the shared committee instance (attestation and sync signing) waits for the round timer or another valid proposal, which is existing behavior for any value-check rejection. Reusing the block-header slot that weighted attestation data fetches was considered and left out: it is opt-in and only helps the timer path.

## Validation

- `cargo test -p ssv_types --lib -- gloas`: 30 passed, including four new cases (reject on the known root, index 0 accepted, other block accepted, no knowledge accepted).
- `cargo test -p anchor_validator_store --lib`: 174 passed, including three cases for the slot guard and one test through the production constructor path, because the committee tests' mock decider discards the validator.
- `cargo fmt --check`, `cargo clippy -p ssv_types -p anchor_validator_store --tests -- -D warnings`, `git diff --check`: clean.

## Rollback

Revert this PR. The field is in memory; no configuration or database change.

https://claude.ai/code/session_019jwWeHCBBac5Ma7Vrfonaw


shane-moore (commented):
Reviewed `bbe6fc46`. No correctness or design findings. The same-slot check uses a fixed instance snapshot and preserves the permitted behavior for unknown roots.

Formatting is fixed. CI passed all 965 tests in both debug and release, including the eight new cases; formatting and Clippy also passed. Still running: coverage, local-testnet.

Reviewed by gpt-6-astra xhigh.

PR #1302, opened 2026-09-22
feat(message_validator): admit batched builder request auth partial signatures

shane-moore:

- SIP-94 §5/§7 currently require every `RequestAuthPartialSig` packet to carry exactly one entry. Matheus proposed relaxing this on the SIP thread so an operator can gossip all of its builder auth shares for a proposal slot in one bounded packet, with per-builder isolation enforced by the receiving runner instead of by packet shape: https://github.com/ssvlabs/SIPs/pull/94#issuecomment-5695296209
- ssv-spec already emits and admits such batches (ssvlabs/ssv-spec#633). Old singleton-only receivers REJECT them, which penalizes the forwarding peer, so receive support has to propagate before any client emits batches.
- This PR is the receive side only. It evaluates the proposed amendment; the SIP text has not been updated yet.

## Change Overview (Required)

- Message validation admits `RequestAuth` packets on `Role::ProposerPreferences` with 1 to 8 entries. Above 8 is a Reject, empty is still a Reject.
- Every entry of such a packet must carry the same signer and validator index. A packet naming two validator indices is a new Reject-class failure, checked before the local-metadata membership check so it cannot be downgraded to an Ignore.
- The per-kind distinct-signing-root budget is now set-based: a packet adding no new root is an Ignore, a packet whose new roots would overflow the budget is an Ignore that records nothing, and otherwise all of its new roots are recorded together. For single-entry packets this is exactly the previous behavior, so `ProposerPreferences` packets need no special case and there is still one enforcement site.
- Read the diff as: one new error variant, one role arm split out of the one-entry rule, one budget site generalized from a root to a set.
- Intentionally unchanged: `ProposerPreferences` packets stay at one entry, every other role's packet rules, the sender (Anchor still emits one singleton packet per auth root), the signature collector, duty counting, and state retention. Configured-root matching stays in local collection, which already requests specific roots; gossip validation does not filter on private builder configuration.

- Blast radius is limited to `RequestAuth` packets on role 8. The risk is a classification slip (Accept/Ignore/Reject) or a budget slip on the batched path.
- Trade-off: the raw entry bound and the distinct-root budget share one constant (8), the SIP §5 configured-entry cap. Repeated roots inside a packet count as raw entries but only once toward the budget.
- Mitigation: the tests below pin every classification boundary, the exact budget fit, that an over-budget or forged packet records nothing, and that the preferences budget is untouched by a full auth batch.

## Validation (Required)

- `cargo test -p message_validator`: 161 passed (20 new). New tests cover singleton, two, and eight entry acceptance, empty and nine entry rejection, nine raw entries with repeated roots, two `ProposerPreferences` entries still rejected, inner signer drift and inner-vs-outer disagreement, mixed validator indices with metadata present, with an unknown index, and with empty local metadata, all-known versus mixed known/new packets, repeated in-packet roots counted once, exact budget fit, over-budget recording nothing, forged outer signature recording nothing, independent kind budgets, one duty slot for a batch plus a preferences packet, and the pre-Gloas fork gate.
- `cargo test -p signature_collector`: 35 passed (1 new). The new test feeds three-entry auth packets for roots A, B, C from each remote operator and registers local collection for A and B only. Both reconstruct; C keeps a share-only collector and no local partial is published for it. This is the concrete case from the SIP thread: configuration agrees on A and B, differs on C, and C must not block A and B.
- `make cargo-fmt-check` and `make lint` pass.

- Revert the commit. No config, database, or wire-format impact. Reverting only restores the singleton-only rule, so it must not happen once any peer emits batches.

- Draft until the SIP-94 amendment wording is agreed. No dependency on #1282, #1285, or #1283; none of them touch message validation.
- Reference state at the time of writing: ssv-spec#633 emits and admits batches; ssvlabs/ssv#2901 is still singleton on both sides and would REJECT a batch.

- Once #1282 lands, an end-to-end check through `sign_request_auth_v1` is worth adding: with local entries for A and B only, both resolve while peer packets also carry C; with a local C entry whose data differs, only that local root fails to reach threshold.


shane-moore:
Reviewed `b54c2c9f`. No blocking findings.

The receive-side change matches the current [SIP-94 §5/§7 text](https://github.com/ssvlabs/SIPs/blob/fc4f965d6b80292790d14cccb14d0fd46773ccb9/sips/epbs_support.md): 1 to 8 auth entries, consistent signer and validator index, atomic distinct-root accounting, and independent preference/auth budgets. The collector test verifies that unmatched root C does not prevent A and B reconstructing.

Checked the hosted CI logs: 986 debug and 986 release tests passed on the merge with current `epbs`, with the reviewed source files unchanged. Current lint, formatting and audit checks also pass.

Two nonblocking notes: the PR description still says the SIP amendment is unwritten, but it is now in the linked text. Receiver support must also reach the whole network before batch sending is enabled; the current go-ssv ePBS branch still rejects multi-entry auth packets. This review establishes receive-side compliance, not mixed-client rollout readiness.

Reviewed by GPT-6 Astra on high

